PR #1228, opened 2014-07-19
PSR-4 Autoloader

splinter89:
Hi!

I've recently found that Codeception doesn't support neither PSR-0 nor PSR-4 autoloading for its Helpers, StepObjects etc. So I took example autoloader with tests from [here](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-4-autoloader-examples.md).

Then I added support for empty namespace prefix, so one can use `Autoload::addNamespace('', 'path')` to load `Foo\Bar\ClassName` from `path/Foo/Bar/ClassName.php`.

Finally, this autoloader is backward compatible with an old autoloader, so all existing code should work as before.

**Old default autoloading:** all classes with `Helper` suffix from `helpers` dir. Namespace doesn't matter, because Autoloader doesn't look in subdirectories.

**New default autoloading:** according to PSR-4, all Helpers in global namespace and those in `Codeception\Module` will be autoloaded. Basically, it's done in runtime like this:
```php
// paths.helpers from codeception.yml
Autoload::addNamespace('', 'path/to/helpers');
Autoload::addNamespace('Codeception\Module', 'path/to/helpers');

// or if 'namespace' is set to 'frontend', for example:
Autoload::addNamespace('frontend', 'path/to/helpers');
Autoload::addNamespace('frontend\Codeception\Module', 'path/to/helpers');
```

The only issue users might face with strict PSR-4 autoloader is that classes in custom namespaces wouldn't be autoloaded, i.e.:
```php
Autoload::registerSuffix('Helper', 'path'); // in existing code
// is now equal to this
Autoload::addNamespace('', 'path');
// so if Foo\Bar\UserHelper is located in 'path/UserHelper.php',
// it will NOT be autoloadable via PSR-4 rules
$a = new \Foo\Bar\UserHelper; // Autoloader will look for it in 'path/Foo/Bar/UserHelper.php'
```

That's why in case of missing file for `Foo\Bar\UserHelper` using strict rules of PSR-4, this autoloader also tries to load it from `path/UserHelper.php`, just like an old autoloader did. Again, this behavior will not introduce any conflicts in existing code.

p.s. Maybe, this hack can be removed in Codeception :three:

### DIFF
--- a/src/Codeception/Util/Autoload.php
+++ b/src/Codeception/Util/Autoload.php
@@ -3,98 +3,149 @@
 namespace Codeception\Util;
 
 /**
- * Custom autoloader to load classes by suffixes: `Helper`, `Page`, `Step`, etc.
- *
+ * Autoloader, which is fully compatible with PSR-4, and can be used to autoload your `Helper`, `Page`, and `Step` classes.
  */
 class Autoload
 {
     protected static $registered = false;
-    protected static $map = array();
+    /**
+     * An associative array where the key is a namespace prefix and the value is an array of base directories for classes in that namespace.
+     * @var array
+     */
+    protected static $map = [];
+
+    private function __construct() {}
 
     /**
-     * A very basic yet useful autoloader, not compatible with PSR-0.
-     * It is used to autoload classes by namespaces with suffixes.
+     * Adds a base directory for a namespace prefix.
      *
      * Example:
      *
-     * ``` php
+     * ```php
      * <?php
-     * // loads UserHelper in 'helpers/UserHelper.php'
-     * Autoload::register('app\Codeception\Helper', 'Helper', __DIR__.'/helpers/');
-     * // loads LoginPage in 'pageobjects/LoginPage.php'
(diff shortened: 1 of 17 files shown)


DavertMik:
So let this show begin!

javigomez:
great! :clap:

javigomez:
Just tested, how does it look http://phptest.club/t/issue-with-codeception-autoload-and-namespaces/192/2?u=javi_gomez?

Would you use it this way, or other way?

src/Codeception/Util/Autoload.php
```diff
@@ -3,98 +3,159 @@
...
+        }
+
+        // fix for empty prefix
+        if (isset(self::$map['\\']) && ($class[0] != '\\')) {
+            return self::load('\\'.$class);
+        }
+
+        // backwards compatibility with old autoloader
+        // :TODO: it should be removed
+        if (strpos($class, '\\') !== false) {
```
Naktibalda, line 115: Removing this code breaks Codeception 5, I will remove TODO comment instead.

Issue #1262, opened 2014-08-01
Test code or tested code did not (only) close its own output buffers


dynasource:
I am getting this error now and then in combination with Yii2. I drilled the issue down to the occurence of an error NOTICE.

When I change error_reporting(E_NOTICE) to error_reporting(E_ERROR), the code continues.

It took quite a while to traverse through the code to get to this error. Isnt there a better response than "Test code or tested code did not (only) close its own output buffers"?

I prefer maximal insights. Why not outputting errors to the log area? Or to the console?

thanx

samdark:
Seems to be fixed.

PR #1325, opened 2014-08-18
Add yii events for app run

TriAnMan:
Add yii2 application events as mentioned in #1262 see: https://github.com/Codeception/Codeception/issues/1262#issuecomment-52208406

### DIFF
--- a/src/Codeception/Lib/Connector/Yii2.php
+++ b/src/Codeception/Lib/Connector/Yii2.php
@@ -77,7 +77,7 @@ class Yii2 extends Client
         $this->statusCode = null;
 
         ob_start();
-        $app->handleRequest($app->getRequest())->send();
+        $app->run();
         $content = ob_get_clean();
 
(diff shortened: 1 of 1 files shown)


samdark:
@TriAnMan is it still valid?

TriAnMan:
I think so. I use my own hacks to overcome the most issues I've posted. Have to recheck my source code for this particular issue.

samdark:
Thanks!

TriAnMan:
@samdark I've reviewed my code and found that I not use any workarounds for this issue.
As I remember I pulled it in case of ivestigation of tests ouptut buffers issue (#1262).
To be more precise: I've found that Codeception workflow differs from the yii's one and decided to make this PR. I have no particular issues with it because I have no events that should be run on an application startup. But someone may have.

Naktibalda:
2.0 is dead, could you make a pull request to 2.1?

PR #1504, opened 2014-11-01
Randomize the order tests run

schmitty42:
To ensure that tests don't depend on one another, randomize the order in which they run.


DavertMik:
`shuffle` option added. Can be configured in global config:

```
settings:
    shuffle: true
```
or in suite:
```
shuffle: true
```

by default shuffle is `false`

PR #1526, opened 2014-11-10
addFailure was firing Events::TEST_AFTER

### LINKED ISSUE #1229: canSee isn't working properly
bash-3.2$ bin/codecept --version
Codeception version 2.0.0

UpdaterCept.php - acceptance test:

$I->amOnPage("update?os=mac&version=6.2.0.666"); 
$I->canSee('some stuff');

fatal error:
Trying to Check old updater.php scenarios (UpdaterCept.php)                                                    [F]PHP Fatal error:  Call to a member function getHistory() on a non-object in /Applications/XAMPP/htdocs/vendor/codeception/codeception/src/Codeception/Lib/InnerBrowser.php on line 184

zbateson:
addFailure, called when can* functions like canSee fail, was firing
Events::TEST_AFTER which is meant to clean up after a test.  This
was causing errors if further tests were called after it as modules
would be in an uninitialized state.

This should fix #1229.  I had a look around to see how a failed can* functions can be tested but couldn't figure it out.  If there's a way, let me know and I'll look at adding something.


zbateson:
Any chance for this one to be included in the next release? Would be nice to have can* cant* working again :)

DavertMik:
I'm working hard on implementing 2.1 features. You can do PR into master without a fear to break some things. As for 2.0 probably I won't have a time to do this change there

zbateson:
Yeah, I'm happy to do a PR into master for this - the actual code change is only a few lines really and I'm fairly confident they're okay.  I'll look into that and post when I get a chance :)

DavertMik:
Thanks!

PR #1551, opened 2014-11-20
Dev 2.0 gsuite invalid chars

ha17:
Add a PHP variable rules check to g:suite so invalid characters such as dashes would not be used. Modifiedy the CLI/g:suite test to attempt to generate such a suite and verify the failure.


DavertMik:
Thanks. I had to merge this manually

Issue #1583, opened 2014-12-08
InnerBrowser doesn't allow passing in a path or domain when handling cookies.


mGuv:
I'm having issues correctly acceptance testing our product as all the cookie related functions on InnerBrowser only expect one parameter: the name of the cookie.

For example, in seeCookie, this ends up being passed to CookieJar::get() which expects name, path and domain. Only the name is used currently so path and domain default to "/" and null.

CookieJar indexes the cookies on their path, meaning it is impossible to specify a key other than "/". This means you can only interact with cookies defined in the top level domain and path, which is not always the case. In our example, we have a cookie that only has relevance on one page and for security, we want to keep it that way, rather than using the workaround of defining all cookies at the top level path. I tried passing extra parameters to seeCookie, just in-case it somehow worked but it did not.

InnerBrowser::seeCookie() (and others) should accept a path and domain parameter too, defaulting to the same as CookieJar::get().

I'm not sure if I'm missing something so I updated InnerBrower.php to allow me to pass in a path and domain, then rebuilt my test helpers and it works as I would expect now. E.g. $I->seeCookie("myCookie", "myPath"); will now work, whilst previously, it would not.

PR #1584, opened 2014-12-08
Add Path/Domain to cookies in InnerBrowser

mGuv:
Previously there was no way to access cookies outside of the top level.

See: https://github.com/Codeception/Codeception/issues/1583


DavertMik:
@mGuv I think I will write my own implementation for this case. 
I'd rather not use functional arguments per each cookie attribute but push them to array. This way we can set domain without changing path, as well as pass additional arguments like expire or secure to cookie.

PR #1599, opened 2014-12-17
Fix an erroneous cookie behavior of functional tests

TriAnMan:
This is something that hard to catch. But it is.

To reproduce you can run any functional test for Yii2 advanced app (installed as an application skeleton not as yii dev playground and having different domains for the frontend and the backend) and run something like `../vendor/bin/codecept run functional -vvv`
You will see a lot of such messages:

```
Scenario:
* I am on page "/index-test.php/site/login"
  [Response] 200
  [Page] http://localhost/index-test.php/site/login
  [Cookies] []
```

You will notice, that your domain is something different then `localhost`

It is hard to achieve an example code that will do a false negative pass.
But I have one in my application and this fix helps me to solve it.


DavertMik:
Ok, looks like I can close this.
Pls send PR if you find the real cause of your issue

PR #1617, opened 2014-12-31
Feature/magento module

betooliveira79:
Hi.

Im working on a Magento (http://magento.com/) Module for unit tests.

Until this moment the module allow easy integration of magento with tests

and  override helpers and model by mock/stubs.

The next steps is allow override blocks and fixtures for the models data.

Best regards.

### DIFF
--- a/src/Codeception/Lib/Magento/App.php
+++ b/src/Codeception/Lib/Magento/App.php
@@ -18,7 +18,8 @@ class App
     public function initMagento()
     {
         if ($file = $this->getBaseDir('app/Mage.php')) {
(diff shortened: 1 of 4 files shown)


DavertMik:
That sounds like a great present for a new year! :)

I don't have experience with Magento so you need to write documentation.

DavertMik:
@betooliveira79 I still have no idea how to use this module ) Please write a documentation for it.
Also as I don't know how to use this module I'd rather not include it into our core, I will ask you to publish your module as standalone package and add it as 3rd party module to [Addons](http://codeception.com/addons) section on site.

Thanks

PR #1618, opened 2015-01-01
Feature, MongoDb Count In Collection

sahanh:
Assert number of docs in a collection.
```
    $I->countInCollection('users', 2);
    $I->countInCollection('users', 1, array('name' => 'miles'));
```


DavertMik:
All assertion methods should start with `see` prefix.
So it should be like `$I->seeNumElementsInCollection` or.. provide better naming.

PR #1623, opened 2015-01-03
Instructions how to make Codecept available in $PATH

ajtrichards:
Just some simple instructions on how to make codecept available in the $PATH


DavertMik:
Thanks!
I think Global Composer installation would be useful as well )

PR #1624, opened 2015-01-04
always eval error level

koenpunt:
The default `errorLevel` is also a string, so it should be evaluated


DavertMik:
Yeah, right. Thanks!

PR #1625, opened 2015-01-05
Fix double slashes in certain forms submitted by InnerBrowser::submitForm

Revisor:
The Codeception\Lib\InnerBrowser method getFormUrl() can under certain
conditions create a form URL where the path begins with double slashes.

Conditions for the error:
1. The page with the form lies on a first-level path, e.g.
example.com/register
2. The form has an action without a leading slash, e.g.
<form action="login">

This is valid and browsers understand what is meant.

On Unix the malformed action URL will look like this:
http://example.com//form-target

On Windows this is even worse because dirname() of a first-level directory
returns a backslash (\) instead of a forward slash.

See http://www.webmasterworld.com/php/3752500.htm

Thus a form URL with Codeception run on Windows will look like this:
http://example.com/\/form-target

Both malformed URLs can cause server errors, depending on the server
configuration mostly 404 or 500.

This commit adds a test that confirms the error and fixes it
in a localized way that shouldn't regress anything else.

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -393,7 +393,13 @@ class InnerBrowser extends Module implements Web
                     $build[$part] = $build[$part] . $value;
                     continue;
                 }
-                $build[$part] = dirname($build[$part]) . '/' . $value;
+                $dir = dirname($build[$part]);
+                
+                // Do not double the slashes and guard against the Windows backslash
+                if ($dir === '/' || $dir === '\\') {
+                    $dir = '';
+                }
+                $build[$part] =  $dir . '/' . $value;
                 continue;
             }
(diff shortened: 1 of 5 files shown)


zbateson:
Nice catch, thanks Revisor.

It could be made to catch more possibilities and leaner with a nicely placed rtrim instead of the conditional though - what do you think of:

[code]
$dir = rtrim(dirname($build[$part]), '\\/') . '/' . $value;
[/code]

That way any extra trailing forward or back slashes would be trimmed regardless.

Thanks again! :+1:

DavertMik:
Thanks! Merging your pull request, and I think I will try so simplify with @zbateson suggestion.

Revisor:
Thank you both for simplifying and accepting the fix.

PR #1626, opened 2015-01-06
Get the id of a Facebook test user

ipalaus:
Add a new method to obtain the user id.

### DIFF
--- a/src/Codeception/Module/Facebook.php
+++ b/src/Codeception/Module/Facebook.php
@@ -190,6 +190,16 @@ class Facebook extends BaseModule
     }
 
     /**
+     * Returns the test user id.
+     *
+     * @return string
+     */
+    public function grabFacebookTestUserId()
+    {
+        return $this->testUser['id'];
+    }
+
+    /**
      * Returns the test user email.
      *
      * @return string
--- a/tests/unit/Codeception/Module/FacebookTest.php
+++ b/tests/unit/Codeception/Module/FacebookTest.php
@@ -56,6 +56,7 @@ class FacebookTest extends \PHPUnit_Framework_TestCase
     }
 
     /**
+     * @covers Facebook::haveFacebookTestUserId
      * @covers Facebook::haveFacebookTestUserAccount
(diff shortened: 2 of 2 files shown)


ipalaus:
PS: The failing tests are not related to my changes. Just checked and validated it's working.

Revisor:
It looks like you reformatted src/Codeception/Module/Facebook.php, since the diff tool is completely confused. Maybe line breaks?

DavertMik:
Yeah, it's hard to understand changes because of reformatting (

Revisor:
Davert, you can prevent line break reformatting globally by using the .gitattributes file.
https://help.github.com/articles/dealing-with-line-endings/

An example
```
# Auto detect text files and perform LF normalization
*       text=auto

# Denote text files explicitly, normalize line endings to LF on checkin and forbid conversion to CRLF on checkout
# Without eol=lf, core.autocrlf might come in effect
*.html  eol=lf diff=html
*.css   eol=lf
*.js    eol=lf
*.sql   eol=lf
*.php   eol=lf diff=php
*.yml   eol=lf
*.xml   eol=lf
*.dist  eol=lf
*.conf  eol=lf
*.json  eol=lf
*.lock  eol=lf

# Denote all files that are truly binary and should not be modified.
# Git would detect them automatically, but this will save the autodetection
*.png   binary
*.jpg   binary
*.gif   binary
*.ico   binary
*.zip   binary
*.jar   binary
*.docx  binary
*.docx  binary
*.dot   binary
*.xls   binary
*.otf   binary
*.eot   binary
*.ttf   binary
*.ttf   binary
*.woff  binary
*.phar  binary
```

DavertMik:
Thanks. Getting this merged!

@Revisor I will add .gitattributes as you suggested

DavertMik:
@ipalaus by the way, this module needs a new maintainer. Could you be the one? ;)
Currently I disabled Facebook tests as they are permanently failing. I don't use Facebook API so I don't have a clue what might go wrong there. Maybe you can help to figure out what is happening and have to get it fixed?

Issue #1628, opened 2015-01-07
Deprecated error when loading config file because of Symfony\Yaml 2.5+


freezy-sk:
Because of `"symfony/yaml": "~2.4"` every `build` and `run` command shows `PHP Deprecated:  Constant ENCODING in class Symfony\Component\Yaml\Unescaper is deprecated since version 2.5 and will be removed in 3.0.`

DavertMik:
Sorry, this commit can't be accepted as it is a very rough hook around runner.
Is there a way I can reproduce this issue on the latest changes of 2.0 branch?
I'm using Symfony 2.6

Issue #1571, opened 2014-12-01
`codecept run` is broken after `composer update`


xt99:
After running `composer update` _Codeception_ seems to be broken due to _Symfony_ components update : 

```
$ ./bin/codecept run --env backend --skip-group frontend --steps
PHP Deprecated:  \Symfony\Component\Console\Helper\DialogHelper is deprecated since version 2.5 and will be removed in 3.0. Use QuestionHelper instead. in /media/work/src/php/dse2/vendor/symfony/console/Symfony/Component/Console/Helper/DialogHelper.php on line 14
PHP Deprecated:  The "ProgressHelper" class is deprecated since version 2.5 and will be removed in 3.0. Use "ProgressBar" class instead. in /media/work/src/php/dse2/vendor/symfony/console/Symfony/Component/Console/Helper/ProgressHelper.php on line 17
PHP Deprecated:  The TableHelper class was deprecated in version 2.5 and will be removed in 3.0. Use Symfony\Component\Console\Helper\Table instead. in /media/work/src/php/dse2/vendor/symfony/console/Symfony/Component/Console/Helper/TableHelper.php on line 38
PHP Deprecated:  The ability to pass file names to Yaml::parse() was deprecated in 2.7 and will be removed in 3.0. Please, pass the contents of the file instead. in /media/work/src/php/dse2/vendor/symfony/yaml/Symfony/Component/Yaml/Yaml.php on line 58
Codeception PHP Testing Framework v2.0.8
Powered by PHPUnit 4.5-ge3692ba by Sebastian Bergmann and contributors.
PHP Deprecated:  The ability to pass file names to Yaml::parse() was deprecated in 2.7 and will be removed in 3.0. Please, pass the contents of the file instead. in /media/work/src/php/dse2/vendor/symfony/yaml/Symfony/Component/Yaml/Yaml.php on line 58
PHP Deprecated:  The ability to pass file names to Yaml::parse() was deprecated in 2.7 and will be removed in 3.0. Please, pass the contents of the file instead. in /media/work/src/php/dse2/vendor/symfony/yaml/Symfony/Component/Yaml/Yaml.php on line 58
PHP Deprecated:  The ability to pass file names to Yaml::parse() was deprecated in 2.7 and will be removed in 3.0. Please, pass the contents of the file instead. in /media/work/src/php/dse2/vendor/symfony/yaml/Symfony/Component/Yaml/Yaml.php on line 58

[...]

PR #1629, opened 2015-01-07
Suppressing E_USER_DEPRECATED errors thrown by vendor packages

freezy-sk:
This should fix problem with Symfony packages throwing E_USER_DEPRECATED errors from functions or classes which are used just inside Symfony functions and not used or extended directly by Codeception.
Solves issues #1628 #1571

### DIFF
--- a/codecept
+++ b/codecept
@@ -8,6 +8,20 @@ require_once dirname(__FILE__).'/autoload.php';
 
 use Symfony\Component\Console\Application;
 
+/**
+ * Suppressing E_USER_DEPRECATED errors thrown by vendor packages
+ */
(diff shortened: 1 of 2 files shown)


freezy-sk:
PS: Failing tests are not related to my changes

etki:
I'm not a part of Codeception team, but i strongly recommend to avoid any error suppression at all.
This error clearly states Symfony component behavior change that **will** result in error after 3.0 release if nobody will fix it.

DavertMik:
Sorry, this commit can't be accepted as it is a very rough hook around runner.
Is there a way I can reproduce this issue on the latest changes of 2.0 branch?
I'm using Symfony 2.6

PR #1633, opened 2015-01-11
allow redirects for guzzle connector

TorbenDaudistel:
This allows phpbrowser, which relies on guzzle, to follow redirects correctly.

### DIFF
--- a/src/Codeception/Lib/Connector/Guzzle.php
+++ b/src/Codeception/Lib/Connector/Guzzle.php
@@ -14,7 +14,7 @@ class Guzzle extends Client
 {
     protected $baseUri;
     protected $requestOptions = [
-        'allow_redirects' => false,
+        'allow_redirects' => true,
         'headers' => [],
     ];
     protected $refreshMaxInterval = 0;


DavertMik:
Not really sure that helps as redirects are handled by Symfony BrowserKit, not Guzzle.
We have bunch of tests for redirects, some are pretty tricky ones, and all they pass

TorbenDaudistel:
Yeah, I saw a strange behavior, maybe you have got an idea: When allow_redirects is set to false, my acceptance test follows the redirect but only passes the first url parameter. When setting allow_redirects to true, all url parameters are forwarded correctly.

DavertMik:
However, your PR fails our redirection tests
https://travis-ci.org/Codeception/Codeception/jobs/46627546#L9059

Can you reproduce your issue in Codeception tests?

Demo app:
https://github.com/Codeception/Codeception/blob/2.0/tests%2Fdata%2Fapp%2Findex.php#L25

Test:
https://github.com/Codeception/Codeception/blob/2.0/tests%2Funit%2FCodeception%2FModule%2FPhpBrowserTest.php#L121

TorbenDaudistel:
I checked your test. Is there a reason why the second url parameter (&sn=testnumber) is not involved in the assertion?

DavertMik:
Not sure ) You may add it there

PR #1635, opened 2015-01-12
Switch between database connections

codemedic:
The MultiDb module is loosely based on the MySQL Db module and has re-implemented some of the features.

Comments, suggestions, improvements ... all much appreciated

### DIFF
--- a/src/Codeception/Module/MultiDb.php
+++ b/src/Codeception/Module/MultiDb.php
@@ -1,5 +1,5 @@
 <?php
-namespace Redmatter\Codeception\Common\Module;
+namespace Codeception\Module;
 
 use Codeception\Exception\TestRuntime as TestRuntimeException;
 use Codeception\Lib\Driver\Db as Driver;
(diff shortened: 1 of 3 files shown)


codemedic:
Not sure why the build is broken.
I looked through the build log; couldn't see anything that I changed caused the breakage.

DavertMik:
Sorry, I feel this module is pretty complex and I'd rather not include it into core. 
All the Db features are hard to maintain, the more code Codeception has, the harder is to manage :(

If you don't mind, could you publish it as standalone package and add it to [Addons](http://codeception.com/addons) section on site. There are already some of 3rd party modules there.

Thanks

codemedic:
@DavertMik as advised ... I have created an extension.
https://packagist.org/packages/natterbox/codeception-multidb

Is it possible for this to be reviewed somehow. It is functional and it is internally used.
Also, is it possible to have a mention of the extension on codeception website?

Issue #1467, opened 2014-10-15
WebDriver module: selectOption and (dont)seeOptionIsSelected are broken with multiple radio button groups


MasonM:
If you have multiple radio button groups with some buttons having the same value, you won't be able to use the selectOption, seeOptionIsSelected, or dontSeeOptionIsSelected methods on anything except the first button with the desired value. This is hard to explain, so I [created a self-contained repository](https://github.com/theasci/codecept_bug) with a test case to reproduce this. To reproduce:
1. Run "git clone https://github.com/theasci/codecept_bug.git"
2. Run "cd code_cept_bug; composer install" to install Codeception
3. Start the Selenium standalone JAR. If you don't have it, [get it from here](https://code.google.com/p/selenium/downloads/detail?name=selenium-server-standalone-2.37.0.jar&can=2&q=selenium-server-standalone-2).
4. Run "vendor/bin/codecept run"
5. Should get a test failure with `Couldn't see option is selected "input[name=first_test_radio]","Yes":`

It appears this bug is in the [Codeception\Module\Webdriver->findCheckable()](https://github.com/Codeception/Codeception/blob/8a2a8fbd813af98371bbc3706c85e12937a94aa3/src/Codeception/Module/WebDriver.php#L761) function. Specifically, when the `$byValue` parameter is true, it appends the XPath `//input[./@type = 'radio'][./@value = $locator]` to the locator. However, that XPath expression will search the _entire_ document, not the children of the context node. Hence, if multiple radio buttons have the same value, it'll only match the first one.

PR #1638, opened 2015-01-12
Allow selectOption() to select options not inside forms.

n8whnp:
This fixes a regression in behavior that was introduced in #1468.
It includes a new unit test to ensure that this behavior can be preserved
going forward in future modifications. I preform two sets of checks so that
if the ancestor::form is catching things we don't risk selecting the wrong
element. Confirmed this works for the tests for issue #1467 as well.


DavertMik:
Thanks!

DavertMik:
@n8whnp please check my commit

n8whnp:
@DavertMik the commit visually looks good. I will run our test suites against it and confirm nothing unexpected shakes out. I captured what was failing in the unit test, so I am confident that it should continue to work.

n8whnp:
@DavertMik I have confirmed this addresses all varieties our test suite comes across and works w/o issue.

PR #1641, opened 2015-01-14
Fixing matchOption to return the option value in case there is no value attribute

synchrone:
Sometimes HTML select element only has value <option>like this</option>, but not the "value" attribute. So we need to return the text, much like we match it initially.


DavertMik:
Good, please resend it into 2.0. Thanks.

PR #1645, opened 2015-01-15
Remote coverage context options and headers fix

synchrone:


DavertMik:
@synchrone Could you provide some description about this PR, so I could mention it in changelog. 
Also please resend it into 2.0 branch so I could include it into next bugfix release

PR #1646, opened 2015-01-15
add a function that lets you provide additional headers before making an...

russellsimpkins:
... HTTP Request.

I noticed you have $this->headers but there's no way to set those. This function lets me say

$I->addHeaders(array("Content-Type"=>"application/jsonp"));

Or add any other headers I may need to test my application.


russellsimpkins:
I may have been too quick - I naively assumed that the logic to set headers would allow me to set headers in the request being made.

PR #1648, opened 2015-01-15
Adds note that explains helper classes must end in 'Helper.php'

Cameronjmayfield:
This note would have saved me some time integrating Codeception into our codebase.


docs/03-ModulesAndHelpers.md
```diff
@@ -42,6 +42,8 @@ Codeception doesn't restrict you to only the modules from the main repository. N
 
 It's a good idea to define missing actions or assertion commands in helpers.
 
+Note: Helpers class names must end with "*Helpers.php"
```
DavertMik: Helper.php not Helper**s**.php

samdark:
Thanks!

PR #1567, opened 2014-11-28
Add SSH key support to the (S)FTP module

mattvot:
Adding support for sftp connections with an RSA SSH key.

PR #1659, opened 2015-01-21
Resubmission of Add SSH key support to the (S)FTP module

mattvot:
For the next release.

Adding support for sftp connections with an RSA SSH key.

See https://github.com/Codeception/Codeception/pull/1567


DavertMik:
Something wrong. Please merge with 2.0 before resending

PR #1663, opened 2015-01-24
Fixing matchOption to return the option value in case there is no value attribute

synchrone:
Sometimes HTML select element only has value <option>like this</option>, but not the "value" attribute. So we need to return the text, much like we match it initially.


DavertMik:
Thanks

DavertMik:
Ok, I [have issues with this PR](https://travis-ci.org/Codeception/Codeception/jobs/48303019
) so I made a patch for it

PR #1664, opened 2015-01-24
Remote context options fix

synchrone:
When running remote coverage with multiple test suites, Coverage class gets multiple calls to *beforeSuite*, hence *applySettings* triggers *$this->c3Access* to be merged over and over again with yaml config. *array_merge_recursive* behaviour for duplicates is to create an array with two or more values, so the result is e.g 
```php
['http' => 'timeout' => [300, 300, ...]
```
, which is invalid http_stream_context options.

Same goes for *addC3AccessHeader*, which concatenates headers over and over again. While there's no harm in that, it's kind of wrong.

The whole remote coverage client would be better off by using a proper symfony http client, but this would be a bigger refactoring.


DavertMik:
Thanks!

Issue #1656, opened 2015-01-20
Symfony: Bug in Symfony2 Connector


hacfi:
I ran into a strange issue where running a single `Cept` works fine but running multiples results in an issue because the User stays logged in. I’ll create a test in the next few days to demonstrate the issue. I think the issue comes from `$this->hasPerformedRequest` (https://github.com/Codeception/Codeception/blob/2.0/src/Codeception/Lib/Connector/Symfony2.php#L13) being false as the new `Cept` starts but the container still contains services.

hacfi:
Adding a `$I->sendGET('/');` at the beginning of the test fixes the problem with the logged in user..still this issue should stay open.

PR #1668, opened 2015-01-25
Fix Symfony2 bug #1656 where kernel is not shutdown before next test

hacfi:
Explanation:
A new `Codeception\Lib\Connector\Symfony2` object is initiated for every test (see https://github.com/Codeception/Codeception/blob/2.0/src/Codeception/Module/Symfony2.php#L104) resetting the hasPerformedRequest property but reusing the old kernel.

In my case that resulted in a user being logged in on the first request of a new test..see #1656 .

### DIFF
--- a/src/Codeception/Lib/Connector/Symfony2.php
+++ b/src/Codeception/Lib/Connector/Symfony2.php
@@ -3,18 +3,18 @@ namespace Codeception\Lib\Connector;
 
 class Symfony2 extends \Symfony\Component\HttpKernel\Client
 {
-    private $hasPerformedRequest = false;
+    private static $hasPerformedRequest;
 
     public $persistentServices = [];
 
     protected function doRequest($request)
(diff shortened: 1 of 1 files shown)


DavertMik:
@hacfi Thanks!

hacfi:
Thanks for the merge @DavertMik

DavertMik:
@hacfi trying to figure out how `$hasPerformedRequest` is set to false?

hacfi:
@DavertMik It never is and it shouldn’t. The kernel should be shutdown for each request besides the first one (where it’s set to true).

DavertMik:
Ok, I see. Thanks

hacfi:
The kernel instance isn’t stateless - it contains all "activated"/accessed services and stores information like login tokens. Calling  `shutdown` is the same as a new request to the webserver.

Issue #1669, opened 2015-01-25
HTML Report, simple filter


raistlin:
I need simple filters in html report, and also improve html report.
Filters
* display all
* toggle view of success
* toggle view of failures
* toggle view of skipped
* toggle view of incompleted

I also think that summary is better to display it at top of html report, so I don't need to navigate to bottom page just to see if any test is failed.

PR #1672, opened 2015-01-26
Filter HTML report.

raistlin:
Added a 'toolbar' for easy filtering.
Click on every statusCode toogle visibility of given status.
Additionally there is a "Show all" button.

This time against 2.0


raistlin:
Some improvements of #1669

DavertMik:
Thanks!

DavertMik:
ups

DavertMik:
@raistlin looks like markup should be fixed a bit

raistlin:
I'm not a a good UI designer, CSS is black magic to me, but will try to fix it.

PR #1676, opened 2015-01-26
removing the @ so that the classloader errors are not swallowed. 

russellsimpkins:
With the @ the code will simply exit with exit code 255 if you have a syntax error and not tell you that you've introduced a syntax error. By removing the @ any syntax errors will properly throw a stack trace making it much quicker and easier to discover your problem.


DavertMik:
You'd better use `class_exists($className, false)` here

russellsimpkins:
I tried that at first, but if you do that, the classes don't get loaded.

DavertMik:
If you use only Composer's autoloader current code works correctly. As autoloader does not throw errors.

russellsimpkins:
Maybe I'm not fully grasping what you're telling me. I had a helper I created and noticed that codeception was not generating my actor class correctly. It took me about an hour to discover that the root cause was this @class_exists function. Because you put the @ infront of class_exists, the php parser exception was gobbled up and vendor/bin/codecept build silently failed. There was no error message, it simply exited with an exit code of 255. 

If I modify the code and remove the @ then if I introduce a PHP parser error - say by not ending a line with a ; or by not ending a function with } or by missing a ) (I happen to make a fair amount of typos) then when I run "build" codeception will just fail silently.

I also do not know why you guys are ignoring these errors, so I have no idea what kind of trouble it causes in other situations.

I'm not using my own custom autoloader, just the one provided by composer.

DavertMik:
@russellsimpkins sorry for misleading you. That was a good question why `@` is used in this case. 
Looks like it was introduced for compatibility with Yii1. 
Codeception runs Yii1 functional tests but its autoloader can throw errors in cases when class not found. Sure it is a shame to have it working that way :( Anyway, I'm not sure how this change affects our Yii users. I think this update should go to next major release, so not to introduce breaking change in minor.

russellsimpkins:
No worries. I really like your framework and just wanted to help out a bit.

Thanks,

Russ

DavertMik:
removed it in master )

PR #1681, opened 2015-01-28
 Anonymous function in PHP5.3 does not have $this

petrusek:
We had to write
@depends Codeception\TestCase\Cest::something
PHP Warning ("undefined variable this") was triggered when I changed the error reporting in the suite.

With this change @depends works as documented.


samdark:
The change is valid for 5.3 support.

tiger-seo:
Codeception 2.0 and higher requires PHP 5.4 installed. Download latest version of Codeception 1.8.x if you run PHP 5.3

samdark:
@tiger-seo it's a pull request for `Codeception:1.8` branch.

DavertMik:
:+1: thanks

PR #1683, opened 2015-01-30
set params on Rest module. Fixes #1682.

### LINKED ISSUE #1682: REST module, params never set
Am I missing something, or is the public property `$rest->params` never actually set? It always returns an empty array, even when I send URL params with a request. I'll work on a PR.

gmhenderson:


DavertMik:
Thanks!

PR #1684, opened 2015-02-02
Type null converted to empty string

borro:
For examlpe: 
```php
$I->sendPATCH('/api/category/1', ['parent_id' => null]);
```
Encode to json:
```json
{"PARENT_CATEGORY_ID":""}
```


DavertMik:
Thanks!

PR #1692, opened 2015-02-04
add "imports" directive to codeception.yml in order to load parameters f...

rchdevel:
add "imports" directive to codeception.yml in order to load parameters (variables by placeholders: %parameter%) from other yml files (so we can use parameters defined in other projects, like in Symfony2)

usage:
### Example (`codeception.yml`) - Symfony 2.x Directory Structure

    imports:
        - { resource: "app/config/parameters.yml" }   # relative to from where the codecept is running

    modules:
        config:
            Db:
                dsn: "%database_driver%:host=%database_host%;dbname=%database_name%_dev;charset=%database_charset%"
                user: "%database_user%"
                password: "%database_password%"

* Importing "parameters" from yml file - like in symfony2, with placeholders: %variable% 
* file (parameters.yml) has to be:
 paramaters
   variable1:    value
   variable2:    value

### DIFF
--- a/docs/modules/Symfony2.md
+++ b/docs/modules/Symfony2.md
@@ -24,6 +24,30 @@ This module uses Symfony2 Crawler and HttpKernel to emulate requests and test re
 * debug: true - turn on/off debug mode
 
 
+### Example (`codeception.yml`) - Symfony 2.x Directory Structure
+
+    imports:
+        - { resource: "app/config/parameters.yml" }   # relative to from where the codecept is running
+    modules:
+        config:
+            Db:
+                dsn: "%database_driver%:host=%database_host%;dbname=%database_name%_dev;charset=%database_charset%"
+                user: "%database_user%"
+                password: "%database_password%"
+                dump: tests/_data/dump.sql
+
+
+* Importing "parameters" from yml file - like in symfony2, with placeholders: %variable% 
+* file (parameters.yml) has to be:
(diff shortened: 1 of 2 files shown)


DavertMik:
That's nice change but I don't quite understand why do you update documentation for Symfony?
This is global change unrelated to Symfony module. 

Code looks nice but it requires a test. We have `ConfigurationTest` in unit, it uses data from original `codeception.yml` which you may update with placeholders for testing purposes.

DavertMik:
Sorry, this PR can no longer be merged automatically, as well as I don't see any responses on it. 
If it is still valid please resend it and add few more tests. If some things are unclear - never hesitate to ask

gmansilla:
How can I load variables from parameters.yml in codeception.yml?

DavertMik:
No way right now. We have separate configs. But you can try to update Codeception's config in runtime by accessing `Codeception\Configuration::append()`

gmansilla:
So, if this pull request had more unit tests would it be merged?
I find it easier to import a file than accessing configuration in run time

DavertMik:
>  if this pull request had more unit tests would it be merged?

at least one :) 
There is no way it can be done as a unit test (as Codeception runs its own tests, and relies on its configuration). But you can create a demo suite in `tests/data/claypit` with parameters loaded, and check that test receives those parameters.

PR #1702, opened 2015-02-06
Update SqlSrv.php

drewsonian:
select() needs the $criteria parameter to be passed by reference to be compatible with Db::select()


DavertMik:
thanks

PR #1707, opened 2015-02-09
fix put method in Yii2 connector

miroslav-chandler:
Hi.
There is a problem:
When i'm using ```sendPut``` in functional tests, ```Yii::$app->getRequest()->getBodyParams()``` is empty in controller.
It's happens because Yii2 connector is not putting ```methodParam``` field into POST array.


samdark:
Thanks!

PR #1711, opened 2015-02-11
Add dontSeeResponseJsonMatchesJsonPath method to REST module.

m14t:
This is the inverse for `seeResponseJsonMatchesJsonPath`

This pull request also updates the comment to specify which version of `flow/jsonpath` to install since `0.2.*` returns an instance of `\Flow\JSONPath\JSONPath` instead of an array.

### DIFF
--- a/src/Codeception/Module/REST.php
+++ b/src/Codeception/Module/REST.php
@@ -689,6 +689,19 @@ class REST extends \Codeception\Module
     }
 
     /**
+     * Checks if json structure does NOT match response [JsonPath](http://goessner.net/articles/JsonPath/).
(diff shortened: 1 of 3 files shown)


DavertMik:
> We can’t automatically merge this pull request.

Can you pull the updates and resend this PR?
Recently I have merged something really similar )

m14t:
@DavertMik Thanks. I hadn't seen that.  Will there be a tag for this?

Also, it may be worth still using the comment that I added (commit  7cc7851) in here since the version that's in there now does not work with `flow/jsonpath` `0.2.*`.

PR #1714, opened 2015-02-12
Fix quiet crash when crawler is null

aivus:
When `amOnPage` is not called and try to call some function like fillField (with XPath-rule) the codeception quiet crashed without any errors.

It happened because `crawler->filterXPath` wrapped by "`@`".

This PR provide additional check on `crawler` exist


DavertMik:
Yeah, that's really helpful. Thank you!

PR #1717, opened 2015-02-13
Fixed problem with extensions being always loaded with default options

### LINKED ISSUE #1716: Extensions are loaded with default options
In loaded extension there is no access to config by options in console. 

For example:

    $this->output->getVerbosity()

always return '1' even if script is started with '-vvv'

As I found in https://github.com/Codeception/Codeception/blob/master/src/Codeception/Codecept.php#L65 extensions are loaded before merging config from file and constructor parameter.

sjableka:
Fixes #1716


DavertMik:
This break current test suite. Please run `codecept run cli` to check how the test differs from your PR

sjableka:
@DavertMik Sorry about that, it is fine now.

src/Codeception/Codecept.php
```diff
@@ -57,12 +57,14 @@ class Codecept
...
+
+        $baseOptions = $this->mergeOptions($options);
+
+        $this->loadExtensions($baseOptions);
 
         $this->config = Configuration::config();
 
-        $this->options = array_merge($this->options, $this->config['settings']);
-        $this->options = array_merge($this->options, $options);
+        $this->options = $this->mergeOptions($options);
```
DavertMik: is there a reason why `mergeOptions` is executed 2 times on the same data?
can merged options be reused?

sjableka:
It's not the same data. It calls Configuration::config() to get current configuration, which may be changed during extenstion loading. Without calling it twice, tests using SimpleOutput class are failing.

src/Codeception/Codecept.php
sjableka: @DavertMik you can find my answer below ;)

DavertMik:
Ok, got it. Thanks.

PR #1722, opened 2015-02-17
Update greater/less asserts to match functionality

### LINKED ISSUE #1721: Asserts module greater/less than documentation is opposite of functionality
In the [Asserts Module documentation](http://codeception.com/docs/modules/Asserts#assertGreaterThan), it is claimed that when using `assertGreaterThan($expected, $actual)`, the expression `$expected > $actual` should evaluate to `true`. The order of parameters is the same for `assertGreaterThanOrEqual`, `assertLessThan`, and `assertLessThanOrEqual`.

However, this is the opposite of actual functionality, as documented in [PHPUnit](https://phpunit.de/manual/current/en/appendixes.assertions.html). This can be tested with the following Cest code:

    public function test_asserts(ApiTester $I)
    {   
        $I->assertLessThan(2,1);
        $I->assertLessThanOrEqual(2,1);

        $I->assertGreaterThan(1,2);
        $I->assertGreaterThanOrEqual(1,2);
    }

Each of these pass, although they contradict the documentation.

apkostka:
Fixes #1721

### DIFF
--- a/docs/modules/Asserts.md
+++ b/docs/modules/Asserts.md
@@ -44,7 +44,7 @@ Checks that condition is negative.
 
 ### assertGreaterThan
  
-Checks that expected is greater than actual
+Checks that actual is greater than expected
 
  * `param`        $expected
  * `param`        $actual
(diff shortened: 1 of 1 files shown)


DavertMik:
this change should be done in `Asserts` module itself, `.md` file is generated from it

Issue #1680, opened 2015-01-28
Laravel middleware not being run


harhoo:
This may be an edge case but:

Running API tests via Codeception. All works nicely except middleware is not run.

I can enable filters using $I->haveEnabledFilters() but this doesn't run middleware.

The service provider that registers the middleware is called, but the middleware itself is not used or run. I have some reasonably complicated, potentially flaky middleware so I'd like to run tests against them.

Set up:

_bootstrap.php:
```php
include __DIR__.'/../../vendor/autoload.php';
$unitTesting = true;
$testEnvironment = 'testing';
$app = require_once __DIR__.'/../../bootstrap/start.php';
\Codeception\Util\Autoload::registerSuffix('Page', __DIR__.DIRECTORY_SEPARATOR.'_pages');
$app->boot(); 
```
api.suite.yml

```yml
class_name: ApiTester
modules:
    enabled: [Laravel4, ApiHelper, PhpBrowser, REST]
    config:
      PhpBrowser:
          url: http://localhost/
      REST:
          url: http://localhost/api/
```

Service provider:

```php
    public function register()
    {
        $this->app->middleware('MyApp\Service\Http\ApiRights', [$this->app]);
    }
```

jotweh:
I had a similar problem and I figured out, the stacked http kernel (which adds the middleware functionality) is not being used. Solution is to update the Laravel4 Connector so that it includes the middleware (see my pull request)

PR #1724, opened 2015-02-17
Laravel4: fix problem with middleware not being loaded

jotweh:
See also https://github.com/Codeception/Codeception/issues/1680


DavertMik:
Thanks! This should be really useful for Laravel users

PR #1725, opened 2015-02-17
More dynamic submitForm

zbateson:
Removed submitForm's reliance on using parse_str and parse_url to
generate params (which caused unexpected side-effects like failing
for values with ampersands).

Modified the css selector for input elements so disabled input
elements don't get sent default values.

Modifications to ensure multiple values get sent correctly.


DavertMik:
Thanks. I will need to take a deeper look into it then I'm merging

zbateson:
No problem @DavertMik - feel free to pass back any problems you see.

DavertMik:
Yeah, looks awesome!

PR #1727, opened 2015-02-18
Ability to have another primary key then 'id'

KennethVeipert:
We have the problem that the primary key isn't named 'id'. That means that the cleanup of the Module 'Db' not delete our entries. This fix helped us and should help everyone else who have the same problem.

### DIFF
--- a/src/Codeception/Module/Db.php
+++ b/src/Codeception/Module/Db.php
@@ -113,6 +113,13 @@ class Db extends \Codeception\Module implements \Codeception\Lib\Interfaces\Db
     protected $insertedIds = [];
 
     /**
+     * associative array with table name => primary-key
+     *
+     * @var array
+     */
+    protected $primaryKeys = [];
+
+    /**
      * @var array
      */
     protected $requiredFields = ['dsn', 'user', 'password'];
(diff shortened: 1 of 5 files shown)


src/Codeception/Module/Db.php
```diff
@@ -157,26 +178,30 @@ protected function removeInserted()
     {
         foreach ($this->insertedIds as $insertId) {
             try {
-            $this->driver->deleteQuery($insertId['table'], $insertId['id']);
+                $pdo  = $this->driver->getDbh();
+                $stmt = $pdo->prepare(
+                  'DELETE FROM `' . $insertId['table'] . '` WHERE `' . $insertId['primary'] . '` = ?'
```
DavertMik: Is there a reason why you use raw query here? I think it would be better to move SQL outside Db module into db drivers we have.

src/Codeception/Module/Db.php
```diff
@@ -239,20 +265,54 @@ public function haveInDatabase($table, array $data)
...
-    public function seeInDatabase($table, $criteria = array())
+    /**
+     * @param string $tableName
+     *
+     * @return string
+     * @throws \Exception
+     */
+    protected function getPrimaryColumn($tableName)
+    {
+        if ('sqlite' === $this->driver->getProvider($this->config['dsn'])) {
```
DavertMik: This case should not be handled inside a module but inside Sqlite driver

DavertMik:
@KennethVeipert thanks for this PR. It looks very useful, however I'd ask you to do some refactorings in order to keep database specific code separated from the module itself. Thanks

KennethVeipert:
Hi, sorry for the late answer.
What's the simplest way to get the current primary key of the table cross all supported database-systems? I know that this SQLite-handling isn't the best way but it's a fast hotfix.

KennethVeipert:
Hi, sorry for the late answer. I submitted a fix for that and added 2 ToDo's for the SQLite-driver(Can't test it currently, sorry). Please review my changes and if everything is okay merge the PR.

DavertMik:
btw, do you use some kind of auto-formatter to replace `array()` to `[]`?

KennethVeipert:
Yes, PhpStorm AutoFormat - Is it a problem?

DavertMik:
No, just wanted to use this as well )
Strange, my PHPStorm did not do this change. Probably I need to play with settings a bit. 
Thank you!

KennethVeipert:
You're welcome ;)
Screenshot

PR #1728, opened 2015-02-18
Fixed issue with domain-specific routes

janhenkgerritsen:
This pull request fixes an issue with domain-specific routes in the Laravel 5 module that was reported to me. I also added a test to the sample app that covers this issue. I will push this test to the sample app repository after the fix is included in Codeception, to prevent failing tests.


DavertMik:
Thanks!

PR #1733, opened 2015-02-20
updated to phpunit 4.5 using next significant release tag ~

rgeraads:


verschoof:
I'm not sure, but @DavertMik did change it to `4.4.*`, I don't know why because he didn't describe it in his commit :)

DavertMik:
@verschoof actually I did this change because PHPUnit 4.6 has some breaking changes in API so I decided to have releases with fixed versions. The next one can go with 4.5, no problems with that )

PR #1734, opened 2015-02-20
HTML entity encode test "name" for HTML reports

### LINKED ISSUE #1712: Test name and dataset text in HTML report is not entity encoded
The report.html that is rendered from ```codeception run unit --html``` may contain _test names_ or _dataset values_ that would trigger browser behavior.  Almost certainly it would be from datasets in most practical occurrences.

If the test (function testFoo($input)) receives input, this input is destined to be in the test's name (line item in html, console or otherwise)...if that ```$input``` happens to contain HTML special characters, then it will be allowed to flow directly to the final rendered file without any treatment, such as ```htmlentities()```.  That is, in ```Codeception\PHPUnit\ResultPrinter\HTML::onTest()```, the ```$name``` argument is trusted and simply passed to the template.  I presume that $name is generated based on a Step (more on that below).

I hit this condition when attempting to generate a friendly HTML report from some unit tests, which included tests for a output filter class
```bash
\Input\FilterTest::testEntityEncode with data set "word and HTML script" ('abc<script>alert('hi');</script>def', 'abc&lt;script&gt;alert(&#039;...gt;def')
```
The most immediate solution would be to ```htmlentities($name)``` [here](https://github.com/Codeception/Codeception/blob/2.0/src/Codeception/PHPUnit/ResultPrinter/HTML.php#L112).

I thought about if it were possible to do it one layer up, in the template (scenario.html.dist).  But the template system appears to be ```\Text_Template``` which is very rudimentary and doesn't allow any variable processing.  Even if it were possible, I have reservations about the limit of a template layer fix vs some internals.

I'm actually concerned that the fix in HTML.php (or potential template layer) is not enough.  I assume that ```onTest()``` is called by the harness and fed data based on a Step or type of Testcase, but I have yet to delve into it.  If it is related to a Step, this is where it gets a little soupy... I don't understand the innards of the Step to make a proper assertion or have a clear expectation of the range of data that would get processed, normally, from ```$action``` and ```$arguments``` (the constructor args), nor do I know enough about what/when the concrete Steps are factory'd, so bear with me.  But, at a glance, if we look in the abstract Step class we have a concrete method, ```Step::getHtmlAction()```, clearly it is providing markup yet implicitly trusts $action and $arguments to be "safe".  The various concrete decedents of this class overload this method, and further confuse me about when "safe" is _safe_.  

I also thought about altering the my tests, but honestly...should test data (particular ones that are testing safety of output) really be concerned with how it _might_ be take literally by the report output layer (in a harness that was clearly designed to handle output formatting)?

Thanks for your time!

davidneimeyer:
Test "name" will render the method test name and any data provided as
an argument to the test method.  If that data is HTML, it can
influence the browser when viewing the HTML report, namely a chance
for XSS.

Fixes #1712

### DIFF
--- a/src/Codeception/PHPUnit/ResultPrinter/HTML.php
+++ b/src/Codeception/PHPUnit/ResultPrinter/HTML.php
@@ -109,7 +109,7 @@ class HTML extends \Codeception\PHPUnit\ResultPrinter
         $scenarioTemplate->setVar(
           array(
             'id'             => ++$this->id,
-            'name'           => $name,
+            'name'           => htmlentities($name),
             'scenarioStatus' => $scenarioStatus,
             'steps'          => $stepsBuffer,
 	        'time' => round($time, 2)


DavertMik:
Hi, thanks for the fix. 
I hope chance of XSS is minimal here. It is just a static page with HTML report.
Still it is better to have HTML output to be formatted properly

PR #1738, opened 2015-02-23
FTP Passive Mode

PH2M:
Hello Michael,

I found an issue of FTP module:

1. Passive mode do not get config value. Line 578 => You compare `$this->config['passive’]` with enabled instead of `true` or `1`
2. (Most important) You define passive mode before login, and you have to set it just after login.

I found the solution here :
- http://stackoverflow.com/a/1612323/3437153
- http://biostall.com/phps-ftp_nlist-returning-false-even-when-files-definitely-exist

With this pull request, FTP Module is fully operational.


DavertMik:
Thanks

PR #1743, opened 2015-02-25
add request headers logging

honzabrecka:
I miss them in output.


DavertMik:
Yep, they could be helpful )

PR #1745, opened 2015-02-26
Support for multiple databases

ovide:
Only a small change to #1727 .
If the table is in another database the string is like \`other_database.table_name`. Removing the backtick works well:

I know that ideally should escape the string, but there are [plenty](https://github.com/Codeception/Codeception/blob/2.0/src/Codeception/Lib/Driver/MySql.php#L26) of [lines](https://github.com/Codeception/Codeception/blob/2.0/src/Codeception/Lib/Driver/Db.php#L168) [where](https://github.com/Codeception/Codeception/blob/2.0/src/Codeception/Lib/Driver/Oci.php#L8) is [not covered](https://github.com/Codeception/Codeception/blob/2.0/src/Codeception/Module/Dbh.php#L91) 

```php
$I->haveInDatabase('other_db.my_table', $data); //was working until 2.0.10
$I->haveInDatabase('my-speçial-table-ñame', $data); //never worked
```
So I think that escaping table name should be fixed in another patch.

### DIFF
--- a/src/Codeception/Lib/Driver/Db.php
+++ b/src/Codeception/Lib/Driver/Db.php
@@ -208,7 +208,7 @@ class Db
     public function getPrimaryColumn($tableName)
     {
         if (false === isset($this->primaryColumns[$tableName])) {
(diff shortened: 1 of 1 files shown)


DavertMik:
Thanks. Waiting for upcoming patch as well! )

Issue #1208, opened 2014-07-15
Laravel4 does not work when running a test suite across multiple Packages


joshbenham:
For example if i do

```
codeception bootstrap workbench/vendor/package1
codeception bootstrap workbench/vendor/package2
```

Then I create a `codeception.yml file on the base of my laravel installation with something like.

```yaml
include:
    - workbench/vendor/package1
     -workbench/vendor/package2
paths:
    log: log
settings:
    colors: true
```

I willl enable the Laravel4 module on package1 and rebuild codeception. Then from the base when I do `codeception run` I get.

```
  [Codeception\Exception\ModuleConfig]                                                                                                        
  Laravel4 module is not configured!                                                                                                          
  Laravel start.php file not found in /media/bravo/Development/foxdigitalweb/bitbucket/foxdigitalweb/workbench/vendor/package1/bootstrap/start.php.  
  Please provide a valid path to it using 'start' config param.   
```

Because https://github.com/Codeception/Codeception/blob/2.0/src/Codeception/Module/Laravel4.php#L71 is actually wrong as when it runs package1 tests it will read `codeception.yml` from `/workbench/vendor/package1/codeception.yml` and not just `/codeception.yml`.

To fix this I just change like #71 to 

```php
$projectDir = explode('workbench', \Codeception\Configuration::projectDir())[0];
```

It could probably be done a nicer way but this is just a quick fix on my side.

PR #1755, opened 2015-02-28
Update project directory to work on workbench tests.

jonathantorres:
Same issue as https://github.com/Codeception/Codeception/issues/1208 but with Laravel 5.


jonathantorres:
Better fix here: https://github.com/Codeception/Codeception/pull/1782

PR #1750, opened 2015-02-27
Fixed "Call to undefined method Symfony\Component\HttpFoundation\Request::route()"

jedrzej:
This is a fix to #1748.

Issue #1748, opened 2015-02-26
Laravel5 Request::route undefined


simnom:
Hey,

Getting the following error when running a functional test for Laravel 5:

```
PHP Fatal error:  Call to undefined method Symfony\Component\HttpFoundation\Request::route() in /var/www/vhosts/buildingcentre.knauf.co.uk/vendor/laravel/framework/src/Illuminate/Foundation/Http/Kernel.php on line 146
```

Associated yaml config is:

```
class_name: ApiTester
modules:
    enabled:
        - REST
        - Laravel5
        - ApiHelper
        - Asserts
    config:
        Laravel5:
            environment_file: .env.testing
```

EKOInternetMarketing:
I also received this error as of this morning.

bcdennis:
Ditto

mnabialek:
I have exact same message. Cannot run tests at the moment

constb:
this is caused by terminable middlewares. they were causing troubles before if you used them - they expected to receive ``\Illuminate\Http\Request`` but Laravel5 Connector was giving them ``\Symfony\Component\HttpFoundation\Request`` instead. but with some workarounds this was managable.

now that laravel application supports terminable *route* middlewares, it tries to gather them itself in ``terminate`` by accessing ``Request::route``. that where it goes down.

EKOInternetMarketing:
Thanks for the fix. It's working perfectly.

mikeerickson:
So far as I can see (updated to latest version of Codeception along with rebuilding the modules so I pull in latest Laravel5 module) this is still having issues.

http://d.pr/i/DcQj

bastiaan89:
Yeah it's only been pulled in the 2.0 branch but not master or any tags. @DavertMik, please merge it in? This is kind of a big deal, having to modify it manually on each update.

PR #1756, opened 2015-03-02
Fix for undefined method Symfony\Component\HttpFoundation\Request::route()

janhenkgerritsen:
This is a fix for #1748, based on the fix proposed in #1750 by @jedrzej. 

I confirmed the issue when I did a `composer update` on the sample application, and this commit fixes the issue.

### DIFF
--- a/src/Codeception/Lib/Connector/Laravel5.php
+++ b/src/Codeception/Lib/Connector/Laravel5.php
@@ -65,6 +65,7 @@ class Laravel5 extends Client implements HttpKernelInterface, TerminableInterfac
 	 */
 	public function terminate(DomRequest $request, Response $response)
 	{
-		$this->httpKernel->terminate($request, $response);
+		$this->httpKernel->terminate(Request::createFromBase($request), $response);
 	}
+
 }


DavertMik:
@janhenkgerritsen Thanks
If this fix is really important I can release 2.0.12 right now

janhenkgerritsen:
@DavertMik it is an important fix, because the issue causes fatal errors even for default Laravel installations.

ssfinney:
I would love a fix for this too!

javichito:
Please we need the 2.0.12 release.

carlituxman:
when the 2.0.12 release???

janhenkgerritsen:
@DavertMik when can you release 2.0.12? I get a lot of complaints of Laravel users, not only here but also by email :)

IsraelOrtuno:
We need this release, cannot test Laravel 5

clockwork-coder:
Afraid I have to agree with the other posters here; it's been making our CI tests fail for over a week now, which really doesn't look good; workarounds are only good for local testing...

janhenkgerritsen:
For everybody waiting for this release, I just updated the old Laravel5 module project you can find at https://github.com/janhenkgerritsen/codeception-laravel5. This now includes the fix for this error.

However, to make this work you should use a version of Codeception in your project that does not include the Laravel 5 module yet, otherwise the incorrect classes will be loaded. You should specify a version prior to `2.0.11`, for example `2.0.10.1`.

This is also a workaround, but one that should also work for CI tests.

mabasic:
Waiting for 2.0.12 release :)

PR #1767, opened 2015-03-10
Remove unnecessary parameter from Webdriver PHPDoc

FnTm:
The tag `@internal` was causing PhpStorm to consider the method as internal. Additionally, it has no meaning for this method, since there is no variable `$element` used in the method the PhpDoc describes - `waitForText`


DavertMik:
thanks

Issue #1768, opened 2015-03-10
Support for absolute log path in config


pfazzino:
Currently log path is assumed as a sub-folder of the configuration folder which means codeception cannot be run when located in a readonly folder.

pfazzino:
pr submitted

PR #1769, opened 2015-03-10
support for log absolute path

pfazzino:
Currently log path is assumed as a sub-folder of the configuration folder which means codeception cannot be run when located in a readonly folder.

### DIFF
--- a/src/Codeception/Configuration.php
+++ b/src/Codeception/Configuration.php
@@ -421,7 +421,11 @@ class Configuration
         if (!self::$logDir) {
             throw new ConfigurationException("Path for logs not specified. Please, set log path in global config");
         }
-        $dir = self::$dir . DIRECTORY_SEPARATOR . self::$logDir . DIRECTORY_SEPARATOR;
+
+        $dir = self::$logDir . DIRECTORY_SEPARATOR;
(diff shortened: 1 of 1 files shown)


pfazzino:
issue - https://github.com/Codeception/Codeception/issues/1768

src/Codeception/Configuration.php
```diff
@@ -421,7 +421,11 @@ public static function outputDir()
         if (!self::$logDir) {
             throw new ConfigurationException("Path for logs not specified. Please, set log path in global config");
         }
-        $dir = self::$dir . DIRECTORY_SEPARATOR . self::$logDir . DIRECTORY_SEPARATOR;
+
+        $dir = self::$logDir . DIRECTORY_SEPARATOR;
+        if(strcmp(self::$logDir[0], "/") !== 0){
```
splinter89: Would it work on Windows?

pfazzino: No. It would fallback to current behaviour and treat the path as relative.
I don't easily have access to windows machine but could spin up a VM and
look at a further PR to refine this.

DavertMik:
Thanks, could you include support for Windows directories as well?

PR #1770, opened 2015-03-10
Fix https support and verbose output

TriAnMan:
This patch enables https support for functional suites.
Also this patch fix verbose output for functional suites.

E.g.: ../vendor/bin/codecept run functional -vvv

Was:

```
Scenario:
* I am on page "/index-test.php/site/login"
  [Response] 200
  [Page] http://localhost/index-test.php/site/login
```

After fix:

```
Scenario:
* I am on page "/index-test.php/site/login"
  [Response] 200
  [Page] http://YOURDOMAIN/index-test.php/site/login
```

### DIFF
--- a/src/Codeception/Module/Yii2.php
+++ b/src/Codeception/Module/Yii2.php
@@ -55,8 +55,11 @@ class Yii2 extends Framework implements ActiveRecord
     {
         $this->client = new \Codeception\Lib\Connector\Yii2();
         $this->client->configFile = \Codeception\Configuration::projectDir().$this->config['configFile'];
-        $this->client->setServerParameter('HTTP_HOST', parse_url(\Codeception\Configuration::config()['config']['test_entry_url'], PHP_URL_HOST));
-        $this->client->setServerParameter('HTTPS', parse_url(\Codeception\Configuration::config()['config']['test_entry_url'], PHP_URL_SCHEME) === 'https');
+        $mainConfig = \Codeception\Configuration::config();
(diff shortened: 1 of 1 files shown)


src/Codeception/Module/Yii2.php
```diff
@@ -55,6 +55,8 @@ public function _before(\Codeception\TestCase $test)
     {
         $this->client = new \Codeception\Lib\Connector\Yii2();
         $this->client->configFile = \Codeception\Configuration::projectDir().$this->config['configFile'];
+        $this->client->setServerParameter('HTTP_HOST', parse_url(\Codeception\Configuration::config()['config']['test_entry_url'], PHP_URL_HOST));
```
DavertMik: I"m not sure, where does `test_entry_url` come from?

TriAnMan: For example from here:
https://github.com/yiisoft/yii2-app-advanced/blob/master/tests/codeception/frontend/codeception.yml#L17

DavertMik: Oh, this is not an official specification, so I'd ask you to wrap your code into `if` block, where you check this parameter (as well as `config` section) actually exists

TriAnMan: Hmm... There are a lot of Yii2 code related to this parameter already e.g. https://github.com/yiisoft/yii2-app-advanced/search?q=test_entry_url
But your idea seems reasonable to users that have their applications installed before this paramater was introduced.

DavertMik:
thank you!

PR #1776, opened 2015-03-13
Fix Yii 2 haveRecord not inserting unsafe fields

### LINKED ISSUE #1775: Yii 2 haveRecord does not insert unsafe attributes
By default, [setAttributes](http://www.yiiframework.com/doc-2.0/yii-base-model.html#setAttributes()-detail) will only set attributes that are safe, i.e. those that are expected to receive user input.  [haveRecord](https://github.com/Codeception/Codeception/blob/1497a8b9fcda64e5687d314e1e3ac522ae7e3cb7/src/Codeception/Module/Yii2.php#L102) uses setAttributes, so it fails to insert unsafe attributes into the database.

nkovacs:
fixes #1775


samdark:
Thanks!

PR #1782, opened 2015-03-16
Add `packages` option for application packages.

jonathantorres:
Issue: https://github.com/Codeception/Codeception/issues/1783

### DIFF
--- a/src/Codeception/Module/Laravel5.php
+++ b/src/Codeception/Module/Laravel5.php
@@ -30,6 +30,7 @@ use Illuminate\Http\Request;
  * * environment_file: `string`, default `.env` - The .env file to load for the tests.
  * * bootstrap: `string`, default `bootstrap/app.php` - Relative path to app.php config file.
  * * root: `string`, default `` - Root path of our application.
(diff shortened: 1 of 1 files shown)


janhenkgerritsen:
I do not develop packages for Laravel 5, so could you please explain to me what the issue is that this change solves?

jonathantorres:
Exactly as https://github.com/Codeception/Codeception/issues/1208  (you can still use the workbench on Laravel5 as a separate package)

When you run your main app tests you'll also want to run the tests on the packages. But since each package must have it's own `codeception.yml` file when the package tests are run, codeception tries to run the tests from it's own `codeception.yml` file and not from the main laravel app.

So I added another option to the module in which you can specify the directory in which you'll store your packages. (Since in laravel5 it can be any name, not necessary "workbench")

DavertMik:
@janhenkgerritsen looks good to me. Should I merge?

janhenkgerritsen:
Also looks good to me, so you can merge this.

DavertMik:
done. thank you!

PR #1786, opened 2015-03-18
add func assertSame

hidechae:
add assert function `assertSame`


samdark:
Looks good to me. @DavertMik what do you think?

DavertMik:
yes, looks ok )

PR #1790, opened 2015-03-18
fix namespace ignoring when console is invoked

lukyer:
./codecept console accept did not work when namespace attribute occured in codeception.yml file. Helper classes could not be in namespaces.

### DIFF
--- a/src/Codeception/Command/Console.php
+++ b/src/Codeception/Command/Console.php
@@ -81,6 +81,7 @@ class Console extends Command
         $this->suite    = $suiteManager->getSuite();
 
         $scenario = new Scenario($this->test);
+        if (isset($config["namespace"])) $settings['class_name'] = $config["namespace"].'\\'.$settings['class_name'];
         $actor      = $settings['class_name'];
         $I        = new $actor($scenario);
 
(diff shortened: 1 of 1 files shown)


src/Codeception/Command/Console.php
```diff
@@ -81,6 +81,7 @@ public function execute(InputInterface $input, OutputInterface $output)
         $this->suite    = $suiteManager->getSuite();
 
         $scenario = new Scenario($this->test);
+        if (isset($config["namespace"])) $settings['class_name'] = $config["namespace"].'\\'.$settings['class_name'];
```
DavertMik: PSR-2 please

src/Codeception/Command/Console.php
```diff
@@ -81,6 +81,7 @@ public function execute(InputInterface $input, OutputInterface $output)
...
         $actor      = $settings['class_name'];
```
DavertMik: Maybe the `$actor` vat can be changed instead of `$settings['class_name']`?

lukyer: Don't think so, class name includes a namespace. Property is unusable without the namespace.

DavertMik:
thanks

PR #1792, opened 2015-03-19
Allow bool values for checkboxes in submitForm

zbateson:
submitForm now accepts boolean values for checkbox fields, mapping them to the actual checkbox field value if found.

Also minor cleanup of conditionals in WebDriver's submitForm.


DavertMik:
That's great, thank you!

PR #1795, opened 2015-03-20
Added seeInFormFields for Web tests

zbateson:
Allows quickly checking multiple form field values with an array with InnerBrowser and WebDriver.

Also updated seeInField functionality to allow comparing an empty value to an unselected combo box.

Created simple tests and updated documentation in Interfaces/Web.php (IIRC that's the only place needed? Let me know if otherwise.)

Example:

```php
$I->seeInFormFields('#myform', [
    'checkbox[]' => [
        'see test one',
        'see test two',
    ],
    'radio1' => 'see test one',
    'checkbox1' => true,
    'checkbox2' => false,
    'select1' => 'see test one',
    'select2' => [
        'see test one',
        'see test two',
        'see test three'
    ]
]);
```
Can be paired with submitForm for easy testing:

```php
$form = [
     'field1' => 'value',
     'field2' => 'another value',
     'checkbox1' => true,
     // ...
];
$I->submitForm('#my-form', $form, 'submitButton');
$I->seeInFormFields('#my-form', $form);
```

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -270,23 +270,64 @@ class InnerBrowser extends Module implements Web
 
     public function seeInField($field, $value)
     {
-        $this->assert($this->proceedSeeInField($field, $value));
+        $crawler = $this->getFieldsByLabelOrCss($field);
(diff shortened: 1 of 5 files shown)


DavertMik:
This looks awesome, thank you!

PR #1796, opened 2015-03-22
Update ZF2.php

Gorp:
I added fix the setting header with key "Content-type" 
That header is ignored during request to Zend if you set its  via
```
$I->haveHttpHeader('Content-Type', 'application/json');
```
It makes problem in Zend\Mvc\Controller\AbstractRestfulController because it checks content type and without right header that class doesn't decode content


DavertMik:
Thanks!

Issue #1785, opened 2015-03-17
mocking Laravel5 models is problematic due to __mocked being set in Stub::make


constb:
hello. I'm trying to mock models in laravel5 application. I replace ``getAttribute`` and ``setAttribute`` to run assertions against which attributes tested class touches. Unfortunately, codeception stubs also always require me to make an extra exception for ``__mocked`` attribute. Because it is being set after methods are already replaced and there's no ``__mocked`` in model object.

this is minimal example of a helper I use in my tests:

```php
private function fakeUserStub(&$fakeId, &$fakeLastSeen)
{
    $mocked = null; // NEED THIS TO PASS TESTS

    return Stub::make(
        'App\User',
        [
            'getAttribute' => function ($key) use ($fakeId, $fakeLastSeen, $mocked) {

                if ($key === '__mocked') { // NEED THIS TO PASS TESTS
                    return $mocked;        // NEED THIS TO PASS TESTS
                }                          // NEED THIS TO PASS TESTS

                $this->tester->assertContains($key, [ 'id', 'last_seen' ]);

                return ${'fake' . studly_case($key)};
            },
            'setAttribute' => function ($key, $value) use (&$fakeId, &$fakeLastSeen, &$mocked) {

                if ($key === '__mocked') { // NEED THIS TO PASS TESTS
                    $mocked = $value;      // NEED THIS TO PASS TESTS
                    return;                // NEED THIS TO PASS TESTS
                }                          // NEED THIS TO PASS TESTS

                $this->tester->assertContains($key, [ 'id', 'last_seen' ]);

                ${'fake' . studly_case($key)} = $value;
            },
        ]
    );
}
```

DavertMik:
I think `__mocked` can be removed. Actually not sure, but probably it is not used anymore

EVODelavega:
@DavertMik I've encountered some issues with this `__mocked` property, too, when trying to test code that depended on the Couchbase extension (brett19's repo, CouchbaseBucket was the class that I needed). This class implemented the `__set` method, and triggered errors whenever the user tried to set an invalid property. Looking at the Laravel source code, they seem to implement the `__set` method, too, which might explain this bug. So for what it's worth:

[...]

PR #1800, opened 2015-03-25
suggested fix for issue #1785: overloading and __set bug

EVODelavega:
The `Stub::*` methods all tend to set a `__mocked` property on the mocked classes. When attempting to mock a class that implements a `__set` method that doesn't allow (as PHP calls it) _"overloading"_, this results in a PHPUnit exception.

This PR addresses this issue by checking the class for an existing `__set` method. If such a method exists, the `__mocked` flag is not set.
This check is also used when attempting to add a non-existent property to classes with a `__set` method. These properties will be added, but a `PHPUnit_Framework_Exception` will be thrown if this fails.

The `__mocked` flag was primarily (only?) used in the `Stub::update` call (to change the mock's behavior on the fly). the `__mocked` property check has been replaced with an `instanceof PHPUnit_Framework_MockObject_MockObject` check. If this check fails, the behavior remains unchanged: calling `Stub::update` on anything that is not an instance of the `MockObject` class cause a `LogicException` to be thrown.

### DIFF
--- a/src/Codeception/Util/Stub.php
+++ b/src/Codeception/Util/Stub.php
@@ -60,12 +60,22 @@ class Stub
         }
 
         self::bindParameters($mock, $params);
-        //only set __mocked if __set is not implemented
-        //to avoid __set implementations that error on overloading
-        if (!$reflection->hasMethod('__set')) {
-            $mock->__mocked = $class;
-        }
 
+        return self::markAsMock($mock, $reflection);
+    }
+
+    /**
+     * Set __mock flag, if at all possible
+     *
+     * @param object $mock
+     * @param \ReflectionClass $reflection
+     * @return object
+     */
+    private static function markAsMock($mock, \ReflectionClass $reflection)
+    {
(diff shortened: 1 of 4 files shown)


DavertMik:
Thanks. This looks like a really serious work done to fix this issue. Probably we can remove `__mocked` at all. Anyway, I'm really glad to get this merged

DavertMik:
Actually the second reason to implement `__mocked` was to show mocked class name in console output. Probably this can be improved as well
https://github.com/Codeception/Codeception/blob/2.0/src%2FCodeception%2FStep.php#L69

EVODelavega:
Happy to see my PR merged, If I find the time, I'll set about removing the `__mocked` property entirely, and look into an alternative approach to display the mocked class name (which probably won't be that hard of a thing to do, there's probably some method on the `PHPUnit_Framework_MockObject_MockObject` instance I could use, or keep a reference to the `ReflectionClass` instance somewhere. If I get 'round too it, I'll make a PR.

PR #1802, opened 2015-03-26
WebDriver submitForm allows array parameter values

### LINKED ISSUE #1797: submitForm doesn't work with checkbox arrays
Say you have the following checkbox inputs on a page:
```html
<input name="array[]" value="1" type="checkbox">
<input name="array[]" value="2" type="checkbox">
<input name="array[]" value="16" type="checkbox">
```

Checking them manually works just fine:
```php
$I->checkOption('#form input[type=checkbox][value="1"]');
$I->checkOption('#form input[type=checkbox][value="2"]');
$I->checkOption('#form input[type=checkbox][value="16"]');
```

However, the equivalent using submitForm does absolutely nothing when using an array, yet it does work when using just one simple value:
```php
$I->submitForm('#form', array(
            'array[]'    => "[1,2,16]" // array value is NULL, server side
        ));
```
```php
$I->submitForm('#form', array(
            'array[]'    => "2" // adds value 2 as if the second checkbox had been checked
        ));
```

Of course, I can't add another array key with the same name and another value...
What would be the solution here? Could this be a bug?

zbateson:
- Fixes #1797
- Updated function to allow submitting array parameter values like
  for PhpBrowser
- Updated PhpBrowser so array of booleans could be sent - separated
  into separate function
- Cleaned up WebDriver's submitForm a bit
- Added some documentation
- Added a few more tests

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -347,12 +347,63 @@ class InnerBrowser extends Module implements Web
         ];
     }
 
+    /**
+     * Strips out one pair of trailing square brackets from a field's
+     * name.
+     *
+     * @param string $name the field name
+     * @return string the name after stripping trailing square brackets
+     */
     protected function getSubmissionFormFieldName($name)
     {
         if (substr($name, -2) === '[]') {
             return substr($name, 0, -2);
         }
         return $name;
+    }
+
+    /**
+     * Replaces boolean values in $params with the corresponding field's
+     * value for checkbox form fields.
+     *
+     * The function loops over all input checkbox fields, checking if a
+     * corresponding key is set in $params.  If it is, and the value is
+     * boolean or an array containing booleans, the value(s) are
+     * replaced in the array with the real value of the checkbox, and
+     * the array is returned.
+     *
+     * @param Crawler $form the form to find checkbox elements
+     * @param array $params the parameters to be submitted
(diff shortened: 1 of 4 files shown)


DavertMik:
@zbateson sounds good. 
For some time I have an idea to propose you to join Codeception team. It looks that you know Wd + InnerBrowser in some ways better than I do ) What would you say? 

We can discuss details in Skype: davert.ua

wasmithee:
The fix for this bug breaks the submitForm function for me. I have a select with the name attribute of: ``` 'Departments[dept_name][]'```. Stripping off the last ```'[]'``` causes the option to not be selected.

How do you suggest we make it so it will work for both cases?

DavertMik:
@wasmithee could you provide example HTML so we could include it into our test suite?

wasmithee:
````
<div class="row">
	<label for="Departments_dept_name" class="required">Department Name <span class="required">*</span>
	</label>
	<select size="5" multiple="multiple" name="Departments[dept_name][]" id="Departments_dept_name">
		<option value="65">BTT</option>
		<option value="68">BTT Demo2</option>
		<option value="70">Demo Companies</option>
		<option value="71">Demo Company</option>
	</select>
</div>
````

zbateson:
Hi @wasmithee 

I've created an issue to reference this - #1976 - it helps me a lot to have it in a new issue instead of a closed pull request :smiley:

PR #1805, opened 2015-03-26
support functional testing of Laravel Artisan commands

bgetsug:
I would like to add support for functional testing of Laravel Artisan commands to the Laravel4 module—at least I think this is a bit more friendly than just using `$I->grabService('artisan')`, etc.


DavertMik:
Thanks!

PR #1806, opened 2015-03-27
Pass body request into yii2 request

m8rge:
This allows to send Xml payload:
```php
$i->haveHttpHeader('Content-Type', 'application/xml');
$i->sendPOST('/v1/door/new-goods', $xmlString);
```

### DIFF
--- a/src/Codeception/Lib/Connector/Yii2.php
+++ b/src/Codeception/Lib/Connector/Yii2.php
@@ -83,8 +83,10 @@ class Yii2 extends Client
 
         ob_start();
 
+        $yiiRequest = $app->getRequest();
+        $yiiRequest->setRawBody($request->getContent());
         try {
-            $app->handleRequest($app->getRequest())->send();
+            $app->handleRequest($yiiRequest)->send();
         } catch (\Exception $e) {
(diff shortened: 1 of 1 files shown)


samdark:
Have you tried it with Yii code from master?

m8rge:
@samdark I tested this code with 2.0.3

samdark:
How to reproduce the issue?

m8rge:
This pull request is not actually bugfix, but feature.

Without this pull, you can't post xml payload like this:
```php
public function addNewItem(FunctionalTester $i) {
    $xml = <<<EOT
<root>
    <goods>
        <productId>1</productId>
        <placeId>1</placeId>
        <boxId>2</boxId>
    </goods>
</root>
EOT;
    $i->sendPOST('/v1/door/new-goods', $xmlString);
}
```

samdark:
What will happen with passing arrays?

m8rge:
It will be assigned to $_POST variable. Then, array will be returned as is in https://github.com/yiisoft/yii2/blob/master/framework/web/Request.php#L370
Please note, that parsers in L382 are not reached.

samdark:
Thanks!

Issue #1354, opened 2014-08-28
PHP Fatal error:  Call to a member function request() on a non-object in Codeception/Module/REST.php on line 381


pantaovay:
PHP Fatal error:  Call to a member function request() on a non-object in phar:///vagrant_data/xueba/codecept.phar/src/Codeception/Module/REST.php on line 381
PHP Stack trace:
PHP   1. {main}() /vagrant_data/xueba/codecept.phar:0
PHP   2. require_once() /vagrant_data/xueba/codecept.phar:7
PHP   3. Symfony\Component\Console\Application->run() phar:///vagrant_data/xueba/codecept.phar/codecept:34
PHP   4. Symfony\Component\Console\Application->doRun() phar:///vagrant_data/xueba/codecept.phar/vendor/symfony/console/Symfony/Component/Console/Application.php:124
PHP   5. Symfony\Component\Console\Application->doRunCommand() phar:///vagrant_data/xueba/codecept.phar/vendor/symfony/console/Symfony/Component/Console/Application.php:193
PHP   6. Symfony\Component\Console\Command\Command->run() phar:///vagrant_data/xueba/codecept.phar/vendor/symfony/console/Symfony/Component/Console/Application.php:887
PHP   7. Codeception\Command\Run->execute() phar:///vagrant_data/xueba/codecept.phar/vendor/symfony/console/Symfony/Component/Console/Command/Command.php:252
PHP   8. Codeception\Codecept->run() phar:///vagrant_data/xueba/codecept.phar/src/Codeception/Command/Run.php:149
PHP   9. Codeception\Codecept->runSuite() phar:///vagrant_data/xueba/codecept.phar/src/Codeception/Codecept.php:145
PHP  10. Codeception\SuiteManager->run() phar:///vagrant_data/xueba/codecept.phar/src/Codeception/Codecept.php:162
PHP  11. Codeception\PHPUnit\Runner->doEnhancedRun() phar:///vagrant_data/xueba/codecept.phar/src/Codeception/SuiteManager.php:153
PHP  12. PHPUnit_Framework_TestSuite->run() phar:///vagrant_data/xueba/codecept.phar/src/Codeception/PHPUnit/Runner.php:100
PHP  13. PHPUnit_Framework_TestCase->run() phar:///vagrant_data/xueba/codecept.phar/vendor/phpunit/phpunit/src/Framework/TestSuite.php:675
[...]

Issue #1111, opened 2014-06-17
HELP!  FATAL ERROR. TESTS NOT FINISHED.


mikeerickson:
I have been battling this issue for weeks and until now, I have just worked around it (ignored testing this part of application) but I have to face it head on now and dont know what to do.

I am simply trying to perform a simple test POST using API suite.  

If there are any suggestions, it would be greatly appreciated!

Error

FATAL ERROR. TESTS NOT FINISHED.
Call to a member function request() on a non-object 

Test method causing error

	public function it_should_create_new_resource(ApiTester $I)
	{
		$I->sendPOST($this->baseUrl().'companies', ['name', 'CodeDungeon'], null);
		$I->canSeeResponseIsJson();
	}

api.suite.yml

class_name: ApiTester
modules:
  enabled: [REST, ApiHelper, PHPBrowser]
  config:
    PHPBrowser:
      url: http://rangladex.app:8000/
  REST:
    url: http://rangladex.app:8000/api/v1/

zbateson:
Should be fixed now with #1834

Issue #111, opened 2013-01-16
TestCase class getTrace() method error (call to undefined method)


Ragazzo:
Error occurs when unit-test fails, for example simple check `$this->assertTrue(false);`. So error on this line https://github.com/Codeception/Codeception/blob/master/src/Codeception/Subscriber/Console.php#L147. Did u removed this method or it was in phpunit < 3.7? I use phpunit 3.7.

PR #1807, opened 2015-03-27
don't fire TEST_AFTER on conditional failure, fixes #1647 #1354 and #111...

### LINKED ISSUE #1647: REST module throws execeptions after using "canSeeResponseContainsJson"
I created a simple test to call a URL and then test if I can:
```php
$I->canSeeResponseContainsJSON(array('status'=>'OK'));
```
When I test for that, and the response contains the array('status'=>'OK') everything is OK

BUT, if I change the code to be:
```php
$I->canSeeResponseContainsJSON(array('status'=>'OK2'));
//OR
$I->canSeeResponseContainsJSON(array('status1'=>'OK'));
```
REST.php throws an exception the next time I try to issue a $I->sendGET
```bash
PHP Fatal error:  Call to a member function request() on a non-object in vendor/codeception/codeception/src/Codeception/Module/REST.php on line 380
```
Here is my test
```
    $params['url'] = 'http://localhost.com/2010/06/01/travel/escapes/dan_is_testing.html';
    $I->sendPOST("http://$host/svc/bitly/shorten", $params);
    $I->canSeeResponseIsJson();
    $I->canSeeResponseContainsJSON(array('status'=>'OK2'));
    unset($params);
    $params['url'] = 'http://localhost.com/2010/06/01/travel/escapes/dan_is_testing.html';
    $params['start_date'] = '2010-06-02';
    $params['end_date'] = '2010-06-02';
    $I->sendGET("http://$host/svc/bitly/urlstats", $params); // **THOWS ERROR HERE**
    $I->seeResponseIsJson();
    $I->seeResponseContainsJSON(array('status'=>'OK'));
```
I'm using codeception 2.0.9
> php -v
PHP 5.5.13 (cli) (built: Sep  8 2014 14:41:26) 
Copyright (c) 1997-2014 The PHP Group
Zend Engine v2.5.0, Copyright (c) 1998-2014 Zend Technologies
    with Xdebug v2.2.5, Copyright (c) 2002-2014, by Derick Rethans
> composer -V
Composer version 1.0-dev (c58b7d917c65692eeb00c22b2fbaaa251cb390dc) 2015-01-05 16:31:16

schmunk42:
I had the same problems like described in #1647 #1354 and #1111 

Please review this PR, as I have no deep understanding of Codeception's internals, but from the explanations in the issues, the fix would look logical to me.

### DIFF
--- a/src/Codeception/PHPUnit/Listener.php
+++ b/src/Codeception/PHPUnit/Listener.php
@@ -5,6 +5,7 @@ use Codeception\Events;
 use Codeception\Event\FailEvent;
 use Codeception\Event\SuiteEvent;
 use Codeception\Event\TestEvent;
+use Codeception\Exception\ConditionalAssertionFailed;
 use Codeception\TestCase;
 use Exception;
 use PHPUnit_Framework_Test;
@@ -42,7 +43,11 @@ class Listener implements \PHPUnit_Framework_TestListener
     {
         $this->unsuccessfulTests[] = spl_object_hash($test);
         $this->fire(Events::TEST_FAIL, new FailEvent($test, $e));
-        $this->fire(Events::TEST_AFTER, new TestEvent($test, $time));
+        if ($e instanceof ConditionalAssertionFailed) {
+            // do not fail test on ConditionalAssertionFailed
+        } else {
+            $this->fire(Events::TEST_AFTER, new TestEvent($test, $time));
+        }
     }
 
     public function addError(\PHPUnit_Framework_Test $test, \Exception $e, $time)


zbateson:
Hi @schmunk42 - can/cant methods are generated, so you won't see tests in unit tests for the modules (because the can/cant methods don't actually exist on them).  I had written some tests for this exact problem but I think the idea is to put this fix in a coming release.  

You can see the tests I had created here if interested: https://github.com/zbateson/Codeception/commit/df2960da6fc515aa6378b701e341405405ee1658 which is pretty much the same as your PR but with some testing, and will most likely be a part of 2.1, not 2.0.

zbateson:
Unless @DavertMik would be happy to merge this, and have the test code only be a part of 2.1 which I'm all for, as it would allow us to use can/cant before 2.1 :)

schmunk42:
@zbateson Thanks for your info :) Is your code already in 2.1?

zbateson:
It isn't yet, I hadn't had a chance to submit a pull request to master - I'm hoping to have some time in the next 4/5 days to do so.  That commit was written against 2.0 so I'll have to check if the test code would need any modification before submitting.

zbateson:
It was decided to merge my pull request against 2.0 after all.  It's the same as yours just with the added tests :)

Thank you

PR #1809, opened 2015-03-27
moved output data to debug section, fixes #1808

### LINKED ISSUE #1808: Feature Request: Show JSON response data on -vv
I tried to patch  the code to be able to see less output from failed JSON responses.
But I could not access the OutputStream from there:
https://github.com/Codeception/Codeception/blob/2.0/src/Codeception/Module/REST.php#L510

Would be nice if the full output is only shown with -vv or -vvv.

schmunk42:


DavertMik:
I think your fix is too specific. I think JSON should be displayed on failure so we can compare both actual and expected inside failure message. `debugSection` will show those jsons while test runs and if there will be lots of lines there text won't be availble in the end. 

Probably you can shorten somehow the output of var_export if you feel there are too much lines there (how many do you have?). Probably we can limit them to 20 lines, for example

schmunk42:
Limiting the output is in general desirable, in my current example I have like 250kB of JSON data downloaded.

PR #1811, opened 2015-03-27
Fix for submitForm default value for a textarea

zbateson:
Textarea node doesn't have 'value' attribute, and an issue with conditions meant it wasn't working correctly.


DavertMik:
:+1:

PR #1814, opened 2015-03-28
Extend SOAP module sendRequest handling

### LINKED ISSUE #1784: Module/SOAP with Symfony2 Framework
I configured Codeception with 
```
class_name: ApiTester                                                                                                                                             
modules:
    enabled: [Symfony2, SOAP, ApiHelper]
```

In my opinion the code in Module/SOAP.php::processInternalRequest does not make sense for the use of the Symfony2 Framework.

Why should the symfony2 output be catched via output buffer? The internal response from symfony is a Symfony\Component\HttpFoundation\Response class.

A return for $this->client->getInternalResponse()->getContent() should be the better choice.

Or do I miss something?

Thanks.
Noles

Noles:
Fixes #1784 - With this changes the output buffer use can be disabled by user.

### DIFF
--- a/docs/modules/SOAP.md
+++ b/docs/modules/SOAP.md
@@ -25,6 +25,9 @@ If you use PHP SoapServer with framework, try to block call to this method in te
 ## Configuration
 
 * endpoint *required* - soap wsdl endpoint
+* schema - namespace (ns) schema (default: *empty*)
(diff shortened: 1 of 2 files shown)


DavertMik:
Thanks. Looks useful

PR #1817, opened 2015-03-29
Update AMQP.php

jistok:
add a routing key to a push to exchange to facilitate proper routing of topics


src/Codeception/Module/AMQP.php
```diff
@@ -114,12 +114,12 @@ public function _before(\Codeception\TestCase $test)
      * @param $exchange
```
DavertMik: could you update docs here as well?
and provide an example of usage. Thanks

DavertMik:
Thanks!

PR #1819, opened 2015-03-30
[dev-master] Fix Dependent Module Injection

### LINKED ISSUE #1818: [dev-master] ModuleContainer::injectDependentModule Injects the Wrong Module
After the latest composer update, running tests that require REST module fail with the following message:
```
 Argument 1 passed to Codeception\Module\REST::_inject() must be an instance of Codeception\Lib\InnerBrowser, instance of Codeception\Module\REST given, called in /vagrant/vendor/codeception/codeception/src/Codeception/Lib/ModuleContainer.php on line 166 and defined in /vagrant/vendor/codeception/codeception/src/Codeception/Module/REST.php on line 111
```
After some digging I noticed that $dependentModule is created from the Module itself and not the dependent module looking code in \Codeception\Lib\ModuleContainer.php  on line 162:
```php
public function injectDependentModule($name, DependsOnModule $module)
    {
        $message = '';
        $dependency = $module->_depends();
        if (is_array($dependency)) {
            $message = reset($dependency);
            $dependency = key($dependency);
        }
        if (!isset($this->config['modules']['depends'][$name])) {
            throw new ModuleRequire($module,
                "\nThis module depends on module of $dependency\n" .
                "Please specify the dependent module inside module configuration section.\n" .
                "\n\n$message");
        }
        $dependentModule = $this->create($name, false);
        if (!method_exists($module, '_inject')) {
            throw new ModuleException($module, 'Module requires method _inject to be defined to accept dependencies');
        }
        $module->_inject($dependentModule);
    }
```

Changing line 162 to the following code seems to fix this:
```php
 $dependentModule = $this->create($this->config['modules']['depends'][$name], false);
```

ShayaTepper:
Fixes #1818. With this update, we are changing the module being passed to the _inject method from the module itself ($name) to the correct 'depends' module as it is configured in the associated .yml file.

### DIFF
--- a/src/Codeception/Lib/ModuleContainer.php
+++ b/src/Codeception/Lib/ModuleContainer.php
@@ -159,7 +159,7 @@ class ModuleContainer
                 "Please specify the dependent module inside module configuration section.\n" .
                 "\n\n$message");
         }
-        $dependentModule = $this->create($name, false);
+        $dependentModule = $this->create($this->config['modules']['depends'][$name], false);
         if (!method_exists($module, '_inject')) {
             throw new ModuleException($module, 'Module requires method _inject to be defined to accept dependencies');
         }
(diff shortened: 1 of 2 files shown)


DavertMik:
already fixed )

Issue #1791, opened 2015-03-18
HTTP header merge/delete


coviex:
Hi,
Inside cept I use phpbrowser module to see() if I'm not logged in, then rest module to login with xhr (I set xhr header explicitly), then again phpbrowser module to see() if I'm logged in.
Problem is in Codeception\Lib\Connector\Guzzle::doRequest() which merges my first xhr header into all subsequent requests, creating header duplicates and resulting in that last check failure. It was OK several months ago, I guess it wasn't merging back then.

It would be great if whether the connector could stop merging recursively or method to delete a header by name was added.

Thanks

```yaml
class_name: AcceptanceHeadlessTester
modules:
    enabled:
        - PhpBrowser
        - REST
        - MyHelper
        - AcceptanceHeadlessHelper
    config:
        PhpBrowser:
            url: "http://my.dev/"
        REST:
            url: "http://my.dev/"
```

```php
namespace Codeception\Module;
class AcceptanceHeadlessHelper extends \Codeception\Module
{
    public function login($username, $password)
    {
        $browser = $this->getModule('REST');
        $browser->haveHttpHeader('X-Requested-With', 'XMLHttpRequest');
        $browser->sendPOST(
            'forum/login/',
            [
                'login[login]'    => $username,
                'login[password]' => $password
            ]
        );
        $browser->seeResponseCodeIs(200);
        $browser->seeResponseIsJson();
        $browser->seeResponseContainsJson(['status' => true]);
        $browser->seeResponseContains('Welcome');

        $browser = $this->getModule('PhpBrowser');
        // this one does not override previously set
        $browser->setHeader('X-Requested-With', '');
    }
}
```

zbateson:
Hi @coviex 

I can confirm that setting the header to a blank value doesn't work - perhaps the solution would be to create a 'deleteHeader' instead though.

coviex:
not only changes by @zbateson do not resolve the problem (because they are in the wrong file) but also show lack of knowledge of PHP basics. (using  "empty()" on arbitrary value)

DavertMik:
@coviex please do better =)

coviex:
@DavertMik sorry for inline

\Codeception\Lib\Connector\Guzzle
```php
    protected function doRequest($request)
    {
        $requestOptions = array(
            'body' => $this->extractBody($request),
            'cookies' => $this->extractCookies($request),
            'headers' => $this->extractHeaders($request)
        );

        // {{{ custom part to delete headers
[...]

PR #1824, opened 2015-04-01
Added deleteHeader and docs for PhpBrowser

zbateson:
- Created deleteHeader
- Allow deleting a header with an empty value when setting (since
  setting to an empty value doesn't make sense anyway)
- Added documentation for setHeader and deleteHeader
- Added simple test

Provides a solution for #1791


DavertMik:
:+1:

Issue #1826, opened 2015-04-01
Silently ignore DROP command denied to user during Codeception\Lib\Driver\MySql::cleanup()


Antnee:
To work around the fact that we only want Codeception to cleanup specific tables, rather than the entire MySQL database, I've created a specific user with privileges that allows access only to the required table. This works perfectly and the table is created and dropped between tests as expected and all other tables remain. I've then added an additional table that I don't want to be dropped and not granted the drop privilege. My tests now fail stating
```
(Exception in Db) SQLSTATE[42000]: Syntax error or access violation: 1142 DROP command denied to user 'test-user'@'hostname' for table 'test_readonly_table'
```
Would it be possible to catch this exception and continue with only a warning? Or add a flag that optionally enables this? I don't want to stop it dropping between tests, only for it to not completely fail if it's unable to delete a table that it can see. Ideally it would look at the current user's own privileges and not even try if it's not allowed to, but why manage it in Codeception when it's already handled in MySQL itself?

Antnee:
I've added this in for MySQL myself and submitted PR #1827 

Appreciate comments

Antnee:
I'm going to close this as we have two *competing* PRs on #1827 and #1831 and discussions therein

PR #1827, opened 2015-04-01
Do not error if unable to drop table during cleanup()

Antnee:
By adding `ignoredrop: true` to the MySQL DB module config, I can use MySQL permissions to prevent tables that I don't want to be dropped from actually being dropped. Previously this would've thrown an exception that would've failed my build. Now I can create a user for the test that has a drop privilege on the tables in my dump.sql, but not on the other tables.

I've not added to the other drivers since I don't have working installs for them, but presumably it would work in a similar fashion.

See #1826

### DIFF
--- a/src/Codeception/Lib/Driver/MySql.php
+++ b/src/Codeception/Lib/Driver/MySql.php
@@ -12,7 +12,7 @@ class MySql extends Db
                 $this->dbh->exec('drop table `' . $row[0] . '`');
             } catch (\PDOException $e) {
                 if ($ignoreDrop) {
-                    trigger_error('Unable to drop table '.$row[0].' during cleanup', E_WARNING);
+                    trigger_error('Unable to drop table '.$row[0].' during cleanup', E_USER_WARNING);
                 } else {
                     throw $e;
                 }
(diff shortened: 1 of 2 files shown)


DavertMik:
From my point of view this commit is too specific. There are few things prevents me from merging pull request. Cleanup method signature should be the same for all DB drivers (yes, its hard, however they all work through PDO), The option can be renamed to something more descriptive and general, like `ignore_errors`.

Maybe we could add `skip_tables` option to achieve the same?
That can be much cleaner solution

Antnee:
Fair points. How do you see `skip_tables` working? Are you suggesting a list of tables that `cleanup()` ignores? That solution seems more backwards to me. I just want to only drop the tables that I'm creating in my dump, so I'd prefer a `clean_tables` list instead.

`ignore_errors` would also suggest that all errors in that module would be ignored, which has the potential to create unexpected results. An `ignore_clean_errors` at least tells me what it's ignoring.

Personally, I like the idea of a list of `clean_tables`. If no list is provided then clean all the things. If all drivers are using PDO then a single method could `exec()` and do the whole try/catch and be called from each driver easily enough. Happy to implement that instead tomorrow if you like the idea?

DavertMik:
@Antnee I agree to your points. Will be waiting for your PR

PR #1831, opened 2015-04-02
Only cleanup specified tables - MySQL

Antnee:
Refs #1827 @DavertMik 

Config can now contain a `clean_tables` sequence of tables that will be cleaned during `cleanup()`. If this list is missing/empty then all tables will be cleaned. If it is not missing/empty then only tables in this sequence will be cleaned up. Other tables will be completely ignored unless also included in a provided dump. **This first implementation is ONLY for MySQL**

Example:
```yaml
modules:
    config:
        Db:
            dsn: 'mysql:host=myhostname;dbname=dbname'
            user: 'username'
            password: 'Pa5Sw0Rd!'
            dump: tests/_data/dump.sql
            populate: true
            cleanup: true
            clean_tables:
                - clean_me_one
                - clean_me_two
```

I've only implemented this in MySQL at the moment because the `cleanup()` methods for other DB drivers are quite different:

* The MySQL driver will get a list of tables that are available and iterate over them and check that either `clean_tables` is missing or that the current table is in the `clean_tables` array. This is checked by `Db::shouldCleanTable()`, which returns `true` if the table should be dropped, and `false` if not.
* The MsSQL and SqlSrv drivers create a cursor that moves through all tables dropping as it goes. This would need that query updating to say something like `if @tablename NOT IN(table1, table2 ...) EXEC('DROP TABLE ' + @tablename) ... ` but I don't know enough about this engine to be able to write suitable SQL.
* The PostgreSQL driver runs three queries, each of which returns a complete DROP statement that can be executed later. That means that `Db::shouldCleanTable()` could not check a simple table name and would instead need to compare against the entire query, else we'll need to regex the query to get the table out and pass that through to `Db::shouldCleanTable()`. That's relatively straightforward, but a bit ugly.
* The Oracle driver also runs `cleanup()` similar to MsSQL where it loops through all triggers, tables and sequences and and drops immediately, never returning anything to PHP for us to process. This will mean that `Db::shouldCleanTable()` can't work.

So I'm submitting this actually as a work-in-progress for further feedback. I'm undecided about what the best solution is. I like this idea because we can easily configure in Codeception which tables we actually want to clean and disregard the rest, but it's difficult to implement in the other engines. I also like the previous idea (on #1827) because I can set the permissions on MySQL and just catch a `\PDOException` in PHP and continue as I was. If you like this solution then we'd need to find people who know a lot more about the various RDBMS engines in order to implement properly in each, or make it a MySQL only implementation for now, which I know you don't like.

Before I go any further, please let me know your thoughts.

Thanks

### DIFF
--- a/src/Codeception/Lib/Driver/Db.php
+++ b/src/Codeception/Lib/Driver/Db.php
@@ -102,6 +102,14 @@ class Db
     {
     }
 
+    protected function shouldCleanTable($table, array $cleanOnly=[])
+    {
+        if (empty($cleanOnly) || in_array($table, $cleanOnly)) {
+            return true;
+        }
+        return false;
+    }
+
     public function load($sql)
     {
         $query           = '';
--- a/src/Codeception/Lib/Driver/MySql.php
+++ b/src/Codeception/Lib/Driver/MySql.php
@@ -3,12 +3,14 @@ namespace Codeception\Lib\Driver;
 
 class MySql extends Db
 {
-    public function cleanup()
+    public function cleanup(array $cleanTables=[])
     {
         $this->dbh->exec('SET FOREIGN_KEY_CHECKS=0;');
         $res = $this->dbh->query("SHOW FULL TABLES WHERE TABLE_TYPE LIKE '%TABLE';")->fetchAll();
-        foreach ($res as $row) {
-            $this->dbh->exec('drop table `' . $row[0] . '`');
+        foreach ($res AS $row) {
+            if ($this->shouldCleanTable($row['0'], $cleanTables)) {
+                $this->dbh->exec('drop table `' . $row['0'] . '`');
+            }
         }
         $this->dbh->exec('SET FOREIGN_KEY_CHECKS=1;');
     }
--- a/src/Codeception/Module/Db.php
+++ b/src/Codeception/Module/Db.php
@@ -92,9 +92,10 @@ class Db extends \Codeception\Module implements \Codeception\Lib\Interfaces\Db
      * @var array
      */
     protected $config = [
-      'populate' => true,
-      'cleanup'  => true,
-      'dump'     => null
+      'populate'     => true,
+      'cleanup'      => true,
+      'dump'         => null,
+      'clean_tables' => [],
     ];
 
     /**
@@ -193,7 +194,7 @@ class Db extends \Codeception\Module implements \Codeception\Lib\Interfaces\Db
             if (!count($this->sql)) {
                 return;
             }
-            $this->driver->cleanup();
+            $this->driver->cleanup($this->config['clean_tables']);
         } catch (\Exception $e) {
             throw new ModuleException(__CLASS__, $e->getMessage());
         }


DavertMik:
Thanks for PR and detailed investigation. Let's hear what other think of this

Antnee:
Worth considering is the [DB module page docs](http://codeception.com/docs/modules/Db) which states the following:

* MySQL: stable
* SQLite: stable
* Postgres: beta
* MSSQL: alpha
* Oracle: alpha

Accordingly, we should prioritise MySQL and SQLite, but the rest could follow later

Antnee:
Apparently nobody has anything to say :) Any further thoughts, @DavertMik?

tiger-seo:
IMO, instead of new property, this should be additional case for `cleanup`, let me show what i mean  in example:
- to cleanup everything
```yaml
modules:
    config:
        Db:
            dsn: 'mysql:host=myhostname;dbname=dbname'
            user: 'username'
            password: 'Pa5Sw0Rd!'
            dump: tests/_data/dump.sql
            populate: true
            cleanup: true
```
- to cleanup only two defined tables
```yaml
modules:
    config:
        Db:
            dsn: 'mysql:host=myhostname;dbname=dbname'
            user: 'username'
            password: 'Pa5Sw0Rd!'
            dump: tests/_data/dump.sql
            populate: true
            cleanup:
                include:
                    - clean_me_one
                    - clean_me_two
```
- to cleanup everything, except defined tables
```yaml
modules:
    config:
        Db:
            dsn: 'mysql:host=myhostname;dbname=dbname'
            user: 'username'
            password: 'Pa5Sw0Rd!'
            dump: tests/_data/dump.sql
            populate: true
            cleanup:
                exclude:
                    - clean_me_one
                    - clean_me_two
```
so `cleanup` will be polimorphic; same approach is used by doctrine for mappings

oytuntez:
I agree with @tiger-seo, this is way cleaner and also backward-compatible.

One thing to note is that cleanup mechanism does not clean database procedures. For now I put a database drop on the dump file, however this feature may be worth looking at. The configuration block could be something like this:

```yaml
modules:
    config:
        Db:
            dsn: 'mysql:host=myhostname;dbname=dbname'
            user: 'username'
            password: 'Pa5Sw0Rd!'
            dump: tests/_data/dump.sql
            populate: true
            cleanup:
                exclude:
                    - dont_clean_me_one
                    - dont_clean_me_two
                procedures: true
```

`procedures` switch would be `false` by default to make it backward-compatible.

DavertMik:
I agree with @tiger-seo as well, I really like proposed configuration format.
Also I agree with proposals of @oytuntez 

But we have one question opened: who is going to implement that? ;)

Antnee:
I don't mind implementing for MySQL, but I'm not well placed to do the others.

@oytuntez also raises a point that somewhat relates to my own thoughts @tiger-seo's solution; if we're simply specifying what to include or exclude, we need to support all appropriate types; triggers, views, tables etc.

These also vary between between RDBMS in the current drivers. PostgreSQL, for example, drops tables, sequences and types. Oracle drops triggers, tables and sequences. MySQL just drops tables. My solution was to specifically state the types that you're deleting. I don't know if it's possible in the various engines to have different types with the same names (ie have a trigger called foo and a table called foo), but if it is then you would need to be explicit about which you want to delete. Hence my solution to have a `clean_tables` sequence which is *clearly* cleaning only the specified tables and nothing else. Maybe the solution is to be more like this:
```yaml
modules:
    config:
        Db:
            dsn: 'mysql:host=myhostname;dbname=dbname'
            user: 'username'
            password: 'Pa5Sw0Rd!'
            dump: tests/_data/dump.sql
            populate: true
            cleanup:
                tables:
                    - clean_me_table_one
                    - clean_me_table_two
                procedures:
                    - clean_me_proc_one
```

It could get very messy if we started putting in includes and excludes as well though. I worry that this is going to get out of hand if we're not careful. At the moment a cleanup wipes everything out, which is a heavy-handed solution IMHO. I'm not looking to create something massively complicated though, just to give a little more control without needing to keep dropping and recreating everything.

What do we think?

oytuntez:
I understand @Antnee's point of view; however, procedures and views are not data-containers, they are static elements to execute. So in practice, I believe nobody would care if they are cleaned and recreated as they won't have any data bound to themselves. I'd say we can safely remove all procedures, views, triggers -basically anything other than tables- and recreate them on test initialization. In this case, `include` and `exclude` switches would specify solely tables, and other elements would only have boolean fields.

How does this sound?

Antnee:
Hmmm, I see where you're coming from... I don't think it's terribly clear to the end user, but I agree with your position on the difference between tables and views/procedures etc.

I'm not entirely convinced that the exclude option is useful though and am concerned that it just adds additional complexity. What happens if someone specifies include and exclude? Exception? Assumption? If you've given a list of included tables then by definition you are excluding everything else so presumably you can ignore the excludes, but what if a table is included and excluded?

I haven't used Codeception for long but a bit of Googling tells me that the problem is that people only want a couple of tables to be cleaned but are finding the entire DB being cleaned, save for the tables specified in the dump. Being able to say "please only clean these tables" seems much more useful than "please clean all tables except..." 

Alternatively, the whole issue becomes more simplified and we just say "only run the dump between tests" and we ensure that the dump includes appropriate drops and creates. This requires a change to cleanup though as it will need to cease dropping anything which breaks BC

Antnee:
Another one that I've just thought of... What if the same DB is used by different test suites, and the Codeception tests are deleting all procedures (because `procedure:true`) and this deletes a procedure that is used by the other suite which isn't creating them? Not everyone has the ideal configuration of one DB per test suite. I don't like forcing a process through software. The software should bend to support the process IMHO

oytuntez:
@Antnee, I agree that  "please clean only these tables" sounds more practical. The other way around would probably be used less. But this also brings a paradigm: we get to decide the use case of the feature (is it inclusion or exclusion decision), instead of the actual user of the test suite. Still, not too bad; just not an "all in one" path.

Regarding the same DB usage... Hmm... Do you think Codeception is responsible with the preservation of this? If the DB is used in another external test suite, the user would need to mark `procedure: false` and handle the deletion externally as things seem messed up a little bit in her scenario.

***[20 minutes later]***

Hmm... What about standardizing everything under `cleanup`? Take a look at the example below. 

Basically, every category accepts two different value types: boolean, dictionary. 
- If boolean:
    - If true, remove all of the entities.
    - If false, leave them all.
- If dictionary:
    - If filter type is "include":
        - Remove the entities listed.
    - If filter type is "exclude":
        - Remove all except the entities listed.

Any other case throws an exception. And for the sake of flexibility, we could provide a magic _raw key to run an SQL file during cleanup process. If _raw is provided, we'll ignore everything else. This _raw idea is a long shot, but just my two cents :)

```yaml
            cleanup:
                tables: true
                tables:
                    filter: include
                    items:
                        - clean_me_table_one
                        - clean_me_table_two
                procedures: false
                procedures:
                    filter: exclude
                    items:
                        - dont_clean_me_proc_one
                        - dont_clean_me_proc_two
                views: true
                views:
                    filter: include
                    items:
                        - clean_me_view_one
                        - clean_me_view_two
                triggers: false
                triggers:
                    filter: exclude
                    items:
                        - dont_clean_me_trigger_one
                        - dont_clean_me_trigger_two
                _raw: tests/_data/cleanup.sql
```

Antnee:
Should we make an assumption about the filter value, or require that it's specified? I'm not sure that it's necessarily clear to an end user, but seems to cover most (all?) possibilities... Should it be possible to do a `cleanup: true` as well for BC?

oytuntez:
I think we should avoid even more optional values or assumptions. Even the scheme I suggested is accepting two different value types for one key, which isn't really the best... so I'd advise against `cleanup: true` for further confusion. Besides, currently `cleanup: true` is not removing procedures and stuff...

With the same thought flow, I'd suggest keeping "filter" as required to avoid confusion, just like you said. It's usually better to specify exact values, so that we can protect them in future versions.

What do you think?

Antnee:
Just worried about breaking BC. It may not currently clear procedures, but what if someone does a `composer update`? Otherwise, good...

oytuntez:
Oh, you're right. I completely forgot that. Do you think we should remove everything when `cleanup: true`, or just tables like in the current version?

Antnee:
Unless you want a gazillion issues raising, I think we should maintain current behaviour with `cleanup: true` and add the additional functionality only when the new structure is utilised

As before, I'm happy to write the foundation code and MySQL implementation, but will need someone else to implement for the other engines. I'd like to see if anyone else has any feedback before I spend a couple of hours coding it up though

oytuntez:
We'd never want them :) I agree we should keep the exact functionality that we have at the moment. The roadmap could seem something like this:

- Let's wait until next Monday for other suggestions. The repo is quite active, so this shorter time could be enough. Not sure though, please correct if too short.
- Implement the new structure. 
- Keep the current functionality. 
- Put deprecation notice for the current functionality.
- Update documentation, with an "unrecommended/deprecated" notice for `cleanup: true`.

I'll take SQLite implementation, including SqliteGeneral.

DavertMik:
Thanks! Sorry for delaying my comments on this. 

Please do the MySQL and SQLite implementations! That would be great. The only thing I'd ask you to push your changes into the new branch which I will create. So we could have those 2 RDBMses in one place. 

For Postges and others I propose to throw `NotImplemented` exceptions for now. What do you think?

DavertMik:
Please send your Pull Requests to https://github.com/Codeception/Codeception/tree/db-cleanup

Thanks

Antnee:
That works for me @DavertMik. I'm a little tied up at the moment so unable to implement now but will get on it ASAP

Issue #1828, opened 2015-04-01
Functional testing: possibility to add form fields


lowerends:
Would it make sense to add the possibility to add form fields during functional testing, in order to simulate Javascript behaviour?

At this moment, one can fill form fields one-by-one using the ```fillField()``` function or fill them all in a ```formSubmit()``` function. In a ```formSubmit()``` one can also add extra fields in the parameter array that will be submitted, but it's not possible to combine ```formSubmit()``` with ```fillField()``` to add extra fields to submitted form, as ```formSubmit()``` overwrites all form fields that were filled using ```fillField()```.

Wouldn't it therefore be useful to have a function like e.g. ```addField()``` where you can define a field name and a value and which would also be submitted when clicking the submit button? This could then be used to add extra fields to the form which would be generated by Javascript in a real browser.

zbateson:
If I'm not mistaken you should be able to add any random submission field names/values you like with submitForm.  Have you tried:

```php
$I->submitForm('#my-form', ['some-non-existent-field' => 'some value']);
```

zbateson:
I assume you're using a framework module.  For selenium (WebDriver) tests that would not work.

lowerends:
I'm using the Laravel5 framework module indeed.

I did try with ```submitForm()```, but it's overwriting the values that were filled before in the testcase using ```fillField()```. That means that *all* the form fields need to be included in the parameter array of ```submitForm()```. This is not ideal since I want to make sure that these form fields actually exist on the page, hence ```fillField()``` will make sure that they do exist.

zbateson:
Yeah, at the moment the two functions (fillField and submitForm) work differently and aren't entirely compatible.  I agree it would be better if they were and thought of that myself.

At some point I may get around to it.  I rarely use fillField myself and instead use $I->seeElement if I need to check for existence, and always submit with submitForm.  Just my preference though.

Feel free to submit a pull request if you have the time to play with it :) in this case the changes would be to src/Lib/InnerBrowser.php

lowerends:
Using ```seeElement()``` in combination with ```submitForm()``` would indeed assert that the form fields are present, so that may be a good alternative for now.

I had looked at the ```InnerBrowser.php``` code already. I thought of modifying the ```submitForm()``` function to take ```fillField()``` values into account, but how can I retrieve these values within the ```submitForm()``` function? Is there some method for that?

zbateson:
[...]

Issue #1689, opened 2015-02-04
Functional Tests refuse to fillField.


jarrettj:
I'm not quite sure why my functional tests refuse to fill a field.

My functional.suite.yml:

```
class_name: TestGuy
modules:
    enabled: 
        - TestHelper
        - ZF2
    config:
        ZF2:
          config: 'config/application.config.php'
```

My login test:
```
$I = new TestGuy($scenario);
$I->amOnPage('/');
$I->fillField('identity', $username);
$I->fillField('credential', $password);
$I->submitForm(".form-signin", array());
```

The verbose output is:
```
php vendor/bin/codecept run functional -vvv
Codeception PHP Testing Framework v2.1.0
Powered by PHPUnit 4.4.5 by Sebastian Bergmann.

Functional Tests (2) ---------------------------------------------------------------------------------------------------------------------
Modules: TestHelper, ZF2
------------------------------------------------------------------------------------------------------------------------------------------
1stLoginCept 
Scenario:
* I am on page "/"

  [Response] 200
  [Page] http://localhost/user/login
  [Cookies] []
  [Headers] {"Location":"/user/login"}
* I fill field "identity","admin-user@naspers.com"
* I fill field "credential","password"
* I submit form ".form-signin",
  [Uri] http://localhost/user/login
  [Method] post
  [Parameters] {"identity":"","credential":""}
  [Response] 200
  [Page] http://localhost/user/login
  [Cookies] []
  [Headers] {"Location":"/user/login"}
* I see in current url "/"
 PASSED 

^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Create an admin user. (Application/AdminUser/CreateCept) 
Scenario:
 PASSED 

------------------------------------------------------------------------------------------------------------------------------------------
```

The fields are found, but nothing is posted when the form gets submitted. Has anyone had this issue with ZF2 and codeception before? 

Any help would be appreciated. Thanks.

zbateson:
Hi @jarrettj 

[...]

PR #1832, opened 2015-04-02
Fix for issue 1828

lowerends:
This fixes issue #1828 and also gets rid of the defaults loop in the submitForm() function.

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -414,43 +414,8 @@ class InnerBrowser extends Module implements Web
             throw new ElementNotFound($selector, 'Form');
         }
 
-        $defaults = [];
-        /** @var  \Symfony\Component\DomCrawler\Crawler|\DOMElement[] $fields */
-        $fields = $form->filter('input:enabled,textarea:enabled,select:enabled,button:enabled,input[type=hidden]');
-        foreach ($fields as $field) {
-            $fieldName = $this->getSubmissionFormFieldName($field->getAttribute('name'));
-            if (($field->getAttribute('type') === 'checkbox' || $field->getAttribute('type') === 'radio') && !$field->hasAttribute('checked')) {
-                continue;
-            } elseif ($field->getAttribute('type') === 'button') {
-                continue;
-            } elseif (($field->getAttribute('type') === 'submit' || $field->tagName === 'button') && $field->getAttribute('name') !== $button) {
-                continue;
-            } elseif ($field->tagName === 'select') {
-                $values = [];
-                $select = new Crawler($field);
-                $options = $select->filter('option:enabled:selected');
-                foreach ($options as $option) {
-                    $values[] = $option->getAttribute('value');
-                    if (!$field->hasAttribute('multiple')) {
-                        break;
-                    }
-                }
-                if (count($values) > 1) {
-                    $defaults[$fieldName] = $values;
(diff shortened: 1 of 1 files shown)


zbateson:
Thanks for the PR @lowerends .

There are some Travis errors for FrameworkTest and PhpBrowserTest - could you look into them please?  It may be easier to run the tests for those two modules locally.  Have a look at https://github.com/Codeception/Codeception/blob/2.0/tests/README.md

lowerends:
The tests for FrameworkTest and PhpBrowserTest are running fine locally (no errors). Am I merging to the correct branch?

As for the order of merging the values, my first commit is giving trouble when using ```uncheckOption()```, while it's working fine with the second commit. I'm looking into that now. Any ideas what could cause this? The same queryParams are submitted, but in a different order.

zbateson:
Hmm, I can't explain why the tests would work for you locally.  Applying your PR as a patch and running tests for PhpBrowser and FrameworkTest locally give me 10 errors.

A quick glance at the errors tells me there are probably a few issues:

1 - the order of params would cause the defaults on the page to be sent, rather than what is sent via $params parameter
2 - checkbox values are a special case for booleans at least, see ```setCheckboxBoolValues``` but I'm not sure this is the issue... I'd have to dig a bit deeper
3 - submit button values are probably not included in what's returned by ```getValues()``` - I'm not sure there's a way to get the value of a submit button from the returned form object returned by ```getFormFor()``` for the $button parameter - if not, may need to get it with $form->filter()

Thanks for all the effort.

lowerends:
Ok, I managed to get the tests failing now (was testing on the wrong branch). There's a couple of issues indeed which don't seem to be that easy to solve. Before digging deeper, I'd like to take one step back at the problem that we're trying to solve here, which is adding fields to the POST data. 

I would like to repeat my former question about adding a ```addField()``` method which would allow to add POST data via ```addField($key, $value)```. This would actually allow to simulate Javascript added fields to a form. Would such a function make sense?

I'll also have a deeper look at the problems I have with boolean checkboxes.

lowerends:
A possible extension would then of course be to have a ```deleteField()``` as well, which would allow to remove POST data from the form. This could then be used to simulate a Javascript delete action on a form field.

zbateson:
No worries - thanks for checking :)

You can already send POST data without relying on addField using submitForm - you can pass any data you like in the second argument and it'll be posted:

```php
$I->submitForm('#my-form', ['some-non-existent-field' => 'some value']);
```

lowerends:
Thanks for your feedback.

The problem I have with ```submitForm()``` is that it requires you to add *all* the data to the ```$params``` array, as the ```$params``` array replaces all the form values. So you can't use ```fillField()``` anymore.

zbateson:
That is true I fully agree, however if this PR is fixed and merged fillField will be usable for exactly that scenario :+1: 

Actually $params doesn't replace all form values, it's just incompatible with fillField at the moment.  In other words, it takes default values set on the form when the page was loaded but not those added with fillField.

zbateson:
This would also affect #1689

zbateson:
Hi @lowerends - please check out #1840 - was a bit more of a change than expected :)

DavertMik:
closed in favor of #1840

lowerends:
@zbateson thanks! This looks great. I was looking into it myself as well but had some other work popping up in between. Thanks for the great improvement!

Issue #1830, opened 2015-04-02
group listener not working


litpuvn:
Hi,

I generate Admin group using generate:group command. This will create AdminGroup listener with event handling method: _before and _after

then i enabled it in global config codeception.yml as suggested in the doc.
Then i made a simple cest class with a test method annotated to Admin group.

When i run test, i didn't see the even handling methods invoked.

Any reason for it please? I tried to follow internal codeception code with debug but i haven't found the cause.

thanks,

-Hoang Long

torkiljohnsen:
Seems like the same problem as I'm having: http://phptest.club/t/cest-cept-groups-and-before-after-issues/465

eXorus:
Hello,

I can't reproduce anymore on Codeception 4.1.9.

The listeners works.

PR #1835, opened 2015-04-03
enable group listener for grouping with annotation

litpuvn:
This is a fix for this issue:
https://github.com/Codeception/Codeception/issues/1830


DavertMik:
Finally I get to this commit and yeah, you are right. I will merge it once I reproduce the issue with a failing test.

PR #1838, opened 2015-04-07
Sync changes

janhenkgerritsen:
Sync changes from my own repository of the Laravel 5 module. 

There is no need to release a new version, these changes can just go out when you plan a new release yourself.

### DIFF
--- a/src/Codeception/Module/Laravel5.php
+++ b/src/Codeception/Module/Laravel5.php
@@ -188,15 +188,15 @@ class Laravel5 extends Framework implements ActiveRecord
         return $app;
     }
 
-	/**
+   /**
      * Provides access the Laravel application object.
      *
      * @return \Illuminate\Foundation\Application
      */
-	public function getApplication()
-	{
-		return $this->app;
-	}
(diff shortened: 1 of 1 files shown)


DavertMik:
@janhenkgerritsen thanks!

I was trying to contact you via Facebook. Maybe that was not a good option, but can we talk somehow. I'd like to propose you to join Codeception team and get all means to maintain Laravel modules. We can talk in FB or in skype: `davert.ua`

DavertMik:
btw, i see only whitespace fixes in commit 0.o

janhenkgerritsen:
Yeah that's why I said you don't have to release a new version. But these changes got lost when I moved some code around between two repositories, so I decided to submit the PR already, so I do not forget them at a later point in time :-)

I just checked Facebook and now I see your message from last Friday. I did not reply because I had not seen it yet, somehow I have not received a notification. But we can continue our conversation there.

DavertMik:
FYI team members are: @samdark, @zbateson , @tiger-seo , and @DavertMik

PR #1840, opened 2015-04-08
Refactoring to merge fillField/submitForm

zbateson:
- submitForm refactored to use DomCrawler's Form object, and grab
  defualt values from it instead of filtering the crawler for
  fields manually.
- getFormFor refactored so it creates its own crawler.  Needed so
  input fields in disabled fieldsets aren't submitted
- Created getFormValuesFor to get an array of values as expected by
  submitForm
- Created proceedSubmitForm which is called instead of the deleted
  submitFormWithButton
- Refactored getFormUrl into getFormUrl, getAbsoluteUrlFor and
  mergeUrl methods
- More docblock comments for protected/private methods
- Minor cleanup

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -539,9 +539,13 @@ class InnerBrowser extends Module implements Web
      */
     protected function getFormFor(Crawler $node)
     {
-        $form = $node->parents()->filter('form')->first();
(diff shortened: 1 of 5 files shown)


zbateson:
Fixes #1828 and #1689 - builds over #1832

DavertMik:
I really like the new refactoring, I hope Scrutinizer will like them as well )

PR #1848, opened 2015-04-10
fix $I->am when User name starts with vowel

javigomez:
This is just a small fix for english output. 

Right now if you do:

```php
		$I->am(' Administrator');
```

You will get:

```
* As a Administrator
```

After the patch you will get:
```
* As an Administrator
```

### DIFF
--- a/src/Codeception/Lib/Actor/Shared/Comment.php
+++ b/src/Codeception/Lib/Actor/Shared/Comment.php
@@ -18,14 +18,13 @@ trait Comment
         return $this->comment('I am going to ' . $argumentation);
     }
 
-    public function am($role) {
-	    $role = (string) trim($role);
-	    $firstLetter = $role[0];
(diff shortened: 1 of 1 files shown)


zbateson:
Aah, nice catch :)

A few issues with the commit though unfortunately - if you could fix them that would be great!

1. PSR-2 issues:
    1. Tabbing - should be spaces instead of tabs + indenting is off in the condition
    2. Opening braces for conditions should be on the same line (in your if condition)
2. No need to cast the return value from trim to string, it returns a string
3. ```preg_match``` seems like overkill in this instance, how about ```stripos('aeiou', $role[0]) !== false```

Thank you

javigomez:
Thanks for your review @zbateson :smile: 

I have added your suggested changes.

DavertMik:
thanks

PR #1854, opened 2015-04-14
Update haveRecord method docblock

burdiyan:
Looks like it is expecting table name, not model name. I see error if I place model name. If I'm right, variables and arguments name should be changed as well, but I was afraid changing those. Thanks a lot.


janhenkgerritsen:
It should indeed be the table name and not the model name. The same was true for some other methods in the Laravel4 module, as well as for the corresponding methods in the Laravel5 module. I fixed this in d899ced7b8060a55272e1c66fbe81738f50fcb43.

Issue #668, opened 2013-11-15
Testing 2 different status codes


shanechiam:
Hi everyone,

Thank you in advance. I am trying to configure a test to see 2 different response codes from the test. I basically want the test to run and pass if it sees a 200 or 404 status code. How can I implement this?

PR #1856, opened 2015-04-14
Issue #668 – REST Module : grabResponseStatus

jbottigliero:
Adds `grabResponseStatus` method to the REST module for obtaining the
request's HTTP status code.

Although issue #668 was closed, this seems like it should be available to Scenarios; related to the following quote:

> P.S. also @DavertMik what about to add method to REST module grabResponseCode so people can handle this situations?

Implemented as `getResponseStatus` to match the module's `client` method for retrieving the status code.

### DIFF
--- a/src/Codeception/Module/REST.php
+++ b/src/Codeception/Module/REST.php
@@ -94,6 +94,32 @@ class REST extends \Codeception\Module
     }
 
     /**
+     * Returns current response status code so that it can be used in next scenario steps.
+     *
+     * Example:
+     *
+     * ``` php
+     * <?php
+     * $status = $I->grabResponseStatus();
+     * switch($status) {
+     *     case 201:
+     *         $I->seeResponseIsJson();
+     *     break;
+     *     case 202:
+     *         // "Accepted" Response Handling
+     *     break;
+     * }
+     * ?>
+     * ```
+     *
+     * @return integer
+     */
+    public function grabResponseStatus()
+    {
+        return $this->client->getInternalResponse()->getStatus();
+    }
+
+    /**
      * Sets HTTP header
      *
      * @param $name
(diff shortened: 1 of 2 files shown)


DavertMik:
Thanks. I'm not against module improvements, but I don't like the example you provided. Using `switch` operator inside test is really a bad idea. I'd rather not show bad examples in official documentation. Can you provide better use case for this method?

Right now it looks to me that you need Helper class for that.

jbottigliero:
Sure, would something like the following suffice:

    $I->assertContains(
      $I->grabResponseCode(),
      [200, 202],
      'The resource must either be immediately updated, or be accepted for the update.'
    );

DavertMik:
In your case I still think you should move that code into Helper, there is no wrong in that :)

``` php
function seeResponseIsOk()
{
    $code = $this->getModule('REST')->client->getInternalResponse()->getStatus();
    $this->assertContains($code,  [200, 202],  'The resource must either be immediately updated, or be accepted for the update.');
}
```

DavertMik:
I'm closing it. I still think you should use helper for such cases.

Grabbers are good when you get something in test and pass that argument into next step. Not for implementing complex assertion logic.

kgkg:
Here's example: I want to display or log some data only if http request code is >=400.
If I use "$this->getModule('REST')->client->getInternalResponse()->getStatus()" (which IS btw example of bad coding style), I get: Fatal error: Call to a member function getStatus() on null in C:\usr\htdocs\game\tests\_support\Helper\Api.php on line 23

michaelarnauts:
I know this issue is ancient, but I did an upgrade of Codeception, and you need to use `getStatusCode()` instead of `getStatus()` now to get the HTTP status code.

```
$this->getModule('REST')->client->getInternalResponse()->getStatusCode();
```

It might save somebody a few minutes of searching and cursing.

Naktibalda:
@michaelarnauts It would recommend using wrapper method instead.
It is not available in REST module, so you have to call connection module, e.g.
```php
$this->getModule('PhpBrowser')->_getResponseStatusCode();
```

PR #1860, opened 2015-04-17
Allow following symlinks when searching for tests

nechutny:
Hello,

in our projects  we have in separated modules packed with tests. When deploying to Linux server automated scripts make symlinks instead of copy of this tests to one directory and run them. Symphony Finderdoesnt follow these symlinks by default, so I'm suggesting this change to allow this behavior.

### DIFF
--- a/src/Codeception/TestLoader.php
+++ b/src/Codeception/TestLoader.php
@@ -112,7 +112,7 @@ class TestLoader
 
     public function loadTests()
     {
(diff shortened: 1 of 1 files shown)


DavertMik:
this looks more like a patch to current stable branch which is 2.0 now
please resend your PR to it

nechutny:
OK, created PR #1862

PR #1862, opened 2015-04-17
Allow following symlinks when searching for tests

nechutny:
Hello,

in our projects we have separated modules packed with tests.
When deploying to Linux server, automated scripts create
symlinks, instead of copying these tests to a directory and
run them. Symphony Finder doesnt follow these
symlinks by default, so I suggest a change to allow such behavior.

(Moved from master to branch 2.0 based on #1860)


DavertMik:
thanks!

PR #1872, opened 2015-04-27
Fix g:scenarios --single-file missing linebreaks between scenarios

### LINKED ISSUE #1866: Incorrect scenarios generation for Cests
When I run: `php ~/codecept.phar g:scenarios acceptance --path tests/doc`

I get the following output (5 tests there):                                                                          
* Checkout generated
* Checkout generated
* Checkout generated
* Checkout generated
* Checkout generated

But in the resulting Checkout.txt only the first test scenario is listed. I'm I doing something wrong or this is an issue?

It looks like `--single-file` option works correctly generating all the scenarios, but the txt formatting is slightly broken, e.g. no linebreaks between tests. I would gladly contribute some fixes for this if the issue is confirmed

Zifius:
Partially resolves #1866

### DIFF
--- a/tests/cli/GenerateScenariosCept.php
+++ b/tests/cli/GenerateScenariosCept.php
@@ -1,6 +1,6 @@
 <?php
 $I = new CliGuy($scenario);
-$I->wantTo('generate scenarios сценарий');
+$I->wantTo('generate test scenario');
 $I->amInPath('tests/data/sandbox');
 $I->executeCommand('generate:scenarios dummy');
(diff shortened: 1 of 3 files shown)


DavertMik:
Thanks!

tests/cli/GenerateScenariosCept.php
```diff
@@ -1,6 +1,6 @@
 <?php
 $I = new CliGuy($scenario);
-$I->wantTo('generate scenarios сценарий');
+$I->wantTo('generate test scenario');
```
sergeyklay: :+1:

PR #1878, opened 2015-04-28
Updated Guides for Codeception 2.1

DavertMik:
Docs were updated and in order to reflect new features of Codeception as well as focus on best usage practices, and demonstrate some popular use cases. 

Chapters were reordered Modules and Helpers were rewritten into more general chapter "Reusing Test Code" with Helpers, Actors, PageObjects, and StepObjects described. This chapter goes after Acceptance/Functional/Unit tests, that's why those chapters moved one position up in the order.

Please help by reviewing the text, and finding stupid typos and strange language constructs :)
Your help on documentation improvement is highly appreciated!


docs/06-ReusingTestCode.md
```diff
@@ -0,0 +1,636 @@
...
+    {
+        $I->loginAsAdmin();
+        $I->amOnPage('/admin/profile');
+        $I->see('Admin Profile', 'h1');        
+    }
+}
+?>
+```
+
+If you have complex interaction scenario you may use several page objects in one test. If you feel like adding too much actions into your Actor class (which is AcceptanceTester in this case) consider to move some of them into separate StepObjects.
```
javigomez: Not sure, but I think when you said `page objects` here, you meant `step objects`

Other topic:
I'm very pleased to see this. Reading this documentation seems that we will not need to create Modules like https://github.com/lucatume/wp-browser/blob/master/src/WPWebDriver.php, but instead we will be able to create Injectable Step Objects, right?

I'm just missing one thing: let's say we are testing my "wordpress e-commerce Extension". To test it, I need:

- install Wordpress CMS in my server
- install my e-commerce wordpress plugin in the Wordpress CMS
- test my e-commerce Wordpress Plugin (this is the only real task to test in my project, the rest are wordpress related and not under my responsibility) 

To cover previous steps 1 and 2 we could use in Codeception 2.0  the WordpressBrowser Module created by @lucatume: https://github.com/lucatume/wp-browser/blob/master/src/WPWebDriver.php. But from what I understood now instead of writing a module (witch have some drawbacks like not having the Actor functions) we can just:

- write generic Step Objects: install wordpress, install plugin.
- Autoload them somehow
- use them in my tests

Could you advise me how to do the autoloading part? should we use the _bootstrap file?

Is there a way we could do a generic Packagist package of Step Objects for the most common CMS's: Joomla, Wordpress, Drupal so they can be easily incorporated to an existing codeception instance?


Thanks @DavertMik ^_^

docs/05-UnitTests.md
```diff
@@ -181,22 +186,106 @@ To enable the database functionality in the unit tests please make sure the `Db`
 The database will be cleaned and populated after each test, as it happens for acceptance and functional tests.
 If it's not your required behavior, please change the settings of `Db` module for the current suite.
 
+### Interacting with the Framework
+
+Probably you should not access your database directly if your project already uses ORM for database interactions.
+Why not to use that ORM inside tests? Let's try to write a test using Laravel's ORM Eloquent, for this we need configured Laravel5 module. We won't need its web interaction methods like `amOnPage` or `see`, so let's enable only ORM part of it:
```
phpcodemonkey: This should be "Why not use the ORM directly inside your tests?"

docs/05-UnitTests.md
```diff
@@ -181,22 +186,106 @@ To enable the database functionality in the unit tests please make sure the `Db`
...
+    $this->assertEquals('bill', $user->getName());
+    // verify data was saved using framework methods
+    $this->tester->seeInRepository('Acme\DemoBundle\Entity\User', ['name' => 'bill']);
+    $this->tester->dontSeeInRepository('Acme\DemoBundle\Entity\User', ['name' => 'miles']);
+}
+?>
+```
+
+In both examples you should not be worried about the data persistance between tests. 
+Doctrine2 module as well as Laravel4 module will clean up created data in the end of a test.
```
phpcodemonkey: will clean up created data **at** the end of a test.

docs/05-UnitTests.md
```diff
@@ -181,22 +186,106 @@ To enable the database functionality in the unit tests please make sure the `Db`
...
+    // verify data was saved using framework methods
+    $this->tester->seeInRepository('Acme\DemoBundle\Entity\User', ['name' => 'bill']);
+    $this->tester->dontSeeInRepository('Acme\DemoBundle\Entity\User', ['name' => 'miles']);
+}
+?>
+```
+
+In both examples you should not be worried about the data persistance between tests. 
+Doctrine2 module as well as Laravel4 module will clean up created data in the end of a test. 
+This is done by wrapping a test into transaction and rolling it back afterwards.
```
phpcodemonkey: wrapping a test in **a transaction**

docs/05-UnitTests.md
```diff
@@ -181,22 +186,106 @@ To enable the database functionality in the unit tests please make sure the `Db`
...
-/**
- * @var Symfony\Component\DependencyInjection\Container
- */
+/** @var Symfony\Component\DependencyInjection\Container */
 $container = $this->getModule('Symfony2')->container;
 ?>
 ```
 
-All public variables are listed in references for corresponding modules.
+The same can be done for all public properties of enabled module. Accessible properties are listed in module reference
```
phpcodemonkey: The same can be done for all public properties of **an** enabled module. Accessible properties are listed in **the** module reference

docs/05-UnitTests.md
```diff
@@ -234,11 +323,21 @@ For unit testing you may include `Asserts` module, that adds regular assertions
...
+        - Asserts
+        - Db
+        - \Helper\Unit
 ```
 
 [Learn more about Cest format](http://codeception.com/docs/07-AdvancedUsage#Cest-Classes).
 
+<div class="alert alert-info">
+It may look like Cest format is too simple for writing tests. It doesn't provide assertion methods,
+methods to create mocks and stubs or even accessing module with `getModule`, as we did in example above.
```
phpcodemonkey: accessing **the** module with

docs/05-UnitTests.md
```diff
@@ -234,11 +323,21 @@ For unit testing you may include `Asserts` module, that adds regular assertions
...
+However Cest format is better in separating concerns. Test code does not interefere with support code, provided by `UnitTester` object. All additional actions you may need in your unit/intergation tests you can implement in `Helper\Unit` class. This is a recommended approach, and allows keeping tests verbose and clean.
```
phpcodemonkey: is better **at** separating

This is **the** recommended approach,

docs/06-ReusingTestCode.md
```diff
@@ -0,0 +1,636 @@
...
+```
+
+However, implementing all actions for a reuse in one actor class may lead to breaking the [Single Responsibility Principle](http://en.wikipedia.org/wiki/Single_responsibility_principle). 
+
+## StepObjects
+
+If `login` method defined in Actor class may be used in 90% of your tests,
+StepObjects are great if you need some common functionality for a group of tests. Let's say you are going to test and admin area of a site. Probably you won't need the same actions from admin area while testing the frontend, so it's a good idea to move those admin-specific into their own class. We will call such class a StepObject.
+
+Lets create a Adnin StepObject with generator, by specifying test suite, and passing method expected names on prompt.
```
phpcodemonkey: Lets create a**n** Ad**m**in StepObject

docs/06-ReusingTestCode.md
phpcodemonkey: If you have complex interaction scenario you may use several **step** objects in one test. If you feel like adding too **many** actions into your Actor class

docs/06-ReusingTestCode.md
```diff
@@ -0,0 +1,636 @@
...
+[PageObject pattern](http://code.google.com/p/selenium/wiki/PageObjects), which is widely used by test automation engineers. The PageObject pattern represents a web page as a class and the DOM elements on that page as its properties, and some basic interactions as its methods.
+PageObjects are very important when you are developing a flexible architecture of your tests. Please do not hardcode complex CSS or XPath locators in your tests but rather move them into PageObject classes.
+
+Codeception can generate a PageObject class for you with command:
+
+```bash
+$ php codecept.phar generate:pageobject Login
+```
+
+This will create a `Login` class in `tests/_support/Page`. The basic PageObject is nothing more then empty class with a few stubs.
```
phpcodemonkey: nothing more th**a**n **an** empty

docs/06-ReusingTestCode.md
```diff
@@ -0,0 +1,636 @@
...
+$I->amOnPage(LoginPage::$URL);
+$I->fillField(LoginPage::$usernameField, 'bill evans');
+$I->fillField(LoginPage::$passwordField, 'debby');
+$I->click(LoginPage::$loginButton);
+$I->see('Welcome, bill');
+?>
+```
+As you see, you can freely change markup of your login page, and all the tests interacting with this page will have their locators updated according to properties of LoginPage class.
+
+But let's move further. A PageObject concept also defines that methods for the page interaction should also be stored in a PageObject class. It now stores passed instance of Actor class. An AcceptanceTester can be accessed via `AcceptanceTester` property of that class. Let's define a `login` method in this class.
```
phpcodemonkey: It now stores **a** passed instance of **an** Actor class

docs/06-ReusingTestCode.md
```diff
@@ -0,0 +1,636 @@
...
+<?php
+$I = new AcceptanceTester($scenario);
+$loginPage = new \Page\Login($I);
+$loginPage->login('bill evans', 'debby');
+$I->amOnPage('/profile');
+$I->see('Bill Evans Profile', 'h1');
+?>
+```
+
+If you write your scenario-driven tests in Cest format (which is recommended approach), you can bypass manual creation of PageObject and delegate this task to Codeception. If you specify which objects you need for a test, Codeception will try to create it using it's dependency injection container. In case of PageObject you should declare a class as a parameter for a test method:
```
phpcodemonkey: (which is **the** recommended approach)

 bypass manual creation of **a** PageObject

In **the** case of **a** PageObject you should declare a class

docs/06-ReusingTestCode.md
```diff
@@ -0,0 +1,636 @@
...
+    {
+        $loginPage->login('bill evans', 'debby');
+        $I->amOnPage('/profile');
+        $I->see('Bill Evans Profile', 'h1');        
+    }
+}
+?>
+```
+
+Dependency Injection container can construct any object that require any known class type. For instance, `Page\Login` required `AcceptanceTester`, and so it was injected into `Page\Login` constructor, and PageObject was created and passed into method arguments. You should specify explicitly types of requried objects for Codeception to know what objects should be created for test. Dependency Injection will be described in next chapter.
```
phpcodemonkey: **The** Dependency Injection container can construct

You should specify explicitly **the** types

objects should be created for **a** test

will be described in **the** next chapter

docs/06-ReusingTestCode.md
```diff
@@ -0,0 +1,636 @@
...
+    }
+}
+?>
+```
+
+Dependency Injection container can construct any object that require any known class type. For instance, `Page\Login` required `AcceptanceTester`, and so it was injected into `Page\Login` constructor, and PageObject was created and passed into method arguments. You should specify explicitly types of requried objects for Codeception to know what objects should be created for test. Dependency Injection will be described in next chapter. 
+
+## Modules and Helpers
+
+In examples above we only grouped actions into one. What happens when we need to create a custom action?
```
phpcodemonkey: In **the** examples above we only grouped actions into one **class**.

docs/06-ReusingTestCode.md
```diff
@@ -0,0 +1,636 @@
...
+
+Dependency Injection container can construct any object that require any known class type. For instance, `Page\Login` required `AcceptanceTester`, and so it was injected into `Page\Login` constructor, and PageObject was created and passed into method arguments. You should specify explicitly types of requried objects for Codeception to know what objects should be created for test. Dependency Injection will be described in next chapter. 
+
+## Modules and Helpers
+
+In examples above we only grouped actions into one. What happens when we need to create a custom action? 
+In this case it's a good idea to define missing actions or assertion commands in custom modules, which are called Helpers. They can be found in the `tests/_support/Helper` directory.
+
+<div class="alert alert-info">
+We already know how to create custom login method in AcceptanceTester class. We used actions from standard modules and combined them to make user easier to log in. Helpers allow us to create **new actions** unrelated to standard modules (or using their internals).
```
phpcodemonkey: know how to create **a** custom login

them to make **it easier for a** user ~~easier~~ to log in

docs/06-ReusingTestCode.md
```diff
@@ -0,0 +1,636 @@
...
+```
+
+### Resolving Collisions
+
+What happens if you have two modules that contain the same named actions?
+Codeception allows you to override actions by changing the module order.
+The action from the second module will be loaded and the action from the first one will be ignored.
+The order of the modules can be defined in the suite config.
+
+However, some of modules may conflict with each others. In order to avoid confusion which module is used in first place, Framework modules, PhpBrowser, and WebDriver can't be used together. `_conflicts` method of module is used to specify which class or interface it conflicts to. Codeception will throw exception if there will be a module enabled which matches the provided criteria.
```
phpcodemonkey: may conflict with each other**~~s~~**

which module is used in **the** first place

**The** `_conflicts` method of **a** module

interface it conflicts **~~to~~with**

Codeception will throw **an** exception

docs/06-ReusingTestCode.md
```diff
@@ -0,0 +1,636 @@
...
+
+### Connecting Modules
+
+It's possible that you will need to access internal data or functions from other modules. For example, for your module you might need access responses or internal actions of modules.
```
phpcodemonkey: you might need **to** access responses or internal

docs/06-ReusingTestCode.md
```diff
@@ -0,0 +1,636 @@
...
+    $dbh = $this->getModule('Db')->dbh;
+    $dbh->close();
+    $dbh->open();
+}
+?>
+```
+
+By using the `getModule` function, you get access to all of the public methods and properties of the requested module. The dbh property was defined as public specifically to be available to other modules.
+
+If you want to extend functionality of standard module you may connect to it and by using public properties and methods your custom action or assertion:
```
phpcodemonkey: extend functionality of **a** standard module

and methods **create** your custom action or assertion

docs/06-ReusingTestCode.md
```diff
@@ -0,0 +1,636 @@
...
+For example, the PhpBrowser module prints the response code and current URL every time it moves to a new page.
+Thus, modules are not black boxes. They are trying to show you what is happening during the test. This makes debugging your tests less painful.
+
+To display additional information, use the `debug` and `debugSection` methods of the module.
+Here is an example of how it works for PhpBrowser:
+
+```php
+<?php
+    $this->debugSection('Request', $params);
+    $client->request($method, $uri, $params);
```
phpcodemonkey: $**this->**client->request

docs/06-ReusingTestCode.md
```diff
@@ -0,0 +1,636 @@
...
+<?php
+$this->getModule('WebDriver')->_reconfigure(array('browser' => 'chrome'));
+?>
+```
+
+At the end of a test, all your changes will be rolled back to the original config values.
+
+### Additional options
+
+Another option to extend standard module functionality is to create helper inherited from the module.
```
phpcodemonkey: is to create **a** helper inherited

docs/06-ReusingTestCode.md
```diff
@@ -0,0 +1,636 @@
...
+```php
+<?php
+namespace Helper;
+
+class MyExtendedSelenium extends \Codeception\Module\WebDriver  {
+}
+?>
+```
+
+In this helper you replace implemented methods with your own realization.
```
phpcodemonkey: with your own **~~realization~~implementation**

docs/07-AdvancedUsage.md
```diff
@@ -67,31 +67,29 @@ Also you can define `_failed` method in Cest class which will be called if test
 
 ## Dependency Injection
 
-Codeception supports simple dependency injection for Cest and \Codeception\TestCase\Test classes. It means that you can specify which classes you need as parameters of special `_inject()` method, and Codeception will automatically create respective objects and invoke this method, passing all dependencies as arguments. This can be useful when working with Helpers, for example:
+Codeception supports simple dependency injection for Cest and \Codeception\TestCase\Test classes. It means that you can specify which classes you need as parameters of special `_inject()` method, and Codeception will automatically create respective objects and invoke this method, passing all dependencies as arguments. This may be useful when working with Helpers, for example:
```
phpcodemonkey: as parameters of **the** special `_inject()` method

automatically create **the** respective objects

docs/07-AdvancedUsage.md
```diff
@@ -141,7 +136,27 @@ class MathTest extends \Codeception\TestCase\Test
 ?>
 ```
 
-It is usually preferable to make `_inject()` method `protected` or `private` for not to confuse it with `public` test methods. `_inject()` will be invoked just once right after creation of TestCase object (either Cest or Test).
+However, Dependency Injection is not limited for this. It allows you to **inject any class**, which can be constructed with arguments known to Codeception.
```
phpcodemonkey: Injection is not limited **~~for~~to** this

docs/07-AdvancedUsage.md
```diff
@@ -141,7 +136,27 @@ class MathTest extends \Codeception\TestCase\Test
...
+
+In order to make auto-wiring work, you will need to implement `_inject()` method with the list of desired arguments. It is important to speicfy the type of arguments, so Codeception could guess which objects are expected to receive. `_inject()` will be invoked just once right after creation of TestCase object (either Cest or Test). Dependency Injection will also work in similar manner for Helper and Actor classes.
```
phpcodemonkey: so Codeception **~~could~~can** guess which objects are expected to **be** receive**d**.

after creation of **the** TestCase object

work in **a** similar manner for Helper

docs/07-AdvancedUsage.md
```diff
@@ -189,6 +204,134 @@ class ModeratorCest {
...
+It's also possible to merge multiple environments into one configuration by using comma as a separator:
+
+```bash
+$ php codecept.phar run acceptance --env dev,phantom --env dev,chrome --env dev,firefox
+```
+
+Configuration is merged in the given order. This way you can easily create multiple combinations of your environment configurations.
+
+Depending on environment you may choose which tests are to be executed.
+For example, you might need some tests to be executed only in Firefox, and few tests only in Chrome.
```
phpcodemonkey: and **a** few tests only in Chrome

docs/07-AdvancedUsage.md
```diff
@@ -189,6 +204,134 @@ class ModeratorCest {
...
+If merged environments are used, then you can specify multiple required environments (order is ignored):
+
+```php
+<?php
+$scenario->env('firefox,dev');
+$scenario->env('dev,phantom');
+?>
+```
+
+This way you can easily control what tests will be executed for which environments.
```
phpcodemonkey: control **~~what~~which** tests will be executed for **~~which~~each** environment**~~s~~**.

docs/07-AdvancedUsage.md
```diff
@@ -331,353 +474,7 @@ groups:
...
-
-This way you can easily control what tests will be executed for which environments.
-
 
 ## Conclusion
 
-Codeception is a framework which may look simple at first glance. But it allows you to build powerful tests with single API, refactor them, and write them faster using interactive console. Codeception tests can be easily organized in groups or Cest classes, locators can be stored in PageObjects and common steps can be combined in StepObjects.
-
-Probably it has too much features for one framework. But nevertheless Codeception follows the KISS principle: it's easy to start, easy to learn, and easy to extend.
+Codeception is a framework which may look simple at first glance. But it allows you to build powerful tests with single API, refactor them, and write them faster using interactive console. Codeception tests can be easily organized in groups or Cest classes.
```
phpcodemonkey: powerful tests with **a** single API

faster using **the** interactive console

docs/10-WebServices.md
```diff
@@ -19,21 +19,43 @@ Configure modules in `api.suite.yml`:
...
-		    url: http://serviceapp/api/v1/
-		    depends: PhpBrowser
+    enabled:
+				- REST:
+				    url: http://serviceapp/api/v1/
+				    depends: PhpBrowser
 ```
 
-The REST module will automatically connect to `PhpBrowser`. In case you provide it with Symfony2, Laravel4, Zend, or other framework module, it will connect to them as well. Don't forget to run the `build` command once you finished editing configuration.
+The REST module will connect to `PhpBrowser` according to this configuration. Depending on web service we may deal with XML or JSON responses. Codeception handles well both data formats, however If you don't need one of them, you can explicitly specify JSON or XML parts of module will be used:
```
phpcodemonkey: Codeception handles **~~well~~** both data formats **well**

you can explicitly specify **that the** JSON or XML parts of **the** module will be used:

docs/10-WebServices.md
```diff
@@ -19,21 +19,43 @@ Configure modules in `api.suite.yml`:
...
+```
+
+Once we configured new testing suite, we can create the first sample test:
 
 ```bash
 $ php codecept.phar generate:cept api CreateUser
 ```
 
-It will be called `CreateUserCept.php`. We can use it to test creation of user via web service.
+It will be called `CreateUserCept.php`. We can use it to test creation of user via REST API.
```
phpcodemonkey: creation of **a** user via **the**

docs/10-WebServices.md
```diff
@@ -41,20 +63,38 @@ $I = new ApiTester($scenario);
...
 $I->seeResponseCodeIs(200);
 $I->seeResponseIsJson();
 $I->seeResponseContains('{"result":"ok"}');
 ?>
 ```
 
-REST module is designed to be used with services that serve responses in JSON format. For example, method `seeResponseContainsJson` will convert provided array to JSON and check whether response contains it.
+### Testing JSON Responses
+
+The latest line checked that the response contained provided string. However we shoudn't rely on it, as dpending on content formatting we can receive different results with the same data. What we actually need it to check that response can be parsed and it contains some of values we expect. In case of JSON we can use `seeResponseContainsJson` method
```
phpcodemonkey: the response contained **the** provided string

as d**e**pending on content formatting

What we actually need **~~it~~is** to check that **the** response can be parsed and it contains some of **the** values we expect. In **the** case of

docs/10-WebServices.md
```diff
@@ -41,20 +63,38 @@ $I = new ApiTester($scenario);
...
+	'user' => [
+			'name' => 'davert',
+			'email' => 'davert@codeception.com',
+			'status' => 'inactive'
+	]
+]);
+?>
+```
+
+You may want to perform even more complex assertions on response. This can be done with writing your own methods in [Helper](http://codeception.com/docs/03-ModulesAndHelpers#Helpers) classes. To access the latest JSON response you will need to get `response` property of `REST` module. Let's demonstrate it with `seeResponseIsHtml` method:
```
phpcodemonkey: complex assertions on **a** response

be done **~~with~~by** writing your

response you will need to get **the** `response` property of **the** `REST` module

docs/10-WebServices.md
```diff
@@ -67,6 +107,54 @@ class ApiHelper extends \Codeception\Module
 
 The same way you can receive request parameters and headers.
 
+### Validate JSON structures
+
+It is pretty common for API tests to not only validate the received data but to check the structure of the response. Response data is not something constant, and may change on each request, however the JSON/XML structure should kept the same for an API version. In order to check response structure REST module have some useful methods.
```
phpcodemonkey: Response data is not **~~something constant~~usually considered to be consistent**, and may change on each request,

structure should **be** kept the same for an API

check response structure **the** REST module **~~have~~has** some useful methods

docs/10-WebServices.md
```diff
@@ -67,6 +107,54 @@ class ApiHelper extends \Codeception\Module
 
 The same way you can receive request parameters and headers.
 
+### Validate JSON structures
+
+It is pretty common for API tests to not only validate the received data but to check the structure of the response. Response data is not something constant, and may change on each request, however the JSON/XML structure should kept the same for an API version. In order to check response structure REST module have some useful methods. 
+
+If we expect JSON response to be received we can check its structure with [JSONPath](http://goessner.net/articles/JsonPath/). It looks and sounds like XPath but is designed to work with JSON data, however we can convert JSON into XML and use XPath to validate the structure. Both approaches are valid and can be used in REST module:
```
phpcodemonkey: If we expect **a** JSON response to be received

and can be used in **the** REST module

docs/10-WebServices.md
```diff
@@ -67,6 +107,54 @@ class ApiHelper extends \Codeception\Module
...
+		'user' => [
+			'name' => 'davert',
+			'email' => 'davert@codeception.com',
+			'status' => 'inactive'
+	]
+));
+?>
+```
+
+We are using XmlUtils class which allows us to build XML structures in a clean manner. `toXml` method may accept string, array and returns \DOMDocument instance. If your XML contains attributes and so can't be represented as PHP array you can create XML using [XmlBulder](http://codeception.com/docs/reference/XmlBuilder) class. We will take a look at it a bit more in next section.
```
phpcodemonkey: **The** `toXml` method may accept **a** string**~~,~~ or** array

represented as **a** PHP array you can create XML using **the**

docs/10-WebServices.md
```diff
@@ -131,18 +219,18 @@ $I->seeSoapResponseContainsXPath('//result/user/name[@id=1]');
 ```
 
 In case you don't want to write long XML strings, consider using [XmlBuilder](http://codeception.com/docs/reference/XmlBuilder) class. It will help you to build complex XMLs in jQuery-like style.
-In the next example we will use `XmlBuilder` (created from SoapUtils factory) instead of regular XMLs.
+In the next example we will use `XmlBuilder` instead of regular XMLs.
```
phpcodemonkey: instead of regular XML**~~s~~**.

docs/12-ParallelExecution.md
```diff
@@ -12,7 +12,12 @@ Codeception does not provide a command like `run-parallel`. There is no common s
 * Will they use different hosts?
 * How should I split my tests across parallel processes?
 
-And we don't have a single answer for you to any of those questions. Instead, we provide you with a customizable solution which can be easily configured to match your needs. Also, we will share some ideas that you might find helpful.
+There are two approaches to achive parallelization. We can use [Docker](http://docker.com) and run each process inside isolated containers, and have those containers executed simultaneously. 
+
+<div class="alert alert-warning">
+Docker works realy well for isolating testing environments. 
+By the time of writing this chapter, we didn't have an awesome tool like it. This chapter demonstrates how to manage parallel execution manually. As you will see we spend too much efforts trying to isolate tests which Docker does for free. Today we **recommend using Docker** for parallel testing.
```
phpcodemonkey: we spend too much effort**~~s~~** trying

docs/06-ReusingTestCode.md
DavertMik: @javigomez 

> Could you advise me how to do the autoloading part? should we use the _bootstrap file?

If you use Composer autoloading should be handled by it. If not, you can rely on bootstrap file. If you wrap StepObjects into package you can install them and use in Codeception tests.

In case of Joomla, WP, Drupal tests, I'd recommend to have PageObjects with all common locators defined. They may suit better than StepObjects, as you can take a particular locator of a page and use it in the test, as well as you can define actions which will be used there.

javigomez: thanks :smile: 

do you know any example that you can point me to of `wrap StepObjects into package`?

Thanks very much.

docs/05-UnitTests.md
```diff
@@ -181,22 +186,106 @@ To enable the database functionality in the unit tests please make sure the `Db`
...
+In both examples you should not be worried about the data persistance between tests. 
+Doctrine2 module as well as Laravel4 module will clean up created data at the end of a test. 
+This is done by wrapping a test in a transaction and rolling it back afterwards. 
+
 ### Accessing Module
 
 Codeception allows you to access properties and methods of all modules defined for this suite. Unlike using the UnitTester class for this purpose, using module directly grants you access to all public properties of that module.
 
-For example, if you use `Symfony2` module, here is the way you can access Symfony container:
+We already demonstrated this case in previous code piece where we accessed Entity Manager from a Doctrine2 module
```
pfaocle: s/in previous code piece/in a previous code example/

docs/10-WebServices.md
```diff
@@ -19,21 +19,43 @@ Configure modules in `api.suite.yml`:
...
 modules:
-    config:
-		- REST:
-		    url: http://serviceapp/api/v1/
-		    depends: PhpBrowser
+    enabled:
+				- REST:
+				    url: http://serviceapp/api/v1/
+				    depends: PhpBrowser
 ```
```
javigomez: right now: `vendor/bin/codecept generate:suite api` do not generate the previous lines but this:

```
class_name: ApiTester
modules:
    enabled: [\Helper\Api]
```

Should generate command be updated to add the "url"?

DavertMik: You should include module name and config manually. You just generate an empty suite, no one knows what modules you expect there

javigomez: oh! very nice :smile: 

screen shot 2015-06-26 at 13 20 05

javigomez: But the example is wrong, it should be:
```
class_name: ApiTester
modules:
  enabled:
    - REST
  config:
    REST:
      url: http://joomla.box/repos/redshopb2b/tests/joomla-cms3/
      depends: PhpBrowser
```

Otherwise you get: `[Symfony\Component\Yaml\Exception\ParseException]  `

DavertMik: yeah, there should be linbreak after `enabled:`

PR #1881, opened 2015-04-28
Fixed array to string conversion in InnerBroser commands.

neochief:
Several commands in InnerBrowser have documentation about [array selectors](http://codeception.com/docs/modules/PhpBrowser#seeElement), however such selectors give you the **[ErrorException] Array to string conversion** exception.

I've fixed this and also added support of array selectors to other commands, which have $selector as optional parameter to make things more consistent.


zbateson:
Thanks @neochief - yes this is needed.

Could you fix code formatting though?  Should be PSR-2 compliant - opening brace for if should be on same line:

```php
if ($blah) {
    doStuff();
}
```

and 'tabs' are always 4 spaces not 2.

Issue #1873, opened 2015-04-27
PhpBrowser see/dontSee with HTML


alexwhitman:
In porting some tests from 1.8 to 2.0 I've found an odd issue with `see` and `dontSee` when there are HTML elements within the string being tested.  For example, take the following HTML
```html
<p>This is a sentence containing a <b>bold</b> word.</p>
```
which is rendered as
This is a sentence containing a __bold__ word.

In 1.8 I could use
```php
$I->see('This is a sentence containing a bold word');
```
and the test would pass.  The same assertion in 2.0 fails.  In 2.0 I have to either include the HTML:
```php
$I->see('This is a sentence containing a <b>bold</b> word');
```
or pass in a selector as the second parameter:
```php
$I->see('This is a sentence containing a bold word', 'body');
```
This seems to be because of way the text is checked depending on whether a selector is used or not in https://github.com/Codeception/Codeception/blob/2.0/src/Codeception/Lib/InnerBrowser.php#L140

zbateson:
@alexwhitman - I too have noticed this.  My preference would be for it to work as it did in 1.8.  @DavertMik - the documentation doesn't say anything specific about this behaviour (that I can find).  Could you clarify if it's intentionally done this way?

DavertMik:
Actually I missed this thing while creating PhpBrowser for Codeception 2.0
I rewrote PhpBrowser with a new engine, removing Mink + Goutte driver, that's why the behavior changed. 

Probably the behavior should be as it was in 1.8 if there are no counterpoints against it. Looks like we need to use strip_tags before making lookup with `see`/`dontSee`

PR #1883, opened 2015-04-29
Strip tags for see/dontSee in InnerBrowser

zbateson:
see/dontSee should test against text rather than HTML content.  Solution for #1873

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -1068,13 +1068,21 @@ class InnerBrowser extends Module implements Web
     protected function assertPageContains($needle, $message = '')
     {
         $constraint = new PageConstraint($needle, $this->_getCurrentUri());
-        $this->assertThat($this->client->getInternalResponse()->getContent(), $constraint,$message);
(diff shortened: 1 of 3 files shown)


alexwhitman:
This doesn't seem to completely fix the problem.  Using the following as a test case
```html
<!doctype html>
<html>
	<head>
		<title>html test</title>
	</head>
	<body>
		<ul>
			<li><a href="#">Link 1</a> something else</li>
			<li><a href="#">Link 2</a> something else</li>
			<li><a href="#">Link 3</a> something else</li>
			<li><a href="#">Link 4</a> something else</li>
		</ul>
	</body>
</html>
```
the following test fails:
```php
    public function testHTMLString(WebGuy $I)
    {
        $I->wantTo('get this thing working');
        $I->amOnPage('resources/index2.html');
        $I->see('Link 2 something else');
    }
```

alexwhitman:
Actually, ignore that. It probably does work but I was using `dev-master` which doesn't include this yet.

zbateson:
No worries, let me know if you find any issues :)

PR #1894, opened 2015-05-03
Fix undefined field in generated Actions trait

sustmi:
PhpStorm IDE reports an undefined field `scenario` in generated *Actions trait.
For example here:
```
public function canSee($text, $selector = null) {
    return $this->scenario->runStep(new \Codeception\Step\ConditionalAssertion('see', func_get_args()));
}
```
I think it makes sense, because there is no such field defined in the trait and you can't be sure that the trait will always be inserted into a class that has the field defined.

There is a possibility to define a field in the trait itself, but if it conflicts with the field of same name in the imported class PHP issues either E_STRICT (if the definition is the same) or E_FATAL (if it's not the same) - see http://php.net/manual/en/language.oop5.traits.php#language.oop5.traits.properties .

Therefore I suggest to declare an abstract method `getScenario()` which is basically like specifying  the interface of the class importing the trait (it imposes the importing class to have such method - see http://php.net/manual/en/language.oop5.traits.php#language.oop5.traits.abstract ).

### DIFF
--- a/src/Codeception/Actor.php
+++ b/src/Codeception/Actor.php
@@ -19,6 +19,13 @@ abstract class Actor
         $this->scenario = $scenario;
     }
 
+    /**
+     * @return \Codeception\Scenario
+     */
+    protected function getScenario()
+    {
+        return $this->scenario;
+    }
 
     public function wantToTest($text)
     {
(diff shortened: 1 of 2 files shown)


sustmi:
rerun travis

DavertMik:
:+1: Thanks

PR #1895, opened 2015-05-04
fix port number in WebDriverTest skip message

sustmi:
In commit 002563602837195015ef1671154ed8775d382a78 @davert forgot to change port number to 4444 in WebDriverTest skip message.


zbateson:
Thank you :+1:

PR #1896, opened 2015-05-04
Fb sdk update

enginvardar:
Facebook SDK used for Facebook module was outdated and abandoned by Facebook, This PR updates updates the Facebook Module to the latest Facebook SDK and adds name option to test user to be created.


enginvardar:
rerun travis

tests/unit/Codeception/Module/FacebookTest.php
```diff
@@ -11,12 +11,13 @@
 class FacebookTest extends \PHPUnit_Framework_TestCase
 {
     protected $config = array(
-        'app_id'    => '460287924057084',
-        'secret'    => 'e27a5a07f9f07f52682d61dd69b716b5',
+        'app_id' => '1584417941835888',
+        'secret' => '17e014121a007dc5d149ada7f9201d07',
```
samdark: Are you sure these should be changed? I guess not.

tests/unit/Codeception/Module/FacebookTest.php
```diff
@@ -11,12 +11,13 @@
...
         'test_user' => array(
             'permissions' => array(
-                'publish_stream'
-            )
+                'public_profile',
```
samdark: There's no need for any special permission for it. Public profile is there by default in v4 API.

samdark:
Overall looks OK except application ID change and unused permission.

enginvardar:
@samdark i've changed back application id and removed redundant permission, also i've updated documentation and find out that if test_user config does not exist test fail because of not found name key and fixed that too.

samdark:
Great. Looks good to me now.

Will be checked once again by @DavertMik when he'll be back from vacation and then merged. Thanks!

src/Codeception/Lib/Driver/Facebook.php
```diff
@@ -15,93 +18,98 @@ public function __construct($config, $logCallback = null)
...
 
     /**
-     * @inheritdoc
+     * @param $name
+     * @param array $permissions
+     *
+     * @return array
      */
-    protected function setPersistentData($key, $value)
+    public function createTestUser($name, array $permissions)
```
tiger-seo: name isn't required by Facebook, so i think it shouldn't be required by Codeception ;)

src/Codeception/Lib/Driver/Facebook.php
```diff
@@ -15,93 +18,98 @@ public function __construct($config, $logCallback = null)
...
+    public function sendPostToFacebook($testUserAccessToken, array $parameters)
     {
-        // TODO: Implement clearAllPersistentData() method.
+        $response = $this->executeFacebookRequest(
+            new FacebookSession($testUserAccessToken),
+            'POST',
+            '/me/feed',
+            $parameters
+        )->getRawResponse();
+        return json_decode($response, true);
```
tiger-seo: one empty line should be before return;  please, fix :)

src/Codeception/Lib/Driver/Facebook.php
enginvardar: Actually name is optional, if it's not set, default is 'Codeception Testuser' in module _initialize, i think defining test users name may help during tests.

DavertMik: If the name is optional then it should be passed as optional param as well, I think.

DavertMik:
As for me this looks OK. @tiger-seo what about you? Should we merge or there are things that should be taken into account?

tiger-seo:
it's ok ;)

DavertMik:
Thanks! Merged

PR #1897, opened 2015-05-04
Phalcon1: Fixed declaration of PhalconMemorySession::get

sergeyklay:
+ Cleanup, minor improvement and CS fixes

### DIFF
--- a/src/Codeception/Lib/Connector/Phalcon1.php
+++ b/src/Codeception/Lib/Connector/Phalcon1.php
@@ -75,14 +75,14 @@ class Phalcon1 extends Client
         }
 
         if (!$application instanceof Application && !$application instanceof MicroApplication) {
-            throw new RuntimeException('Unsupported application class');
+            throw new RuntimeException('Unsupported application class.');
         }
 
         $_COOKIE                   = $request->getCookies();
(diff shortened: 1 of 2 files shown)


src/Codeception/Lib/Connector/Phalcon1.php
```diff
@@ -93,7 +93,7 @@ public function doRequest($request)
         $_SERVER['QUERY_STRING'] = http_build_query($_GET);
         $_SERVER['REMOTE_ADDR']  = '127.0.0.1';
 
-        $di['request'] = Stub::construct($di->get('request'), [], array('getRawBody' => $request->getContent()));
+        $di['request'] = Stub::construct($di->get('request'), array(), array('getRawBody' => $request->getContent()));
```
DavertMik: why `array()`? We are in modern PHP era )

DavertMik:
Looks good to merge

src/Codeception/Lib/Connector/Phalcon1.php
sergeyklay: Oops! :smile_cat:  I've been thinking that you are still on 5.3

PR #1915, opened 2015-05-12
Added optional EntityManager configuration to Doctrine2 module

danieltuwien:
Added an option to set a specific EntityManager for the Doctrine2 module. By default the EntityManager will be 'doctrine.orm.entity_manager'.


DavertMik:
This option is about using Doctrine2 with Symfony2 module, so it should be named so user could get that it is used with Symfony only. I'd recommend to provide option to specify exact service name in DIC for entity_manager. 

For example: 

```
symfony_service: doctrine.orm.entity_manager
```

danieltuwien:
Good idea, is done.

DavertMik:
Thanks, however for better readability I'll rename it to `symfony_em_service`

PR #1921, opened 2015-05-14
ZF1 connector improvements

Naktibalda:
1. Set raw request body, it is particularly necessary when using REST module with application/json content-type.
2. Catch exception and close output buffer to avoid 'risky test' mark, allows testing exceptions thrown by controllers.


DavertMik:
Could you resend this into 2.0 branch?

PR #1922, opened 2015-05-15
Appveyor

DavertMik:
Windows environment is now tested as well (not fully) using Appveyor CI


DavertMik:
OK, Windows tests are in!

Issue #1925, opened 2015-05-15
Double slashes issue in a namespace


eugene-manuilov:
Create a test project with namespace using following command:

    codecept bootstrap --namespace My\Nested\Ns

Then create a cest file:

    codecept generate:cest acceptance JustTest

You will see following cest file with double slashes:

```php
<?php
namespace My\Nested\Ns;
use My\Nested\Ns\\AcceptanceTester;

class JustTestCest
{
    public function _before(AcceptanceTester $I)
    {
    }

    public function _after(AcceptanceTester $I)
    {
    }

    // tests
    public function tryToTest(AcceptanceTester $I)
    {
    }
}
```

Same thing for cept files:

```php
<?php use My\Nested\Ns\\AcceptanceTester;
$I = new AcceptanceTester($scenario);
$I->wantTo('perform actions and see result');
```

eugene-manuilov:
Close this issue since pull request has been merged.

PR #1926, opened 2015-05-15
Fix double slashes issue in the namespaces

eugene-manuilov:
This pull request fixes issue described in the #1925 ticket.


DavertMik:
Is this happens only in `master` or in `2.0` this error happens as well?

eugene-manuilov:
It happens just in `master` and works fine in `2.0`

DavertMik:
Thanks

PR #1927, opened 2015-05-15
Improved console log message for generated cept files.

eugene-manuilov:
Currently when we generate cept file using `generate:cept` command we see following message in the terminal:

    Test was created in JustTestCept.php

This message looks less informative then `generate:cest` command's message which looks like this:

    Test was created in /path/to/the/project/folder/tests/acceptance/JustTestCest.php


DavertMik:
thanks

Issue #1928, opened 2015-05-15
State of support for raw request body in framework connectors


Naktibalda:
REST module sends JSON encoded parameters in request body if request method is not GET and Content-Type header is set to application/json

Encoding:
https://github.com/Codeception/Codeception/blob/2.0/src/Codeception/Module/REST.php#L407

Client request:
https://github.com/Codeception/Codeception/blob/2.0/src/Codeception/Module/REST.php#L393

If $parameters is not an array(it is a JSON string now), REST sets $parameters to empty array and passes string as 6th argument ($content).

A lot of framework connectors do not expect to get parameters in request body.

Guzzle - OK,
Kohana - no support for request body,
Laravel4 - OK,
Laravel5 - OK,
Phalcon1 - OK,
SocialEngine - no support, but it uses ZF1, so I will patch it,
Symfony - OK,
Universal - no support and  there is no way to mock php://stdin for it, $HTTP_RAW_POST_DATA could provide a partial cover, but its' use is discouraged,
Yii1 - no support,
Yii2 - OK,
ZF1 - I submitted a patch for it,
ZF2 - partial support, does not match the logic of REST module.

DavertMik:
according to your list all major frameworks support parameters. 
Kohana as SocialEngine are not actively maintained and planned to be removed in 2.1.
Yii1 is really hard to maintain and I hope most of Yii users already switched to Yii2.

I think ZF2 support is the only important here.

Naktibalda:
One problem with my patches was that there were no tests for ZF1 module in Codeception.

Today I created a testcase https://github.com/Naktibalda/codeception-rest-testcase
and a minimal ZF1 app to run these tests in https://github.com/Naktibalda/codeception-zf1-tests

master branch of codeception-zf1-tests uses 2.0.13 version of Codeception and 3 tests fail

Make get request with custom request header (REST/GetMethodWithRequestHeaderCept)                                                                                  Fail
Make post request with raw body (REST/PostMethodWithRawBodyCept)                                                                                                   Fail
Make put request with raw body (REST/PutMethodWithRawBodyCept)                                                                                                     Fail

develop branch uses codeception with all my patches and all tests pass.

I will run these tests against ZF2 app next.

DavertMik:
@Naktibalda Thank you, this is awesome! ZF tests will be really helpful

Naktibalda:
https://github.com/Naktibalda/codeception-zf2-tests

[...]

PR #1929, opened 2015-05-15
Pass raw body to ZF2 if it is present in request

Naktibalda:
Related to issue #1928

This change allows to use raw body in POST and other methods.

I haven't tested this change because I am not using ZF2, but I haven't introduced any new method calls and it should just work.


DavertMik:
Works with https://github.com/Naktibalda/codeception-zf2-tests

Thanks!

Issue #1933, opened 2015-05-16
$_POST is not cleared among GET requests in Yii1 module


xt99:
Seems like in Yii1 module `$_POST` superglobal persists once `POST` data was sent and until next `POST` request comes. Inside all `GET` requests between `POST` requests will appear data from the first `POST` request. 

Suppose I have this test code : 

```php
// submit POST form
$I->amOnPage('/site/login/');
$I->fillField('LoginForm[email]', 'user@test.com');
$I->fillField('LoginForm[password]', '1111');
$I->click('Войти', 'button');

Debug::debug('$_GET is :');
Debug::debug(print_r($_GET, true));
Debug::debug('$_POST is :');
Debug::debug(print_r($_POST, true));

// just a GET request with params
$I->amOnPage('/?foo=bar');

Debug::debug('$_GET is :');
Debug::debug(print_r($_GET, true));
Debug::debug('$_POST is :');
Debug::debug(print_r($_POST, true));

// just another GET request
$I->amOnPage('/products/');

Debug::debug('$_GET is :');
Debug::debug(print_r($_GET, true));
Debug::debug('$_POST is :');
Debug::debug(print_r($_POST, true));
```

This is what I get : 

```
Scenario:
* I am on page "/site/login/"

  [Response] 200
  [Page] http://localhost/site/login/
  [Cookies] []
  [Headers] []
* I fill field "LoginForm[email]","user@test"
* I fill field "LoginForm[password]","1111"
* I click "Войти","button"
  [Uri] http://localhost/test.php/site/login
  [POST] {"LoginForm[email]":"user@test","LoginForm[password]":"1111","yt0":""}
  [Response] 200
  [Page] http://localhost/
  [Cookies] []
  [Headers] []
  $_GET is :
  Array
  (
  )
  
  $_POST is :
  Array
  (
      [LoginForm] => Array
          (
              [email] => user@test
              [password] => 1111
          )
  
      [yt0] => 
  )
  
* I am on page "/?foo=bar"
  [Response] 200
  [Page] http://localhost/?foo=bar
  [Cookies] []
  [Headers] []
  $_GET is :
  Array
  (
      [foo] => bar
  )
  
  $_POST is :
  Array          <--- this is wrong and should be empty, I've just send a GET request
  (
      [LoginForm] => Array
          (
              [email] => user@test
              [password] => 1111
          )
  
      [yt0] => 
  )
  
* I am on page "/products/"
  [Response] 200
[...]

PR #1934, opened 2015-05-16
$_POST is now cleared among GET requests

xt99:
Fixed issue #1933.
Based on [Yii2 code](https://github.com/Codeception/Codeception/blob/2.0.13/src/Codeception/Lib/Connector/Yii2.php#L51).


DavertMik:
thanks

Issue #1803, opened 2015-03-26
Example of Doctrine2 with Symfony2 module setup


mhightower:
I'm having problems setting up a simple test for Codeception v1.8.7 using the Symfony2 and Doctrine2 modules. Is there a working example somewhere?

Context: I'm setting up a base project that's using Codeception to handle the testing. For this base project I want to write a simple test and I'm using the WelcomeCept.php but I'm not sure what else I would need to setup to make it work.
 The error I get is:
```
Codeception\Exception\ModuleConfig]
  Doctrine2 module is not configured!
  Doctrine2 module requires EntityManager explicitly set.
  You can use your bootstrap file to assign the EntityManager:
  \Codeception\Module\Doctrine2::$em = $em 
```

I've added the \Codeception\Module\Doctrine2::$em = $em to the bootstrap but where do I get the EntityManager to populate $em?

Thanks

pitpit:
Actually it's a bug that I fixed here https://github.com/Codeception/Codeception/pull/1939

You can bypass it temporarly giving a value for config `connection_callback`

```
modules:
    enabled:
        Doctrine2:
            depends: Symfony2
            connection_callback: [AFakeValueHere]
```

PR #1939, opened 2015-05-19
when setting Symfony2 as dependency of Doctrine2, whe should not have an exception

pitpit:
The exception was :

```
  [Codeception\Exception\ModuleConfigException]
  Doctrine2 module is not configured!
  Doctrine2 module requires EntityManager explicitly set.
  You can use your bootstrap file to assign the EntityManager:
  \Codeception\Module\Doctrine2::$em = $em
```

it should answer to question #1803


DavertMik:
thanks. Looks valid to me

PR #1941, opened 2015-05-20
Win1251

MikhailovMV:
Доработка над Win1251


DavertMik:
What branch are you pushing it into? It looks like you've chosen 1.8 but you are putting commits from 2.0

Issue #1942, opened 2015-05-20
Db::haveInDatabase with foreign keys


satahippy:
When you try to `Db::haveInDatabase` in tables with foreign keys some rows can't be deleted.
It occurs when you insert into first table, then insert into second table, that has a foreign key to first.
First inserted row can't be deleted.

PR #1943, opened 2015-05-20
Db::haveInDatabase. Reverse order of removing inserted records.

satahippy:
For foreign keys compatibility.
Issue: #1942


DavertMik:
Simple and elegant fix, IMHO. Thanks, I don't mind merging it

PR #1945, opened 2015-05-20
silex must extend Framework

arduanov:
Sliex module must extend \Codeception\Lib\Framework as well as Symfony 2 module.
In order to Rest module set $this->isFunctional to true only for Framework instance.

 ```
        if ($this->connectionModule instanceof Framework) {
            $this->isFunctional = true;
        }
```


DavertMik:
thanks!

PR #1949, opened 2015-05-21
remove merged duplicated (and older) line

scottaubrey:
This causes an error during test run, I think it was added during a merge of older branches and is not needed.


DavertMik:
yes, looks like a bad merge. However, my local version does not contain this line. Strange

PR #1956, opened 2015-05-22
Update InnerBrowser.php

sergeyklay:
In accordance with `Codeception\Lib\Interfaces\Web` to avoid PHPDoc generation issue


DavertMik:
thanks

PR #1957, opened 2015-05-22
Update InnerBrowser.php

sergeyklay:
Explicit definition of the parameters in order to avoid `the variable xxxxx is undefined`.
Added support for `httpOnly` & `encodedValue`


DavertMik:
Thanks!

PR #1960, opened 2015-05-24
Partial patch for amOnSubdomain - wierdness #1757

### LINKED ISSUE #1757: amOnSubdomain - wierdness
Hi

I have a URL, such as `dev.mysite.co.uk`. This is the top level website, i.e. `mysite.co.uk` will redirect to `dev.mysite.co.uk`

I also have `admin.mysite.co.uk`

In my functional tests, I define the URL (using PhantomJS) as follows:

`url: 'http://dev.mysite.co.uk'`

If I run the main tests, all is good, until I get to testing the admin

I have something along the following:

```
$I->amOnSubdomain('admin')
$I->amOnPage('login');
```

Pretty straight forward, until I get an exception and a failed image.

It seems that Codeception, thinks that becuase I am using `admin` as the subdomain (and .co.uk) it thinks I want `admin.co.uk` which is wrong.

I'm using version `1.8.7` (due to using an older version of PHP)

Has anyone else seen this?

Naktibalda:
Fixed #1757

I made subdomain removal ungreedy.

This patch enables transition from dev.mysite.co.uk to admin.mysite.co.uk

It doesn't enable transition from mysite.co.uk to admin.mysite.co.uk, because that functionality requires knowledge about all composite top level domains.

### DIFF
--- a/src/Codeception/Module/PhpBrowser.php
+++ b/src/Codeception/Module/PhpBrowser.php
@@ -168,7 +168,7 @@ class PhpBrowser extends InnerBrowser implements Remote, MultiSession
     public function amOnSubdomain($subdomain)
     {
         $url = $this->config['url'];
-        $url = preg_replace('~(https?:\/\/)(.*\.)(.*\.)~', "$1$3", $url); // removing current subdomain
+        $url = preg_replace('~(https?:\/\/)(.*?\.)(.*\.)~', "$1$3", $url); // removing current subdomain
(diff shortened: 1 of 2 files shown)


Naktibalda:
I blame travis, the test passed after patch on my pc.

Naktibalda:
Actually my change is backwards incompatible for anyone who is currently using $I->amOnSubdomain('admin.mysite');

I will propose the alternative solution in the issue.

Issue #1779, opened 2015-03-14
[PHPUnit_Framework_Exception] Undefined variable: file


mipon:
Hello,

I've started seeing this error and the tests failing.

```
$ php codecept.phar run
```
> [PHPUnit_Framework_Exception] 
Undefined variable: file 

If `functional` is specified, it gave the following warnigns and the tests completed.
```
$ php codecept.phar run functional
```

```
PHP Notice: Undefined variable: file in phar:///codecept.phar/src/Codeception/Lib/Parser.php on line 105

PHP Stack trace:

PHP 1. {main}() /codecept.phar:0

PHP 2. require_once() /codecept.phar:7

PHP 3. Symfony\Component\Console\Application->run() phar:///codecept.phar/codecept:34

PHP 4. Symfony\Component\Console\Application->doRun() phar:///codecept.phar/vendor/symfony/console/Symfony/Component/Console/Application.php:126

PHP 5. Symfony\Component\Console\Application->doRunCommand() phar:///codecept.phar/vendor/symfony/console/Symfony/Component/Console/Application.php:195

PHP 6. Symfony\Component\Console\Command\Command->run() phar:///codecept.phar/vendor/symfony/console/Symfony/Component/Console/Application.php:874

PHP 7. Codeception\Command\Run->execute() phar:///codecept.phar/vendor/symfony/console/Symfony/Component/Console/Command/Command.php:253

PHP 8. Codeception\Command\Run->runSuites() phar:///codecept.phar/src/Codeception/Command/Run.php:191

PHP 9. Codeception\Codecept->run() phar:///codecept.phar/src/Codeception/Command/Run.php:262

PHP 10. Codeception\Codecept->runSuite() phar:///codecept.phar/src/Codeception/Codecept.php:149

PHP 11. Codeception\SuiteManager->loadTests() phar:///codecept.phar/src/Codeception/Codecept.php:165

PHP 12. Codeception\TestLoader->loadTests() phar:///codecept.phar/src/Codeception/SuiteManager.php:94

PHP 13. call_user_func:{phar:///codecept.phar/src/Codeception/TestLoader.php:118}() phar:///codecept.phar/src/Codeception/TestLoader.php:118

PHP 14. Codeception\TestLoader->addTest() phar:///codecept.phar/src/Codeception/TestLoader.php:118

PHP 15. Codeception\Lib\Parser::getClassesFromFile() phar:///codecept.phar/src/Codeception/TestLoader.php:125

[...]

PR #1963, opened 2015-05-25
Reproduced issue #1779

Naktibalda:
Failing test for issue #1779


Naktibalda:
The simplest possible fix.

samdark:
Looks valid.

Naktibalda:
Actually getClassesFromFile shouldn't include, it violates SRP.

src/Codeception/Lib/Parser.php
```diff
@@ -101,7 +101,7 @@ protected function addCommentStep($comment)
 
     public static function getClassesFromFile($file)
     {
-        include_once $file;
+        self::includeFile($file);
```
DavertMik: I think anonymous function can be better here. At least it does just what needed - creates a new context without implementing a new method

Naktibalda: But it would look much uglier and be much slower.

DavertMik:
Ok, thanks!

Issue #1964, opened 2015-05-25
No "feature" attribute is added for TestCase test report


tankist:
For Cest and Cept formats `feature` attribute is added to the report.xml:

```xml
<testcase file="tests/unit/Library/ArrayQueryCest.php" name="testArrayQueryTokenize" class="Unit\Library\ArrayQueryCest" feature="test array query tokenize" assertions="1" time="0.006296"/>
```

But not for Test format:

```xml
<testcase name="testRobotsAction" class="Unit\Controller\IndexControllerTest" file="tests/unit/Controller/IndexControllerTest.php" line="82" assertions="3" time="0.134337"/>
```

Issue is happened because Codeception\TestCase\Test class does not implement Interfaces\Reported interface.

tankist:
This issue is related to https://github.com/Block8/PHPCI/issues/988. I've fixed it there but I think that fixing it there is more logically.

tankist:
Fixed in https://github.com/Codeception/Codeception/pull/1973

PR #1965, opened 2015-05-25
Fix for #1964. Added Interfaces\Reported implementation to the Codeception\TestCase\Test class

tankist:
Added getReportFields to the Codeception\TestCase\Test class


DavertMik:
thanks. Looks valid. You have sent this PR into master branch so your changes will be applied starting from Codeception 2.1 (which is planned for final release soon)

DavertMik:
@tankist  Sorry, could you resend this PR into `2.0` branch? I think this should go as minor change

PR #1967, opened 2015-05-25
Phalcon1 DI

sergeyklay:
* Improved dependency injector container check 
* Used short syntax for arrays

**Important**
* `Phalcon1::_after` now uses `isset($this->di['db'])` instead `$this->di->has('db')` — in some cases di not initialized here (for example when test fails)

### DIFF
--- a/src/Codeception/Module/Phalcon1.php
+++ b/src/Codeception/Module/Phalcon1.php
@@ -2,6 +2,7 @@
 
 namespace Codeception\Module;
 
+use Codeception\Exception\Module;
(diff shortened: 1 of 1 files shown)


sergeyklay:
@DavertMik Current 2.0 branch is broken? Why tests fails?

DavertMik:
Yeah, some stupid MySQL test :(
I will fix those test and merge valid PRs afterwards

Thanks for waiting

DavertMik:
Thanks, merging

PR #1948, opened 2015-05-21
Add a note about how to contribute documentation

javigomez:
I got the idea from https://github.com/Codeception/codeception.github.com/issues/17#issuecomment-104038409

PR #1969, opened 2015-05-26
Update contributing instructions

javigomez:
This pull updates the readme and contribution files based on @Ragazzo  suggestion at #1948

@DavertMik I'm sending this only to Master. If this works for you, I will prepare the same for 2.0 branch.


DavertMik:
Plese send this directly to 2.0 as it will be displayed on repo page

javigomez:
@DavertMik thanks for merging #1971 on `2.0`. I'm not sure however why the current pull got closed, I guess that you would like to have the same contributing instructions in `master` than at `2.0`, isn't it?

PR #1971, opened 2015-05-26
Update Contributing instructions in 2.0

javigomez:
This pull updates the readme and contribution files based on @Ragazzo suggestion at #1948

This is the same contribution sent at #1969 this time against 2.0 branch as suggested by @DavertMik


DavertMik:
Thanks!

PR #1973, opened 2015-05-26
Fix for #1964 Added getReportFields to Codeception\TestCase\Test class

tankist:
Added getReportFields to Codeception\TestCase\Test class


DavertMik:
Thanks

Issue #1923, opened 2015-05-15
seeInField() has been broken for input with associative arrays in the name since 2.0.13 


Leammas:
I've used next code snippet for testing it out (changing selector does not help):

```html
<html>
<head>
</head>
<body>
<form id="foo">
<input type="text" value="baz" name="FooBar[bar]">
</form>
</body>
</html>
```

```php
<?php
$I = new AcceptanceTester($scenario);
$I->amOnUrl('http://localhost:8080/foo.html');
$I->seeInField('input[name="FooBar[bar]"]', 'baz');
```

Results for different versions below:
2.0.12
```
vendor/bin/codecept run
Codeception PHP Testing Framework v2.0.12
Powered by PHPUnit 4.5.1 by Sebastian Bergmann and contributors.
Acceptance Tests (1) -----------------------------------------------------
FooCept                                                             Ok
--------------------------------------------------------------------------
Time: 3.61 seconds, Memory: 5.75Mb
OK (1 test, 1 assertion)
```

2.0.13
```
vendor/bin/codecept run
Codeception PHP Testing Framework v2.0.13
Powered by PHPUnit 4.6.6 by Sebastian Bergmann and contributors.

Acceptance Tests (1) -----------------------------------------------------
FooCept                                                             Fail
--------------------------------------------------------------------------
Time: 8.6 seconds, Memory: 6.75Mb
There was 1 failure:
---------
1) FooCept (/var/www/srv.lol/foo/tests/acceptance/FooCept.php)
Couldn't see in field "input[name="FooBar[bar]"]","baz":
Failed testing for 'baz' in FooBar[bar]'s value:
Failed asserting that an array contains 'baz'.

Scenario Steps:
2. I see in field "input[name="FooBar[bar]"]","baz"
1. I am on url "http://localhost:8080/foo.html"

FAILURES!
Tests: 1, Assertions: 1, Failures: 1.
```

aniolekx:
I have similar issue under version: v2.1.0-beta

aniolekx:
any updates on above?

DavertMik:
yes, this was fixed in 2.0 branch

DavertMik:
but not yet released nor merged into master

PR #1983, opened 2015-05-27
Alternative fix for regression #1923

zbateson:
- Reverted changes to getFormValuesFor
- Created a different method to get the singular value for
  "seeInField" testing
- Added test for multi-dimensional array submission in
  PhpBrowserTest (functionality not yet implemented in WebDriver)

@DavertMik - I wanted to keep the getFormValuesFor function returning multi-dimensional arrays, so they can be overridden by calls to submitForm.  I plan to update seeInField so it accepts multi-dimensional arrays as well, so one can call submitForm/seeInFormFields using the same array (not quite there yet.)

Please have a quick look when you have a moment, and let me know if you have any concerns!

Thanks


DavertMik:
Looks good. Thank you

PR #1988, opened 2015-05-28
Fixes REST module when POST a JsonSerializable.

andersonamuller:
Because of https://github.com/Codeception/Codeception/commit/dcfc8e303b1260a9b6937448be991620d20e5a64#diff-d8e5babfee3fd0e8675337cf8ff58b50, when I post/put/patch a `JsonSerializable` object, is always sending a request without body.

### DIFF
--- a/src/Codeception/Module/REST.php
+++ b/src/Codeception/Module/REST.php
@@ -390,7 +390,7 @@ class REST extends \Codeception\Module
                 $this->client->request($method, $url, $parameters, $files);
             } else {
                 $this->debugSection("Request", "$method $url ".json_encode($parameters));
-                $this->client->request($method, $url, array(), $files, array(), $parameters);
+                $this->client->request($method, $url, array(), $files, array(), json_encode($parameters));
             }
         } else {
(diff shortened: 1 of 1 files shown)


DavertMik:
Could you look into failing tests and fix them?

Naktibalda:
Hi @andersonamuller,

All encoding logic is in [encodeApplicationJson](https://github.com/Codeception/Codeception/blob/dcfc8e303b1260a9b6937448be991620d20e5a64/src/Codeception/Module/REST.php#L409) method.

As you can see, REST module only send JSON in request body when Content-Type header is set to application/json.
Is it a part that you missed?

Could you create a failing test to demonstrate your issue before trying to fix it?

Naktibalda:
You can send JSON in request body in 2 ways:

1. let REST module handle encoding
  ```php
  $I->haveHttpHeader('Content-Type','application/json');
  $I->sendPOST('/url', $object);
  ```

2. encode it yourself
  ```php
  $I->sendPOST('/url', json_encode($object));
  ```

Issue #1997, opened 2015-06-01
mongodb driver does not allow whitespace in the filename of its dumps


christoph-hautzinger:
I had the problem within jenkins where we have paths like this:

`jenkins@build01:~/workspace/00 - conspecton - Build - Build/products`

setting config

```
MongoDb:
  dsn: 'localhost:27017/conspecton_test'
  user: ''
  password: ''
  dump: 'tests/_data/product_platform.js'
  populate: true
  cleanup: false
```

only cleared the whole database without loading the fixtures

PR #1998, opened 2015-06-01
fixed a problem with whitespaces in filenames when loading mongodb dumps

christoph-hautzinger:
fixed a problem with whitespaces in filenames when loading mongodb dumps #1997


DavertMik:
Thanks

PR #2005, opened 2015-06-03
Configure the client host for the laravel5 module

CristianGiordano:
This will set the correct base url of the application which resolves
issues when routing, and provides better error outputs as the URL
will be the correct application URL and not localhost or empty


JonoB:
:+1:

janhenkgerritsen:
Fixed this in another commit: 82ee3b3ece71f357d672c0b0d47a6fa19c317d36. Thanks anyway!

Issue #2008, opened 2015-06-03
The Symfony\Component\EventDispatcher\Event::getName method is deprecated since version 2.4 and will be removed in 3.0.


Naktibalda:
I ran `composer update` and SuiteManagerTest started failing.

There was 1 error:

---------
1) SuiteManagerTest::testRun
PHPUnit_Framework_Exception: The Symfony\Component\EventDispatcher\Event::getName method is deprecated since version 2.4 and will be removed in 3.0. The event name can be received in the listener call instead.

#1  Codeception\Subscriber\ErrorHandler->errorHandler
#2  /home/gintas/workspace/codeception/vendor/symfony/event-dispatcher/Event.php:116
#3  /home/gintas/workspace/codeception/tests/unit/Codeception/SuiteManagerTest.php:37
#4  SuiteManagerTest->{closure}
#5  /home/gintas/workspace/codeception/vendor/symfony/event-dispatcher/EventDispatcher.php:164
#6  /home/gintas/workspace/codeception/vendor/symfony/event-dispatcher/EventDispatcher.php:53
#7  /home/gintas/workspace/codeception/tests/unit/Codeception/SuiteManagerTest.php:39
#8  SuiteManagerTest->testRun

--

There was 1 failure:

---------
1) SuiteManagerTest::testGroupEventsAreFired
Failed asserting that an array contains 'test.before'.

#1  /home/gintas/workspace/codeception/tests/unit/Codeception/SuiteManagerTest.php:82
#2  SuiteManagerTest->testGroupEventsAreFired

DavertMik:
Yeah, that what happened to me in `master` branch. I already applied fixes there. I think this should be backported to 2.0

PR #2009, opened 2015-06-03
Event dispatcher upgrade

Naktibalda:
Patch for issue #2008


DavertMik:
Thanks!

DavertMik:
I have some differences in my implementation but actually there is no major difference what to use

PR #2010, opened 2015-06-03
Phalcon1: Fixed getting fields

sergeyklay:
In some cases Phalcon's model message has more than one field.
For example:

```php
    public function validation()
    {
        $this->validate(new Uniqueness(
            [
                'field' => ['phone', 'email'],
                'message' => 'User already exists!'
            ]
        ));

        if ($this->validationHasFailed() == true) {
            return false;
        }

        return true;
    }
```

In this case `$message->getField()` returns:

```
Array
(
    [0] => phone
    [1] => email
)

```

### DIFF
--- a/src/Codeception/Module/Phalcon1.php
+++ b/src/Codeception/Module/Phalcon1.php
@@ -241,11 +241,19 @@ class Phalcon1 extends Framework implements ActiveRecord
     {
         $record = $this->getModelRecord($model);
         $res = $record->save($attributes);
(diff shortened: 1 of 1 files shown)


Naktibalda:
Merge the latest changes from 2.0 and GenerateStepObjectTest tests will pass, I don't know why did the AuthCest::authAsUser failed.

sergeyklay:
@Naktibalda Done

PR #2002, opened 2015-06-06
multibyte message was broken

nojimage:
I write an acceptance test as below,

```
<?php
$I->wantTo('日本語のメッセージを書きたい');
```

I run this testcase, results expect that
```
Acceptance Tests (1) -------------------------------------------------------------------------------------------------
日本語のメッセージを書きたい (MultibyteCept)                                                      Ok
----------------------------------------------------------------------------------------------------------------------
```

but, actual get
```
Acceptance Tests (1) -------------------------------------------------------------------------------------------------
Ɨ�本語のメッセージを書きたい (MultibyteCept)                                                      Ok
----------------------------------------------------------------------------------------------------------------------
```

The reason for this problem is using `strtoupper` at output messages.

The `strtoupper` will be broken multibyte string.
I think it is better that you use instead of the `mb_strtoupper`.


nojimage:
Please merge PR, if there is no problem.

src/Codeception/Lib/Console/Message.php
```diff
@@ -32,7 +33,7 @@ public function style($name)
 
     public function width($length, $char = ' ')
     {
-        $message_length = strlen(strip_tags($this->message));
+        $message_length = Multibyte::strwidth(strip_tags($this->message));
```
samdark: Better to name it strlen still.

src/Codeception/Util/Multibyte.php
```diff
@@ -0,0 +1,130 @@
...
+namespace Codeception\Util;
+
+/**
+ * Multibyte handling methods
+ */
+class Multibyte
+{
+
+    /**
+     * strlen
```
samdark: Should be properly described.

src/Codeception/Util/Multibyte.php
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    /**
+     * strwidth
+     *
+     * @param string $string
+     * @param string $encoding
+     * @return string
+     */
+    public static function strwidth($string, $encoding = 'UTF-8')
```
samdark: Why not strlen?

src/Codeception/Util/Multibyte.php
```diff
@@ -0,0 +1,130 @@
...
+     * @param string  $string
+     * @param integer $start
+     * @param integer $length
+     * @param string  $encoding
+     * @return string
+     */
+    public static function substr($string, $start, $length = null, $encoding = 'UTF-8')
+    {
+        if (!function_exists('mb_substr')) {
+            $length = is_null($length) ? strlen($length) : $length;
```
samdark: `is_null` should be replaced with `$length === null`

sergeyklay: why?

kkopachev:
What's the point in introducing new class with mb_functions existence check, if in composer.json ext-mbstring is required anyway, so all these functions are available? Plus test will fail if there is no mbstring extension installed.

src/Codeception/Lib/Console/Message.php
nojimage: mb_strwidth is return width of string. multibyte characters are usually twice the width of single byte characters.

nojimage:
@kkopachev I had overlooked ext-mbstring required in composer.json.  :p
The Multibyte class does not need.

src/Codeception/Util/Multibyte.php
samdark: There's no need to use function in this case.

sergeyklay: `is_null($length) ? strlen($length)` :smiley_cat: I just noticed on this design

DavertMik:
this PR can't be merged automatically. Please resend it

PR #2020, opened 2015-06-06
Create a new Symfony2 kernel for every test

quaninte:
To fix entity manager closed error when running all tests together.


DavertMik:
Thanks

Issue #2021, opened 2015-06-07
Buttons or Input submit fields with an associative array name do not get included in the request paramters as part of the array


efrag:
I am currently using Codeception for writing functional tests against a Symfony2 application and am trying to test a form that is similar to the following:
```
<form id="foo" method="POST" action="/form/example18">
    <label for="example18_foo">Foo</label>
    <input type="text" id="example18_foo" name="example18[foo]" value="" />

    <label for="example18_bar">Bar</label>
    <input type="text" id="example18_bar" name="example18[bar]" value="" />

    <input type="submit" name="example18[submit]" value="Submit Foo" />
    <button type="submit" name="example18[cancel]" id="cancelButton" value="cancel">Cancel</button>
</form>
```
If I try to submit the form (from within a codeception test) the submit button is not included in the request parameters in the same way as the foo and bar fields.
```
        $this->module->amOnPage('/form/example18');
        $this->module->fillField('Foo', 'foo value');
        $this->module->fillField('Bar', 'bar value');
        $this->module->click('Submit Foo');
```
More specifically printing the debug output for this request for the latest version of the 2.0 branch I get the following:
```
[Parameters] {"example18":{"foo":"foo value","bar":"bar value"},"example18[submit]":"Submit"}
```
When Symfony2 reads the submitted values it will identify that the form has been submitted but it will not pick up on the fact that the submit button was clicked. 

This gets a bit trickier if you try to identify whether the cancel button has been clicked on a given form while using the Symfony2 form component for the application. 

For the form shown above if the cancel button does not have a value specified then it is not included at all in the request parameters 
```
[Parameters] {"example18":{"foo":"foo value","bar":"bar value"}}
```
and if it does have a value the request parameters are as follows
```
[Parameters] {"example18":{"foo":"foo value","bar":"bar value"},"example18[cancel]":"cancel"}
```
The reason this is happening is because of the way the name of the button is defined. It is read as a string "example18[cancel]" and it is not transformed into example18['cancel']. 

I will create a pull request with the code that I believe would resolve this for the click and clickByLocator functions of the InnerBrowser.

Could you please consider this for review?

Kind regards,
Eleni

PR #2022, opened 2015-06-07
Allow associative array names for buttons and input type submit

efrag:
Code and pull request following #2021

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -102,8 +102,12 @@ class InnerBrowser extends Module implements Web
         $button = $this->crawler->selectButton($buttonText);
         if (count($button)) {
             $buttonValue = [];
-            if (strval($button->attr('name')) !== '' && $button->attr('value') !== null) {
-                $buttonValue = [$button->attr('name') => $button->attr('value')];
+            if (strval($button->attr('name')) !== '') {
+                if (strpos($button->attr('name'), '[') !== false) {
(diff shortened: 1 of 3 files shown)


zbateson:
Hi @efrag - 

Thanks for the pull request.  I'm a bit surprised by this one to be honest, your test works without your proposed changes to InnerBrowser (if you undo the three changes in InnerBrowser, run the tests, all is good).

Could this be an issue with Symfony, and not InnerBrowser/PhpBrowser?

zbateson:
For clarification, actually field names with 'named' associative arrays (like foo[bar]) can be passed as strings or arrays.  The only exception to that is when you want sequenced array elements without passing the actual keys, because you can't have two elements named 'foo[]' passed in the array (only one could exist with that key in the array.)

So you could have:
```php
[
    'foo' => ['value1', 'value2'],    // sends 'foo[]'
    'bar[blah]' => 'value3',
    'bar[argh]' => 'value4',
    // or
    'baz' => [
        'key1' => 'value5',
        'key2' => 'value6',
    ]
]
```
and it would submit fine.

efrag:
Hi @zbateson,

Thanks for looking into this and for your quick response. 

After your comment I re-run the tests on the Symfony2 app and noticed that I was using Codeception version 2.0.13 in which the test was failing. After upgrading to 2.0.14 the tests run successfully after adding a value on the submit and cancel buttons. 

So, again thank you for looking into this, shall we close this pull request?

zbateson:
Aah, right - there was a fix somewhere related to this :) I always forget what's already been released.

PR #2026, opened 2015-06-09
Upgraded to Guzzle 6

DavertMik:
There are lots of chanes due to complete rework of Guzzle with new Psr7 classes in it.

* Files are sent as stream
* Form requests body are formed manually
* Urls use Psr7 classes
* Cookies are moved to middleware

...and other notable changes

Please review this code.


src/Codeception/Lib/InnerBrowser.php
```diff
@@ -516,20 +500,20 @@ public function submitForm($selector, array $params, $button = null)
...
-                $add['path'] = $dir . '/' . $add['path'];
-            }
-        }
-        return array_merge($base, $add);
-    }
+//    private function mergeUrls(array $base, array $add)
+//    {
+//        if (!empty($add['path']) && strpos($add['path'], '/') !== 0 && !empty($base['path'])) {
+//            // if it ends with a slash, relative paths are below it
+//            if (preg_match('~/$~', $base['path'])) {
```
samdark: Should be deleted.

src/Codeception/Lib/Connector/Guzzle.php
```diff
@@ -224,25 +216,53 @@ protected function extractHeaders(BrowserKitRequest $request)
...
+    protected function extractFormData(BrowserKitRequest $request)
     {
-        if (in_array(strtoupper($request->getMethod()), array('GET','HEAD'))) {
+        if (!in_array(strtoupper($request->getMethod()), ['POST', 'PUT', 'PATCH'])) {
             return null;
         }
+
+        // guessing if it is a form data
+        $headers = $request->getServer();
+        if (isset($headers['HTTP_CONTENT_TYPE'])) {
```
Naktibalda: It seems that the content type can be passed as $server['CONTENT_TYPE'].
extractHeaders() expects to get it in that format, extractFormData should do the same thing, right?

Naktibalda:
guzzlehttp/guzzle 6.0.1 requires php >=5.5.0

Are you moving the minimum supported version to 5.5.0 ?

DavertMik:
> guzzlehttp/guzzle 6.0.1 requires php >=5.5.0

Oh, yet another problem :(

Actually this makes sense, as new PHPUnit 5.0 coming in October will also require php 5.5
We can keep Codeception 2.0 on PHP 5.4 and move to PHP 5.5 for 2.1. Major release is a good point to do that.

However lots of users still may use PHP5.4 as well as Guzzle 4-5 and we can think on making them use Codeception 2.1 as well. I have an idea to create `codeception-base` repo where we will mirror Codeception repo, leaving only minimal set of dependencies in `composer.json`. If anyone like to have Codeception running on PHP 5.4 or on Guzzle 5 they can install `codeception-base` as well as `codeception-phpbrowser` package which is PhpBrowser from 2.0 branch. 

But yeah, this sounds like a complex transition plan

DavertMik:
On third thought we could have 2 connectors for Guzzle (depending on version) and not require any Guzzle in `composer.yml` by default. 

Instead of creating `codeception-base` repo we can create `codeception-full` packe which will include Guzzle6 + WebDriver

zbateson:
:+1: so much better.

PR #2027, opened 2015-06-09
Fix docs sample code to use the correct function

thinkspill:
It appears as though the old function was left as a typo. Please reject this commit if I'm mistaken. :)


DavertMik:
No, you are right. Thanks!

PR #2028, opened 2015-06-09
Bug fix for Guzzle Connector headers

valeriyaslovikovskaya:
For now it is impossible to send Content-Type Http request header


zbateson:
I do prefer headers to be in proper-case... why is that a bug fix though, headers should be case-insensitive anyway?

Naktibalda:
It fixes line 216.

Before fix: http://codepad.viper-7.com/3lQ0hY
After fix: http://codepad.viper-7.com/ogvTRG

zbateson:
Right you are, thanks @valeriyaslovikovskaya @Naktibalda - I will move the foreach post-merge

Issue #2025, opened 2015-06-08
Symfony2 and Doctrine2 entity manager issue after first test


AlexStansfield:
Been having an issue using the Doctrine2 and Symfony2 modules together.

It looks like even though it's persisting the entity manager between requests something is going wrong.

Basically the first test runs fine, the Entities I've created using `$I->haveInRepository` are found when we make the symfony request. However the second test (even though in the example I give below is the exact same test) fails to find them.

I know the Doctrine2 module is creating the entities inside the transaction as I `find()` one of them as part of the data setup. All I can think is there is some extra service that needs to be persisted between symfony requests that isn't being so any doctrine requests being done in the following tests are outside of the transaction that the Doctrine2 module has created.

**EDIT:** To save reading through my code examples I created a project to reproduce the issue, see the 3rd comment (https://github.com/Codeception/Codeception/issues/2025#issuecomment-109953529)

My test file is:

```
<?php
use \ApiTester;

class TokenCest
{
    public function _before(ApiTester $I)
    {
        $em = \Codeception\Module\Doctrine2::$em;
        $customerId = $I->haveInRepository(
            'Ubiquity\Bundle\ConnectBundle\Entity\Customer',
            [
                'id' => 'codeception-customer',
                'name' => 'Codeception Test',
                'added' => new \DateTime('2015-01-01'),
                'details' => 'Customer Details'
            ]);
        $customer = $em->find('UbiquityConnectBundle:Customer', $customerId);
        $I->haveInRepository(
            'Ubiquity\Bundle\ConnectBundle\Entity\Client',
            [
                'name' => 'codeception-client',
                'secret' => 'thisissecret',
                'customer' => $customer,
                'allowedGrantTypes' => ['password']
            ]);
        $I->haveInRepository(
            'Ubiquity\Bundle\ConnectBundle\Entity\User',
            [
                'username' => 'codeception-user',
                'email' => 'user@codeception.com',
                'customer' => $customer,
                'salt' => '478c1c486c1a8bc36d888f1f6de25749',
                'password' => 'I8YhzmKq77ewX4aGpt5iUyC96lCAnazvoDQpi0irkxbqfghRLw/4QRDkyEsjoykkD2/Wj4F8XoitJhdWyfXsCA=='
            ]
        );
    }

    public function _after(ApiTester $I)
    {
    }

    // tests
    public function passwordGrantTest(ApiTester $I)
    {
        $I->wantTo('get an access token with a password');
[...]

PR #2042, opened 2015-06-12
Fix doctrine issues after first request is made.

AlexStansfield:
Related to #2025 

I found this change fixed all the problems I was having when using doctrine to create data in a transaction.

Unsure if it's the right solution, but means my tests now work.

### DIFF
--- a/src/Codeception/Lib/Connector/Symfony2.php
+++ b/src/Codeception/Lib/Connector/Symfony2.php
@@ -12,7 +12,7 @@ class Symfony2 extends \Symfony\Component\HttpKernel\Client
         $services = [];
         if (self::$hasPerformedRequest) {
             $services = $this->persistServices();
-            $this->kernel->shutdown();
+            $this->kernel = clone $this->kernel;
         } else {
             self::$hasPerformedRequest = true;
         }


AlexStansfield:
Any chance any one has had a chance to look at this? I don't know if it's the right solution, but it works for us. It would be great to get this bug fixed so we can reliably run our tests without needing to patch codeception.

DavertMik:
If it works for you and tests pass why not to merge then )

AlexStansfield:
Haha ok, I don't claim to know anywhere near enough about the internals of the symfony kernel but if it doesn't break anyone else's tests then I'm happy :grinning:

vyarmolenko:
The solution to clone kernel between requests causes memory leak in functional tests. It can be seen on tests with multiple requests and/or multiple tests in a suite.

I had 'out of memory' errors when ran my test. I changed the code of Lib/Connector/Symfony2.php to its previous version and it works now.

In relation to [https://github.com/Codeception/Codeception/issues/2025]:
In it's current implementation when 'cleanup: true' codeception initializes Doctrine transaction after it creates Symfony2 web client. It may be the cause of problems with the first transaction related routines.
In my test I needed to run different requests against the same db so I implemented a helper method  (mostly from codeception sources). It first rolls back codeception's transaction and initializes its own transaction which is being rolled back between requests:
```php
   /*
     *  Rollback codeception's or any other external transaction(s) and set own  
     */
    public function prepareDatabase($isTransactional = true)
    {
        while ($this->em->getConnection()->isTransactionActive()) {
            try {
                $this->em->getConnection()->rollback();
            } catch (\PDOException $e) {
            }
        }
        $em = $this->em;
        $reflectedEm = new \ReflectionClass($em);
        if ($reflectedEm->hasProperty(' repositories')) {
            $property = $reflectedEm->getProperty('repositories');
            $property->setAccessible(true);
            $property->setValue($em, []);
        }
        $this->em->clear();
        if ($isTransactional) {
            $this->em->getConnection()->beginTransaction();
        }
    }
```

AlexStansfield:
Would changing
```
$this->kernel = clone $this->kernel;
```
to 
```
$tmpKernel = $this->kernel;
$this->kernel = clone $tmpKernel;
unset($tmpKernel);
```
have enabled php to garbage collect when it needs the memory?  

I will revert the change here to see if our tests run successful (and this was fixed else where) or if the initial problem reappears.

AlexStansfield:
Just reread you comment, and will look into what you think might be causing the transaction issue

raistlin:
@vyarmolenko Could you prepare a pull request with your proposed solution?

Issue #2046, opened 2015-06-12
Codeception 2.0.14 - Fatal error: Call to a member function takeScreenshot() on null in phar:///export/web/sb/trudolph/svn/core70/tests/codeception/codecept.phar/src/Codeception/Module/WebDriver.php on line 240


TroyRudolph:
Our tests do not take any screenshots "manually".  This error appeared after running a dozen or so successful tests including the last one. In other words, there were no test failures, just this internal error.

Update:  Just reran the acceptance suite.  The same error occurred, but much earlier in the suite and again, immediately after a passed test.  No other tests failed.

This is a blocking issue for our project to move forward.

I tried to post the full log, but the end of it wasn't visible so I'll just post the end.

<pre>
* I see element "//td[@class='column_Name']//a[starts-with(.,'Standalone Test 1.43412833452E+12')]/../*[normalize-space(.)='T']"
* I click "//a[@id="top_menu_item_logout"]"
* I wait for js "return (typeof $ != 'undefined') && ($.active == 0) && (typeof $.ci != 'undefined') && ($.ci.loadingcount <= 0) && ($.ci.documentReady == 1) && ($.ci.windowReady == 1);",180
* I wait for element not visible "//div[@class='throbber']",180
* I ws set tunable "FORCE_PASSWORD_CHANGE",false
[37;45m PASSED [39;49m
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Fatal error: Call to a member function takeScreenshot() on null in phar:///export/web/sb/trudolph/svn/core70/tests/codeception/codecept.phar/src/Codeception/Module/WebDriver.php on line 240

FATAL ERROR. TESTS NOT FINISHED.
Call to a member function takeScreenshot() on null 
in phar:///export/web/sb/trudolph/svn/core70/tests/codeception/codecept.phar/src/Codeception/Module/WebDriver.php:240
</pre>

TroyRudolph:
More research indicates that this issue is caused by incompatibilities among Codeception, Firefox, and Selenium Standalone Server.  (Due to concurrent testing that we are doing, we use Selenium Standalone Server as a browser manager. )

After playing the permutation game for a while :-(, I got it to work.  

Thank you.

TroyRudolph:
Just as I closed the issue, the tests crashed again with the same error.  

[...]

PR #2048, opened 2015-06-13
Fixed issue 2046

Naktibalda:
Prevent fatal error and log stacktrace when webDriver is not set #2046


DavertMik:
:+1: 

Great that all similar issues were fixed in master. Please merge it (if you don't suppose to add more commits)

PR #2056, opened 2015-06-16
Reconnect option for Db module

Naktibalda:
Codeception 2.1 executes tests in random order.
If percentage of db tests is low and tests are being execute in slow environment, like Travis, database connection often gets lost.

To solve this issue, I implemented reconnect option which creates a new connection before each scenario and drops it after scenario.


DavertMik:
Looks ok to me. You can merge it if you think it is ready

sergeyklay:
:+1:

PR #2058, opened 2015-06-16
Codeception issue #2057 - relative redirect URIs

### LINKED ISSUE #2057: Redirect URIs relative to the previous URI break
According to [RFC 7231](https://tools.ietf.org/html/rfc7231#section-7.1.2), relative redirect URIs (e.g. not beginning with a /) should be "computed by resolving it against the effective request URI."

"The [effective request URI](https://tools.ietf.org/html/rfc7230#section-5.5) is the same as the request URI in most cases." ([source](http://stackoverflow.com/a/25643550))

In other words, if a server returns a `Location` header containing `edit.php`, and the request URI was `/admin/index.php`, the final URI should be `www.domain.com/admin/edit.php`.

WordPress uses these relative redirect URIs in some cases. When submitting the form to create a new user, you submit to `/wp-admin/user-new.php`, which redirects via the header: `Location:users.php?update=add&id=2`, which should send you to `/wp-admin/users.php?update=add&id=2`. With the current code, you get redirected to `/users.php?update=add&id=2`.

firejdl:
Codeception\Lib\Connector\Guzzle::getAbsoluteUri() should compute relative URIs against the effective request URI when there is a redirect.

Fixes #2057.


zbateson:
Excellent, thank you. :+1:

PR #2063, opened 2015-06-18
Add optional `message` parameter to seeInDatabase and dontSeeInDatabase

goulvench:
I was writing a batch of tests for several objects having the same interface and would have needed a way to know which object failed the test. This change does not alter the current behavior. If there's a better way, I'd be happy to know about it! :-)

### DIFF
--- a/src/Codeception/Module/Db.php
+++ b/src/Codeception/Module/Db.php
@@ -270,16 +270,16 @@ class Db extends \Codeception\Module implements \Codeception\Lib\Interfaces\Db
         return $lastInsertId;
     }
 
-    public function seeInDatabase($table, $criteria = [])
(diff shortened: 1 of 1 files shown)


DavertMik:
This could be useful for any assertions in Codeception framework, however we don't provide one. 
We can think on implementing a better way of writing custom messages, but not to change the method signature for that. 

Probably you can use `$I->expect()` methods to write your comment before the assertion

goulvench:
Indeed, but `expect` is not available when running unit tests.

DavertMik:
Sorry, I don't have better ideas. Yeah there is probably a room for improvement, but as I said I'd probably introduce such change globally

PR #2065, opened 2015-06-18
Allow request_timeout_in_ms and connection_timeout_in_ms for WebDriver

n8whnp:
This allows you to pass all possible arguments to php-webdriver's create()
from settings in the suite.yml file.

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -51,6 +51,8 @@ use Codeception\PHPUnit\Constraint\Page as PageConstraint;
  * * clear_cookies - Set to false to keep cookies, or set to true (default) to delete all cookies between tests.
  * * wait - Implicit wait (default 0 seconds).
  * * capabilities - Sets Selenium2 [desired capabilities](http://code.google.com/p/selenium/wiki/DesiredCapabilities). Should be a key-value array.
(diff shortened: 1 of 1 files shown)


DavertMik:
Personally I don't like `_in_ms` suffix. I agree that this should be taken into account, as all other timeouts in configuration are seconds, but I'd rather write that in documentation instead of naming parameter this way

n8whnp:
@DavertMik I am happy to change the parameter name to match conventions within codeception. I see two possible ways forward here:

1) Remove the suffix and just mention the timeout is in miliseconds in the documentation.
2) Remove the suffix and convert the time from seconds to miliseconds in the __initialize() function. 

Which would you prefer going forward?

zbateson:
Hi @n8whnp

Let's keep it in seconds and handle it in ```__initialize``` I think that is cleanest :) then in documentation it can say "default is 30 seconds".

Thanks

n8whnp:
@zbateson thanks for the quick reply, updated to reflect agreed changes.

zbateson:
Excellent, thank you :)

n8whnp:
Given the branch name, will this be going out in a future 2.0 release or is it going out as part of 2.1?

zbateson:
I will leave that for @DavertMik to respond to

DavertMik:
This is going to be included in 2.0.15 which will be released today

PR #2069, opened 2015-06-19
Implemented assertRegExp and assertNotRegExp

RogerRussel:
It was implemented assertRegExp and assertNotRegExp from PHPUnit_Framework_Assert.


DavertMik:
Thanks

PR #2074, opened 2015-06-20
Message cut length fix

Naktibalda:
Modified Message::cut method to cut text at length specified in parameter.


DavertMik:
Thanks!

PR #2076, opened 2015-06-22
Add Sqlsrv database type a primary column function

sitawit:
Sqlsrv uses a different way to get primary column. This update add an override function to use a proper function in case of Sqlsrv database driver.

### DIFF
--- a/src/Codeception/Lib/Driver/SqlSrv.php
+++ b/src/Codeception/Lib/Driver/SqlSrv.php
@@ -76,4 +76,34 @@ class SqlSrv extends Db
         
         $this->sqlQuery($query);
     }
+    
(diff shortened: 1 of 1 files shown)


DavertMik:
Thanks, I think your PR may go to 2.0 as it is minor enhancement. Please resend

sitawit:
@DavertMik Ok, I'll submit to 2.0. So I need to close this request and create another for 2.0 right? (I've never done it before)

DavertMik:
Send to 2.0. I can close this any time )

sitawit:
It's done. I created a new pull here https://github.com/Codeception/Codeception/pull/2107.

PR #2089, opened 2015-06-24
Implemented assertFileExists and assertFileNotExists.

RogerRussel:
Implemented assertFileExists and assertFileNotExists, as did their tests.
I also put some more tests other asserts that were not being tested.


DavertMik:
Thanks!

PR #2101, opened 2015-06-27
Fix output cept file path

junichi11:
Change to the same format as the others(test, phpunit, cest) because there is a difference in output when we run generate commands.

- test, phpunit, cest : full file path
- cept : not full file path

e.g.
```
$ codecept generate:test suite Foo
Test was created in /home/junichi/Projects/myproject/tests/suite/FooTest.php
$ codecept generate:cept suite Foo
Test was created in FooCept.php
```


junichi11:
It seems that this problem is not occurred in version 2.1.0. So close this PR.

PR #2104, opened 2015-06-28
Fixed PhpBrowser::setCookie() method

enumag:
The cookies were later [removed by guzzle](https://github.com/guzzle/guzzle/blob/9c43893a8e580d8401e4210cebe038d850341012/src/Cookie/CookieJar.php#L217) because the domain did not match.

### DIFF
--- a/src/Codeception/Lib/Connector/Guzzle6.php
+++ b/src/Codeception/Lib/Connector/Guzzle6.php
@@ -172,7 +172,7 @@ class Guzzle6 extends Client
         );
 
         $options = $this->requestOptions;
-        $options['cookies'] = $this->extractCookies();
+        $options['cookies'] = $this->extractCookies($guzzleRequest->getUri()->getHost());
         $multipartData = $this->extractMultipartFormData($request);
         if (!empty($multipartData)) {
             $options['multipart'] = $multipartData;
@@ -304,7 +304,7 @@ class Guzzle6 extends Client
         return $files;
     }
     
-    protected function extractCookies()
+    protected function extractCookies($host)
     {
         $jar = [];
         $cookies = $this->getCookieJar()->all();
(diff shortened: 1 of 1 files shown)


DavertMik:
Yeah, thanks. Didn't have tests to match this case

enumag:
@DavertMik Can you add some test for it then? I would do it myself but I didn't manage to run codeception tests without a ton of fails and I have no idea what's causing them.

DavertMik:
Unfortunately I don't have idea how to test other hosts than localhost locally. Do you have ideas?

As for running tests: you should not run all them locally. Here is a small guide to run them https://github.com/Codeception/Codeception/tree/2.0/tests#testing-a-module

enumag:
I use [Acrylic DNS Proxy](http://mayakron.altervista.org/wikibase/show.php?id=AcrylicHome) to host all *.l domains on localhost. That's because I have lots of projects so I can't run them all on a single domain called localhost.

I understand that writing a functional or acceptation test for this would be difficult. Maybe we could add a unit test for it then?

And for running tests - why should I not run them locally? I mean when I do some refactoring I need to run all tests to I know I didn't break anything.

DavertMik:
Well, you can run all them locally, but you need lots of services to be started like selenium, mysql, mongodb. 
As you see I tried to describe everything in a guide I linked above. 

If you want to execute all tests you may test them via Travis. Just add your fork to TravisCI and watch for them passing. Also from Travis config you can see why it is so hard to running them locally. So I'd prefer to test only particular things and everything else push to CI. 

> I understand that writing a functional or acceptation test for this would be difficult. Maybe we could add a unit test for it then?

Ok, this can work

enumag:
Thanks for the tips. :-)

Maybe tests that are missing some requirements should be skipped by default?

DavertMik:
they do so. However, some of them like `web` tests just need webdriver. Also running tests locally may take a long time

DavertMik:
Finally added a test here.

enumag:
:+1:

Issue #766, opened 2013-12-22
WebDriver setCookie() do not set cookie


lordlele:
Hi, maybe related with #765 but I have see a strange behavior:

In my **index.php** i have:

```
<?php
print_r($_COOKIE);
```

In a Cept I write:

```
<?php
$I = new WebGuy($scenario);
$I->wantTo('test application start');
$I->setCookie("MyCookie", "My Cookie content");
$I->amOnPage("index.php");
$I->see("My Cookie content");
```

Test results with **PhpDriver**:

```
Acceptance Tests (1) ---------------------------------------------
Trying to test application start (ApplicationCept.php)       Ok
------------------------------------------------------------------
```

Test results with **WebDriver** or **Selenium2**:

```
Acceptance Tests (1) ---------------------------------------------
Trying to test application start (ApplicationCept.php)       Fail
------------------------------------------------------------------


Time: 33.89 seconds, Memory: 8.75Mb

There was 1 failure:

---------
1) Failed to test application start in ApplicationCept.php
Sorry, I couldn't see "My Cookie content":
Failed asserting that 
-->  /index.php<html xmlns="http://www.w3.org/1999/xhtml"><head></head><body>Array
(
)
</body></html>
--> contains "my cookie content".

Scenario Steps:
3. I see "My Cookie content"
2. I am on page "index.php"
1. I set cookie "MyCookie","My Cookie content"


FAILURES!
Tests: 3, Assertions: 3, Failures: 1.
```

In Selenium logs, I have this row:

```
21:00:44.211 INFO - Executing: [add cookie: MyCookie=My Cookie content; path=/] at URL: /session/8533f595-7e93-4c04-ba52-35085cb88f17/cookie)
```

So apparently cookie was set… however I can't get out of this headache.

enumag:
@DavertMik Does that mean that there is no way to set cookie for the first visited page? The setCookie call after amOnPage works only on following amOnPage calls, the first is uneffected. So when I want to write test with a cookie I have to do this:

```php
<?php
$I = new AcceptanceTester($scenario);
$I->wantTo('see page');
$I->amOnPage('/blank.html'); // I'd like to get rid of this line but it doesn't work without it.
$I->setCookie('key', 'Cookie value');
$I->amOnPage('/cookie-value-as-title');
$I->seeInTitle('Cookie value');
```

PR #2105, opened 2015-06-28
Fixed handling multiple cookies

enumag:
My god, which idiot wrote the Guzzle6 class???

The flattenHeaders method completely broke my tests because it threw away every cookie except the first and since PHPSESSID was the second cookie in response, session did not work at all.

Then there is #2104 because of which setCookie did not work at all so I couldn't set PHPSESSID manually either. The bug is again in Guzzle6 class.

Oh and by the way there are problems with cookies even when using WebDriver instead of PhpBrowser. I was not able to fix that. https://github.com/Codeception/Codeception/issues/766#issuecomment-115005827

### DIFF
--- a/src/Codeception/Lib/Connector/Guzzle6.php
+++ b/src/Codeception/Lib/Connector/Guzzle6.php
@@ -94,17 +94,17 @@ class Guzzle6 extends Client
     protected function createResponse(Psr7Response $response)
     {
         $body = (string) $response->getBody();
-        $headers = $this->flattenHeaders($response->getHeaders());
+        $headers = $response->getHeaders();
 
         $contentType = null;
-        if (isset($headers['Content-Type'])) {
-            $contentType = $headers['Content-Type'];
-        }
-        if (!$contentType or strpos($contentType, 'charset=') === false) {
-            if (preg_match('/\<meta[^\>]+charset *= *["\']?([a-zA-Z\-0-9]+)/i', $body, $matches)) {
-                $contentType .= ';charset=' . $matches[1];
+        if (isset($headers['Content-Type']) && count($headers['Content-Type']) === 1) {
+            $contentType = reset($headers['Content-Type']);
+            if (strpos($contentType, 'charset=') === false) {
+                if (preg_match('/\<meta[^\>]+charset *= *["\']?([a-zA-Z\-0-9]+)/i', $body, $matches)) {
+                    $contentType .= ';charset=' . $matches[1];
+                }
+                $headers['Content-Type'] = $contentType;
             }
-            $headers['Content-Type'] = $contentType;
         }
 
         $status = $response->getStatusCode();
@@ -120,7 +120,7 @@ class Guzzle6 extends Client
             // match by header
             preg_match(
                 '~(\d*);?url=(.*)~',
-                (string) $headers['Refresh'],
+                (string) reset($headers['Refresh']),
                 $matches
             );
         }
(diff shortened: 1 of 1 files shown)


src/Codeception/Lib/Connector/Guzzle6.php
```diff
@@ -94,17 +94,17 @@ public function setAuth($username, $password)
...
+        if (isset($headers['Content-Type']) && count($headers['Content-Type']) === 1) {
+            $contentType = reset($headers['Content-Type']);
+            if (strpos($contentType, 'charset=') === false) {
+                if (preg_match('/\<meta[^\>]+charset *= *["\']?([a-zA-Z\-0-9]+)/i', $body, $matches)) {
+                    $contentType .= ';charset=' . $matches[1];
+                }
+                $headers['Content-Type'] = $contentType;
             }
-            $headers['Content-Type'] = $contentType;
         }
```
enumag: I thried to fix the magic above to the best of my ability but I actually think it should be removed completely.

src/Codeception/Lib/Connector/Guzzle6.php
```diff
@@ -120,7 +120,7 @@ protected function createResponse(Psr7Response $response)
             // match by header
             preg_match(
                 '~(\d*);?url=(.*)~',
-                (string) $headers['Refresh'],
+                (string) reset($headers['Refresh']),
```
enumag: I just hope this will work, did not test it.

src/Codeception/Lib/Connector/Guzzle6.php
```diff
@@ -94,17 +94,17 @@ public function setAuth($username, $password)
...
-        if (!$contentType or strpos($contentType, 'charset=') === false) {
-            if (preg_match('/\<meta[^\>]+charset *= *["\']?([a-zA-Z\-0-9]+)/i', $body, $matches)) {
-                $contentType .= ';charset=' . $matches[1];
+        if (isset($headers['Content-Type']) && count($headers['Content-Type']) === 1) {
+            $contentType = reset($headers['Content-Type']);
+            if (strpos($contentType, 'charset=') === false) {
+                if (preg_match('/\<meta[^\>]+charset *= *["\']?([a-zA-Z\-0-9]+)/i', $body, $matches)) {
+                    $contentType .= ';charset=' . $matches[1];
+                }
+                $headers['Content-Type'] = $contentType;
```
enumag: I thried to fix the magic above to the best of my ability but I actually think it should be removed completely.

Ragazzo:
> My god, which idiot wrote the Guzzle6 class???

yeah, they following their own vision of standarts in php-fig ;D

DavertMik:
> My god, which idiot wrote the Guzzle6 class???

It's not hard to guess, especially as we use Git. And well, you know, that was me )
Ok, will try to work on it. Anyway, it is just RC now, so we could test such issues before the release

enumag:
@DavertMik Sorry I was rude, I was just pretty upset after many hours of debuging this thing. Let me know if I can help further.

DavertMik:
I'm writing tests for this cases...

DavertMik:
@enumag Thanks for your time and good review of my connector. Guzzle 6 introduced lots of major changes, so that was hard to write connector from scratch and make it work for all possible cases. At least now we know the issue, we can write tests, fix it, and never see again.

DavertMik:
Accroding to Travis your fixes didn't break anything. I added some tests, so I'm getting it merged.

PR #2107, opened 2015-06-29
Update SqlSrv.php for a primary column function

sitawit:
SqlSrv uses a proper function to get primary column.


sitawit:
This pull failed during the checks. However, it fails due to the composer. Not about the code itself. So what should I do?

DavertMik:
yeah, those CI issues are depressing. Nevermind, let me restart build and merge everything when it goes green

PR #2112, opened 2015-06-30
added HTTPS to server parameters only when it required

onedevlink:
https://github.com/symfony/BrowserKit/blob/master/Client.php#L557 symfony check if HTTPS is isset. Not true or false.

Also, may be make check if HTTP_HOST is not empty? For backward compatibility in this config case:
```
config:
    # the entry script URL (without host info) for functional and acceptance tests
    # PLEASE ADJUST IT TO THE ACTUAL ENTRY SCRIPT URL
    test_entry_url: /frontend/web/index-test.php
```

### DIFF
--- a/src/Codeception/Module/Yii2.php
+++ b/src/Codeception/Module/Yii2.php
@@ -76,13 +76,11 @@ class Yii2 extends Framework implements ActiveRecord, PartedModule
                     PHP_URL_HOST
                 )
             );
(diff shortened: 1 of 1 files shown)


Naktibalda:
Was there a good reason not to merge this pull request?

Naktibalda:
Please make a new PR if this change is still necessary.

PR #2115, opened 2015-07-01
Fixed error in \Codeception\Extension::getSubscribedEvents

yurijbogdanov:
Fixed error in \Codeception\Extension::getSubscribedEvents


DavertMik:
thanks

PR #2132, opened 2015-07-07
Fix for fillField using values that contain ampersands

GawainLynch:
Text fields, such as passwords or fields that contain free text, can use
ampersands and DOMNode property textContent should be set instead of
nodeValue to prevent "unterminated entity reference" errors

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -290,17 +290,17 @@ class InnerBrowser extends Module implements Web, PageSourceSaver
         $crawler = $this->getFieldsByLabelOrCss($field);
         $this->assertNot($this->proceedSeeInField($crawler, $value));
     }
-    
+
     public function seeInFormFields($formSelector, array $params)
     {
         $this->proceedSeeInFormFields($formSelector, $params, false);
     }
-    
+
     public function dontSeeInFormFields($formSelector, array $params)
     {
         $this->proceedSeeInFormFields($formSelector, $params, true);
     }
-    
+
     protected function proceedSeeInFormFields($formSelector, array $params, $assertNot)
     {
         $form = $this->match($formSelector)->first();
@@ -332,7 +332,7 @@ class InnerBrowser extends Module implements Web, PageSourceSaver
     /**
      * Returns an array of values for the field with the passed name.  Usually
      * the array consists of a single value.  Used by proceedSeeInField
-     * 
+     *
      * @param Form $form
      * @param string $fieldName
      * @return array
@@ -357,7 +357,7 @@ class InnerBrowser extends Module implements Web, PageSourceSaver
         }
         return $values;
     }
-    
+
     protected function proceedSeeInField(Crawler $fields, $value)
     {
         $form = $this->getFormFor($fields);
@@ -508,7 +508,7 @@ class InnerBrowser extends Module implements Web, PageSourceSaver
 
     /**
      * Returns the form action's absolute URL.
-     * 
+     *
      * @param \Symfony\Component\DomCrawler\Crawler $form
      * @return string
      * @throws \Codeception\Exception\TestRuntimeException if either the current
@@ -563,7 +563,7 @@ class InnerBrowser extends Module implements Web, PageSourceSaver
     /**
      * Returns the DomCrawler\Form object for the form pointed to by
      * $node or its closes form parent.
-     * 
+     *
      * @param \Symfony\Component\DomCrawler\Crawler $node
      * @return \Symfony\Component\DomCrawler\Form
      */
@@ -583,7 +583,7 @@ class InnerBrowser extends Module implements Web, PageSourceSaver
         }
         return $this->forms[$action];
     }
-    
+
     /**
      * Returns an array of name => value pairs for the passed form.
      *
@@ -613,13 +613,13 @@ class InnerBrowser extends Module implements Web, PageSourceSaver
         $input = $this->getFieldByLabelOrCss($field);
         $form = $this->getFormFor($input);
         $name = $input->attr('name');
-        
+
         $dynamicField = $input->getNode(0)->tagName == 'textarea'
             ? new TextareaFormField($input->getNode(0))
             : new InputFormField($input->getNode(0));
         $formField = $this->matchFormField($name, $form, $dynamicField);
         $formField->setValue($value);
-        $input->getNode(0)->nodeValue = $value;
+        $input->getNode(0)->nodeValue = htmlspecialchars($value);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


GawainLynch:
Any clues as to why this would pass on 5.6, but fail on 5.4 & 5.5?

zbateson:
My best guess would be that prior to 5.6 setting textContent doesn't update the value of the node.  Does your test (and the other tests) pass if you have both lines? Like this:

```php
$input->getNode(0)->nodeValue = $value;
$input->getNode(0)->textContent = $value;
```

It may need both - if it does, please add a comment as well explaining why :) the reason why it errors is related to how nodes are selected for 'array' nodes (like nodes ending in [])... the node in the DOM and the node in the form have to be identical or the node can't be found in matchFormField.

GawainLynch:
Ah… Derp!

OK, `nodeValue` will always fail for a string that contains an `&` with an "unterminated entity reference" that hit us very near to 2.1.0 after 3dcf0bda6ccb4c4eb13eb8a75bd1838caef08ad9 was merged.

In turn I have made the same mistake and assumed too much, let update the PR with an array check… and a comment ;-)

GawainLynch:
I have gone for the simplest approach, and am happier to hear dissent :grin: 

Basically, if I understand correctly, the arrays appear for checkboxes… and are numeric. So, I have used a try/catch that defaults to `->nodeValue` and then falls back to `->textContent` if there is an `\Exception`.

zbateson:
Ooh, I see... that seems too complicated - I think you should be able to replace that with a single line:

```php
$input->getNode(0)->nodeValue = htmlspecialchars($value);
```

:smile:

GawainLynch:
Nope, that just cause 7,852 errors instead of the 13 I was trying to resolve.

GawainLynch:
Thing is, the escaping happens later in DOM object's life time… this is what I was battling with as all we want to do is handle the raw text that's needs to be entered into the field. 

Again, I am troubled myself with my own approach.

zbateson:
You mean 7,852 errors in your application?  I just tried it and the test case in your PR and all the Codeception PhpBrowser/Framework tests are passing for me if I change it to htmlspecialchars - is it a test case we're missing somewhere?  What's the error that's happening for you?

Although your code works there is one scenario it's missing - you see it's not just checkboxes that could be arrays, it can be any field.  So it's possible someone would call fillField on an "array" field, and with an ampersand in its value... which would have the same problem as your initial commit.

GawainLynch:
> You mean 7,852 errors in your application?

Yes

> Although your code works there is one scenario it's missing

Understood… It's just not what I have a lot of visibility over and would love help… We've not been able to run Codeception after 2.0.12 due to various (and continually differing per version) BC breaks that have appeared. We were about to move to 2.1, as dev-master was solid for a while until just before release… So I am trying to dig in and get it sorted :grin: 

> it's not just checkboxes that could be arrays

Absolutely understand. In our case it is password fields (many of them, this one can be worked around), and many YAML input fields that *very* validly contain `&`.

zbateson:
Yeah, I know what you mean - to be honest much of these BC breaks are my fault... my hope is over time things will be a lot better though.  As it is a lot of the class is still much messier than I'd like it to be (which is making our lives difficult) so apologies and thanks for the effort! :)

What does it error on if you don't mind me asking though?  I'm wondering if the error is also something that should be fixed/accounted for since it's not something currently covered by our available tests.

GawainLynch:
> Yeah, I know what you mean - to be honest much of these BC breaks are my fault

No stress… Over on the [Bolt](https://github.com/bolt/bolt) project we just blame the :koala: for these things. :grin: 

I am not sure what started breaking from 2.0.13 onwards as we just immediately locked onto 2.0.12 to keep Travis builds working. But I was tracking the dev-master branch as was preparing to cut us over to 2.1.0 upon release, but as I mentioned #2077 re-introduced problems.

> What does it error on if you don't mind me asking though? 

The first failure I found was with the UI creation and logging on of the "[Developer User](https://github.com/bolt/bolt/blob/master/tests/codeception/acceptance/_bootstrap.php#L31-L36)" 
 as it has a complex password that contains an `&` and then throws several errors:
```                                                                                                                     
[PHPUnit_Framework_Exception] Codeception\Lib\InnerBrowser::fillField(): unterminated entity reference          nGu3$$  
```

Obviously that can be worked around, and when I did, similar error messages also occured during the acceptance tests for our backend UI, as one of the features of Bolt is the ability to edit your YAML configuration files, and Twig templates in a UI editor that has input in a `textarea` field

On [this test](https://github.com/bolt/bolt/blob/master/tests/codeception/acceptance/102-BackendDeveloper/BackendDeveloperCest.php#L119-L145) the test is to modify [this line of Twig](https://github.com/bolt/bolt/blob/master/theme/base-2014/_footer.twig#L6) (text) that also contains an ampersand.

Another failure, which I intend on updating our test code for as I am not sure we were Doing it Right&trade; or was/is related, is that `$I->amOnPage('bolt');` in 2.0.x would assume that the URL was relative to base, i.e. `/bolt` so I will PR something to our project soon.

All can be witnessed with a clone of the repo, a composer install and then we have a script at `./tests/scripts/run-codeception-tests` if you want to have a look directly.

zbateson:
Hi @GawainLynch -

Installed and ran your project with "codeception/codeception" : "dev-master@dev" in composer.json, then ran composer install.

With ``` $input->getNode(0)->nodeValue = $value; ``` I get:

Tests: 53, Assertions: 153, Errors: 10, Failures: 2.

With ``` $input->getNode(0)->nodeValue = htmlspecialchars($value); ``` I get:

OK (53 tests, 193 assertions)

So it's all looking good to me with my proposed change :) am I missing something?

GawainLynch:
At a guess you are missing a :koala:

I'm out at the moment, I'll be home in a few hours and test/push something. Thanks for  the collaboration.

zbateson:
No worries :)

src/Codeception/Lib/InnerBrowser.php
```diff
@@ -613,13 +613,19 @@ public function fillField($field, $value)
...
         $dynamicField = $input->getNode(0)->tagName == 'textarea'
             ? new TextareaFormField($input->getNode(0))
             : new InputFormField($input->getNode(0));
         $formField = $this->matchFormField($name, $form, $dynamicField);
         $formField->setValue($value);
-        $input->getNode(0)->nodeValue = $value;
+
+        // If the value contains an ampersand, assigning nodeValue will fail
+        try {
+            $input->getNode(0)->nodeValue = $value;
```
DavertMik: why not just check that ampersand is there with `strpos`?

zbateson:
I think the correct solution will be to call htmlspecialchars on $value before assigning it to nodeValue - I'm pretty sure that works which avoids having an exception be thrown and having to check for it, and may also work for other characters that need to be escaped also :) just waiting to hear back.

GawainLynch:
OK, updated and yes, it passes here… it still throws a whole lot of errors, but what I missed initially is that they are coming from Twig and not looking to be a problem with Codeception directly.

This is good to go from my end.

zbateson:
Excellent, thanks @GawainLynch

PR #2133, opened 2015-07-07
Corected all instances of out to our in documentation

grantlucas:
Corrected "Do you agree to **out** terms?" to "Do you agree to **our** terms?" where ever it showed up.

I wasn't 100% sure if some of these files were auto-generated, and if so, how to regenerate them. Figured I'd submit all the fixes for now.


DavertMik:
Thanks! All of those files except src/Codeception/Lib/Interfaces/Web.php are autogenerated. Good that you didn't miss that one. Thanks!

Issue #2120, opened 2015-07-02
Update facebook/webdriver to version 1.0.0


execut:
I need new version, but in codeception used version 0.5 and change it i broke codeception =( bad. How about migrate to new version?

PR #2140, opened 2015-07-09
Updated to use facebook/webdriver 1.0

DavertMik:


javigomez:
closes #2120

DavertMik:
they have changed naming of every class. This may look like a BC break :(

javigomez:
ouch!

Here is the compare view: 

https://github.com/facebook/php-webdriver/compare/v0.5...1.0.0

DavertMik:
@zbateson could you look into it? I implemented a bit more than just updating webdriver. Actually I added few new methods to WebDriver and PhpBrowser to make them simpler to use in helpers

composer.json
```diff
@@ -19,7 +19,7 @@
         "ext-json": "*",
         "ext-mbstring": "*",
         "phpunit/phpunit": "~4.7.0",
-        "facebook/webdriver": "~0.4|~0.5",
+        "facebook/webdriver": "~0.4|~0.5|~1.0",
         "guzzlehttp/guzzle": ">=4.0|<7.0",
```
DavertMik: oh, probably this should be changed to `~1.0` only

composer.json
```diff
@@ -19,7 +19,7 @@
         "ext-json": "*",
         "ext-mbstring": "*",
         "phpunit/phpunit": "~4.7.0",
-        "facebook/webdriver": "~0.4|~0.5",
+        "facebook/webdriver": "~0.4|~0.5|~1.0",
```
DavertMik: oh, probably this should be changed to `~1.0` only

Issue #2117, opened 2015-07-01
Absolute url problem with REST


tmp77:
There is a problem with hosts using absolute url
$I->sendGET('http://example.com/123/');

```modules:
    enabled: [PhpBrowser, REST]
    config:
        PhpBrowser:
            url: http://example.local
        REST:
            depends: PhpBrowser
            url: http://example.local
```

Real request is sent to http://example.local/123/. See Util\Uri::mergeUrls()

withremote:
I'm having a similar issue with Acceptance tests. Tests with absolute urls that passed in 2.0.15 return 404s in 2.1. Mine are all https urls if that would make a difference.

DavertMik:
@withremote please could you be more specific. I was trying to reproduce something like this in acceptance tests, but I didn't see any issues in acceptance tests.

@tmp77 test for your issue was written and it was fixed

PR #2141, opened 2015-07-09
Missing HTTPS when trying to access protected pages

andreytkachenko-home24:
So, in config main site's URL is http://google.com/
Some of my pages are https://
But trying to access them, I get errors.
When dig into, found that all new URLs are build based on base_uri, adding path, query, if any.
Adding also scheme solved the issue.

Actual:
Uri::mergeUrls('http://google.com/';, 'https://google.com/account/')  === 'http://google.com/account/';

Expecting:
Uri::mergeUrls('http://google.com/';, 'https://google.com/account/')  === 'https://google.com/account/';

### DIFF
--- a/src/Codeception/Util/Uri.php
+++ b/src/Codeception/Util/Uri.php
@@ -43,6 +43,9 @@ class Uri
             }
             $base = $base->withPath($path);
         }
+        if (isset($parts['scheme'])) {
+            $base = $base->withScheme($parts['scheme']);
+        }
         if (isset($parts['query'])) {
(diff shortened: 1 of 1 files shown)


zbateson:
Looks good, thank you

Naktibalda:
This change was not necessary, because this issue was fixed by an earlier fix to issue #2117 

https://github.com/Codeception/Codeception/commit/105af4e05d1d25b987c1cf17437d68ea110896ea#diff-081a985cd797805b8f44b78e3d71e6faR26

PR #2142, opened 2015-07-09
grabNumberOfElement Action added to PhpBrowser and WebDriver to grab …

valeriyaslovikovskaya:
…number of elements matched by the given locator on the page

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -1088,6 +1088,11 @@ class InnerBrowser extends Module implements Web
         }
     }    
 
+    public function grabNumberOfElements($selector)
+    {
+        return count($this->match($selector));
+    }
+
     public function seeOptionIsSelected($selector, $optionText)
     {
         $selected = $this->matchSelectedOption($selector);
--- a/src/Codeception/Lib/Interfaces/Web.php
+++ b/src/Codeception/Lib/Interfaces/Web.php
@@ -750,6 +750,20 @@ interface Web
     public function seeNumberOfElements($selector, $expected);    
     
     /**
+     * Grab number of elements matched by the given locator on the page.
+     *
+     * ``` php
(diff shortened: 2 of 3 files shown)


DavertMik:
Could you provide a use-case for this?

valeriyaslovikovskaya:
Web application I'm testing displays several items from the product catalog that fit certain criteria. 
I never know exactly how many products user will see on the page: at least one, at most 12, depends
on the result of catalog request query.  
User proposed to choose one of them (clicking on the item image or selecting radio button).
I want to:
1. grab the number of elements from the page N
2. choose one element randomly (rand(1, N))
3. click it
4. assert that total sum of order increased with sum of product, that order description contains selected product
5. choose another product on the page (steps 1-3)
6. check that order summary changed accordingly

And for this test I need the possibility to access any element on on the page with
"my_test_marker_class:nth-of-type(n)" selector, where 0 < n < N, so
I need to grab N.

DavertMik:
I'd not recommend you to do anything randomly, especially in acceptance tests. They are not as stable in any use, and any random things makes them even more fragile. If such test fails how would you reproduce it? 

But I think we can have a compromise here :) Let's say you will be available to call `_findElements` function from your Helper class, like this:

```php
namespace Helper;
class Acceptance {
     function chooseElements($locator)
     {
           $els = $this->getModule('WebDriver')->_findElements($locator);
           // ....
     }
} 
```

This feature will land in Codeception 2.1.1. For now you can locate elements using WebDriver API:

```
namespace Helper;
class Acceptance {
     function chooseElements($xpath)
     {
           $els = $this->getModule('WebDriver')->webDriver->findElements(WebDriverBy::xpath($xpath));
           // ....
     }
} 
```

Anyway, I think you have a very specific case so you should use Helper class to implemnt such logic.

valeriyaslovikovskaya:
Ok. Thanks for comment.

PR #2148, opened 2015-07-10
Cleanup should happen only if enabled

gammamatrix:
The `Db::_initialize()` method calls `$this->cleanup()` regardless of the value of `$this->config['cleanup']`.

I added a check to see if `cleanup` is set to `true`.

### DIFF
--- a/src/Codeception/Module/Db.php
+++ b/src/Codeception/Module/Db.php
@@ -143,7 +143,9 @@ class Db extends CodeceptionModule implements DbInterface
 
         // starting with loading dump
         if ($this->config['populate']) {
-            $this->cleanup();
+            if ($this->config['cleanup']) {
+                $this->cleanup();
+            }
             $this->loadDump();
             $this->populated = true;
         }


DavertMik:
Looks reasonable, however this PR should be sent to 2.0 (which is patch branch) or 2.1 (which is current stable). `master` is kept for edge development

gammamatrix:
@DavertMik I sent 2 pull requests for the 2.0 and 2.1 branches.

https://github.com/Codeception/Codeception/pull/2152 is for 2.1

https://github.com/Codeception/Codeception/pull/2153 is for 2.0

DavertMik:
this can be ignore as master branch will be merged with 2.1

blasic:
I think this is not a good idea. 

I'm running a scenario where I'm not cleaning after every test. My dump contains schema creation statemets, wich I want to run before suite starts. Cleanup before load ensures, that load is successfull. 

With introduction of this change load fails if cleanup is set to false. Making the combination of populate=true and cleanup = false unusable.

I could include drop statements in the dump, but that breaks starting from an empty database.

 Please revert this change, It's not introducing anything new just breaking things.

PR #2153, opened 2015-07-12
Cleanup should happen only if enabled

gammamatrix:
Issuing a request for the 2.0 branch.

This is a follow up for https://github.com/Codeception/Codeception/pull/2148


DavertMik:
Thanks

Issue #1245, opened 2014-07-25
Module DB parsing issue?


Salvner:
My code requires some sql to be on multiple lines as having them on the same line results in errors due to weird reasons, this works fine when run as a query through sql, but is counted as a syntax error when run through codeceptions db, my current workaround is to use HaveInDataBase but thats a bit annoying as the form is different to that of my database dumps.
Is there something I could alter to change this?

gammamatrix:
If you have your semicolon delimiter on a separate line, you will get a syntax error if there are subsequent queries in your dump.sql file.

For example, this fails:

```sql
REPLACE INTO `Contacts` (`created`, `modified`, `status`, `contact`, `first`, `last`) VALUES
(NOW(), NOW(), 1, 'Bob Ross', 'Bob', 'Ross'),
(NOW(), NOW(), 1, 'Fred Flintstone', 'Fred', 'Flintstone')
;

DELETE FROM `Order`;
```

Moving the semicolon to the last data row does not fail:

```sql
REPLACE INTO `Contacts` (`created`, `modified`, `status`, `contact`, `first`, `last`) VALUES
(NOW(), NOW(), 1, 'Bob Ross', 'Bob', 'Ross'),
(NOW(), NOW(), 1, 'Fred Flintstone', 'Fred', 'Flintstone');

DELETE FROM `Order`;
```

Is this the same issue you had @Salvner ?

PR #2154, opened 2015-07-12
Documenting format of dump.sql

gammamatrix:
Updated docs to show how to use data dump file. See issue https://github.com/Codeception/Codeception/issues/1245

NOTE: This section should also note that if `cleanup` is set to `true`, then the entire schema could/should be placed in this file.

Likewise, if `cleanup` is set to `false`, then the schema is not required.

The pull request https://github.com/Codeception/Codeception/pull/2148 affects this configuration.

### DIFF
--- a/docs/modules/Db.md
+++ b/docs/modules/Db.md
@@ -65,6 +65,22 @@ Check out drivers if you get problems loading dumps and cleaning databases.
              cleanup: false
              reconnect: true
 
+### SQL data dump
+
+ * Comments are permitted.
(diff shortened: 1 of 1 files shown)


DavertMik:
Thank you, this looks good, however the documentation of modules are generated from their docblock comments. Could you updated source of Db module instead?

gammamatrix:
@DavertMik no problem :)

gammamatrix:
@DavertMik There were extra spaces added to the file on those first two patches.

PR #2159, opened 2015-07-14
Documenting format of dump.sql

gammamatrix:
Moving the documentation to the source so it can be generated.

See pull request https://github.com/Codeception/Codeception/pull/2154


DavertMik:
Thanks!

Issue #2052, opened 2015-06-16
CheckOptions method to check multiple options identified by the selector


pkeatingcrossflow:
While putting together a performance test, I wanted there to be a method that checks options by class name. I created this method in InnerBrowser.php, which works for our particular case. I thought it might provide a base for a future addition. I make no claims for its quality. It is a basic hack. Let me know whether you think this would be a useful addition.

``` php
    /**
    * Checks all nodes matched by the selector. 
    *
    * @param $selector
    * @return void
    */
    public function checkOptions($selector)
    {
        $nodes = $this->match($selector);

        foreach($nodes as $nodeIndex => $node){
            if($node->hasAttribute('id')){
                $nodeId = $node->getAttribute('id');
                $this->checkOption('#'.$nodeId);
            }
            elseif($node->hasAttribute('name')){
                $nodeName = $node->getAttribute('name');
                $this->checkOption($nodeName);
            }       
        }
    }
```

zbateson:
Hmmm... does ```$I->checkOption('.className'); ``` not work for some reason?

pkeatingcrossflow:
When I ran the test in debug mode using checkOption, the POST data only contained one item that had been checked. The ChoiceFormField returned by proceedCheckOption appears to match a single form field, i.e. the first one. This item is then ticked.

```php
     
    public function checkOption($option)
    {
        $this->proceedCheckOption($option)->tick();
    }

     /**
     * @param $option
     * @return ChoiceFormField
     */
    protected function proceedCheckOption($option)
    {
        $form = $this->getFormFor($field = $this->getFieldByLabelOrCss($option));
        $name = $field->attr('name');

        if ($field->getNode(0) === null) {
            throw new TestRuntime("Form field $name is not located");
        }
        // If the name is an array than we compare objects to find right checkbox
        $formField = $this->matchFormField($name, $form, new ChoiceFormField($field->getNode(0)));
        if (!$formField instanceof ChoiceFormField) {
            throw new TestRuntime("Form field $name is not a checkable");
        }
        return $formField;
    }

```

zbateson:
Aaah, I see what you're saying.  I'll consider improvements to proceedCheckOption on that basis... perhaps instead of simply returning one, it could return and check "all" options returned by the selector.  I'm not sure I like the idea of introducing another function for that at this point simply because the class is getting a bit overloaded and I'm hoping to simplify.

[...]

PR #2166, opened 2015-07-15
Issue #2052, changed checkOption method to check all options

pkeatingcrossflow:
This is a suggested enhancement for PhpBrowser's checkOption method, which retrieves all nodes matched by the $option parameter and then ticks each one individually. The checkOption method currently finds the first checkbox matched by $option and then ticks it. This method is intended to check all checkboxes matched by $option.

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -727,9 +727,26 @@ class InnerBrowser extends Module implements Web, PageSourceSaver, ElementLocato
         return $option;
     }
 
+    /**
+    * Checks all options matched by the variable $option
+    * @param $option
+    */
+
     public function checkOption($option)
     {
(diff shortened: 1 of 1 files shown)


DavertMik:
@zbateson are we ok with it?

DavertMik:
Build is failing. Please try to run PhpBrowser or Framework tests locally and try to fix them

zbateson:
Hi @pkeatingcrossflow 

Thanks for the contribution - yes, it makes things much easier if you test locally first.  There's documentation in tests/README.md explaining how.

I think the problem you're having is you're using match() instead of getFieldByLabel or CSS.  Also, I would recommend modifying 'proceedCheckOption' itself instead of in checkOption.  This will make it work the same for both checkOption and uncheckOption.

Naktibalda:
I am closing this pull request because issues weren't fixed.

Issue #2169, opened 2015-07-15
Class not found when generating Test tests


roland-jungwirth:
### Codeception 2.1.1
After generating a test with the command
```
php codecept.phar generate:test unit Issue
```
the running
```
php codecept.phar run
```
fails with the error
> Fatal error: Class 'frontend\\UnitTester' not found in phar://.../codecept.phar/src/Codeception/TestCase/Test.php on line 26
> FATAL ERROR. TESTS NOT FINISHED.
> Class 'frontend\\UnitTester' not found in phar://.../codecept.phar/src/Codeception/TestCase/Test.php:26

roland-jungwirth:
Actually, the error above should read
> Class 'frontend\\UnitTester' not found 

This is not related to issue #2163. The generated code looks like this

> <?php
> namespace frontend;
> class IssueTest extends \Codeception\TestCase\Test
> {
>     /**
>      * @var \frontend\\UnitTester
>      */
>    protected $tester;
>    ...
> }
Even after changing the annotation to
```
@var \frontend\UnitTester
```
I get this error:
> Trying to test me (IssueTest::testMe)... 
> Fatal error: Class 'frontend\\UnitTester' not found in phar:///.../codecept.phar/src/Codeception/TestCase/Test.php on line 26
> FATAL ERROR. TESTS NOT FINISHED.
> Class 'frontend\\UnitTester' not found 
> in phar://.../build/codecept.phar/src/Codeception/TestCase/Test.php:26

DavertMik:
Did you try to run `codecept build` before executing tests? 
Does `frontend\Unit` class exist in `tests/_support` directory?

roland-jungwirth:
Running
> php codecept.phar build
produces
```
Building Actor classes for suites: acceptance, functional, unit
 -> AcceptanceTesterActions.php generated successfully. 0 methods added
frontend\\AcceptanceTester includes modules: PhpBrowser, \frontend\Helper\Acceptance
 -> FunctionalTesterActions.php generated successfully. 0 methods added
frontend\\FunctionalTester includes modules: \frontend\Helper\Functional
 -> UnitTesterActions.php generated successfully. 0 methods added
frontend\\UnitTester includes modules: Asserts, \frontend\Helper\Unit
```

roland-jungwirth:
There is no `frontend\Unit` class exist in `tests/_support` but there is a `UnitTester.php`:

```php
<?php
namespace frontend;

/**
 * Inherited Methods
[...]

PR #2173, opened 2015-07-15
Issue #2169, 2163: double backspace fix

tomtomsen:
Only uses namespace if in settings defined.
removes trailing backslash in namespace.

(base on Cept.php)


DavertMik:
Please send it into 2.1 branch

PR #2184, opened 2015-07-18
Use hhvm-serve instead of php -S on hhvm

Naktibalda:


LionsAd:
Nice, you might also be interested in my project:

https://github.com/LionsAd/drupal_ti

While it started as something drupal specific it is very extensible and you could probably just define a custom codeception "runner" and make use of the inbuilt functions, that would make your .travis.yml again somewhat cleaner.

Thanks for using my hhvm-serve fork! :)

Issue #2016, opened 2015-06-05
Symfony2 include problem


skillberto:
Hello!

I tried to figure how can I run my unit & functional tests from Symfony2 bundles, and acceptance tests from global directory, but I can't solve this.
Could you please tell me the best solution?

The old way:

```
include:
   - src/*/*Bundle
```
have looked good, but now I have to create a lot's of row for each bundle..

Or do you have a good solution?
Neither understand I the *.dist files, why they are good?
In my opinion, Configuration class should be rewritten with a polished structure.

Best regards,
Norbert

raistlin:
Should be fixed by #2189 , could you provide feedback ?

raistlin:
ping @skillberto

skillberto:
Yes, I think so. #2189 fixed this issue too.

PR #2189, opened 2015-07-20
support simple wildcard includes

nzod:
Issue #2016 

Allow giving wildcards in codeception.yml include paths to avoid hand-listing them.


DavertMik:
Looks good. Please send this to 2.1 branch. 
Also could you add some tests into `tests/unit/Codeception/ConfigurationTest`
Thanks

PR #2190, opened 2015-07-20
waitForElementChange fatal error

stipsan:
Upgrading from 2.0.14, got this error:
`PHP Fatal error:  Call to undefined method Codeception\Module\WebDriver::elementOrFail()`


DavertMik:
could you resend this into 2.0 or 2.1 branch? Master is for edge development only and won't be merged until Codeception 2.2

stipsan:
Of course, I'll do that right away!

stipsan:
@DavertMik see #2194

Issue #2185, opened 2015-07-18
Dollar quoting ($$) and single quoting (') errors when populating PostgreSQL dumps


YasserHassan:
The error `Syntax error: 7 ERROR:  unterminated dollar-quoted string at or near "$$` (reported in #1128) that was fixed in #1450 seems did not account for all cases. It works when the dollar quoting is at the beginning of the line like this:
```
...
    AS
$$
    BEGIN
...
```
but produces the mentioned error when the same code above is written like below, which is generated by `pg_dump` and is a valid syntax as shown in the examples at [CREATE FUNCTION](http://www.postgresql.org/docs/9.4/static/sql-createfunction.html):
```
...
    AS $$
    BEGIN
...
```
Also it does not work at all with the following quoting variations:
- Dollar quoting with a tag such as `$BODY$`.
- Single quoting instead of dollar quoting as produced by the '--disable-dollar-quoting' pg_dump option.

I tested that on version 2.0.16 but by looking into the code of version 2.1.1, it seems it's valid there too.

DavertMik:
Thanks. Looks like you deeply discovered this issue. Could you send PR for it? 
Please send it to 2.1.x branch (or to 2.0 if you still use it)

YasserHassan:
@DavertMik, done for branch 2.0 for the case when $$ is not at the beginning of the line. Please notice that the tagged dollar quoting (e.g. `$function$`) and single quoting mentioned above are still unhandled. This fix at least can help when the development database is automatically dumped using `pg_dump' then populated to the testing database before testing.

I suggest adding these two points as known issues and limitations until if/when they are implemented.

YasserHassan:
@DavertMik could you please review this commit? The testing errors seem irrelevant.

PR #2191, opened 2015-07-20
Fix Codeception/Codeception#2185 when $$ not at line start ONLY

YasserHassan:
This commit is to fix the the point of undetected dollar quoting when $$ is not at the beginning of the line. `trim` has been removed from `strpos(trim($sqlLine), '$$')` as it's no longer required.


Naktibalda:
#2185

PR #2194, opened 2015-07-21
Fix waitForElementChange fatal error

stipsan:
Fixes `PHP Fatal error:  Call to undefined method Codeception\Module\WebDriver::elementOrFail()`
@DavertMik


DavertMik:
Thanks

PR #2204, opened 2015-07-23
Check if value is set with empty() instead of !isset()

sascha-egerer:
empty() should be used instead of isset() if a value should be checked
for an array as isset() returns also true if only the array key is set.

### DIFF
--- a/src/Codeception/Configuration.php
+++ b/src/Codeception/Configuration.php
@@ -341,7 +341,7 @@ class Configuration
     {
         $settings = self::suiteSettings($suite, self::config());
 
-        if (!isset($settings['env']) || !is_array($settings['env'])) {
+        if (empty($settings['env']) || !is_array($settings['env'])) {
             return [];
         }
 


DavertMik:
I'm not sure you are right in this case. `env` parameter may not exists in `settings` so it should be checked. 
If it is empty it is ok, the code will work in a normal way then

sascha-egerer:
`$settings['env']` is an array, right?
So the `isset()` checks if the array key  `env` is set. 
That means the current implementation is like 
```
$settings['env'] = array();
if (isset($settings['env'])) {
  echo 'this will be executed'; 
  foreach($settings['env'] as $foo) {
    echo 'this will not be executed as the array is empty';
   };
}
```
But it does not make sense to go into the if condition as there are no elements in the array.

```
$settings['env'] = array();
if (!empty($settings['env'])) {
  echo 'this will NOT be executed'; 
  foreach($settings['env'] as $foo) {
   };
}
```

So this change just prevents exectuion of useless code.

Issue #1944, opened 2015-05-20
Environment config file ignored


monkeymon:
Hi,

Not sure its the generation having issue or the Finder object filtering bug,
In Codeception/Configuration.php line 289 to 293

`
        $envFiles = Finder::create()
            ->files()
            ->name('*{.dist}.yml')
            ->in($path)
            ->depth('< 1');
`
It will only return files that suffix *.dist.yml and not *.yml. And the codeception generate:env will create file name *.yml.  

I'm renaming the environment config file after I generated it for the time being.

I'm using codeception 2.1.*

Cheers,
Chris Sim

Dalamar:
Hi,
Have the same question.
Is it made by design or by mistake?

Naktibalda:
Documentation says "(e.g. chrome.yml or chrome.dist.yml for environment named chrome). "

The pattern should look like this: ->name('*{,.dist}.yml')

sascha-egerer:
thing or dist" does bot make sense. What is the benefit checking for ".dist"?

sascha-egerer:
IMHO the pattern "anything or dist" does bot make sense. What is the benefit checking for ".dist"?

PR #2205, opened 2015-07-23
Find also environment files without `.dist` in its name

sascha-egerer:
Files in the '_envs' folder are currently only recognized
if they are ending with `.dist.yml`.
The environment builder creates those files without the
`.dist` in its name wich is very confusing.

Fixes Codeception/Codeception#1944

### DIFF
--- a/src/Codeception/Configuration.php
+++ b/src/Codeception/Configuration.php
@@ -288,7 +288,7 @@ class Configuration
 
         $envFiles = Finder::create()
             ->files()
-            ->name('*{.dist}.yml')
+            ->name('*.yml')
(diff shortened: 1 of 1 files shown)


Naktibalda:
Documentation says "(e.g. chrome.yml or chrome.dist.yml for environment named chrome). "

The pattern should look like this: ->name('*{,.dist}.yml')

sascha-egerer:
IMHO the pattern "anything or dist" does bot make sense. What is the benefit checking for ".dist"?

Naktibalda:
I apologize, I was wrong about this.

Your patch is correct.

PR #2224, opened 2015-07-29
Escape string in ModuleRequireException correctly

sascha-egerer:


DavertMik:
Can you explain me why it should be changed, for me it works in both ways:

```
>>> $module = "Codeception\Module\WebDriver"
=> "Codeception\\Module\\WebDriver"
>>> str_replace('Codeception\\Module\\', '', ltrim($module, '\\'));
=> "WebDriver"
>>> str_replace('Codeception\Module\\', '', ltrim($module, '\\'));
=> "WebDriver"
```

sascha-egerer:
It's best practice to escape the escape character.

In some cases it also works with one backlash but this depends on the case.
* whether your string is surrounded by single quotes or double quotes and
* the character immediately following the backslash.

To prevent any unexpected results the escaping charcter should always be escaped.

DavertMik:
ok, thank you for PR and explanation

PR #2228, opened 2015-07-29
Replace backticks with double quotes for standard compatibility

subdee:
Backticks are not compatible with SQL Server, using double quotes seems to be universally accepted.

### DIFF
--- a/src/Codeception/Lib/Driver/Db.php
+++ b/src/Codeception/Lib/Driver/Db.php
@@ -150,7 +150,7 @@ class Db
     public function select($column, $table, array &$criteria)
     {
         $where  = $criteria ? "where %s" : '';
-        $query  = "select %s from `%s` $where";
+        $query  = "select %s from \"%s\" $where";
         $params = [];
         foreach ($criteria as $k => $v) {
             $k = $this->getQuotedName($k);
--- a/src/Codeception/Lib/Driver/Db.php
+++ b/src/Codeception/Lib/Driver/Db.php
@@ -150,7 +150,7 @@ class Db
     public function select($column, $table, array &$criteria)
     {
         $where  = $criteria ? "where %s" : '';
-        $query  = "select %s from \"%s\" $where";
+        $query  = "select %s from %s $where";
         $params = [];
         foreach ($criteria as $k => $v) {
             $k = $this->getQuotedName($k);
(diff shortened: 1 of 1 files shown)


Naktibalda:
Correct fix is to use getQuotedName method in  the same manner as it is used in insert() and deleteQuery() method.

Do you use SqlSrv or MsSql driver? MsSql driver does not implement getQuotedName.

subdee:
Youa re right, I just saw the getQuotedName, the driver I'm using however is dblib. I'm not sure that supports it either since it's updated. 

In any case, I've changed this to use the getQuotedName() method, and I've written a helper specifically for my situation.

Naktibalda:
There is no Dblib class in https://github.com/Codeception/Codeception/blob/master/src/Codeception/Lib/Driver/Db.php#L52

So you are using the Db class itself and getQuotedName returns unquoted name.
https://github.com/Codeception/Codeception/blob/master/src/Codeception/Lib/Driver/Db.php#L181

Adding double quotes to getQuotedName looks like a quick fix and will not break supported databases.
But if more changes are necessary to support Dblib, please create a Dblib driver class.

src/Codeception/Lib/Driver/Db.php
```diff
@@ -162,7 +162,7 @@ public function select($column, $table, array &$criteria)
         }
         $sparams = implode(' AND ', $params);
 
-        return sprintf($query, $column, $table, $sparams);
+        return sprintf($query, $column, $this->getQuotedName($table), $params);
```
Naktibalda: You changed $sparams to $params, please change it back.

subdee:
Oops sorry, fixed.

I will look into writing a Dblib driver class indeed.

Naktibalda:
I can merge it, unless you want to add double quotes to getQuotedName first.

subdee:
I think you can merge it, I'll make a different PR for the dblib driver specifically

DavertMik:
Hey, guys, thanks for the PR, however it is about merging into `master` branch. Master to contain only new breaking features that would come in Codeception 2.2 and thet would happen in about 5 months only (not earlier). Please resend it to stable `2.1` branch.

Naktibalda:
I made a pull request to 2.1 branch.

PR #2232, opened 2015-08-02
Added grabServiceFromContainer function

InVeX:


src/Codeception/Module/ZF2.php
```diff
@@ -71,14 +74,17 @@ public function _initialize()
         if (isset($applicationConfig['module_listener_options']['config_cache_enabled'])) {
             $applicationConfig['module_listener_options']['config_cache_enabled'] = false;
         }
-        Console::overrideIsConsole(false);
+        Console::overrideIsConsole(false); 
+        
+        $this->application = Application::init($this->applicationConfig);  
+        $this->serviceLocator = $this->application->getServiceManager();
+        $this->entityManager =  $this->serviceLocator->get('Doctrine\ORM\EntityManager');
```
DavertMik: by default ZF2 doesn't have Doctrine

DavertMik:
@Naktibalda what do you think?

Naktibalda:
I can't comment on necessity of grabServiceFromContainer, because I am not a ZF2 user.
But $this->entityManager =  $this->serviceLocator->get('Doctrine\ORM\EntityManager');  line broke my ZF2 REST tests, so it isn't a good thing.

https://travis-ci.org/Codeception/Codeception/jobs/73775687#L7179

InVeX:
Is there a way to check if Doctrine module is enabled in codeception? I could add this check then everything should work fine.

InVeX:
Ok I have added a check which tries to load entity manager only if it exists.

src/Codeception/Module/ZF2.php
```diff
@@ -71,14 +73,25 @@ public function _initialize()
...
+    }  
 
     public function _before(TestCase $test)
-    {
-        $this->application = Application::init($this->applicationConfig);
+    {      
+        $this->application = Application::init($this->applicationConfig);               
+        $this->serviceLocator = $this->application->getServiceManager();
+        if($this->serviceLocator->has('Doctrine\ORM\EntityManager')) {
+            $this->entityManager =  $this->serviceLocator->get('Doctrine\ORM\EntityManager');
```
DavertMik: why do you ever need this `entityManager` to be set before each test? Is that used somehow anywhere? I don't see it.

src/Codeception/Module/ZF2.php
```diff
@@ -107,8 +120,28 @@ public function _after(TestCase $test)
     }
 
     public function _getEntityManager()
-    {
-        $serviceLocator = Application::init($this->applicationConfig)->getServiceManager();
-        return $serviceLocator->get('Doctrine\ORM\EntityManager');
+    {   
+        return $this->grabServiceFromContainer('Doctrine\ORM\EntityManager');
```
DavertMik: once again, why did you define `entityManager` property if we have this nice method? ;)

DavertMik:
@InVeX ping ) Small changes required to get it merged. Thanks

src/Codeception/Module/ZF2.php
InVeX: It was defined before, reason is in Doctrine2 Module, line 102:
$this->em = $this->dependentModule->_getEntityManager();
I just changed function contents

DavertMik: @InVeX yes, i have nothing to this method. I have concerns about `entityManager` property (see above). Please remove that property.

PR #2243, opened 2015-08-05
Fixed file uploads and multipart form submission with Guzzle6

Naktibalda:


Naktibalda:
I will fix it in a different way

Issue #2244, opened 2015-08-05
file_get_contents error for --coverage-html


jnowisz:
running tests with --coverage-xml works fine, but --coverage-html produces this error:

[PHPUnit_Framework_Exception]
file_get_contents(...path to tmp file...): failed to open stream: No such file or directory

I use EasyPHP which installs in directory with dot (".") in the name. 
In Coverage\Subscriber\RemoteServer in method retrieveAndPrintHtml() before writing temp file all dots are striped from the path. If the path contains dots then parameter passed to file_get_contents is not valid.

PR #2245, opened 2015-08-05
Fixed removing dots from path and temp file, causing errors in html coverage generation

jnowisz:
find_put_contents couldn't open stream because of wrong path. Problems appeared only for those having temp dirs with dots like for users of EasyPHP default installation.

Fixes issue: #2244


DavertMik:
Thanks

PR #2247, opened 2015-08-06
Improves performance of getClassesFromFile method of the parser

Green-Cat:
Due to the way for loops are implemented in php, the count function will run every iteration of the loop. Therefore, moving the count to a variable, calculated before the start of the loop, improves performance (less function calls).


DavertMik:
Thanks!

PR #2252, opened 2015-08-08
Removed disabled and file fields from form values

Naktibalda:


lowerends:
I think this PR creates a problem when using the `seeInField` function on disabled form fields or form fields within a disabled fieldset. Such fields are no longer considered to be part of the form and therefore their value is always empty.

PR #2269, opened 2015-08-12
fix "undefined index config" in 2.1.2 and Yii

quantum13:
if "config" section not exists, all tests have failed. This commit fix it (and also style fix for long line)


samdark:
Thanks!

PR #2279, opened 2015-08-14
Add example to amLoggedAs.

cjaoude:
Add example to amLoggedAs. Clarify $driver param.


cjaoude:
I agree, not a good one. Goes to show it's not clear without an example. Provided alternate examples, as well as, the method which can be used to verify that it did work. Missed that method for a long while. Makes things a lot easier.

cjaoude:
This pull had checked out OK. Accidentally pushed more stuff to this branch. How to get the integration tests run again? :/

Issue #1788, opened 2015-03-18
Strict mode for REST module seeResponseContainsJson


atorian:
It might be useful to have strict mode flag for this kind of methods.
Json might contain:

    ['key'=> '1'] // string value

and it could be compared to: 

    ['key' => 1]

Currently this check will fail. Strict mode by default is good, but sometimes we'd like to disable it.

DavertMik:
If you can implement non-strict matching for JSON that would be great)

mrmarkfrench:
It's actually even worse than this, as numeric comparisons will fail even in cases where the JSON is a character-for-character match. If, for example, my JSON response is:
{"number":5}

This test will pass:
$I->seeResponseContainsJson(array('number'=>(int)5));
But this one will fail:
$I->seeResponseContainsJson(array('number'=>(float)5));

Comparing a float with the same numeric value to the one returned in the JSON only passes when the JSON text contains a decimal point, and if there are no non-zero digits after the decimal point, json_encode() will not write 5.0, it will just write 5.

atorian:
fixed

PR #2283, opened 2015-08-17
fixed issue1788: removed strict check

atorian:
Issue: https://github.com/Codeception/Codeception/issues/1788
i've investigated a bit more and found out that it would be better be go with php way - make it array_diff does http://php.net/manual/ru/function.array-diff.php

    Note:

    Two elements are considered equal if and only if (string) $elem1 === (string) $elem2. In words: when the string representation is the same.

but with current code it will look like simple non-strict check.
So i've changed that condition and applied tests, to proof new behaviour. 
Please verify.

### DIFF
--- a/src/Codeception/Util/JsonArray.php
+++ b/src/Codeception/Util/JsonArray.php
@@ -145,7 +145,7 @@ class JsonArray
                     continue;
                 }
 
-                if ($value1 === $value2 && !isset($matchedKeys[$key2])) {
+                if ($value1 == $value2 && !isset($matchedKeys[$key2])) {
                     $ret[$key1] = $value1;
                     $matchedKeys[$key2] = true;
                     break;
@@ -173,8 +173,7 @@ class JsonArray
                 $ret[$key] = $return;
                 continue;
             }
-
-            if ($arr1[$key] === $arr2[$key]) {
+            if ($arr1[$key] == $arr2[$key]) {
                 $ret[$key] = $arr1[$key];
             }
         }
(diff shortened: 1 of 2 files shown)


Naktibalda:
It looks good to me.

Naktibalda:
Actually the check is way too loose now.

1. ==  casts string to integer if it starts with 0, examples: http://3v4l.org/tT4l8
2. Values of different types have the same string representation - false == null == "", true == "1" == 1
  false, null and true values are very easy to distinguish in both JSON and PHP, I think that we should not use a loose comparison for them,

atorian:
The issue could be partially fixed with json_numeric_check flag passed to json_encode in the app. But this is not always possible, or requires to many effort.
But it won't fix issues with dynamic expected values. For example if value was grabbed from db or file - it will have type === string.
I'll investigate more.

Naktibalda:
Since this issue is specific to numbers, it could be solved by casting all integers and floats to string and then comparing strictly, right?

atorian:
i believe so.
I'll do this today.

atorian:
please verify

src/Codeception/Util/JsonArray.php
```diff
@@ -212,4 +211,17 @@ private function arrayToXml(\DOMDocument $doc, \DOMNode $node, $array)
             }
         }
     }
+
+    private function isEqualValue($val1, $val2)
+    {
+        if (is_numeric($val1)) {
```
sergeyklay: mb [is_scalar](http://php.net/manual/en/function.is-scalar.php) better way?

Naktibalda: It would cast boolean values to strings, which is not necessary.

PR #2296, opened 2015-08-20
Add Interface DependsOnModule to SOAP module

relaxart:
I hope it will be in 2.1, because SOAP is broken without DependsOnModule interface.


DavertMik:
Strange it was not there :(

PR #2304, opened 2015-08-24
WebDriver->seeNumberOfElements should only count visible elements

### LINKED ISSUE #2303: WebDriver->seeNumberOfElements should only count visible elements
The behavior of `seeNumberOfElements` and `seeElement` is currently not the same as `seeNumberOfElements` does also count invisible elements.
To check for elements in DOM there is a seperate method `seeElementInDOM`. This should also be available for `seeNumberOfElementsInDOM`.

sascha-egerer:
The behavior of seeNumberOfElements and seeElement is not the same
as seeNumberOfElements does also count invisible elements.
To check for elements in DOM there is a seperate method
seeElementInDOM.

The method `seeNumberOfElementsInDOM` has been added and the behavior
of `seeNumberOfElements` has been changed to not respect invisible
elements.

Fixes #2303

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -1139,6 +1139,24 @@ class WebDriver extends CodeceptionModule implements
 
     public function seeNumberOfElements($selector, $expected)
     {
+        $counted = count($this->matchVisible($this->webDriver, $selector));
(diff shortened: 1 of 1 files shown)


sascha-egerer:
Sorry - Pushed an unfinished version. New PR will follow

PR #2326, opened 2015-09-01
Filter invalid cookie domains in session snapshots

### LINKED ISSUE #2250: InvalidCookieDomainException when (automatically) adding a 3rd-party cookie via a session snapshot
When loading a session snapshot, **every** previously saved cookie will be added individually. When adding, let´s say a 3rd-party ad cookie with the domain ```.foo.com``` (and my configured URL is ```.bar.com```), an ```InvalidCookieDomainException``` will be thrown since the cookie domain does not match the current site url. The [HttpCommandExecutor](https://github.com/facebook/php-webdriver/blob/community/lib/Exception/WebDriverException.php#L139) throws a ```\Facebook\WebDriver\Exception\WebDriverException``` based on the return code ```24```.

I´m not sure what to do here. Am I misunderstanding something or should the [WebDriver](https://github.com/Codeception/Codeception/blob/2.1/src/Codeception/Module/WebDriver.php#L2144) only accept Cookies that match the host part of the configured URL? The [WebDriverOptions](https://github.com/facebook/php-webdriver/blob/community/lib/WebDriverOptions.php#L42-L43) class documents to only add cookies with a domain value of ```null``` or the same host as the configured URL.

In my case the ad cookie prevents me from loading session snapshots. Please let me know what you think about this, I would gladly create a PR if something I wrote here makes any sense :)

venyii:
This PR filters cookie domains that don´t match the config URL and therefore prevents an `InvalidCookieDomainException` (which prevents all cookies from being saved).

Fixes #2250

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -35,6 +35,7 @@ use Facebook\WebDriver\WebDriverElement;
 use Facebook\WebDriver\WebDriverExpectedCondition;
 use Facebook\WebDriver\WebDriverKeys;
 use Facebook\WebDriver\WebDriverSelect;
+use GuzzleHttp\Cookie\SetCookie;
 use Symfony\Component\DomCrawler\Crawler;
 
 /**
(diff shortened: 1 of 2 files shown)


venyii:
Thank you very much for merging this so quickly @Naktibalda !

Could we get a new tag (`2.1.3`) for this?

sharpner:
:+1: will this get a release?

venyii:
ping @DavertMik 

Any plans to tag a release soon?

Naktibalda:
We really need nightly or weekly builds.

PR #2335, opened 2015-09-03
Add details about required var_path configuration parameter for Symfony2 module

bizmate:
The description of var_path is also required for symfony2 applications.

As a generic good practice tests are located in symfony2 bundles under src. Thus it is required to specify var_path too so the Symfony2 module is aware of where the bootstrap cache is loaded from. app_path alone just tells where the AppKernel is supposed to be loaded from.


Naktibalda:
Could you make the same change in 2.1 branch?

PR #2336, opened 2015-09-03
Guzzle connector override status code to 302 if a meta refresh exists in body

pitpit:
I disabled followRedirect in PhpBrowser (`$this->getModule('PhpBrowser')->client->followRedirects(false)`) to test 301 response with REST module like:

    $I->seeResponseCodeIs(301);

But the endpoind I want to test returns a 301 response with `<meta http-equiv="refresh" />` in body (the Symfony2 way).
So Guzzle recode the status code to 302 and my test fails.

### DIFF
--- a/src/Codeception/Lib/Connector/Guzzle.php
+++ b/src/Codeception/Lib/Connector/Guzzle.php
@@ -28,16 +28,16 @@ class Guzzle extends Client
     {
         $this->baseUri = $uri;
     }
-    
+
     /**
      * Sets the maximum allowable timeout interval for a meta tag refresh to
      * automatically redirect a request.
-     * 
+     *
      * A meta tag detected with an interval equal to or greater than $seconds
      * would not result in a redirect.  A meta tag without a specified interval
      * or one with a value less than $seconds would result in the client
      * automatically redirecting to the specified URL
-     * 
+     *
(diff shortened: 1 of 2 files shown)


src/Codeception/Lib/Connector/Guzzle.php
```diff
@@ -107,38 +107,40 @@ protected function createResponse(Response $response)
...
+            if ((!empty($matches)) && (empty($matches[1]) || $matches[1] < $this->refreshMaxInterval)) {
+                $uri = $this->getAbsoluteUri($matches[2]);
+                $partsUri = parse_url($uri);
+                $partsCur = parse_url($this->getHistory()->current()->getUri());
+                foreach ($partsCur as $key => $part) {
+                    if ($key === 'fragment') {
+                        continue;
+                    }
+                    if (!isset($partsUri[$key]) || $partsUri[$key] !== $part) {
+                        $status = 302;
```
Naktibalda: What exactly have you changed?
I see many formatting changes, but it still overrides status code.

venyii: FYI: You can remove the whitespace changes by appending `?w=0` to the URL.

https://github.com/Codeception/Codeception/pull/2336/files?w=0

PR #2339, opened 2015-09-07
When Tests has NS, throws class not found for actor.

pecinaon:
When i create a tests with own NS:
php codecept bootstrap --namespace "Project\Test"
php codecept run throws fatal error:

Class 'Project\Tests\\UnitTester' not found.

This fix can solve problem.

Or you can navigate me, where is correct point to fix it ;)


DavertMik:
Thanks, this fix should have be done in `bootstrap` generator. I did it )

PR #2343, opened 2015-09-07
Typo change in example

sandinosaso:
Typo used deprecated function instead of actul grabDataFromResponseByJsonPath


DavertMik:
thanks

PR #2345, opened 2015-09-07
[WIP] Strip the cookie domain in session snapshots

venyii:
This came up while I was testing #2250 

Sending a cookie with a domain `.foo.bar.com` will cause a selenium exception
if the current domain is `http://foo.bar.com` (which should match). The test `testSaveSessionSnapshotsExcludeInvalidCookieDomains` will fail without this fix.

Removing the cookie domain should not cause any problems since it matches the
configured url.


venyii:
Ok, this is a bit weird, but I found the solution to this problem. You may only add `.foo.bar.com` ***after*** you opened the site once... `foo.bar.com` works flawlessly.

> NOTE: You have to go to the same domain page (html page) before adding cookie.

So there´s no problem here, closing. Sorry for the noise.

PR #2350, opened 2015-09-08
Allow other HTTP_HOSTs then localhost

JakobReiter:
It was not possible to change the host from localhost to something else.
I now use the URL config to extract the host and set the HTTP_HOST for the client.


DavertMik:
@samdark looks good to me. Merging?

PR #2363, opened 2015-09-11
Symfony2: Fail when profile is not set.

### LINKED ISSUE #652: ErrorException: Argument 1 passed to
I just insalled the codeception with composer and I get the following error during my functional tests.

```
ErrorException: Argument 1 passed to Symfony\Component\HttpKernel\Profiler\Profiler::loadProfileFromResponse() must be an instance of Symfony\Component\HttpFoundation\Response, instance of Symfony\Component\BrowserKit\Response given, called in /vendor/codeception/codeception/src/Codeception/Module/Symfony2.php on line 150 and defined
```

raistlin:
In Symfony2, when a method that need the profiler is called, it will fail instead of crashing.

Fixed #652

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 #### 2.1.3
 
+* [Symfony2] Fixed issue when accesing profiler when no request has been performed #652.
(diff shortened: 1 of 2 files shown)


DavertMik:
Thanks. Btw, I added you to committers so you can merge such simple cases on your own.

Issue #1592, opened 2014-12-10
Add feature to remove code coverage output from console


Ragazzo:
Would be great to have ability to remove code coverage dump from console, option like `--coverage-console` to dump coverage to console would be useful. Will handle this one, once i have  more time

EspadaV8:
Just wondering if any work was started on this? If not I might be able to look at it, but I have no experience with the Codeception code base.

sergeyklay:
It seems it would be enough extend `--quiet` option. Something like this:
```sh
$ vendor/bin/codecept run --coverage-html --quiet


Time: 217 ms, Memory: 5.25Mb

HTML report generated in coverage
OK (33 tests, 144 assertions)
```

Ragazzo:
Currently i am not working on this issue, and dont think that i will have any free time in nearest future, so feel free to handle this one

EspadaV8:
@sergeyklay Implemented it using `--quiet` but as mentioned in the PR not sure if that's really the 'right' way or not, since ideally I'd like to see the testing progress, just not see the coverage report in the console.

PR #2370, opened 2015-09-15
Suppress console coverage report with `--quiet`

EspadaV8:
PR for #1592 

This should work as-is, although I'm not sure it's the 'right' way of doing it. When `--quiet` is used it also suppresses the progress of the tests, which isn't really what I would find useful, ideally I'd like to see the test progress but just hide the report. That would require a new option (something like `--no-coverage-console` or `--coverage-quiet`) and I wasn't sure if that would be okay so I thought I'd check first.

### DIFF
--- a/src/Codeception/Codecept.php
+++ b/src/Codeception/Codecept.php
@@ -52,7 +52,8 @@ class Codecept
         'fail-fast'     => false,
         'verbosity'     => 1,
         'interactive'   => true,
-        'no-rebuild'    => false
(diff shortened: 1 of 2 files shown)


Naktibalda:
@DavertMik No comments in 8 months, can I remove RFC label and merge it?

DavertMik:
Oh, absolutely forgot about it. Sorry. Probably I thought this was something different than that... Anyway, seems reasonable

Issue #2328, opened 2015-09-01
[InnerBrowser] Gracefully handle redirecting to external sites


Naktibalda:
Related to #2327 

When the framework module is used for functional testing, redirecting to external sites fails like this one:

[Zend_Controller_Dispatcher_Exception] Invalid controller specified (http:)

Links and form submission ought to fail in the same way.

I see 2 possible solutions:
a) don't follow redirect, but provide some methods for inspecting it.
b) transparently replace framework client with PhpBrowser.

DavertMik:
> transparently replace framework client with PhpBrowser.

NOOOOOOOOOOOOOOOOOOOOOOOOOOO

>  don't follow redirect, but provide some methods for inspecting it.

sounds like a plan to me. Looks like we need to add methods for inspecting redirects (we didn't have them before, as `seeCurrentUrl*` methods are about that.

Naktibalda:
How about throwing exception?

DavertMik:
Probably this can be a simple solution. To create new `ExternalRedirectException` and use it with InnerBrowser

PR #2371, opened 2015-09-15
Throw exception when framework module tries to open an external URL 

Naktibalda:
#2328

Intercept click() for debugging
Throw exception when framework module tries to open an external URL

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -9,6 +9,7 @@ use Codeception\Exception\ModuleException;
 use Codeception\Exception\TestRuntimeException;
 use Codeception\Lib\Interfaces\ElementLocator;
 use Codeception\Lib\Interfaces\PageSourceSaver;
+use Codeception\Lib\Interfaces\SupportsDomainRouting;
 use Codeception\Lib\Interfaces\Web;
 use Codeception\Module;
 use Codeception\PHPUnit\Constraint\Crawler as CrawlerConstraint;
@@ -1293,7 +1294,7 @@ class InnerBrowser extends Module implements Web, PageSourceSaver, ElementLocato
                 $content = http_build_query($parameters);
             }
 
-            if (preg_match('#^(//|https?://(?!localhost))#', $uri)) {
+            if (preg_match('#^(//|https?://(?!localhost))#', $uri) && (!$this instanceof SupportsDomainRouting)) {
                 throw new ExternalUrlException(get_class($this) . " can't open external URL: " . $uri);
             }
         }
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -1354,7 +1354,7 @@ class InnerBrowser extends Module implements Web, PageSourceSaver, ElementLocato
      *
      * @return Crawler
      */
-    public function clientClick(Link $link)
+    protected function clientClick(Link $link)
     {
         if ($link instanceof Form) {
             return $this->proceedSubmitForm($link);
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -18,6 +18,7 @@ use Codeception\Util\Locator;
 use Codeception\Util\PropertyAccess;
 use Codeception\Util\Uri;
 use Symfony\Component\BrowserKit\Cookie;
+use Symfony\Component\DomCrawler\Link;
 use Symfony\Component\DomCrawler\Crawler;
 use Symfony\Component\DomCrawler\Field\ChoiceFormField;
 use Symfony\Component\DomCrawler\Field\FileFormField;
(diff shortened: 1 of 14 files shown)


Naktibalda:
Some Laravel4, Laravel5 and Yii2 tests failed, because they use domain based routes.
Some framework connectors don't support them, my guess is that this restriction should be limited to frameworks that remove http://localhost prefix from request URI:
https://github.com/Codeception/Codeception/search?utf8=%E2%9C%93&q=%24uri+%3D+str_replace%28%27http%3A%2F%2Flocalhost%27%2C+%27%27%2C+%24request-%3EgetUri%28%29%29%3B&type=Code

Phalcon, Universal, ZF1 and ZF2,

I think that a good solution would be to make them implement interface DoesntSupportDomains
or make others implement SupportsDomains

src/Codeception/Lib/InnerBrowser.php
```diff
@@ -1338,4 +1346,20 @@ protected function redirectIfNecessary($result, $maxRedirects, $redirectCount)
...
     }
+
+    /**
+     * Clicks on a given link.
+     *
+     * @param Link $link A Link instance
+     *
+     * @return Crawler
+     */
+    public function clientClick(Link $link)
```
DavertMik: it should not be public

DavertMik:
@janhenkgerritsen @samdark what you think?
I think `SupportsDomains` can be implemented 

However, even domain-supported frameworks can't do redirects to external sites. But can we guess what sites are external and what are not?

Naktibalda:
I think that such frameworks can handle it themselves.

janhenkgerritsen:
I also think `SupportsDomains` can be implemented, but maybe change the name of the interface to `SupportsDomainRouting`?

For Laravel it will probably be possible to detect which URLs are external and which are not by simply looking through the registered routes and see if they have a domain specified. All domains that are specified can then be considered internal and all others external.

What currently happens for external routes with Laravel is that if there are no routes matching that domain the path of the URL wil be matched against the routes that are defined. This is actually very confusing behavior, because sometimes a match will be found and that page will be the response of the redirect (for example the homepage in case of a redirect to an URL without a path) and sometimes no match is found and you will get a 404 response. So this should be changed, but before I make any changes I will wait to see which solution will be implemented for this issue.

Naktibalda:
any objections?

janhenkgerritsen:
I still do think that `SupportsDomainRouting` or something like that would better communicate the intention of the interface, `SupportsDomains` is very generic.

And another suggestion, maybe give the interface a single method named something like `getInternalDomains` so the check for external redirects can be kept in a single location, instead of implementing it in each of the modules.

DavertMik:
@samdark can we easily receive all internal domains in Yii? So to implement `getInternalDomains` method

Naktibalda:
If you want to add a method to interface, it must be implemented in 6 framework modules.
And I can't do that.
It could be easier to implement it in each module.

janhenkgerritsen:
Can't you just add a default implementation that returns an empty array? Then the maintainers can update the modules after you are finished. I can do the Laravel4,  Laravel5 and Lumen modules.

Naktibalda:
@janhenkgerritsen empty array will break Laravel4 and Laravel5 tests:
https://travis-ci.org/Codeception/Codeception/jobs/80535452#L7134

 [ExternalUrlException] Codeception\Module\Laravel4 can't open external URL: http://l4.app/domain-route 

[ExternalUrlException] Codeception\Module\Laravel5 can't open external URL: http://myapp.com/posts/create

And we can't know what domains are used in tests of our users, so implementation with empty array can't be merged into 2.1 branch.

How about merging this pull request and then implementing getInternalDomains in a separate branch?

janhenkgerritsen:
Merging this PR and implementing `getInternalDomains` in another branch is fine with me. Just notify me when you are finished with this PR and I will create the new branch and a first implementation of the changes.

raistlin:
The more time I think about this, the more I dislike this.
Let me explain:
In my projects with Symfony2 I never set a domain in my Symfony2 code. I just setup apache virtual hosts to point the web path of my symfony2 app, and that is all. Testing domains are 'myapp.test', and production ones are 'myapp.com', application don't need to know on what domain is being running on.

In my testing environments, I just had configured in /etc/hosts files that 'myapp.test' points to 127.0.0.1, and in codeception.yml or acceptance.suite.yml I also have the same domain as base url for PhpBrowser and WebDriver. In production environment is setup simp

As far as I understand, you would like to intercept clicks of the tests to domains outside of what is defined in the framework, and raise an exception. 

What happens with people that have my same environment and never filter by hosts at their frameworks? Will I need to inject domains to my routes simply because of that feature? Did you first check if domain used is the same as stated in PhpBrowser or WebDriver? I just search for this fact but did not found this filtering. If you don't do this, then will force me to inject domain name to my framework, and I don't want to inject this just to prevent going outside my app while testing!

janhenkgerritsen:
This functionality only applies in cases where you try to access an absolute URL through Codeception, for example by doing `$I->amOnPage('http://subdomain.example.com')`. But in an application that only runs on a single domain you won't do that, you will use only paths in your tests, for example `$I->amOnPage('contact')`.

You can take a look at the calling code here: https://github.com/Codeception/Codeception/blob/feature-get-internal-domains/src/Codeception/Lib/InnerBrowser.php#L1268

raistlin:
Ok, then I'm ok with this.

Thanks for explanation, my mind was blowing :)

PR #2379, opened 2015-09-18
fix redirect when base url has path

tiger-seo:


Naktibalda:
Why does it fail on PHP 5.4?

Naktibalda:
ok, the answer is obvious - because you made your change in Guzzle6 connector. Could you make the same change in Guzzle connector?

Naktibalda:
Codeception does not follow regular absolute path rules.
If the path starts with / it should replace path set in base_uri instead of being appended.

tiger-seo:
> Codeception does not follow regular absolute path rules.
> If the path starts with / it should replace path set in base_uri instead of being appended.

good catch :)

DavertMik:
> If the path starts with / it should replace path set in base_uri instead of being appended.

I'm not sure here but what about If we have url:

```
http://site.com/api/v1
```

and we'd like to get users there, then we should be able to use 

```
sendGET('/users')
```

I think a similar example can be provided for PHPBrowser as well. Like accessing `/users` for 

```
http://site.com/blog
```

should move us to 

```
http://site.com/blog/users
```

DavertMik:
@tiger-seo you can skip this test for old Guzzle. Please use `skipForOldGuzzle` method

Naktibalda:
Please don't abandon PHP 5.4, I'm stuck with it for another year at least

DavertMik:
@Naktibalda that's terrible :(
@tiger-seo ok, so could you get it implemented for Guzzle v.5? You can downgrade Guzzle in your composer.json

tiger-seo:
@DavertMik i thought the scope of
> If the path starts with / it should replace path set in base_uri instead of being appended.

is only redirects

Naktibalda:
Link clicks and forms should behave in the same way. Test code like amOnPage and sendPOST can keep current behaviour for BC. We need to support <base> tags.

Naktibalda:
Base tags

DavertMik:
> Link clicks and forms should behave in the same way.

yeah, you are right. 

But what do you mean about base tags?

Naktibalda:
https://developer.mozilla.org/en/docs/Web/HTML/Element/base

tiger-seo:
@DavertMik i've done fix for the old guzzle

JarJak:
Happened to me with PHP 5.6 and latest Codeception / Guzzle 6.0. See: https://github.com/Codeception/Codeception/issues/2402

tiger-seo:
@Naktibalda  @DavertMik  fixing the issue with absolute urls globally would be a major BC break, so for now my suggestion would be to fix only this problem
What's your take on this, guys ?

JarJak:
:+1: fixed my issue

Naktibalda:
That's fine with me.

DavertMik:
thanks

PR #2384, opened 2015-09-20
[Guzzle6] Set Location header to string instead of Uri

Naktibalda:


DavertMik:
Looks like a really good cleanup. Thank you!

DavertMik:
Ok, I'm merging if you don't mind?

PR #2388, opened 2015-09-21
Environment subfolders support

Zifius:
Sorry, first take was a blindly patch. This is actually tested

### DIFF
--- a/src/Codeception/Codecept.php
+++ b/src/Codeception/Codecept.php
@@ -187,7 +187,10 @@ class Codecept
             if (empty($config)) {
                 continue;
             }
-            $suiteToRun = "{$suite}-{$envList}";
+            $suiteToRun = $suite;
+            if (!empty($envList)) {
+                $suiteToRun .= ' (' . implode(', ', $envArray) . ')';
+            }
             $this->runSuite($config, $suiteToRun, $test);
         }
     }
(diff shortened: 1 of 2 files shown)


DavertMik:
Can you provide a small background. What was the issue?

Zifius:
Hi @DavertMik, sure - I implemented the feature initially in #2381, but by testing today the head of 2.1 branch I discovered that changing search depth didn't really work as the subfolder wasn't accounted for config loading so actually environments in subfolders werenn't loaded at all so this PR fixes that. Not sure if totally ok code wise so let me know if you have any notes. Thanks

DavertMik:
why not just increase the search depth?

```
            ->depth('< 2');
```

Zifius:
@DavertMik that's what I did in the initial PR but $path here https://github.com/Zifius/Codeception/commit/72e50e14982699e6aea53012f902c7d6446eeaae#diff-a14abaeb14e2165d64e4de51c4168d5bL309 is the absolute path to _envs folder, it doesn't include subfolders thus no subfolders are used when merging the configs

Zifius:
But I agree that it's complicated, perhaps just need to collect envs with subfolder path in this loop https://github.com/Zifius/Codeception/commit/72e50e14982699e6aea53012f902c7d6446eeaae#diff-a14abaeb14e2165d64e4de51c4168d5bR300. Perhaps that regex is also overkill. I'll look into it and rework

Zifius:
Hi @DavertMik, please review the latest updates for this PR. Thank you

DavertMik:
Yeah, latest fix is much more better than previous one

DavertMik:
I will write a failing test and merge it afterwards )

Zifius:
Cool, thanks :+1:

DavertMik:
Works as expected. Thank you!

Zifius:
Cheers, happy to contribute!

PR #2395, opened 2015-09-22
Interface SupportsDomainRouting is for modules, not for connectors

Naktibalda:
I made a blunder in a previous pull request #2371


Naktibalda:
I keep doing mistakes.

PR #2400, opened 2015-09-23
Fix JUnit XML Format:

SunDrop:
> testcase's attribute "class" must be called "classname"
> !xs:attribute name="classname" type="xs:token" use="required"!

### DIFF
--- a/src/Codeception/PHPUnit/Log/JUnit.php
+++ b/src/Codeception/PHPUnit/Log/JUnit.php
@@ -6,7 +6,7 @@ use Codeception\TestCase\Interfaces\Reported;
 
 class JUnit extends \PHPUnit_Util_Log_JUnit
 {
-    protected $strictAttributes = ['file', 'name', 'class'];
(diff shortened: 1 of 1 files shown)


Naktibalda:
@DavertMik it can be merged, right?

DavertMik:
Sorry, not yet. Let me review the reasons. Looks like the JUnit format may differ between CI systems

drazisil:
Going to :+1: this, since the only known schema (generated from JUnit) clearly states it should be `classname`. Differences between CI system for not be a reason for not following spec.

Ref: https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd

Naktibalda:
@DavertMik Have you reviewed reasons for not merging this PR?

DavertMik:
I think it doesn't work on Jenkins. It is [`class` in PHPUnit](https://github.com/sebastianbergmann/phpunit/blob/master/src/Util/Log/JUnit.php#L318) so we will use it as well. You can write your own reporter and change it in config

Issue #2398, opened 2015-09-23
[Feature Request] Iframe support for InnerBrowser


Naktibalda:
I just discovered that I need to test website which uses an iframe and it seems that this feature is missing in InnerBrowser.

Naktibalda:
WebDriver has switchToIFrame method, so it seems like a good choice for a name.

The simplest implementation could operate in a manner similar to click() or amOnPage().
Switching to another frame or back to parent frame would involve keeping multiple responses in memory.

DavertMik:
We have `sendAjaxRequest` methods that switch contexts as well. After you send a request the context is changed and now you work with ajax response. The original idea was to work with ajax-popups - to receive them and interact. However, there is no way to switch back to original page. 

So I think it is a good idea to have implemented the mechanism for context switching.

PR #2403, opened 2015-09-24
Implemented switchToIframe method for InnerBrowser

Naktibalda:
#2398

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -1346,4 +1346,35 @@ class InnerBrowser extends Module implements Web, PageSourceSaver, ElementLocato
 
         return $this->clientRequest($link->getMethod(), $link->getUri());
     }
+
+    /**
+     * Switch to iframe or frame on the page.
+     *
+     * Example:
+     * ``` html
+     * <iframe name="another_frame" src="http://example.com">
+     * ```
+     *
+     * ``` php
+     * <?php
+     * # switch to iframe
+     * $I->switchToIframe("another_frame");
+     * ```
+     *
+     * @param string $name
+     */
+
+    public function switchToIframe($name)
+    {
+        $iframe = $this->match("iframe[name=$name]")->first();
+        if (!count($iframe)) {
+            $iframe = $this->match("frame[name=$name]")->first();
+        }
+        if (!count($iframe)) {
+            throw new ElementNotFound("name=$name", 'Iframe');
+        }
+
+        $uri = $iframe->getNode(0)->getAttribute('src');
+        $this->clientRequest('GET', $uri);
+    }
 }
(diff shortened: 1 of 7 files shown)


DavertMik:
Yes, we need to do something to switch back. The solution of WebDriver module is not the something I really like - if name is empty we are switching to main page. Maybe a new method like `switchToPage` sounds good?

It will work with `sendAjaxRequest` and `switchToIFrame`...

Actually I don't know if someone really uses `sendAjaxRequest` methods to load HTML pages partially. But probably it is used and we can improve it.

Naktibalda:
That's the meaning of Page? Do you mean getOutOfFrames?

How about implementing moveBack() method?
Since Javascript is not supported, there is not much state to be kept.

But if we wanted to keep multiple frames in memory, we would have to keep separate instances of module and connector for each frame (to have separate histories, form fields and whatever state there is).

DavertMik:
> But if we wanted to keep multiple frames in memory, we would have to keep separate instances of module and connector for each frame (to have separate histories, form fields and whatever state there is).

No.We should keep only one state and only for cases where we narrow context: like we do for iframes or ajax requests. That would be reasonable as I think of it

*

src/Codeception/Lib/InnerBrowser.php
```diff
@@ -1300,12 +1295,16 @@ protected function clientRequest($method, $uri, array $parameters = array(), arr
         }
 
         if (!PropertyAccess::readPrivateProperty($this->client, 'followRedirects')) {
-            return $this->client->request($method, $uri, $parameters, $files, $server, $content, $changeHistory);
+            $result = $this->client->request($method, $uri, $parameters, $files, $server, $content, $changeHistory);
+            $this->debugResponse();
+            return $result;
+        } else {
```
DavertMik: else is not needed

tests/unit/Codeception/Module/TestsForWeb.php
```diff
@@ -1292,4 +1292,12 @@ public function testFillFieldWithAmpersand()
         $form = data::get('form');
         $this->assertEquals('this & that', $form['name']);
     }
+
+    public function testSwitchToIframe()
+    {
+        $this->module->amOnPage('/iframe');
+        $this->module->switchToIframe('content');
+        $this->module->see('Is that interesting?');
+        $this->module->click('Ссылочка');
```
sergeyklay: > Ссылочка

:)

Naktibalda: I tried to click "Sign in!", but it failed with exception.

sergeyklay:
```sh
$ ~/projects/php/Codeception grep -nir 'Ссылочка'
tests/data/app/view/info.php:28:<a href="/">Ссылочка</a>
tests/unit/Codeception/Module/PhpBrowserTest.php:67:        $this->module->seeLink('Ссылочка');
tests/unit/Codeception/Module/PhpBrowserTest.php:68:        $this->module->click('Ссылочка');
```

Naktibalda: I got a very misleading exception: [MalformedLocatorException] XPath or CSS locator is malformed: Sign In! 
I can see why it failed now.

DavertMik:
I think this can be merged. Please do it if you don't find anything else should be added

PR #2406, opened 2015-09-26
ZF1, ZF2: call parent::_after in _after to cleanup after each test

Naktibalda:
I had an issue with client carrying over domain from test which uses domain to test which doesn't.

### DIFF
--- a/src/Codeception/Module/ZF1.php
+++ b/src/Codeception/Module/ZF1.php
@@ -108,11 +108,12 @@ class ZF1 extends Framework
 
         require_once 'Zend/Loader/Autoloader.php';
         \Zend_Loader_Autoloader::getInstance();
-        $this->client = new ZF1Connector();
     }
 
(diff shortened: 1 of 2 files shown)


src/Codeception/Module/ZF1.php
```diff
@@ -147,6 +148,8 @@ public function _after(TestCase $test)
         \Zend_Session::$_unitTestEnabled = true;
         $this->queries = 0;
         $this->time = 0;
+
+        parent::_after($test);
```
DavertMik: but why do you call it in `before`?

Naktibalda: Miscommunication between Github and you.
_after starts in line 134

DavertMik: yeah. right. That was how github showed it.

DavertMik:
yeah. sorry. Please merge then

PR #2413, opened 2015-09-28
humanize WebDriverBy step arguments

oehlschl:
Although WebDriverBy are acceptable arguments to many of the WebDriver module steps taken by the AcceptanceTester, the arguments are not very human-readable when output during test execution:

* I click "Facebook\WebDriver\WebDriverBy"
* I click "Facebook\WebDriver\WebDriverBy"

This change uses the Locator util to humanize the arguments, making them more readable:

* I click "css selector '.something'"

A unit test (StepTest::testGetArguments()) asserting correctness is also included.

### DIFF
--- a/src/Codeception/Step.php
+++ b/src/Codeception/Step.php
@@ -3,6 +3,7 @@ namespace Codeception;
 
 use Codeception\Lib\ModuleContainer;
 use Codeception\Step\Meta;
+use Codeception\Util\Locator;
 
 abstract class Step
 {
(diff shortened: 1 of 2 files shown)


src/Codeception/Step.php
```diff
@@ -106,6 +107,9 @@ protected function getArgumentsAsString(array $arguments)
 
     protected function parseArgumentAsString($argument)
     {
+        if (class_exists('\Facebook\WebDriver\WebDriverBy') && $argument instanceof \Facebook\WebDriver\WebDriverBy) {
```
DavertMik: this can be simplified with `get_class($argument) == 'Facebook\WebDriver\WebDriverBy' `

DavertMik:
Looks good to me. Please merge with latest 2.1 change to have tests pass

src/Codeception/Step.php
oehlschl: For what it's worth, I followed the same paradigm as https://github.com/Codeception/Codeception/blob/4eb7ccf2f1369998fc3bfdeed896ac4e76f7148a/src/Codeception/Util/Locator.php#L195, but I'll change it if you prefer.

oehlschl:
I rebased, made the change to use get_class() as you requested, and moved the WebDriverBy check into the existing is_object() condition (as get_class() apparently yields an error on non-object arguments). Thanks.

DavertMik:
Thank you

PR #2414, opened 2015-09-29
Refactored grabValueFrom and seeInFields

### LINKED ISSUE #2378: seeInField doesn't see value of disabled form fields
Using the `seeInField` function on a disabled form field or a form field within a disabled field set always leads to an empty value for that form field. This makes it impossible to test for a given value in a disabled field. This did work in version 2.0.12 but is not working anymore in the latest version 2.1.2.

DavertMik:
This should reduce code duplication and allow using `seeInField` and `dontSeeInField` with disabled form fields. Fixes #2378 

@zbateson @Naktibalda please review

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -318,14 +318,14 @@ class InnerBrowser extends Module implements Web, PageSourceSaver, ElementLocato
 
     public function seeInField($field, $value)
     {
-        $crawler = $this->getFieldsByLabelOrCss($field);
-        $this->assert($this->proceedSeeInField($crawler, $value));
+        $nodes = $this->getFieldsByLabelOrCss($field);
+        $this->assert($this->proceedSeeInField($nodes, $value));
     }
 
     public function dontSeeInField($field, $value)
     {
-        $crawler = $this->getFieldsByLabelOrCss($field);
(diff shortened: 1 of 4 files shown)


tests/data/app/db
```diff
@@ -1 +1 @@
-a:0:{}
```
Naktibalda: This file should be added to .gitignore, right?

Naktibalda:
I don't see anything wrong in this change.

tests/data/app/db
DavertMik: probably

src/Codeception/Lib/InnerBrowser.php
```diff
@@ -996,35 +968,45 @@ public function grabValueFrom($field)
...
-                    $val = $option->getAttribute('value');
-                    if (!$isMultiple) {
-                        return $val;
-                    }
-                    $results[] = $val;
-                }
+            $field = new ChoiceFormField($nodes->filter('select')->getNode(0));
+            $options = $nodes->filter('option[selected]');
+            foreach ($options as $option) {
+                $values[] = $option->getAttribute('value');
```
DavertMik: the strange thing that I couldn't use `$field->getValue()` here as it returned all options no matter if they were selected or not :(

Issue #2425, opened 2015-10-03
[InnerBrowser] sendAjaxRequest displays incorrect URL in verbose output


Naktibalda:
It was broken by this change: https://github.com/Codeception/Codeception/commit/57927b1eec8a5bde014f348f29d042c5afa45a2d#diff-fde80abc0dc33fc4bca3b43f571f8934L841

```
Running PhpBrowserTest::testAjax... 
  [Page] http://localhost:8000/
  [Response] 200
  [Cookies] []
  [Headers] {"Host":["localhost:8000"],"Connection":["close"],"X-Powered-By":["PHP/5.5.29-1+deb.sury.org~trusty+3"],"Content-Type":["text/html; charset=UTF-8"]}
  [Page] http://localhost:8000/
  [Response] 200
  [Cookies] []
  [Headers] {"Host":["localhost:8000"],"Connection":["close"],"X-Powered-By":["PHP/5.5.29-1+deb.sury.org~trusty+3"],"Content-Type":["text/html; charset=UTF-8"]}
  [Page] http://localhost:8000/
  [Response] 200
  [Cookies] []
  [Headers] {"Host":["localhost:8000"],"Connection":["close"],"X-Powered-By":["PHP/5.5.29-1+deb.sury.org~trusty+3"],"Content-Type":["text/html; charset=UTF-8"]}
PhpBrowserTest::testAjax  
```

Naktibalda:
Correct output:

```
Running PhpBrowserTest::testAjax... 
  [Page] /
  [Response] 200
  [Cookies] []
  [Headers] {"Host":["localhost:8000"],"Connection":["close"],"X-Powered-By":["PHP/5.5.29-1+deb.sury.org~trusty+3"],"Content-Type":["text/html; charset=UTF-8"]}
  [Page] /info
  [Response] 200
  [Cookies] []
  [Headers] {"Host":["localhost:8000"],"Connection":["close"],"X-Powered-By":["PHP/5.5.29-1+deb.sury.org~trusty+3"],"Content-Type":["text/html; charset=UTF-8"]}
  [Page] /form/complex
  [Response] 200
  [Cookies] []
  [Headers] {"Host":["localhost:8000"],"Connection":["close"],"X-Powered-By":["PHP/5.5.29-1+deb.sury.org~trusty+3"],"Content-Type":["text/html; charset=UTF-8"]}
PhpBrowserTest::testAjax                                                                    Ok
```

PR #2426, opened 2015-10-03
Fixed URL of sendAjaxRequest in verbose output

Naktibalda:
#2425

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -122,13 +122,13 @@ class InnerBrowser extends Module implements Web, PageSourceSaver, ElementLocato
 
         if (!PropertyAccess::readPrivateProperty($this->client, 'followRedirects')) {
             $result = $this->client->request($method, $uri, $parameters, $files, $server, $content, $changeHistory);
-            $this->debugResponse();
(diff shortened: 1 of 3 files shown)


Naktibalda:
Build failed because [Symfony2](https://github.com/Codeception/Codeception/blob/d6cc0cdb73207456717973bee6d78cdea60b4be2/src/Codeception/Module/Symfony2.php#L282) (and [ZF1](https://github.com/Codeception/Codeception/blob/d6cc0cdb73207456717973bee6d78cdea60b4be2/src/Codeception/Module/ZF1.php#L152)) override debugResponse() to provide some extra information.

I also noticed that neither calls parent::debugResponse,
so Symfony2 doesn't print Response code, Cookies and Headers,
and ZF1 doesn't print Page, Response code, Cookies and Headers.

Verbose output from ZF1 looks like this:
```
* I am on route "posts.create"

  [Session] []
```

Observation 2. REST module bypasses framework module and prints its' own debug information, so framework specific information is not printed.

Naktibalda:
Output after change:
```
* I am on route "posts.create"

  [Page] /posts-create
  [Response] 200
  [Cookies] []
  [Headers] []
  [Session] []
```

Why there is a blank line between step and debug info?

PR #2430, opened 2015-10-05
Implemented moveBack method for InnerBrowser

Naktibalda:
@DavertMik You asked me for a way to get out of iframe when I implemented switchToIframe method.
This is one simple way to do that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+#### 2.1.4
+
+* [PhpBrowser][Frameworks] Added moveBack method. By @Naktibalda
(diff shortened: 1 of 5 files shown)


Naktibalda:
Oh, the build failed because TestsForWeb.php is executed by WebDriver too.

I have 3 options:
a) Move tests to another file,
b) Implement multi-step moveBack for WebDriver,
c) I really don't want to do this, but I could remove parameter from InnerBrowser::moveBack  to restrict it to moving only one step at a time,

Naktibalda:
@DavertMik could you restart build in AppVeyor?

PR #2434, opened 2015-10-07
Require phpunit/phpunit ~5.0.0

kornrunner:
PHPUnit 5.0 is the current stable release series.

So, composer.json is updated:
- bump phpunit to ~5.0.0
- bump php to >=5.6.0 (due to phpunit)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,10 +15,10 @@
     "minimum-stability": "stable",
 
     "require": {
-        "php": ">=5.4.0",
+        "php": ">=5.6.0",
         "ext-json": "*",
         "ext-mbstring": "*",
-        "phpunit/phpunit": "~4.8.0",
+        "phpunit/phpunit": "~5.0.0",
         "facebook/webdriver": ">=1.0.1",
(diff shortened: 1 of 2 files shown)


GawainLynch:
In a stable branch?

kornrunner:
You're right, it's a breaking change. I don't mind sending another PR, I guess this should go to 3.0? However, I don't see that branch (and I don't know the plans)

GawainLynch:
Hehe… I am "just a consumer" and my comment sounded a bit rude re-reading it… I have not place to sound like that. Humble appolgies!

DavertMik:
It's not that easy to update to PHPUnit, at least current builds won't pass  https://travis-ci.org/Codeception/Codeception/jobs/84162586
We will update to it once we feel we are ready for it. 

Actually, just wondering, is there any particular feature of PHPUnit 5.0 you want to upgrade for?

devinfd:
+1 to this.
I need to use assertObjectHasAttribute https://phpunit.de/manual/current/en/appendixes.assertions.html#appendixes.assertions.assertObjectHasAttribute which was introduced in version 5

PR #2441, opened 2015-10-12
Save the search path when importing Postgres dumps

EspadaV8:
Currently the search path isn't saved when importing a PostgreSQL dump
which causes problems when trying to fill a database table using `COPY`.
If the search_path isn't reset for the new connection then the import
will try to `COPY` the data into the default schema (usually `public`)
and cause the import to fail with
`pg_query(): Query failed: ERROR:  relation "..." does not exist`

I would consider this a bug fix since my imports are currently failing
without the patch, but it could change some behaviour for badly created
database dumps that always assume the search path for `COPY`s
will be set to the `public` schema. If you'd prefer I could change the
base for the merge to be `master` and not the `2.1` branch.

### DIFF
--- a/src/Codeception/Lib/Driver/PostgreSql.php
+++ b/src/Codeception/Lib/Driver/PostgreSql.php
@@ -9,6 +9,8 @@ class PostgreSql extends Db
 
     protected $connection = null;
 
+    protected $searchPath = null;
+
     public function load($sql)
     {
         $query = '';
@@ -33,6 +35,10 @@ class PostgreSql extends Db
                 if (($pos !== false) && ($pos >= 0)) {
                     $dollarsOpen = !$dollarsOpen;
                 }
+            }
+
+            if (preg_match('/SET search_path = .*/i', $sqlLine, $match)) {
+                $this->searchPath = $match[0];
             }
 
             $query .= "\n" . rtrim($sqlLine);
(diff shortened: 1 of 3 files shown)


Naktibalda:
:+1: 
There are no reports about this or similar issue.
It is unlikely to affect any users, and if someone is affected, they can quickly fix their dump files, right?

EspadaV8:
Yeah, they can add in a `SET search_path = public, pg_catalog;` line before the `COPY` commands. I'm not sure how people would get a broken SQL file since `pg_dump` includes all the `SET search_path` commands.

As an aside, I was trying to create a test for this with an SQL file that includes different `search_path`s but I was unable to run the tests because they were being skipped. Is there some documentation on how to run the PostgreSQL tests?

Naktibalda:
Make sure that your postgres server is running and create a passwordless user named postgres and codeception_test database.

Or set valid credentials in https://github.com/Codeception/Codeception/blob/2.1/tests/unit/Codeception/Lib/Driver/PostgresTest.php#L10

EspadaV8:
I've just added a simple test to make sure that the dump imports correctly where there's a `SET search_path` other than `public`. I've tried to keep the dump changes to a minimum but if there's anything you'd like me to remove/update just let me know.

DavertMik:
@Naktibalda I hope I could know anything about Postgres ) If you feel it is ok - please merge this PR.

PR #2445, opened 2015-10-12
Reset leaves previous file uploads in UploadedFile

tibee:
If `$_FILES` is emptied, `\yii\web\UploadedFile::reset()` must be also called. It has a static property `$_files` that keeps previous upload even if `$_FILES` is reset.


DavertMik:
@samdark please review

Naktibalda:
The change looks good to me and matches [documentation](http://www.yiiframework.com/doc-2.0/yii-web-uploadedfile.html#reset%28%29-detail).

PR #2452, opened 2015-10-15
Corrected Specify code example

Naktibalda:
Copied from https://github.com/Codeception/codeception.github.com/pull/49


DavertMik:
yeah, you are right

PR #2454, opened 2015-10-16
Fix #2453 default parameter value

### LINKED ISSUE #2453: Wrong default parameter value for Actor methods'
Irrespective of what is defined in the module from which these methods are derived, the value assumed in the parameter lost of those grafted methods always assume null.
Going a little further with the reflection based grafting that occurs during the build step, it is possible to make it better. ( I have a working proof of concept ).

codemedic:


codemedic:
Please close this one; PR was based on the wrong branch.

PR #2455, opened 2015-10-16
Fix #2453 default parameter value

### LINKED ISSUE #2453: Wrong default parameter value for Actor methods'

codemedic:
Added reflection based parameter inspection to derive default values

### DIFF
--- a/src/Codeception/Lib/Generator/Actor.php
+++ b/src/Codeception/Lib/Generator/Actor.php
@@ -100,7 +100,7 @@ EOF;
         foreach ($refMethod->getParameters() as $param) {
 
             if ($param->isOptional()) {
-                $params[] = '$' . $param->name . ' = null';
+                $params[] = '$' . $param->name . ' = '.$this->getDefaultValue($param);
             } else {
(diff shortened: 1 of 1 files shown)


codemedic:
The fix here is very similar for v2.1 and v2.0.
Would appreciate if I am given guidance on how to go about fixing this for those versions, especially 2.0, which I use for my projects.

Naktibalda:
There is no difference between src/Codeception/Lib/Generator/Actor.php in 2.1 and master.
Please create a new branch based on 2.1 and make a new pull request.

There will be no more releases of 2.0 branch, but if you make a pull request for it, it will be accepted.

codemedic:
Hi Gintautas

Thanks ... I have now submitted two different PRs, one each for the two
versions.

Regards
DK

Issue #2334, opened 2015-09-02
Display provided data in test report


Naktibalda:
When PhpUnit uses data provider, it displays parameters in test result,
```
1) DataTest::testAdd with data set #3 ('1', '1', '3')
Failed asserting that 2 matches expected '3'.
```
https://phpunit.de/manual/current/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers

When Codeception uses data provider, it displays this:
```
1) Test add  |  #3  (DataTest::testAdd))   Fail
Failed asserting that 2 matches expected '3'.
```
It would be useful to see parameters in report.

DavertMik:
I disabled showing data on test running: sometimes text with data was to long to display, also it is not readable when we use complex data structures: array or objects. But yes, we can improve that for results

Naktibalda:
Since then I learned that the key of data provider array is displayed in the test result.
Giving meaningful names test items is good enough to me.

PR #2456, opened 2015-10-17
Fix for output string length

mkudenko:


Naktibalda:
The change looks like the right thing to do.
But could you describe what issue does it fix?

mkudenko:
The issue is that the output string is being cut off when using data providers.

Before:
screenshot from 2015-10-18 22 07 42

After:
screenshot from 2015-10-18 22 08 04

Naktibalda:
What kind of tests produce such output?

mkudenko:
Those are PSR-4 namespaced PHPUnit tests.

Naktibalda:
I am asking, because I raised this issue: https://github.com/Codeception/Codeception/issues/2334

I created this PhpUnit test now:
```
class dataProvidersTest extends \PHPUnit_Framework_TestCase
{
    public function provideData()
    {
        return [
            ["aaa"],
            ["bbb"],
            ["ccc"],
        ];
    }

    /**
     * @dataProvider provideData
     */
    public function testMe($a)
    {
    }
}
```

and got this output:
```
Experiment Tests (1) ------------------------
dataProvidersTest::testMe | #0         Ok
dataProvidersTest::testMe | #1         Ok
dataProvidersTest::testMe | #2         Ok
---------------------------------------------
```

It doesn't look like yours.

mkudenko:
The output looks different because I'm adding descriptions to each data set, like this:

```php
class dataProvidersTest extends \PHPUnit_Framework_TestCase
{
    public function provideData()
    {
        return [
            'triple a' => ["aaa"],
            'triple b' => ["bbb"],
            'triple c' => ["ccc"],
        ];
    }

    /**
     * @dataProvider provideData
     */
    public function testMe($a)
    {
    }
}
```

The output without the fix:

```
Unit Tests (1) ------------------------------
UnitTests\dataProvidersTest::testMe |  Ok
UnitTests\dataProvidersTest::testMe |  Ok
UnitTests\dataProvidersTest::testMe |  Ok
---------------------------------------------

```

The output with the fix:

```
Unit Tests (1) -------------------------------------------------------------------------
UnitTests\dataProvidersTest::testMe | "triple a"                                  Ok
UnitTests\dataProvidersTest::testMe | "triple b"                                  Ok
UnitTests\dataProvidersTest::testMe | "triple c"                                  Ok
----------------------------------------------------------------------------------------
```

PR #2458, opened 2015-10-19
Fixing issue #2453 - Default param is always null

codemedic:
Added reflection based parameter inspection to derive default values; falls back to null.


DavertMik:
Good work, thanks!

Issue #2459, opened 2015-10-19
Wrong default parameter value for Actor methods' (clone of #2453 for v2.0)


codemedic:
See #2453 for more details

PR #2460, opened 2015-10-19
Fixing issue #2459 - Default param is always null (v2.0)

codemedic:
Added reflection based parameter inspection to derive default values; falls back to null.

### DIFF
--- a/src/Codeception/Lib/Generator/Actor.php
+++ b/src/Codeception/Lib/Generator/Actor.php
@@ -237,15 +237,8 @@ EOF;
 
     private function getDefaultValue(\ReflectionParameter $param)
     {
-        $isPlainArray = function ($value) {
-            return ((count($value) === 0)
-                || (
-                    (array_keys($value) === range(0, count($value) - 1))
-                    && (0 === count(array_filter(array_keys($value), 'is_string'))))
-            );
-        };
         if ($param->isDefaultValueAvailable()) {
-            if (method_exists($param, 'isDefaultValueConstant ') && $param->isDefaultValueConstant()) {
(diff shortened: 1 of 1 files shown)


codemedic:
Can see some red marks here ... are they OK?
Is there anything I can / should do to fix them?

Naktibalda:
It isn't your fault.

src/Codeception/Lib/Generator/Actor.php
```diff
@@ -236,5 +235,45 @@ public function getNumMethods()
...
+                    if (in_array($class, ['self', 'static'])) {
+                        $constName = $param->getDeclaringClass()->getName().'::'.$const;
+                    }
+                }
+
+                return $constName;
+            }
 
+            $paramValue = $param->getDefaultValue();
+            switch (true) {
```
Naktibalda: Is there a difference between this switch and var_export() ?

codemedic: I think you are right!

Only possible difference is the array and string formatting.
An array gets put into one line and gets simplified if it was a plain simple one dimensional array.
In the case of string, it is escaped; all escape sequences get converted to ASCII by `var_export` as well as `print_r`.

I can change it if you prefer.

I think that function could do with some unit testing. Is there a specific way doing unit tests in codeception?

codemedic: Wonder if it would be worth just to use one of the already existing php-encoder libraries.
I was playing around with `riimu/kit-phpencoder` seems to cover most cases.

If you are OK with using it, I can redo the PR with this library.

codemedic:
@Naktibalda thanks for your comments; I have updated the PR with a new version, which I think, will sit in the middle ground.

PR #2462, opened 2015-10-19
Fix a root cause hiding error when Symfony2 is configured wrong

mhightower:
@todo Should leave it up to the user to update the php.ini instead of
setting it in code. This overwrote my correct settings.

This ini change caused an error when I used Codeception/Specify library.


DavertMik:
Makes sense. Thank you

PR #2465, opened 2015-10-20
seeInSource method for InnerBrowser/WebDriver

zbateson:
Checks the raw response rather than the tag stripped/html decoded
result.


DavertMik:
:+1: thanks

PR #2471, opened 2015-10-22
Print Selenium logs on failure

MasonM:
WebDriver provides access to the logs for both the client and server, which are sometimes necessary
to debug test failures. This commit modifies Codeception\Module\WebDriver->_failed() to grab all
available logs and print them in human-readable format. The format is identical to what Selenium
Server uses for displaying logs in the console.


DavertMik:
Wow, looks great. Thank you! I will try it at home )

DavertMik:
I'd cut that to 10 lines and not to display all the logs on end. It's just too much for a screen.... Especially if this is a long and complex test we have the output buffer to be filled overfilled with them

DavertMik:
10 is not enough. I moved that to configuration. You can change the value with

`debug_log_entries`: 15

PR #2473, opened 2015-10-23
Improved documentation for PhpBrowser->submitForm()

SimonEast:
I found that `submitForm()` was a little confusing to me at first, so I've updated the documentation to hopefully help out others.

### DIFF
--- a/docs/modules/PhpBrowser.md
+++ b/docs/modules/PhpBrowser.md
@@ -1017,30 +1017,19 @@ $I->amOnPage('test-headers.php');
 Submits the given form on the page, optionally with the given form
 values.  Give the form fields values as an array.
 
-Although this function can be used as a short-hand version of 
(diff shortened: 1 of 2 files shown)


SimonEast:
Oh, actually, perhaps I need to make changes to the actual source code of the function, not the doc file, is that correct?  I'll have to update my PR.

DavertMik:
> Oh, actually, perhaps I need to make changes to the actual source code of the function, not the doc file, is that correct? I'll have to update my PR.

Yes, you need to change that description in phpdoc of WebInterface

SimonEast:
OK, I've updated the patch and put the documentation into the correct place this time. :-)  Please let me know if it's all OK.

Should I also regenerate the MD doc files somehow?  Or can someone do that for me?

Simon.

zbateson:
Yeah, the MD files get generated automatically when a release is pushed out - thanks again I think it's much better.

PR #2478, opened 2015-10-23
Fix for broken paths introduced after Symfony upgraded to 2.7.5.

supre:
We upgraded our symfony package to 2.7.5, only to find out that our codeception is broken.
After a whole day of debugging it turns out the issue was related to the recent modifications made to the Finder component for Symfony ([#15826](https://github.com/symfony/symfony/issues/15826), [#15802](https://github.com/symfony/symfony/issues/15802), [#15803](https://github.com/symfony/symfony/issues/15803)).

In the path to the test in the suite (in our case, we have a test for each bundle), an additional `/` is added between the test and the suite, thus causing an issue in GroupLoader. i.e.:

Line 109 of file Lib/GroupManager:
````
if ($filename == $testPattern) {
    $groups[] = $group;
}
````
We were getting the following scenario:
$filename: /path/to/suite//path/to/test
$testPattern: /path/to/suite/path/to/test

And, the if condition was failing in all cases, which meant that none of our groups was working.

The provided fix will fix the path in case it's broken but will leave it unchanged in case it's fine.


raistlin:
Looks good to me.

DavertMik:
Actually I didn't understand where actually this additional `/` is added to path. 
To merge this PR I need clearly understand what this fix does and is that the optimal one. From what I see it is better to find where `//` is added and remove this redundant `/` in place

raistlin:
The additional "/" seems to be added by Syfmony, I've also seen this in earlier versions of Syfmony, but never caused a problem to me.

DavertMik:
Ok, @raistlin then we are merging it. Thanks

PR #2486, opened 2015-10-25
getInternalDomains optimization - cache domains, reset in _beforeSuite

Naktibalda:


janhenkgerritsen:
I'll look into that Laravel error when I have some time.

Naktibalda:
Don't they happen because this branch is behind 2.1 ?

janhenkgerritsen:
I think so yes, but haven't checked it.

PR #2487, opened 2015-10-25
Add support for nested helpers

luka-zitnik:
My team uses nested helpers. What running `codecept g:helper "A\B"` creates is tests/_support/Helper/A/B.php with the following content

```php
<?php
namespace Helper;

class A\B extends \Codeception\Module
{

}
```
It would be good if it could read like this instead,

```php
<?php
namespace Helper\A;

class B extends \Codeception\Module
{

}
```

### DIFF
--- a/src/Codeception/Command/GenerateHelper.php
+++ b/src/Codeception/Command/GenerateHelper.php
@@ -13,6 +13,7 @@ use Symfony\Component\Console\Output\OutputInterface;
  * Creates empty Helper class.
  *
  * * `codecept g:helper MyHelper`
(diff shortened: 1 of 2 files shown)


luka-zitnik:
I believe this belongs to 2.1 instead. I will make another PR.

luka-zitnik:
From which branch should I be branching off? I see people mostly use 2.1, but docs say

> Bugfixes should be sent to to current stable branch, which is the same as major version number.

which I understand is 2.0.

samdark:
2.1

luka-zitnik:
Thanks!

Issue #2490, opened 2015-10-27
Container (and kernel) is not loaded in the _beforeSuite() hook


Franua:
I faced a situation when I needed to load and use some of my company's Symfony2 project's services in _beforeSuite() to prepare some needed data, obviously, before suit is running. I couldn't do that because current implementation loads kernel and set up container only before a test (in _beforeTest() method).

Naktibalda:
Implemented in  #2491

PR #2491, opened 2015-10-27
Move kernel booting and container set up into _initialize() method

Franua:
I faced a situation when I needed to load and use some of my Symfony2 project's services in _beforeSuite() to prepare some needed data, obviously, before suit is running. I couldn't do that because current implementation loads kernel and set up container only before a test (in the _beforeTest() method).
I also created an issue about that (according to https://github.com/Codeception/Codeception/wiki/Git-workflow-for-Codeception-contributors) - https://github.com/Codeception/Codeception/issues/2490
Tests passed with my changes.

### DIFF
--- a/src/Codeception/Module/Symfony2.php
+++ b/src/Codeception/Module/Symfony2.php
@@ -114,12 +114,13 @@ class Symfony2 extends Framework implements DoctrineProvider, SupportsDomainRout
         if (ini_get($xdebugMaxLevelKey) < $maxNestingLevel) {
             ini_set($xdebugMaxLevelKey, $maxNestingLevel);
         }
-    }
 
-    public function _before(\Codeception\TestCase $test) 
-    {
         $this->bootKernel();
         $this->container = $this->kernel->getContainer();
+    }
+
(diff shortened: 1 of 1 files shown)


tiger-seo:
Afaiu, with beforeTest kernel is a brand new for each test, so tests doesn't influence each other, but with beforeSuite there will be only one kernel for all tests, so it might lead to potential issues

tiger-seo:
Ok, it's not about kernel, but client, nonetheless, the idea is the same

Franua:
Client's initialization remains in _before(). I moved only kernel loading and container's initialization into __initialize().

Franua:
Kernel loading is based (dependent) on config. I don't think config is being changed every test.

Naktibalda:
The tests pass, so this change is not breaking them.

Laravel modules instantiate app and client in _initialize.

If this change is accepted, I will do the same for ZF1 and ZF2.

Naktibalda:
All tests are using the same instance of kernel already,
bootKernel only creates a new instance if $this->kernel is null.

PR #2494, opened 2015-10-27
Add support for nested helpers

luka-zitnik:
This is the same patch from the abandoned #2487, only applied to proper branch.


DavertMik:
Thank you!

PR #2495, opened 2015-10-27
Implemented getInternalDomains for ZF1

Naktibalda:
Includes #2486


Naktibalda:
ZF1 change is good, but I actually lost the "cache internal domains between requests" change during the merge.

PR #2500, opened 2015-10-29
BugFix: string|null as well as null|string can be used to specify the value type

vslovik:
BugFix: In \Codeception\Module\REST::seeResponseMatchesJsonType now can be used 'string|null' as well as 'null|string' to specify the type


dogawaf:
:+1:

DavertMik:
Good. Thank you

PR #2503, opened 2015-10-29
Silex Doctrine provider

arduanov:
It's work for me, but I think code needs some testing.

### DIFF
--- a/src/Codeception/Module/Silex.php
+++ b/src/Codeception/Module/Silex.php
@@ -4,6 +4,7 @@ namespace Codeception\Module;
 use Codeception\Configuration;
 use Codeception\Exception\ModuleConfigException;
 use Codeception\Lib\Framework;
+use Codeception\Lib\Interfaces\DoctrineProvider;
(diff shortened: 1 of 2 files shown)


src/Codeception/Module/Silex.php
```diff
@@ -23,6 +24,7 @@
  * ## Config
  *
  * * app: **required** - path to Silex bootstrap file.
+ * * em_service: 'db.orm.em' - use the stated EntityManager to pair with Doctrine Module.
  *
  * ### Bootstrap File
```
DavertMik: Could you provide here example configuration for Silex + Doctrine2? This would be good for newbies to understand how to use them together.

DavertMik:
Thanks. This would be very useful. 
For additional testing... Can you get some really basic Silex + Doctrine app to be created so we could have it as example, and use it for our Travis tests?

DavertMik:
Anyway, I'm merging. But once again, If you will have some time, please get a sample app of Silex + Doctrine running. 

Thanks

PR #2504, opened 2015-10-29
fix double cache directories

arduanov:


DavertMik:
Oh, could you write something about that. What was the issue?

PR #2512, opened 2015-11-02
Make generate:suite respect bootstrap setting

dmitrivereshchagin:
Suite generation command always creates bootstrap script named `_bootstrap.php` regardless of user settings. Fortunately it's easy to fix.


DavertMik:
Thanks!

sergeyklay:
Good catch :+1:

PR #2526, opened 2015-11-09
Clarified documentation on the inner workings of `see()` and `dontSee()`

SimonEast:
Just added some better documentation to clarify how `see()` and `dontSee()` work, especially in PhpBrowser.  I hope this is the right place to put them.


DavertMik:
@zbateson I think it works in WebDriver just the same way, right?

zbateson:
Yeah, I think part of those was bringing them in line with how it worked in WebDriver.  Should be all good :) thanks for the contribution.

SimonEast:
Thanks so much!

Can you also let me know if #2473 is now suitable for merging, or if it needs any tweaks?

Issue #2519, opened 2015-11-05
ZF2 + Doctrine2: Entity is saved but finding it returns no result


dranzd:
ZF2 + Doctrine2: Entity is saved but finding it returns no result

I have a sample test below which fails when I try to retrieve the persisted entity
using the entity manager taken from ZF2's service manager.  The entity is persited
and verified that the ID is generated along with the data included.  But when finding
for that entity either by ID lookup or other field lookup, it fails to return the
persisted entity.

    // configuration
    class_name: FunctionalTester
    modules:
        enabled:
            # add framework module here
            - \Helper\Functional
            - Asserts
            - ZF2
            - Doctrine2:
                depends: ZF2

    // test
    class UserCest
    {
        ...

        public function testSaveAndRetrieveUser(FunctionalTester $I)
        {
            $app = $I->getModuleOf('ZF2')->application;

            // mapper that points to My\Entity\User, table is using InnoDB
            $user = new User();
            $I->persistEntity($user, [
                'name'       => 'User Name',
                'createdOn'  => new \Datetime(),
                'modifiedOn' => new \Datetime(),
            ]);

            $I->seeInRepository('My\Entity\User', [
                'name' => 'User Name',
            ]); // this passes with ID generated

            // $em here is a different instance from Doctrine2::$em
            $em = $app->getServiceManager()->get('Doctrine\ORM\EntityManager');
            $userRepository = $em->getRepository('My\Entity\User');

            $users = $userRepository->findBy(['name' => 'User Name']);

            $I->assertEquals(count($users), 1);  // this fails and returns 0
        }
    }

I was trying to debug it and found out that the entity manager ZF2 holds is different
from the one Doctrine2 holds.  I don't know if having 2 instances of entity manager 
makes a difference when one is stored in the Doctrine 2 instance and looking up with 
ZF2's instance.

I tried to update the code at ZF2 modules (ZF2::_getEntityManager) from: 

    public function _getEntityManager()
    {
        return $this->grabServiceFromContainer('Doctrine\ORM\EntityManager');
    }

to: 

    public function _getEntityManager()
    {
        if ($this->application) {
            return $this->application->getServiceManager()->get('Doctrine\ORM\EntityManager');
        }
        return $this->grabServiceFromContainer('Doctrine\ORM\EntityManager');
    }

and this makes the test pass.

I don't know if this is a bug or I just don't have my configurations entered correctly.  If
anyone has any idea where I am off, any help is appreciated.

[...]

PR #2529, opened 2015-11-09
Grab service from the current app and not initialize a new one

dranzd:
* fix issue #2519 where Doctrine2 gets different instance of the
   entity manager everytime ZF2's grabServiceFromContainer method
   is called


Naktibalda:
Thank you, this time it looks good.

Issue #2537, opened 2015-11-12
lambda function


neronmoon:
If i pass any array with 2 elements codeception step will be:
I blabla "lambda function".

I see, that Step class checks only syntax. 
```php
if (is_callable($argument, true)) {
    return 'lambda function';
}
```

Why?

Naktibalda:
Why would you prefer to get instead?

Naktibalda:
Put is_array check first, make a pull request.

Naktibalda:
Answer - because nobody reported an issue before.

I think that changing that condition to `if ($argument instanceof \Closure)` would be the best solution.

Naktibalda:
I have a crazy issue (maybe two) with "Lambda function".

My test code is:
`$I->seeResponseContainsJson(['08712222233', '08712222244']);`

The step output is:
`* I see response contains json "lambda function"`

Then I added `codecept_debug(is_callable(['08712222233', '08712222244']));` before assertion
and got this error: `[PHPUnit_Framework_Exception] include_once(08712222233.php): failed to open stream: No such file or directory`

It actually tries to autoload the first item as a class.
I can explain why I got warning in second case, but I can't explain "lambda function" in the first one.

Naktibalda:
Ah, the difference was in the second argument - is_callable($array, true) returns true for any array containing two strings, is_callable($array, false) tries to load the class.

I was actually using an older version of Codeception than I expected, 2.1.0 prints "lambda function", 2.1.5 kills my tests.

I will change it to return "lambda function" only if the first item of array is an object, in other cases it is better to output content of array (and avoid using is_callable).

Naktibalda:
I changed "lambda function" to "Closure" and "callable function" to content of array (object replaced with class name) for 2.2: https://github.com/Codeception/Codeception/pull/2777/files#diff-c1176bf119a1fb6370de596de823689eL20

I will backport this change to 2.1

Naktibalda:
https://github.com/Codeception/Codeception/pull/2778

PR #2538, opened 2015-11-12
Lambda function step output

neronmoon:
#2537 fix

[Review on Reviewable](https://reviewable.io/reviews/codeception/codeception/2538)

### DIFF
--- a/src/Codeception/Step.php
+++ b/src/Codeception/Step.php
@@ -115,8 +115,11 @@ abstract class Step
                 return Locator::humanReadableString($argument);
             }
         }
-        if (is_callable($argument, true)) {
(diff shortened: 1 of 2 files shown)


src/Codeception/Step.php
```diff
@@ -115,9 +115,12 @@ protected function parseArgumentAsString($argument)
                 return Locator::humanReadableString($argument);
             }
         }
-        if (is_callable($argument, true)) {
+        if ($argument instanceof \Closure) {
```
neoascetic: Is it enough to check `is_callable`, isn't it?

neronmoon: Yeah, but i did this to get more detail about function. Why not?

Issue #2501, opened 2015-10-29
\Codeception\Util\JsonType filter delimiter ":" does not allow to use ":" in regex filter


vslovik:
`````` $I->seeResponseMatchesJsonType([
"created_at" => "string:regex(~\d{2}\/\d{2}\/\d{4} \d{2}:\d{2}:\d{2}~)"
]);```

will cause an error for {"created_at":"29/10/2015 08:15:44"}
because of explode by ":" that will break regex pattern,
but it's wrong.

I can use custom filter like a workaround, but the documentation confuses
me hiding the fact that ":" can not be used in regex.
``````

niclopez:
Hi,

I've not been able to use the "|" symbol, too inside a regular expression. In my case, the problem is located around line 120: (\Codeception\Util\JsonType)
...
$matchTypes = explode ('|', $type);
...

I'm trying to do something similar to this with no success:

$I->seeResponseMatchesJsonType(array(
  'code' => 'integer:regex(/100|300/)', // unable to use '|' inside regex...
));

Thanks Davert for your great job!!

niclopez:
Hi again Davert & Vslovik,

I found a solution for this issue. Replacing "explode" statement for "preg_split" alternative works well, too. Doing this, string located inside regex() parenthesis is excluded and regex patterns does his job.

To solve this, just replace following lines at "\Codeception\Util\JsonType" file:

line 120 (pipe character):
// $matchTypes = explode('|', $type);
$matchTypes = preg_split("#(?![^]\(]*\))\|#", $type);

line 127 (colon character):
// $filters = explode(':', $matchType);
$filters = preg_split("#(?![^]\(]*\))\:#", $matchType);


NLopez

niclopez:
My bad, I copy/paste the wrong regex version:

line 120 (pipe character):
// $matchTypes = explode('|', $type);
$matchTypes = preg_split("#(?![^]\(]*\))\|#", $type);

line 127 (colon character):
// $filters = explode(':', $matchType);
$filters = preg_split("#(?![^]\(]*\))\:#", $matchType);

Sorry

niclopez:
Hi @DavertMik,

is there any option to include this fix in your repository code?

thanks¡

Naktibalda:
Make a pull request.

https://github.com/Codeception/Codeception/blob/2.1/CONTRIBUTING.md

Naktibalda:
@niclopez your patch is an improvement, but it doesn't solve all issues.

[...]

PR #2548, opened 2015-11-17
Replacing explode functions with preg_split alternative

niclopez:
Replacing explode functions with preg_split alternative. 
https://github.com/Codeception/Codeception/issues/2501

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name":"codeception/codeception",
+    "name":"niclopez/codeception",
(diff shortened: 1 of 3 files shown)


Naktibalda:
Could you add some tests to https://github.com/Codeception/Codeception/blob/2.1/tests/unit/Codeception/Util/JsonTypeTest.php to demonstrate the issue that your change fixes?

niclopez:
Of course!

Some tests have been done. My proposed fix allows to pass them successfully.

Issue #2550, opened 2015-11-18
Groups by direcories doesn't work correctly.


kimvladis:
I'm trying to add these groups: 
```
groups:
    mobile: [tests/acceptance/Mobile]
    frontend: [tests/acceptance/Frontend]
```

Next I try to execute tests by one of this groups
```
codecept run -g frontend
```

But codecept skips this group, as though there was nothing there.

PR #2551, opened 2015-11-18
Bugfix groups by directory

kimvladis:
Bugfix issue #2550
Replaces double slash by one.

### DIFF
--- a/src/Codeception/Lib/GroupManager.php
+++ b/src/Codeception/Lib/GroupManager.php
@@ -96,7 +96,7 @@ class GroupManager
             if (isset($info['class'])) {
                 $groups = array_merge($groups, \PHPUnit_Util_Test::getGroups($info['class'], $info['name']));
             }
-            $filename = $info['file'];
+            $filename = preg_replace('{//}', '/', $info['file']);
         } else {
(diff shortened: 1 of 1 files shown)


Naktibalda:
Thanks @kimvladis 

@DavertMik Why are these double slash issues happening all over the place?

DavertMik:
Thanks!

> Why are these double slash issues happening all over the place?

Looks like  it is related to Symfony Finder. I didn't have chance to investigate this deeper

kimvladis:
@DavertMik yes it is. Symfony Finder adds one excess slash when concat $rootPath and $filePath. Seems to bug in Symfony Finder.

VoidWalker:
Looks like there is a rework, guys
The fix is good for Unix systems, but fro Windows, where there is a backward slashes, it doesn't
Path still has double slashes after preg_replase:
D:\autotests\projects\uv-fashions-autotests\tests\acceptance\\LoginCest.php

As a meantime workaround you may specify path to tests like:
groups:
    custom: [tests/acceptance//LoginCest.php:loginLogout]

d__autotests_projects- _codecept phar_src_codeception_lib_groupmanager php - phpstorm 10 0 1 2015-12-29 11 34 00

PR #2572, opened 2015-11-23
Memcache module: bug fix "memcache" extension init

pfz:
Prevent this from happening : 
[PHPUnit_Framework_Exception]
  MemcachePool::close(): No servers added to memcache connection


Naktibalda:
Please close connection in `_after`.

pfz:
added. Thanks for the review.

DavertMik:
Wow, it was used actually!

pfz:
indeed :)

PR #2578, opened 2015-11-25
Use json_last_error_msg in seeResponseIsJson

Naktibalda:
Use the same message as in https://github.com/Codeception/Codeception/blob/4ee596129d5d41dc69b0ddf13c162a86037b3b7f/src/Codeception/Util/JsonArray.php


DavertMik:
Thank you

PR #2588, opened 2015-11-28
allow appendField to work with content editable div

nsanden:


nsanden:
No idea why the test in the build is failing because its seeing the text "moonrisesunrise" instead of "sunrisemoonrise" like I get when i test locally. Any ideas?

ss

DavertMik:
This may be related to Selenium or Firefox version. Which version of Selenium Server do you use locally?

nsanden:
I'm using selenium-server-standalone-2.48.2 and Firefox 38.0.5

I see the build is using 2.42.2. I will see if I can downgrade (just to test).

nsanden:
Seems to be Firefox. Travis build is using 31.0esr and when I downgrade to that locally (regardless of selenium version) my test does fail.

DavertMik:
You could press key "END" before entering keys into area, to ensure the caret is in the end

nsanden:
Your END key idea worked great. When I add that in, the test passes fine (locally). Remove it, test fails. As expected.

Once again though, I'm stumped as to why Travis is failing. 

I'm using exactly the same Firefox and Selenium versions. If I add in wait times, I can clearly see the test do exactly as I expect it to. Caret goes to end of div. Appends the text, etc.

DavertMik:
Didn't work for me locally as well. I will try to play around to find a way to make this work for everyone.
Thanks

DavertMik:
I tried different combinations but still not working on my firefox. Let's leave it as experimental feature and skip the test for now.

DavertMik:
Looks like I was using old Selenium Server 2.47. I think this should work fine on 2.48

DavertMik:
Yes!

PR #2590, opened 2015-11-28
PHP7 support

DavertMik:
Actually everything worked for PHP7 with no changes. This PR enables testing for PHP7 and adds syntax checking for loaded tests. For PHP < 7 we use simple "php -l" linter.


DavertMik:
squashed the commits. Please review

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 #### 2.1.5
 
+* **PHP7 support** verified )
```
samdark: `verified )`?

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
...
+* Syntax check for tests. If PHP7 is used, `ParseException` handles syntax error, otherwise linting happens with `php -l`. HHVM - check is skipped. @davertmik
```
samdark: Why? HHVM should support `php -l` as well.

src/Codeception/Lib/Parser.php
```diff
@@ -125,9 +124,33 @@ protected function addCommentStep($comment)
         $this->scenario->addStep(new \Codeception\Step\Comment($comment, []));
     }
 
+    public static function validate($file)
+    {
+        if (defined('HHVM_VERSION') // not for HHVM
+            || (strtoupper(substr(PHP_OS, 0, 3)) === 'WIN') // not for WIN
```
samdark: `php -l` works well under Windows.

tests/unit/Codeception/Lib/ParserTest.php
```diff
@@ -150,4 +150,43 @@ public function testParseFileWhichUnsetsFileVariable()
...
     }
-}
\ No newline at end of file
+    /**
+     * @group core
+     * @throws \Codeception\Exception\TestParseException
+     */
+    public function testCeptValidation()
+    {
+        if (defined('HHVM_VERSION') // not for HHVM
```
samdark: It should work both with HHVM and Windows. `php -l` is supported in both cases.

CHANGELOG.md
DavertMik: I was not sure about this... Is that really so?

src/Codeception/Lib/Parser.php
DavertMik: Yes, it works but Appveyor fails there. Somehow it doesn't handle well the error from another process
https://ci.appveyor.com/project/DavertMik/codeception/build/1.0.1020#L504

CHANGELOG.md
samdark: Yes according to docs. Travis will show you.

DavertMik: Ok, HHVM is not that evil )

src/Codeception/Lib/Parser.php
samdark: https://stackoverflow.com/questions/9114574/php-exec-does-not-return-output

DavertMik:
Enabled linting for HHVM as suggested by @samdark 
If no objections - will megre that in next hours

samdark:
It's still unclear what https://github.com/Codeception/Codeception/pull/2590/files#r46268974 means.

samdark:
Docs here https://github.com/Codeception/Codeception/pull/2590/files#r46269016 should be adjusted.

samdark:
And... how about Windows?

DavertMik:
For Windows we have 2 options:

* skip the appveyor test if it is failing
* skip windows validation

If we will have linter issues (like "php -l" not getting executed) that probably will make Codeception unusable on Windows. If we just ignore linting for windows it will work as it is now...

samdark:
@DavertMik it's possible to solve linting on Windows: https://stackoverflow.com/questions/9114574/php-exec-does-not-return-output

firodj:
@DavertMik We are using HHVM 3.9.1 which are defined as `PHP_MAJOR_VERSION=5`.
As our tests is growing, the start up time when running tests also increasing significantly.
After investigated we found that the lint validation took up about ~120 ms per test file, and 4 minutes total for all files just for linting.

Can this validation as optionally disabled?

Naktibalda:
@firodj Could you raise it as a separate issue?

PR #2593, opened 2015-11-30
Fix Laravel5 module seeCurrentRouteIs crash

maddhatter:
`seeCurrentRouteIs()` will throw an exception if the routes don't match:

`[ErrorException] Object of class Illuminate\Routing\Route could not be converted to string`

It should use the `$currentRouteName` instead of the route object when creating the error message.


janhenkgerritsen:
Thanks!

PR #2595, opened 2015-11-30
Fix bad call to PHPUnit_Framework_Assert::pageContains

LeeKevin:
Updated with PHPUnit_Framework_Assert::contains


DavertMik:
ok

PR #2598, opened 2015-12-01
Add ability to control redirect following in the REST module

brutuscat:
Enables a pair of methods to control automatic redirection following by the REST client, like so:

```php
$I->stopFollowingRedirects();
// ...
$I->startFollowingRedirects();
```

### DIFF
--- a/src/Codeception/Module/REST.php
+++ b/src/Codeception/Module/REST.php
@@ -1087,4 +1087,21 @@ EOF;
     {
         throw new ModuleException($this, "This action was deprecated in Codeception 2.0.9 and removed in 2.1. Please use `grabDataFromResponseByJsonPath` instead");
     }
+
+    /**
+     * Prevents automatic redirects to be followed by the client
+     */
+    public function stopFollowingRedirects()
+    {
+        $this->client->followRedirects(false);
+    }
+
+    /**
+     * Enables automatic redirects to be followed by the client
+     */
+    public function startFollowingRedirects()
+    {
+        $this->client->followRedirects(true);
+    }
+
 }


Naktibalda:
Don't use `$this->getRunningClient()`, it checks if request was made already and makes no sense in your new methods.
Use `$this->client` directly.

It would be good if you could implement some tests for your methods in https://github.com/Codeception/Codeception/blob/2.1/tests/unit/Codeception/Module/PhpBrowserRestTest.php

brutuscat:
@Naktibalda test what? that true is true? or false is false? no need to test I guess... All the functional test have been done by https://github.com/symfony/browser-kit/blob/465db80e663fff1506c4128b272789bfddedce65/Tests/ClientTest.php#L328-L497

I'll fix the `getRunningClient`.

brutuscat:
@Naktibalda updated

Naktibalda:
It probably is good now.
But you thought so the first time.

I actually wanted a full functional test with redirect, but checking that true is true would be enough.

hubbyist:
Is this pul request merged yet. I need this one.

hubbyist:
Codeception version 2.1.4

Defining these functions in acceptance helper is not working.
file : /tests/_support/Helper/Acceptance.php
```php
namespace Helper;

class Acceptance extends \Codeception\Module {

	/**
	 * Prevents automatic redirects to be followed by the client
	 */
	public function stopFollowingRedirects(){
		$this->client->followRedirects(false);
	}

	/**
	 * Enables automatic redirects to be followed by the client
	 */
	public function startFollowingRedirects(){
		$this->client->followRedirects(true);
	}
}
```
also it seems as if Acceptance helper is not being build by build command

```
user@host $ php codecept.phar build
Building Actor classes for suites: functional, acceptance, unit
 -> FunctionalTesterActions.php generated successfully. 0 methods added
\FunctionalTester includes modules: \Helper\Functional
 -> NoGuyActions.php generated successfully. 0 methods added
\NoGuy includes modules: PhpBrowser, \Helper\Acceptance, REST
 -> UnitTesterActions.php generated successfully. 0 methods added
\UnitTester includes modules: Asserts, \Helper\Unit
```

janhenkgerritsen:
Naktibalda is not available this week, but if you want to use this code in a Helper for now you have to change it a bit, something like this will probably work:

    /**
     * Prevents automatic redirects to be followed by the client
     */
    public function stopFollowingRedirects()
    {
        $this->getModule('REST')->client->followRedirects(false);
    }

    /**
     * Enables automatic redirects to be followed by the client
     */
    public function startFollowingRedirects()
    {
        $this->getModule('REST')->client->followRedirects(true);
    }

hubbyist:
Thank you @janhenkgerritsen. This is working as expected even in UnitTester class.

Naktibalda:
The same methods are necessary in InnerBrowser - for controlling redirection in PhpBrowser and framework modules.

DavertMik:
Just interested, maybe automatic redirection should be completely disabled for REST module?

brutuscat:
@DavertMik I was going to answer that no, but then I went to checkout what CURL does: [CURLOPT_FOLLOWLOCATION](http://curl.haxx.se/libcurl/c/CURLOPT_FOLLOWLOCATION.html)

> DEFAULT: 0, disabled

So I guess that it make sense. It seems that http clients like "browsers" do follow the location header without further ceremony.

So to merge this I would have to add:

- Support in all the "http client" modules and frameworks
- Some sort of unit tests
- Off by default for non browser-like clients

When we agree on these items I'll add them to the body of the PR as a checklist

DavertMik:
> Support in all the "http client" modules and frameworks

No, we have one unified client class which is `InnerBrowser` it abstracts Guzzle and Framework clients through BrowserKit component. So you just need to set BrowserKit Client to disable redirects in `_before`. 

However, for PhpBrowser that won't be enough, as we are using our own redirection https://github.com/Codeception/Codeception/blob/2.1/src/Codeception/Lib/Connector/Guzzle6.php#L115 

And please notice, that in this case BrowserKit's redirection is disabled: https://github.com/Codeception/Codeception/blob/2.1/src/Codeception/Lib/Connector/Guzzle6.php#L22) 
PhpBrowser can be used with Guzzle 5 and Guzzle 6 and we have connectors for both. 

So probably you will need to add method `followRedirects` to both connectors and disable redirection inside a connector when this value was set.

Naktibalda:
I would like to point out that InnerBrowser intercepts all redirects:

https://github.com/Codeception/Codeception/blob/2.1/src/Codeception/Lib/InnerBrowser.php#L151

Now I have to admit that it probably doesn't work with Symfony2, Laravel4 and Laravel5 connectors as expected, because they don't implement BrowserKit/Client but implement Symfony\Component\HttpKernel\Client instead.

Nobody complained about it so far.

DavertMik:
Thanks

Issue #2599, opened 2015-12-01
What is the @internal param?


Naktibalda:
https://github.com/Codeception/Codeception/blob/2.1/src/Codeception/Lib/Interfaces/Web.php

```
    /*
     * @param null $uri
     *
     * @internal param $url
     * @return mixed
     */
    public function grabFromCurrentUrl($uri = null);

    /*
     * @param $cssOrXpath
     * @param $attribute
     * @internal param $element
     * @return mixed
     */
    public function grabAttributeFrom($cssOrXpath, $attribute);
```

DavertMik:
To be honest, I don't know. I suppose it is added by PHPStorm somehow. I may miss the generated docblocks when they are created by IDE

Naktibalda:
Oops, I submitted too quickly, I added the link and the code the first comment.

marcverney:
I've just submitted #2600 to fix this.

DavertMik:
Thank you!

PR #2600, opened 2015-12-01
Globally removed @internal phpdoc tags

marcverney:


Naktibalda:
Thanks.

It fixes #2599

DavertMik:
Thanks!

Issue #2396, opened 2015-09-22
Implement getInternalsDomains methods for all frameworks


Naktibalda:
This is a follow up to https://github.com/Codeception/Codeception/pull/2371

I created a new branch: https://github.com/Codeception/Codeception/tree/feature-get-internal-domains
All frameworks that support multiple domain names in tests must implement [getInternalDomain](https://github.com/Codeception/Codeception/blob/feature-get-internal-domains/src/Codeception/Module/Yii1.php#L179) method. 

## ToDo

- [x] Laravel4
- [x] Laravel5
- [x] Lumen
- [x] Silex
- [x] Symfony2
- [x] Phalcon1
- [x] Phalcon2
- [x] Yii1
- [x] Yii2

janhenkgerritsen:
Thanks for setting up the branch, I will try to implement this as soon as possible.

Naktibalda:
A better place for default implementation of getInternalDomain method would be a [Codeception\Lib\Framework](https://github.com/Codeception/Codeception/blob/2.1/src/Codeception/Lib/Framework.php) class or a new common parent of all Connector classes.

And we could get rid of interface.

Naktibalda:
I implemented getInternalDomains for ZF2,
code: https://github.com/Codeception/Codeception/pull/2405
tests: https://github.com/Naktibalda/codeception-zf2-tests/pull/2

An important observation is that domains are not limited to literal values, so in_array is not enough.
ZF2 supports placeholders in domain route and uses regex for matching.

My implementation of getInternalDomains returns array of patterns.
Can you do something similar for Laravel?

janhenkgerritsen:
I implemented the `getInternalDomains()` method for the Laravel 5 module, and I also added tests for the functionality to the sample application: https://github.com/janhenkgerritsen/codeception-laravel5-sample/commit/068714847a21738414406ca0c2929d374eb45bb6.

To get this to work I had to use reflection. So I introduced a `Codeception\Util\ReflectionHelper` class. This class has two static methods. The first method is the `readPrivateProperty` method of the `PropertyAccess` class. The second method is a new method called `invokePrivateMethod`. I deprecated the `PropertyAccess::readPrivateProperty` method for now, and replaced the implementation, it now delegates to the method on the `ReflectionHelper` class.

[...]

PR #2606, opened 2015-12-03
rebased feature-get-internal-domains

Naktibalda:


Naktibalda:
#2396
I tried `git rebase 2.1` on this branch.
Have I done the right thing?
What should I do next? `git push -f upstream feature-get-internal-domains` ?

DavertMik:
Did you try to squash commits? 
Probably you did something wrong as there are too much commits here...

This is how I squash them http://blog.steveklabnik.com/posts/2012-11-08-how-to-squash-commits-in-a-github-pull-request

janhenkgerritsen:
I think a rebase will be difficult because the 2.1 branch has been merged into the feature-get-internal-domains branch before, and there was work done on this branch before and after these merges. I think the easiest solution will be to just merge the feature branch into 2.1.

Another option is to just copy/paste the changes into another branch and merge that into the 2.1 branch. The changes for this feature are pretty isolated, so that should not be too much work. This approach has the benefit that we can group the commits of the changes for this feature together.

What do you guys think?

Naktibalda:
@DavertMik Majority of commits in this PR are commits made in 2.1 branch in the last few months.

janhenkgerritsen:
This  branch will be merged instead of rebased. PR of merge can be found at #2611.

PR #2608, opened 2015-12-04
Introduced Aerospike module

sergeyklay:
To use this module on your machine, you need at least:

- [Aerospike Server][1] >= 3.5.3
- [Aerospike PHP Extension][2]

Probably this will work with older server versions, but I have not tested.

[1]: http://www.aerospike.com/
[2]: http://www.aerospike.com/docs/client/php/install/

### DIFF
--- a/src/Codeception/Module/Aerospike.php
+++ b/src/Codeception/Module/Aerospike.php
@@ -1,0 +1,198 @@
+<?php
+
+namespace Codeception\Module;
+
+use Codeception\Exception\ModuleException;
+use Codeception\Module as CodeceptionModule;
+use Codeception\TestCase;
+
+/**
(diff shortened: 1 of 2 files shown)


src/Codeception/Module/Aerospike.php
```diff
@@ -0,0 +1,199 @@
...
+        $this->assertEquals($value, $actual['bins']['value']);
+    }
+
+    /**
+     * Checks item in Aerospike doesn't exist or is the same as expected.
+     *
+     * @param $key
+     * @param mixed $value
+     */
+    public function dontSeeInMemcached($key, $value = false)
```
Naktibalda: Memcached?

sergeyklay:
@Naktibalda fixed

sergeyklay:
Moved here https://github.com/Codeception/Aerospike-module

PR #2609, opened 2015-12-04
Symfony2: Cache router

raistlin:
Router can be cached on Symfony2 framework. This option is disabled by default.
It can be enabled with:
modules:
  - Symfony2:
    cache_router: true
    ...

It's also provided a method to force cache invalidation on desire named
'invalidateCachedRouter()', to be used on the rare cases you want router to be
cached but need to manually invalidate cache before of after certain tests.

### DIFF
--- a/src/Codeception/Module/Symfony2.php
+++ b/src/Codeception/Module/Symfony2.php
@@ -83,6 +83,7 @@ class Symfony2 extends Framework implements DoctrineProvider, SupportsDomainRout
         'var_path' => 'app',
         'environment' => 'test',
         'debug' => true,
+        'cache_router' => false,
         'em_service' => 'doctrine.orm.entity_manager'
     ];
 
@@ -90,6 +91,9 @@ class Symfony2 extends Framework implements DoctrineProvider, SupportsDomainRout
      * @var
      */
     protected $kernelClass;
+
+    /** @var */
+    protected static $router;
 
     public $permanentServices = [];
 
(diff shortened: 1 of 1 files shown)


src/Codeception/Module/Symfony2.php
```diff
@@ -91,6 +92,9 @@ class Symfony2 extends Framework implements DoctrineProvider, SupportsDomainRout
      */
     protected $kernelClass;
 
+    /** @var */
+    protected static $router;
+
     public $permanentServices = [];
```
DavertMik: actually you can use this array to cache router

DavertMik:
I think this implementation can be improved if router will be cached not only for the module itself but for the whole application. As I see it: we get a router instance after kernel booted the first time, save it (don't use static properties for it, probably you can have `permanentServices` array), then we inject it in every `_before`.

raistlin:
I left amOnRoute and seeCurrentRouteIs because I like having checks on one function and simply call it when I need.

src/Codeception/Module/Symfony2.php
```diff
@@ -142,6 +143,13 @@ protected function bootKernel()
         }
         $this->kernel = new $this->kernelClass($this->config['environment'], $this->config['debug']);
         $this->kernel->boot();
+        if ($this->config['cache_router'] === true) {
+            if (null === $this->permanentServices['router']) {
```
DavertMik: maybe `isset($this->permanentServices['router'])`?

src/Codeception/Module/Symfony2.php
```diff
@@ -174,6 +178,46 @@ protected function getKernelClass()
...
+            return self::$router;
+        }
+
+        return $this->getNonCachedRouter();
+    }
+
+    /**
+     * Invalidate previously cached router.
+     */
+    public function invalidateCachedRouter()
```
DavertMik: I'd rather not make it public. This method is useless if you have disabled cached routes

raistlin: Yes, but if you have enabled it, this is the only way to manually invalidate the cached routes.
I am not sure if this method could be useful or not, it would depend on specific use case.

So I think that is safe to give to the user the option to force cache invalidation, we don't know all scenarios that our users are executing, may be some users have dynamic routes and need a way to invalidate cache only at certain tests... This way it is safe to enable cached routes and invalidate cache on specific tests.

If nobody will use this feature, then it is safe to delete the method.

src/Codeception/Module/Symfony2.php
raistlin: Nop, If router has not been cached previously, then get router from container (from the current kernel) and store instance at ````$this->permanentServices````.

I think it's clear, but I can swap the conditional body to not compare against null, and improve readability.

src/Codeception/Module/Symfony2.php
```diff
@@ -174,6 +182,28 @@ protected function getKernelClass()
...
+            $this->fail('Router not found.');
+        }
+
+        return $this->kernel->getContainer()->get('router');
+    }
+
+    /**
+     * Invalidate previously cached routes.
+     */
+    public function invalidateCachedRouter()
```
DavertMik: should be private or protected

DavertMik:
If you don't mind I will merge it and try benchmarks on our symfony app

raistlin:
Please, do it.

PR #2614, opened 2015-12-08
can add ssl proxy setting

mjntan35:
This is to allow to add ssl proxy settings.

https://github.com/facebook/php-webdriver/wiki/HowTo-Work-with-proxy


DavertMik:
Thank you. 
Looks like we duplicate passing of `httpProxy` here, it is passed as proxy capabilities and as parameter when creating `RemoteWebDriver`. But I think this should not be an issue here

Issue #2060, opened 2015-06-17
Too many connections error when running large test suite (ZF2, Doctrine2)


dranzd:
This is kind of similar to issue #1591 but using ZF2 and Doctrine2.

If run in small number of test suites, it works fine.  But when all test suites are run then I get the error:

```
[Doctrine\DBAL\Exception\DriverException] An exception occured in driver: SQLSTATE[08004] [1040] Too many connections
```

A similar work around as #1591 is to close the connection after each test
```
    public function _after(FunctionalTester $I)
    {
        \Codeception\Module\Doctrine2::$em->getConnection()->close();
    }
```

This also fixes the problem if the code below is added in the Doctrine2 module but I'm not sure if that's the correct solution for this problem.

```
class Doctrine2 extends \Codeception\Module
{
    ...
    public function _after(\Codeception\TestCase $test)
    {
        ...
        self::$em->getConnection()->close();
    }
```

dynasource:
try PDO::ATTR_PERSISTENT => true
 https://github.com/Codeception/Codeception/issues/1363

janthurau:
I'm having the same problem (just with Symfony2 + Doctrine) :(

        foreach(self::getDoctrine()->getManagers() as $name=>$manager){
            $manager->getConnection()->close();
            $manager->close();
        }

I'm closing all open connections after each test (the code above is from _after method). Still, I get too many connection errors :-( I've already tried PDO::ATTR_PERSISTENT => same result.

![image](https://cloud.githubusercontent.com/assets/896179/10712799/9f8c99d4-7aa4-11e5-98da-a60d1f36cf26.png)

dranzd:
Can this be closed since the fix was already merged?

sitaprajapati:
Even in  codeception 2.1.5, Too many connections errors are getting

dranzd:
@sitaprajapati After the fix was merged I've never seen that error.  If you are still experiencing the issue I suggest you open another issue and provide more details or if you can provide a PR would be great.

PR #2625, opened 2015-12-10
Fix issue #2060 - Too many connections error

dranzd:
Close Doctrine2 connection after each test to prevent throwing an
exception "[Doctrine\DBAL\Exception\DriverException] An exception occured
in driver: SQLSTATE[08004] [1040] Too many connections" with a large
test suite

### DIFF
--- a/src/Codeception/Module/Doctrine2.php
+++ b/src/Codeception/Module/Doctrine2.php
@@ -157,7 +157,7 @@ EOF;
 
         $this->em->getConnection()->connect();
     }
-    
+
     public function _after(TestCase $test)
     {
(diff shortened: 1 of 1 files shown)


dranzd:
For some reason the test in the build fails for unrelated fix.  I've tested it locally with a browser and it shows the "Nothing special" after submitting the form.  Am I missing something here?

Naktibalda:
WebDriver tests often fail for no good reason. It passed when I restarted the build on PHP 5.5

Link to the issue: #2060

DavertMik:
Thanks. This is really important fix.

Issue #2627, opened 2015-12-11
No error given when having a conflicting helper method name when multiple modules are loaded


fc:
I had both the ZF1 and REST module enabled in my functional.suite.yml.

In my FunctionalTester helper class in \Helper\Functional.php I had a method called `sendPOST`.  When calling `sendPOST` from my *Cest class it would execute the sendPOST in the REST module.  After renaming sendPOST to something else it worked.  

It's true I probably don't need to load both the ZF1/REST module at the same time (still new to this codeception thing), but even so it never triggered an error that there was a duplicate sendPOST method and led to confusion.  Ideally it'll throw an exception about the duplicate methods.

Naktibalda:
Related issue: #2604

Naktibalda:
I think that this issue was overlooked earlier.

Fixing it will cause a backwards compatibility break - `amHttpAuthenticated` and `grabAttributeFrom` methods must be renamed either in `REST` module or in `InnerBrowser`.
Custom helpers could be affected too.

fc:
For maintaining backwards compatibility, perhaps it could at the very least throw a warning messaging but only when it is calling a method that has been marked as a duplicate.

Naktibalda:
@DavertMik What's your opinion about this issue?

DavertMik:
Starting from 2.2 Codeception is strict about conflicts and makes modules with possible conflicts to throw the error. However, REST+Zend are not considered to be conflictful. It is still possible to have duplicate methods in order you want to override them from helpers.

On methods mentioned by @Naktibalda:

* `grabAttributeFrom` from REST is renamed to `grabAttributeFromXmlElement`
* `amHttpAuthenticated` are both in REST and PhpBrowser but they work on the same client so there shouldn't be problems.

PR #2631, opened 2015-12-13
Throw exception when methods are redeclared by either another module or helper

fc:
Refer to original issue details in #2627

### DIFF
--- a/src/Codeception/Lib/ModuleContainer.php
+++ b/src/Codeception/Lib/ModuleContainer.php
@@ -152,6 +152,10 @@ class ModuleContainer
                 }
             }
 
+            if (isset($this->actions[$method->name])) {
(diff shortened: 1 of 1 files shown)


fc:
Ah... so it failed the check.  I definitely do not know enough about codeception.  I am a little confused then, we can load in multiple modules?  I think my original issue occurred when it was only a helper that conflicted but showed no error at all.  I'll need to do some more exploring later to get a grasp on it haha

Naktibalda:
I think that this issue was overlooked earlier.

Fixing it will cause a backwards compatibility break - `amHttpAuthenticated` and `grabAttributeFrom` methods must be renamed either in `REST` module or in `InnerBrowser`.
Custom helpers could be affected too.

PR #2633, opened 2015-12-14
"enabled: true" was missing

mindplay-dk:
copy/pasting the example didn't work - adding "enabled: true" fixes that


sergeyklay:
Good catch :+1:

Naktibalda:
Is `enabled: true` necessary in Remote CodeCoverage examples too?

DavertMik:
yes, it is necessary for any kind of CC reports. Thank you

PR #2644, opened 2015-12-19
Added support of dblib PDO driver

IvanZuev:
Db module should handle "dblib" PDO driver as well. 
I tested it and figured out that populate and cleanup works well (without taking into account the fact that sql dump from mssql should be converted to appropriate fromat first).


Naktibalda:
Thanks

PR #2645, opened 2015-12-19
Added warning to docs that Asserts module does not work on HHVM

Naktibalda:


DavertMik:
Thanks. I'm working on fixing PHP5.4 tests and publishing a release just after.

PR #2646, opened 2015-12-20
--prefer-dist in .travis.yml to reduce installation time and enable caching

Naktibalda:


Naktibalda:
```
 - Installing symfony/translation (v2.5.12)

    Downloading: Connecting...    Failed to download symfony/translation from dist: Could not authenticate against github.com

    Now trying to download from source
```

Personal access token is required to avoid Github rate limiting
https://codeship.com/documentation/languages/php/#github-api

.travis.yml
```diff
@@ -32,6 +32,9 @@ services:
 
 sudo: false
 
+before_install:
+  - composer config -g github-oauth.github.com $CI_USER_TOKEN
```
javigomez: I hope I'm wrong but I think all this features are only available for private repos. I had similar doubt about having private tokens in public repos in travisCI http://stackoverflow.com/questions/31209313/in-travis-public-repository-how-to-add-a-secure-variable-that-works-on-pull-requ but no luck.

javigomez: mmm, actually it works but you are publishing the token:

screen shot 2015-12-21 at 12 20 40

Naktibalda: It doesn't work,
https://travis-ci.org/Codeception/Codeception/jobs/98095911#L179

    Failed to download bower-asset/jquery.inputmask from dist: Could not authenticate against github.com
    Now trying to download from source

But I still benefit from composer cache when I am using --prefer-dist, right?

Naktibalda:
http://blog.wyrihaximus.net/2015/09/github-auth-token-on-travis/

This article confirmed that I was on the right track.
The only difference was that I was setting the token before `composer self-update`.

PR #2648, opened 2015-12-21
Check if Throwable interface already exists

truffo:
If another composer package already define a compat hack, Codeception can't be installed.


Naktibalda:
A separate if statement for ParseError class would eliminate another possible issue when Throwable is defined, but ParseError is not.

DavertMik:
I agree. Please add same hook for `ParseError`

truffo:
Ok, i make this feature this afternoon

Issue #2647, opened 2015-12-21
Lib\Parser::validate() fails on paths containing whitespace


gkuhlmann:
$file parameter is not properly escaped in line 129:

The line
`exec("php -l $file 2>&1", $output, $code);`

should probably read
`exec("php -l " . escapeshellarg( $file ) . " 2>&1", $output, $code);`

Naktibalda:
Could you fix it and make a pull request?

PR #2652, opened 2015-12-22
Fixed Parser::validate to escape file paths

acuthbert:
Tests on file paths with spaces failed as the new Parser::validate() call wasn't escaping file paths.


Naktibalda:
Thanks,

it fixes #2647

PR #2656, opened 2015-12-22
make carousel slide animation configurable

vml-rmott:
in config set `animate_slides` to false to disable left-right sliding animation between screenshots in output.html


DavertMik:
Thanks. Is that really annoying?

PR #2668, opened 2015-12-25
Postgres driver: cleanup() delete all custom types

### LINKED ISSUE #2666: Db module: Postgres cleanup() does not delete all types
Hello,
Currently, the Postgres module cleanup() only deletes `enumerated` types. This ignores other kinds of types, including `composite` and `base` ([kinds of types can be seen here](http://www.postgresql.org/docs/current/static/sql-createtype.html)).

I am brand new to Postgres - so I don't want to create a PR before this gets some scrutiny - but I found that if you change the `$types` query (around line 65 of the Postgres driver) to the following, cleanup() is able to delete `composite` and `base` types as well:
`"SELECT 'DROP TYPE IF EXISTS \"' || t.typname || '\" cascade;' FROM pg_catalog.pg_type t LEFT JOIN pg_catalog.pg_namespace n ON n.oid = t.typnamespace WHERE (t.typrelid = 0 OR (SELECT c.relkind = 'c' FROM pg_catalog.pg_class c WHERE c.oid = t.typrelid)) AND NOT EXISTS(SELECT 1 FROM pg_catalog.pg_type el WHERE el.oid = t.typelem AND el.typarray = t.oid) AND n.nspname <> 'pg_catalog' AND n.nspname <> 'information_schema' AND pg_catalog.pg_type_is_visible(t.oid);"`

The above query is adapted from the psql that powers the `\dT` command ([source](http://stackoverflow.com/a/17778566/627772)).

cc/ @DavertMik (Db module maintainer)

airdrummingfool:
This is a PR to fix #2666. To test, I'm adding one of each kind of pgsql `type` and then individually checking that they exist before and don't exist after `cleanup()`. I'm doing the checks individually because it doesn't make sense to use the same query to check for type existence that I use to determine which types to delete (and that's the only query I have to list custom types).

Note that `range type`s aren't supported in pgsql below 9.2, so this will break the tests on pgsql 9.1 and below.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,9 @@ matrix:
     - php: 5.6 # Symfony 2.8
       env: SYMFONY=2.8.0
 
+addons:
+  postgresql: "9.2"
+
 branches:
   except:
     - gh-pages
@@ -29,6 +32,7 @@ cache:
 services:
   - mongodb
   - rabbitmq
+  - postgresql
 
 sudo: false
 
--- a/src/Codeception/Lib/Driver/PostgreSql.php
+++ b/src/Codeception/Lib/Driver/PostgreSql.php
@@ -62,7 +62,7 @@ class PostgreSql extends Db
             ->fetchAll();
 
         $types = $this->dbh
-            ->query("SELECT 'DROP TYPE IF EXISTS \"' || pg_type.typname || '\" cascade;' FROM pg_type JOIN pg_enum ON pg_enum.enumtypid = pg_type.oid GROUP BY pg_type.typname;")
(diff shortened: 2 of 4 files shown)


Naktibalda:
Travis runs version 9.1 by default.

Please change the version of Postgres in .travis.yml file as described in https://docs.travis-ci.com/user/using-postgresql/

airdrummingfool:
@Naktibalda I've fixed the tests on Travis, but for some reason an unrelated test failed on semaphoreci. I'm not sure how to trigger a re-run of the tests to confirm that it was a fluke.

PR #2675, opened 2015-12-28
Made Asserts module work on HHVM

Naktibalda:
Working module is better than elegant module.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,9 +83,6 @@ install:
 
 before_script:
   - '[[ "$TRAVIS_PHP_VERSION" == "hhvm" ]] || [[ "$TRAVIS_PHP_VERSION" == "7.0" ]] || echo "extension = mongo.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini'
-  # enable xdebug for hhvm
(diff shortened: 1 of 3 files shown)


sergeyklay:
@Naktibalda can you please change from 
```
(cd cphalcon/ext; export CFLAGS="-g3 -O1 -std=gnu90"; phpize &> /dev/null && ./configure --silent --enable-phalcon &> /dev/null && make --silent -j4 &> /dev/null && make --silent install && phpenv config-add ../unit-tests/ci/phalcon.ini &> /dev/null && cd ../..;)'
```

to 

```
(cd cphalcon/ext; export CFLAGS="-g0 -O0 -std=gnu90"; phpize &> /dev/null && ./configure --silent --enable-phalcon &> /dev/null && make --silent -j2 &> /dev/null && make --silent install && phpenv config-add ../unit-tests/ci/phalcon.ini &> /dev/null && cd ../..;)'
```

to reduce Phalcon compile time.

> CFLAGS="-g3 -O1 -std=gnu90";
> make --silent -j2

DavertMik:
Thanks. This looks less compact than it was before but let's make it work on HHVM!

PR #2687, opened 2016-01-01
Run tests on hhvm-nightly in travis

Naktibalda:


Naktibalda:
Oops, HHVM nightly is no longer supported on Ubuntu Precise.

PR #2694, opened 2016-01-05
Refactored test formats

DavertMik:
Test formats were reworked in order to simplify them. For instance, Cept and Cest formats may not be extended from `PHPUnit_Framework_TestCase`. New folder structure was introduced. 

* `Test\Loader\FormatName` - a class which is responsible for parsing and loading tests from a file
* `Test\Format\FormtatName` - a class representing a test
* `Test\Feature\*` - set of mixins which are included into the most basic test file (`Codeception\Test\Test`)

Each Codeception test implements `Codeception\Testable` interface. Test name, filename, used modules, actors and other properties now stored in `Metadata`. So to get test name you can access `$test->getMetadata()->getName()`. In the same manner test currents can be accessed too: `$test->getMetadata()->current('env')`. So `Scenario` class was simplified to work only with scenarios and not to store additional data in it.


Naktibalda:
Are you going to merge this feature to 2.1 or to 2.2?

DavertMik:
This will go to 2.2 branch, just before the gherkin branch (as it relies on new test formats).

There is one ugly moment: `Codeception\TestCase\Test` is moved to `Codeception\Test\Format\TestCase` but the previous name is left as an alias.

PR #2695, opened 2016-01-05
Updated gitignore and fixed 2688

Muzical84:
Added an annotation to make sure an XML method is pulled in when only part xml is specified.


Naktibalda:
Thanks

PR #2697, opened 2016-01-06
Fixing Assertation message from null to expected value

zondor:
Fix.
if in you use :
```
$I->seeHttpHeader('X-CustomHeader', "X009");
```
In report you will see : 
```
 Fail  Failed asserting that X009 matches expected null.
```
But should be :
```
 Fail  Failed asserting that null matches expected X009.
```


Naktibalda:
Thank you.

PR #2700, opened 2016-01-07
Add some specification in grabFromDatabase to return all values

felipeAraujo:
Sometimes i miss the option to get all the values that follow some criteria in grabFromDatabase. Most of the time if a table has a HasAndBelongsToMany(n:m) relation i want to check if All the values is there, so returning only the first one will not resolve my problem.

### DIFF
--- a/src/Codeception/Module/Db.php
+++ b/src/Codeception/Module/Db.php
@@ -363,18 +363,22 @@ class Db extends CodeceptionModule implements DbInterface
         return (int) $this->proceedSeeInDatabase($table, 'count(*)', $criteria);
     }
 
-    protected function proceedSeeInDatabase($table, $column, $criteria)
+    protected function proceedSeeInDatabase($table, $column, $criteria, $allValues = false)
     {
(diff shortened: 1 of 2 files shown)


DavertMik:
I'm against using booleans in arguments as they do not give provide infromation about what is that parameter is about. It is hard to guess what `true` or `false` mean without looking into documentation.

If you need such functionality you can have it as a method in custom helper class.

Naktibalda:
@felipeAraujo You can get all values without modifying Db module.

```php
$emails = $I->grabFromDatabase('users', 'GROUP_CONCAT(email)', array('name' => 'Davert'));
```
Result will be a comma separated string which can be used in assertions.

felipeAraujo:
Thank you @DavertMik and @Naktibalda. I understood all the points, this was my first pull request and I think it was valid as an experience and acquired knowledge, I really enjoyed the experience, I will no longer use boolean as argument in my codes in the future. Thanks to show me another solution to my problem, I think that solution will be enough to solve my problems. Thanks :-).

PR #2705, opened 2016-01-09
Fix replace comment in load dump file for mysql

scsidisk:
comment look like this dump from Sequel Pro:

    /*!40101 SET @OLD_COLLATION_CONNECTION=@@COLLATION_CONNECTION */;

and don't replace this:

    ('/admin/*',2,NULL,NULL,NULL,1442237501,1442237501),

but set names shoud be keep.
replace

    /*!40101 SET NAMES utf8 */;

to

    SET NAMES utf8;

### DIFF
--- a/src/Codeception/Module/Db.php
+++ b/src/Codeception/Module/Db.php
@@ -73,21 +73,21 @@ use Codeception\TestCase;
  *              reconnect: true
  *
  * ### SQL data dump
- * 
+ *
  *  * Comments are permitted.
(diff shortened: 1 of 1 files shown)


Naktibalda:
I think that the right way to set charset for connection is by using DSN:
` dsn: 'mysql:host=localhost;dbname=testdb;charset=utf8'`
http://php.net/manual/en/ref.pdo-mysql.connection.php

PR #2709, opened 2016-01-11
Master: Fix replace comment in load dump file for mysql

scsidisk:
comment look like this dump from Sequel Pro:
    
        /*!40101 SET @OLD_COLLATION_CONNECTION=@@COLLATION_CONNECTION */;
    
and don't replace this:
    
        ('/admin/*',2,NULL,NULL,NULL,1442237501,1442237501),


Naktibalda:
Original code does not remove conditional comments, your code does.
Why do you want to remove them?

Naktibalda:
Please raise an issue with a clear description first.

Issue #2380, opened 2015-09-18
Bug in seeInRepository when using @Embedded in Doctrine Mapping


jachim:
I have Embedded objects in my doctrine mapping and I`m trying to use seeInRepository:

 $I->seeInRepository('MyEntity', [
            'embeddedObject.field1' => 'val1',
            'embeddedObject.field2' => 'val2'    
]);

This is causing an error: 
[Doctrine\ORM\Query\QueryException] [Syntax Error] line 0, col 107: Error: Expected end of string, got '.' 

I`ve found a solution. 
Fix this in Doctrine Module by replacing this lines:

$paramname = "s__$key";
$qb->andWhere("s.$key = :$paramname");
$qb->setParameter($paramname, $val);

with:

$paramname = str_replace(".", "_", "s__$key");
$qb->andWhere("s.$key = :$paramname");
$qb->setParameter($paramname, $val);

DavertMik:
> I`ve found a solution. 

Sounds good, could you submit it as pull request?

PR #2710, opened 2016-01-11
Bug in seeInRepository when using @Embedded

jachim:
Related to issue #2380

### DIFF
--- a/src/Codeception/Module/Doctrine2.php
+++ b/src/Codeception/Module/Doctrine2.php
@@ -424,7 +424,7 @@ EOF;
             if ($val === null) {
                 $qb->andWhere("s.$key IS NULL");
             } else {
-                $paramname = "s__$key";
+                $paramname = str_replace(".", "", "s_$key");
                 $qb->andWhere("s.$key = :$paramname");
                 $qb->setParameter($paramname, $val);
             }


src/Codeception/Module/Doctrine2.php
```diff
@@ -424,7 +424,7 @@ protected function buildAssociationQuery($qb, $assoc, $alias, $params)
             if ($val === null) {
                 $qb->andWhere("s.$key IS NULL");
             } else {
-                $paramname = "s__$key";
+                $paramname = str_replace(".", "", "s_$key");
```
DavertMik: I didn't work with Doctrine for a while but why `s__` is changed to `s_`. Is that ok?

jachim: When you are working with Embeddable object in Doctrine you must use notation with dot, for example:
$I->seeInRepository("Post", [
    'author.email' => "email@example.com",
]);

A variable $paramname is build based on $key so in above example will be: "s__author.email". A dot in this variable is a reason for exception:
[Doctrine\ORM\Query\QueryException] [Syntax Error] line 0, col 93: Error: Expected end of string, got '.' 

With this replace everything works great, I haven`t noticed any problems. I see all checks have passed so I hope you will merge this pull request.

raistlin:
Seems safe enough to be merged

DavertMik:
Right now we don't have Doctrine-related tests in core, we use our fork of [symfony-demo app](http://github.com/Codeception/Symfony-demo) for it. @jachim  If you will time, could you add a test there to prove embeddeds work fine? 

But anyway, I'm merging it.

PR #2711, opened 2016-01-12
Refactored formats

DavertMik:
Test formats were reworked in order to simplify them. For instance, Cept and Cest formats may not be extended from `PHPUnit_Framework_TestCase`. New folder structure was introduced. 

* `Test\Loader\FormatName` - a class which is responsible for parsing and loading tests from a file
* `Test\FormatName` - a class representing a test
* `Test\Feature\*` - set of mixins which are included into the most basic test file (`Codeception\Test\Test`)

Each Codeception test implements `Codeception\TestInterface`. 

Test name, filename, used modules, actors and other properties now stored in `Metadata`. So to get test name you can access `$test->getMetadata()->getName()`. In the same manner test currents can be accessed too: `$test->getMetadata()->current('env')`. So `Scenario` class was simplified to work only with scenarios and not to store additional data in it.


DavertMik:
I will squash commits before merging

DavertMik:
Is anyone so brave to review it? ;)

tests/unit/Codeception/TestLoaderTest.php
```diff
@@ -54,25 +59,23 @@ public function testLoadAllTests()
...
     }
 
     protected function getTestNames($tests)
     {
         $testNames = [];
         foreach ($tests as $test) {
-            if ($test instanceof \PHPUnit_Framework_TestCase) {
-                $testNames[] = \Codeception\TestCase::getTestSignature($test);
-            }
+            $testNames[] = \Codeception\Test\Descriptor::getTestSignature($test);
```
tiger-seo: why don't you use aliases ?

DavertMik: because I'm very lazy?

DavertMik:
Let's show the concept of new Loader/Format thing:
Loader loads a Cest file. There are several tests. For each test loader creates Format\Cest instance

Naktibalda:
If you leave it open, I will review it in a few days.

DavertMik:
@Naktibalda Great. Thanks!

Naktibalda:
When are you going to make the tests pass?

DavertMik:
I hope today or in next days. There are minor issues, which I made according to proposals in dev chat.

codeception.yml
```diff
@@ -5,7 +5,7 @@ paths:
     support: tests/support
     envs: tests/envs
 settings:
-    shuffle: true
```
Naktibalda: was this change intentional or accidental?

Naktibalda:
I would prefer small meaningful commits over one giant "refactoring".

src/Codeception/Event/StepEvent.php
```diff
@@ -11,7 +11,7 @@ class StepEvent extends TestEvent
      */
     protected $step;
 
-    public function __construct(TestCase $test, Step $step)
+    public function __construct($test, Step $step)
```
Naktibalda: Why is the type hint missing now?

src/Codeception/Exception/TestRuntimeException.php
```diff
@@ -1,7 +1,7 @@
 <?php
 namespace Codeception\Exception;
 
-use Codeception\TestCase;
+use Codeception\TestInterface;
```
Naktibalda: Unused use

src/Codeception/Lib/Actor/Shared/Comment.php
```diff
@@ -6,7 +6,7 @@
 trait Comment
 {
     /**
-     * @return Scenario
+     * @return \Codeception\Scenario
```
Naktibalda: Unnecessary change, there is a use Codeception\Scenario above the trait.

src/Codeception/Lib/Parser.php
```diff
@@ -42,36 +51,22 @@ public function parseScenarioOptions($code)
...
-                call_user_func([$this->scenario, $annotation], $value);
-            }
-        }        
+        $this->metadata->setGroups(Annotation::fetchAllFromComment('group', $comments));
+        $this->metadata->setEnv(Annotation::fetchAllFromComment('env', $comments));
+        $this->metadata->setSkip($this->flatten(Annotation::fetchAllFromComment('skip', $comments)));
+        $this->metadata->setIncomplete($this->flatten(Annotation::fetchAllFromComment('incomplete', $comments)));
+    }
+
+    private function flatten($array)
```
Naktibalda: wrong name for method

src/Codeception/Module/Phalcon2.php
```diff
@@ -3,7 +3,7 @@
 
 use Phalcon\Di;
 use Phalcon\Mvc\Model as PhalconModel;
-use Codeception\TestCase;
+use Codeception\TestInterface;
```
Naktibalda: There is no code in this class, so all `use` statements should be removed.

src/Codeception/SuiteManager.php
```diff
@@ -92,7 +92,7 @@ public function initialize()
 
     public function loadTests($path = null)
     {
-        $testLoader = new TestLoader($this->settings['path']);
+        $testLoader = new Loader($this->settings);
         $path
             ? $testLoader->loadTest($path)
             : $testLoader->loadTests();
```
Naktibalda: if() statement should be used instead of ternary.

src/Codeception/SuiteManager.php
```diff
@@ -129,14 +128,14 @@ protected function addToSuite($test)
...
-            $test->getScenario()->group($groups);
+        if (!empty($groups) && $test instanceof TestInterface) {
+            $test->getMetadata()->setGroups($groups);
         }
     }
 
     protected function createSuite($name)
     {
-        $suite = new Lib\Suite();
+        $suite = new \Codeception\Suite();
```
Naktibalda: $suite = new Suite();

src/Codeception/Test/Metadata.php
```diff
@@ -0,0 +1,195 @@
...
+    protected $skip = null;
+    protected $incomplete = null;
+
+    protected $current = [];
+    protected $services = [];
+
+    /**
+     * @return mixed
+     */
+    public function getEnv()
```
Naktibalda: Is Env the same thing as Environment?

src/Codeception/SuiteManager.php
```diff
@@ -178,28 +177,28 @@ protected function getActor()
...
             return;
         }
         if (!isset($this->settings['env'])) {
-            Notification::warning("Environments are not configured", TestCase::getTestFullName($test));
+            Notification::warning("Environments are not configured", Descriptor::getTestFullName($test));
             return;
         }
         $availableEnvironments = array_keys($this->settings['env']);
-        $listedEnvironments = explode(',', implode(',', $test->getEnvironment()));
+        $listedEnvironments = explode(',', implode(',', $envs));
```
Naktibalda: This code should be in Metadate::setEnv() method

src/Codeception/SuiteManager.php
```diff
@@ -178,28 +177,28 @@ protected function getActor()
...
     }
 
-    protected function isExecutedInCurrentEnvironment(\Codeception\TestCase $test)
+    protected function isExecutedInCurrentEnvironment(TestInterface $test)
     {
-        $envs = $test->getEnvironment();
+        $envs = $test->getMetadata()->getEnv();
         if (empty($envs)) {
             return true;
         }
```
Naktibalda: If explode-implode is moved to setEnv, the code below this line must be updated too.

src/Codeception/Test/Cept.php
```diff
@@ -0,0 +1,78 @@
...
+     */
+    protected $parser;
+
+    public function __construct($name, $file)
+    {
+        $this->setMetadata(new Metadata());
+        $this->getMetadata()->setName($name);
+        $this->getMetadata()->setFilename($file);
+        $this->createScenario();
+        $this->parser = new Parser($this->getScenario(), $this->getMetadata());
```
Naktibalda: How about assigning metadata to local variable and passing it to all methods?
$metadata = new Metadata();
$metadata->setName($name);
$metadata->setFilename($file);
$this->setMetadata($metadata);

src/Codeception/Test/Cept.php
```diff
@@ -0,0 +1,78 @@
...
+
+    public function preload()
+    {
+        $this->getParser()->prepareToRun($this->getRawBody());
+    }
+
+    public function test()
+    {
+        $scenario = $this->getScenario();
+        $testFile = $this->getMetadata()->getFilename();
```
Naktibalda: scenario and metadata are protected properties of this object.
Would it be wrong to use $this->scenario and $this->metadata instead of getters?

src/Codeception/Test/Cept.php
```diff
@@ -0,0 +1,78 @@
...
+    {
+        return $this->getMetadata()->getName() . 'Cept';
+    }
+
+    public function toString()
+    {
+        return $this->getFeature() . " (" . $this->getSignature() . ")";
+    }
+
+    public function getRawBody()
```
Naktibalda: It looks like a wrong name to me. getRawBody associates with HTTP request.
This method should be named getTestCode or getFileContents.

src/Codeception/Test/Cest.php
```diff
@@ -101,32 +102,27 @@ protected function executeContextMethod($context, $I)
...
-            $this->di->set($class);
+            $this->getMetadata()->getService('di')->set($class);
         }
-        $this->di->injectDependencies($this->testClassInstance, $methodName, $context);
+        $this->getMetadata()->getService('di')->injectDependencies($this->testClassInstance, $methodName, $context);
     }
-
     protected function executeTestMethod($I)
     {
         $testMethodSignature = [$this->testClassInstance, $this->testMethod];
```
Naktibalda: This method should use method_exists instead of is_callable.

src/Codeception/Test/Feature/BlockByMetadata.php
```diff
@@ -0,0 +1,37 @@
+<?php
+namespace Codeception\Test\Feature;
+
+use Codeception\Test\Metadata;
+
+trait BlockByMetadata
```
Naktibalda: This name is not self-descriptive. Please add a short description.

DavertMik:
Thanks for the review. 

> I would prefer small meaningful commits over one giant "refactoring".

Sorry. I had small commits few days ago but squashed them few days ago. I will do squash only after complete review next time. However I don't think I could get it with "meaningful" commits as too much things were changed in complex.

src/Codeception/Test/Feature/BlockByMetadata.php
DavertMik: Or to choose a better name...

* SkipByMetadata
* IgnoreFromMetadata

I will think on it

src/Codeception/Test/Cest.php
DavertMik: :+1:

src/Codeception/Test/Feature/Dependencies.php
```diff
@@ -0,0 +1,15 @@
...
+
+trait Dependencies
+{
+    protected $dependencies;
+    protected $dependencyInput = [];
+
+    abstract public function getTestResultObject();
+    abstract public function getTestClass();
+    abstract public function getName();
+    abstract public function getSignatuire();
```
Naktibalda: mistype in the name

src/Codeception/Test/Loader.php
```diff
@@ -0,0 +1,134 @@
...
+
+    protected function relativeName($file)
+    {
+        return str_replace([$this->path, '\\'], ['', '/'], $file);
+    }
+
+    protected function findPath($path)
+    {
+        if (!file_exists($path)
+            && substr(strtolower($path), -strlen('.php')) !== '.php'
```
Naktibalda: Why do you use strtolower here if all other filename matching is case-sensitive?
I would use -4  instead of -strlen().

Naktibalda:
Measuring test execution time is a new feature, right?
Have you implemented any other new features?

tests/cli/OrderCest.php
```diff
@@ -43,7 +43,7 @@ public function checkSimpleFiles(CliGuy $I)
         $I->amInPath('tests/data/sandbox');
         $I->executeCommand('run order --no-exit --group simple');
         $I->seeFileFound('order.txt','tests/_output');
-        $I->seeFileContentsEqual("BIB({{{[ST][STFFT][STF][ST])}}}");
```
Naktibalda: I would like to see explanation what happens here.

tests/data/claypit/tests/_support/OtherHelper.php
```diff
@@ -4,7 +4,7 @@
 
 class OtherHelper extends \Codeception\Module
 {
-    public function _before(\Codeception\TestCase $test)
+    public function _before(\Codeception\TestInterface $test)
     {
         if (strpos(PHP_VERSION, '5.3')===0) $test->markTestSkipped();
```
Naktibalda: We no longer need this code, 5.3 is not supported.

tests/support/CoverHelper.php
```diff
@@ -6,7 +6,7 @@
 class CoverHelper extends \Codeception\Module
 {
     
-    public function _before(\Codeception\TestCase $test) {
+    public function _before(\Codeception\TestInterface $test) {
         if (floatval(phpversion()) == '5.3') $test->markTestSkipped();
```
Naktibalda: Another 5.3 check.

tests/unit/Codeception/Command/BaseCommandRunner.php
```diff
@@ -54,6 +55,7 @@ protected function makeCommand($className, $saved = true)
                 $self->filename = $file;
                 $self->content = $output;
                 $self->log[] = ['filename' => $file, 'content' => $output];
+                $self->saved[$file] = $output;
```
Naktibalda: $self->log and $self->saved store the same data in different structure. Do we really need both?

tests/unit/Codeception/Command/GenerateScenarioTest.php
```diff
@@ -37,29 +37,33 @@ protected function setUp()
...
+        $this->assertContains('I WANT TO CHECK CONFIG EXISTS', $content);
+        $this->assertContains('I see file found "$codeception"', $content);
         $this->assertContains('* File_Exists generated', $this->output);
     }
 
     public function testMultipleTestsGeneration()
     {
         $this->execute(['suite' => 'dummy']);
-        $fileNames = array_map(function ($f) { return $f['filename']; }, $this->log);
+        $fileNames = array_keys($this->log);
```
Naktibalda: How does this work? keys of $this->log are integers.

codeception.yml
DavertMik: I used this for debugging. Thanks, will revert back

DavertMik:
> Measuring test execution time is a new feature, right?

Actually it's not, it was built-in to PHPUnit but was not used. So it's not a feature to me )
Not sure of other features, It's really hard to recap all the changes. However, the big major feature is coming once we merge this branch. I will send PR for Gherkin branch. So a new test format will be added.

src/Codeception/Test/Metadata.php
DavertMik: yes

src/Codeception/SuiteManager.php
DavertMik: Actually, no. We check that we environments in tests are valid but we don't set them. So `setEnv` is not called

src/Codeception/Test/Cept.php
DavertMik: I'd rather not have public `setMetadata` method. It can be used in wrong way.

src/Codeception/Test/Cept.php
DavertMik: I was thinking on making Metadata private... To avoid side effects if someone introduce such property in test of `Codeception\Test\TestCase` and change it.

tests/unit/Codeception/Command/BaseCommandRunner.php
DavertMik: Thanks, removed `log`

DavertMik:
@Naktibalda  Done. Updated everything according to your proposals. If everything else is fine I can't wait to get it merged so we can go on to 2.2

src/Codeception/Lib/Actor/Shared/Comment.php
DavertMik: Now there is not )

src/Codeception/Test/Test.php
```diff
@@ -0,0 +1,101 @@
...
+        $result->endTest($this, $time);
+        return $result;
+    }
+
+    public function getTestResultObject()
+    {
+        return $this->testResult;
+    }
+
+    abstract public function test();
```
Naktibalda: What does it do? This class is central in your new order and it would really benefit from docblocks

src/Codeception/Test/TestCase.php
```diff
@@ -0,0 +1,141 @@
...
+    {
+        return [
+            'name'    => $this->getName(),
+            'class'   => get_class($this),
+            'file'    => $this->getFileName(),
+            'feature' => $this->getFeature()
+        ];
+    }
+
+    protected function handleDependencies()
```
Naktibalda: You moved dependency handling to Test format, is it really impossible to implement dependencies in other formats?

tests/data/claypit/tests/_support/OtherHelper.php
```diff
@@ -1,12 +1,6 @@
 <?php
 namespace Codeception\Module;
 
-
 class OtherHelper extends \Codeception\Module
```
Naktibalda: OtherHelper can be deleted now, right?

tests/unit/Codeception/Command/GenerateScenarioTest.php
```diff
@@ -37,29 +37,33 @@ protected function setUp()
     public function testBasic()
     {
         $this->execute(array('suite' => 'dummy'));
-        $this->assertEquals(codecept_root_dir().'tests/data/scenarios/dummy/File_Exists.txt', $this->filename);
-        $this->assertContains('I WANT TO CHECK CONFIG EXISTS', $this->content);
-        $this->assertContains('I see file found "$codeception"', $this->content);
+        $this->assertArrayHasKey($file = codecept_root_dir().'tests/data/scenarios/dummy/File_Exists.txt', $this->saved);
```
Naktibalda: Inline assignment is ugly.

tests/unit/Codeception/Command/GenerateScenarioTest.php
```diff
@@ -37,29 +37,33 @@ protected function setUp()
...
+        $this->assertContains('I WANT TO CHECK CONFIG EXISTS', $content);
+        $this->assertContains('I see file found "$codeception"', $content);
         $this->assertContains('* File_Exists generated', $this->output);
     }
 
     public function testMultipleTestsGeneration()
     {
         $this->execute(['suite' => 'dummy']);
-        $fileNames = array_map(function ($f) { return $f['filename']; }, $this->log);
+        $fileNames = array_keys($this->saved);
```
Naktibalda: $fileNames could be eliminated by using assertArrayHasKey instead of assertContains

src/Codeception/Test/Test.php
DavertMik: ok, let's do it!

src/Codeception/Test/TestCase.php
DavertMik: yeah, I should do that for Cest too. Thnx

DavertMik: I wish I could understand how PHPUnit dependencies work :(

DavertMik:
About dependencies. Looks like PHPUnit dependencies are very limited: 

> PHPUnit does not change the order in which tests are executed, you have to ensure that the dependencies of a test can actually be met before the test is run.

so they just not work with shuffled tests, nor they work across classes in any smart way. It's not clean that the order of tests matter to make them work. And it's really hard to maintain that order, as well as rely on it in future run. 

Looks like we need our own implementation of dependencies which would ensure that test order is correct, as well there are no cyclic dependencies in tests

DavertMik:
So this will take another day from me to implement this :(

tests/data/claypit/tests/_support/OtherHelper.php
DavertMik: Nope. It is set in corresponding suite config. Sure, this file is empty and does not do anything, but some other empty Helpers of are required in tests to run.

DavertMik:
Ok, I'm done here. I implemented cool dependency mechanism with **smart reordering of tests to match dependencies**, so practically any test can be dependent on other. I.e. Cest method can depend on test from a TestCase and even Cept. Delcarations can be declared with `@depends` annotation. A class name is required if it is a test from other class. In case of Cept you need to pass its name:

```
@depends createUser
@depends UserServiceTest:createUser
@depends Service\UserTest:createUser
@depends UserCept
@depends UserCest:createUser
```

To unify the way dependent tests are handled I disabled PHPUnit's feature in `Codeception\TestCase\Test`, so dependent tests can't receive values from a previous (as it was in PHPUnit). Actually, I don't think this will cause some inconvenience. If someone needs to pass values between dependent tests you can store data in helpers.

DavertMik:
@Naktibalda @tiger-seo is there something more I should take a look at?

Naktibalda:
How about renaming Codeception\Test\TestCase to Codeception\Test\Unit or something similar?

DavertMik:
> How about renaming Codeception\Test\TestCase to Codeception\Test\Unit or something similar?

:+1: good idea

Naktibalda:
Please rename Test/Loader/TestCase.php to Test/Loader/Unit.php for consistency.
Are there any other per-format files?

DavertMik:
> Please rename Test/Loader/TestCase.php to Test/Loader/Unit.php for consistency.

Thanks. I will check if there any other places where it can be.

DavertMik:
Squashed commits for the last time and ready to merge!

sergeyklay:
Good job guys! :+1:

Green-Cat:
:+1:

Issue #2630, opened 2015-12-12
seeResponseContainsJson doesn't work properly


Vishnja:
Codeception ver. 2.1.12
1) Json response:
```
{
    "flattened_array": [
        [
            "2015-09-10",
            "unknown-date-1"
        ]
    ]
} 
```
Codeception test (passes successfully):
```
$I->seeResponseContainsJson([
     "flattened_array" => [
           ["2015-09-10", "unknown-date-1"],
      ]
 ]);
```
2) But when Json response is:
```
{
    "flattened_array": [
        [
            "2015-09-10",
            "unknown-date-1"
        ],
        [
            "2015-10-10",
            "unknown-date-1"
        ]
    ]
}
```
Codeception fails:
```
$I->seeResponseContainsJson([
    "flattened_array" => [
        ["2015-09-10", "unknown-date-1"],
        ["2015-10-10", "unknown-date-1"]
    ]
]);
```
In console:
```
Step  I see response contains json {"flattened_array":[["2015-09-10","unknown-date-1"],["2015-10-10","unknown-date-1"]]}
 Fail  Response JSON contains provided

- array (
  'flattened_array' =>
  array (
    0 =>
    array (
      0 => '2015-09-10',
      1 => 'unknown-date-1',
    ),
    1 =>
    array (
      0 => '2015-10-10',
      1 => 'unknown-date-1',
    ),
  ),
)
+ array (
  'flattened_array' =>
  array (
    0 =>
    array (
      0 => '2015-09-10',
      1 => 'unknown-date-1',
    ),
    1 =>
    array (
      0 => '2015-10-10',
      1 => 'unknown-date-1',
    ),
  ),
)
Failed asserting that false is true.
```

Naktibalda:
Assertion failed because [JsonArray:sequentialArrayIntersect](https://github.com/Codeception/Codeception/blob/b83d16801120d9449447afee6c5dfc4e31b584fe/src/Codeception/Util/JsonArray.php#L144) returned the last partial match:
```
- array (
  'flattened_array' =>
  array (
    0 =>
    array (
      0 => '2015-09-10',
      1 => 'unknown-date-1',
    ),
    1 =>
    array (
      0 => '2015-10-10',
      1 => 'unknown-date-1',
    ),
  ),
)
+ array (
  'flattened_array' =>
  array (
    0 =>
    array (
      1 => 'unknown-date-1',
    ),
    1 =>
    array (
      0 => '2015-10-10',
[...]

PR #2719, opened 2016-01-13
Added simplified testcase for the last issue in #2630

Naktibalda:


Naktibalda:
I will rename this test later.

PR #2722, opened 2016-01-14
facebook module - add support for facebook/php-sdk-v4 version 5

orhan-swe:
all test are passing now but the changes do break support for version 4 unfortunately

### DIFF
--- a/src/Codeception/Lib/Driver/Facebook.php
+++ b/src/Codeception/Lib/Driver/Facebook.php
@@ -112,13 +112,13 @@ class Facebook
         }
 
         switch ($method) {
-            case 'GET' :
+            case 'GET':
                 $response = $this->fb->get($endpoint, $token);
                 break;
-            case 'POST' :
+            case 'POST':
                 $response = $this->fb->post($endpoint, $parameters, $token);
                 break;
-            case 'DELETE' :
+            case 'DELETE':
                 $response = $this->fb->delete($endpoint, $parameters, $token);
                 break;
             default:
--- a/src/Codeception/Lib/Driver/Facebook.php
+++ b/src/Codeception/Lib/Driver/Facebook.php
@@ -4,16 +4,20 @@
  */
 namespace Codeception\Lib\Driver;
 
-use Facebook\Facebook as Facebook_SDK;
+use Facebook\Facebook as FacebookSDK;
 
 class Facebook
 {
     protected $logCallback;
 
     /**
-     * @var Facebook_SDK
+     * @var FacebookSDK
      */
     protected $fb;
+
+    protected $appId;
+    protected $appSecret;
+
 
     public function __construct($config, $logCallback = null)
     {
@@ -21,7 +25,7 @@ class Facebook
             $this->logCallback = $logCallback;
         }
 
-        $this->fb = new Facebook_SDK([
+        $this->fb = new FacebookSDK([
             'app_id' => $config['app_id'],
             'app_secret' => $config['secret'],
             'default_graph_version' => 'v2.5', //TODO add to config
@@ -38,11 +42,11 @@ class Facebook
      */
     public function createTestUser($name, array $permissions)
     {
-        $app_token = $this->appId . '|' . $this->appSecret;
+        $appToken = $this->appId . '|' . $this->appSecret;
         $response = $this->executeFacebookRequest(
             'POST',
             $this->appId . '/accounts/test-users',
-            $app_token,
+            $appToken,
             [
                 'name' => $name,
                 'installed' => true,
@@ -55,11 +59,11 @@ class Facebook
 
     public function deleteTestUser($testUserID)
     {
-        $app_token = $this->appId . '|' . $this->appSecret;
+        $appToken = $this->appId . '|' . $this->appSecret;
         $this->executeFacebookRequest(
             'DELETE',
             '/' . $testUserID,
-            $app_token
+            $appToken
         );
     }
 
--- a/src/Codeception/Lib/Driver/Facebook.php
+++ b/src/Codeception/Lib/Driver/Facebook.php
@@ -89,6 +89,17 @@ class Facebook
         return $response->getDecodedBody();
     }
 
+    public function getVisitedPlaceTagForTestUser($placeId, $testUserAccessToken)
+    {
+        $response = $this->executeFacebookRequest(
+            'GET',
+            "/$placeId",
+            $testUserAccessToken
+        );
+
+        return $response->getDecodedBody();
+    }
+
     public function sendPostToFacebook($testUserAccessToken, array $parameters)
     {
         $response = $this->executeFacebookRequest(
--- a/src/Codeception/Lib/Driver/Facebook.php
+++ b/src/Codeception/Lib/Driver/Facebook.php
@@ -4,12 +4,16 @@
  */
 namespace Codeception\Lib\Driver;
 
-use Facebook\FacebookSession;
-use Facebook\FacebookRequest;
+use Facebook\Facebook as Facebook_SDK;
 
 class Facebook
 {
     protected $logCallback;
+
+    /**
+     * @var Facebook_SDK
+     */
+    protected $fb;
 
     public function __construct($config, $logCallback = null)
     {
@@ -17,9 +21,13 @@ class Facebook
             $this->logCallback = $logCallback;
         }
 
-        FacebookSession::setDefaultApplication($config['app_id'], $config['secret']);
-        $this->appSession = FacebookSession::newAppSession($config['app_id'], $config['secret']);
+        $this->fb = new Facebook_SDK([
+            'app_id' => $config['app_id'],
+            'app_secret' => $config['secret'],
+            'default_graph_version' => 'v2.5', //TODO add to config
+        ]);
         $this->appId = $config['app_id'];
+        $this->appSecret = $config['secret'];
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/Codeception/Module/Facebook.php
```diff
@@ -265,19 +262,19 @@ public function postToFacebookAsTestUser($params)
      *
      * Please, note that you must have publish_actions permission to be able to publish to user's feed.
      *
-     * @param string $placeId Place identifier to be verified against user published posts
+     * @param string $message published post to be verified against the actual post on facebook
      */
-    public function seePostOnFacebookWithAttachedPlace($placeId)
+    public function seePostOnFacebookWithAttachedPlace($message)
```
Naktibalda: Shouldn't it be named seePostOnFacebookWithMessage?

src/Codeception/Module/Facebook.php
```diff
@@ -265,19 +262,19 @@ public function postToFacebookAsTestUser($params)
...
             foreach ($posts['data'] as $post) {
-                if (array_key_exists('place', $post) && ($post['place']['id'] == $placeId)) {
+                if (array_key_exists('message', $post) && ($post['message'] == $message)) {
                     return; // success
                 }
             }
         }
 
-        $this->fail('Failed to see post on Facebook with attached place with id ' . $placeId);
+        $this->fail('Failed to see post on Faceboo: ' . $message);
```
Naktibalda: a half of message is missing.
this assertion would generate a nicer result if it used some assertion for passing.

tests/unit/Codeception/Module/FacebookTest.php
```diff
@@ -61,6 +61,7 @@ protected function tearDown()
      */
     public function testHaveFacebookTestUserAccount()
     {
+        //$this->markTestSkipped();
```
Naktibalda: Please delete all markTestSkipped lines

Naktibalda:
@tiger-seo Please review this pull request.

orhan-swe:
@Naktibalda Thanks, I have added the changes, I am not sure why the WebDriverTest::testTextFieldByNameFirstNotCss failed in the previous commit, it is passing on my local machine

Naktibalda:
WebDriver fails in random place in 1 run out of 10 or 20.

orhan-swe:
ok thanks, good to know

src/Codeception/Lib/Driver/Facebook.php
```diff
@@ -4,22 +4,30 @@
  */
 namespace Codeception\Lib\Driver;
 
-use Facebook\FacebookSession;
-use Facebook\FacebookRequest;
+use Facebook\Facebook as Facebook_SDK;
```
tiger-seo: this is not consistent with PSR,  please, change to BaseFacebook or FacebookSDK

src/Codeception/Lib/Driver/Facebook.php
```diff
@@ -30,83 +38,96 @@ public function __construct($config, $logCallback = null)
      */
     public function createTestUser($name, array $permissions)
     {
+        $app_token = $this->appId . '|' . $this->appSecret;
```
tiger-seo: since Codeception is using PSR-2 coding standard, please fix the code accordingly, thanks

src/Codeception/Lib/Driver/Facebook.php
DavertMik: Why not just FacebookSDK? That looks cleaner to me. It is not common to use underscores in class names (unless they come from a pre-namespace era).

src/Codeception/Lib/Driver/Facebook.php
DavertMik: Nitpick didn't show this as an issue. And we have lots of places where we have underscores in variable names. 
There is no standard actually. But yes, `$appToken` looks better to me

tiger-seo: wow, i was sure it is there :smiley: nonetheless, we don't use underscore for php entities

src/Codeception/Lib/Driver/Facebook.php
```diff
@@ -4,22 +4,30 @@
...
         }
 
-        FacebookSession::setDefaultApplication($config['app_id'], $config['secret']);
-        $this->appSession = FacebookSession::newAppSession($config['app_id'], $config['secret']);
+        $this->fb = new Facebook_SDK([
+            'app_id' => $config['app_id'],
+            'app_secret' => $config['secret'],
+            'default_graph_version' => 'v2.5', //TODO add to config
+        ]);
         $this->appId = $config['app_id'];
```
tiger-seo: please, declare these properties

src/Codeception/Module/Facebook.php
```diff
@@ -265,19 +261,20 @@ public function postToFacebookAsTestUser($params)
      *
      * Please, note that you must have publish_actions permission to be able to publish to user's feed.
      *
-     * @param string $placeId Place identifier to be verified against user published posts
+     * @param string $message published post to be verified against the actual post on facebook
      */
-    public function seePostOnFacebookWithAttachedPlace($placeId)
+    public function seePostOnFacebookWithMessage($message)
```
tiger-seo: place API is not available anymore ?

tests/unit/Codeception/Module/FacebookTest.php
```diff
@@ -117,21 +116,9 @@ public function testLoginToFacebook()
         // preconditions: #3 test user logged in on facebook
         $this->module->haveTestUserLoggedInOnFacebook();
 
-        // go to our page with facebook login button
-        $browserModule->amOnPage('/facebook');
```
tiger-seo: why you've deleted those tests statements ?
 now it's not testing that facebook login work

orhan-swe: Thanks for the feedback, I have added the changes. About the facebook login test, it did not work for me and it was marked with markTestSkipped(). I did try to fix it but I see now that I misunderstood it a little bit. I try to see if I can fix this first thing tomorrow morrning.
Also for seePostOnFacebookWithAttachedPlace the new API did not return placeId, but I will check if it is possible to do this in some other way so that seePostOnFacebookWithAttachedPlace($placeId) is not removed.

DavertMik:
Restarted tests. I hope this PR doesn't bring a BC break, right? If not, I think we can merge it.

src/Codeception/Lib/Driver/Facebook.php
```diff
@@ -30,83 +42,107 @@ public function __construct($config, $logCallback = null)
...
             [
                 'name' => $name,
                 'installed' => true,
                 'permissions' => $permissions
             ]
-        )->getRawResponse();
+        );
 
-        return json_decode($response, true);
+        return $response->getDecodedBody();
```
tiger-seo: is this response is the same as it was with  fb v4 ?

src/Codeception/Lib/Driver/Facebook.php
```diff
@@ -30,83 +42,107 @@ public function __construct($config, $logCallback = null)
...
-            new FacebookSession($testUserAccessToken),
             'GET',
-            '/me'
-        )->getRawResponse();
+            '/me',
+            $testUserAccessToken
+        );
 
-        return json_decode($response, true);
+        return $response->getDecodedBody();
```
tiger-seo: is this response is the same as it was with  fb v4 ?

src/Codeception/Lib/Driver/Facebook.php
```diff
@@ -30,83 +42,107 @@ public function __construct($config, $logCallback = null)
...
     /**
-     * @param FacebookSession $session
      * @param string $method
      * @param string $endpoint
      * @param array $parameters
+     * @param string $token
      * @return \Facebook\FacebookResponse
      */
-    private function executeFacebookRequest(FacebookSession $session, $method, $endpoint, array $parameters = [])
+    private function executeFacebookRequest($method, $endpoint, $token = null, array $parameters = [])
```
tiger-seo: IMO, optional `$token` is a potential issue; according to the FacebookSDK code, if `$accessToken` is null, then it will use `$defaultAccessToken`, which is not being set, so my suggestion here would be to use app token, if `$token` is null;  additionaly we can swap `$token` and `$parameters`, so that methods `createTestUser()` and `deleteTestUser()` can be simplified. what do you think ?

src/Codeception/Lib/Driver/Facebook.php
orhan-swe: yes

src/Codeception/Lib/Driver/Facebook.php
orhan-swe: no, now only id and name are returned, and before it seems we had these properties:
id, first_name, gender, last_name, link, locale, name, timezone, updated_time and verified. It seems api has changed a bit here, although we are only using the 'name' property so that is why tests are passing.

src/Codeception/Lib/Driver/Facebook.php
orhan-swe: This is good, actually I though about it earlier but did not do it for some reason, I can do the changes. 

There is something I am not sure about. What do you think about removing haveTestUserLoggedInOnFacebook()? As it seems it is not possible to send post anymore to log in the user here but I had to go to the facebook page and log in by finding the elements. It may be better if user writes this. There are two reasons for this: 
1. The test may not pass all the time but may fail in every 10 runs, I am not sure why this is. 
2. If one is doing acceptance test with WebDriver he will have to add PhpBrowser as dependency even though he will not need PhpBrowser and also haveTestUserLoggedInOnFacebook() will not work by default for the user if the user is using WebDriver.

src/Codeception/Lib/Driver/Facebook.php
tiger-seo: yeah, but someone may have been using the other data, so if it get dissapearead it will break the test. @DavertMik what do you suggest ? should we introduce BC breack or should we try have it as it was before ?

tiger-seo:
@orhan-swe please, resend your PR to branch `Codeception:feature/facebook-upgrade`, so that we can finish it and merge to the 2.1.  thanks

orhan-swe:
This was wrong branch. A new pull request has been created for feature/facebook-upgrade branch: #2772

PR #2746, opened 2016-01-27
New feature adding Custom Commands to codeception

solutionDrive-Alt:
Hello codeception-Team, we love your project and in order to be able to write custom commands and register them in codecption, we added this possibility.
With custom commands everyone has the ability to write commands for his special needs. So every module, like [cakephp/codeception](https://github.com/cakephp/codeception) could register their custom commands, e.g for generating custom test-classes.


sergeyklay:
As a new feature I like it

src/Codeception/Lib/Interfaces/CustomCommands.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Codeception\Lib\Interfaces;
+
+interface CustomCommands
```
Naktibalda: The name should be singular - CustomCommand

src/Codeception/Application.php
```diff
@@ -0,0 +1,62 @@
...
+     * Register commands from a config file
+     *
+     *  extensions:
+     *      commands:
+     *          - project\MyNewCommand
+     *
+     */
+    public function registerCustomCommands()
+    {
+        $aCommands = Configuration::config();
```
Naktibalda: Please get rid of Hungarian notation

src/Codeception/Application.php
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        try {
+            foreach ($aCommands['extensions']['commands'] as $sCommandClass) {
+                $sCommandName = $this->_getCommandName($sCommandClass);
+                $this->add(new $sCommandClass($sCommandName));
+            }
+        } catch (\Exception $e) {
+            $this->renderException($e, new ConsoleOutput());
+            exit;
```
Naktibalda: You should only exit if autoExit is enabled - BaseApplication::isAutoExitEnabled()

tests/unit/Codeception/ApplicationTest.php
```diff
@@ -0,0 +1,27 @@
...
+
+        $application = new Application();
+        $application->registerCustomCommands();
+
+        $command = null;
+
+        try {
+            $command = $application->find('myProjekt:myCommand');
+        } catch (\Exception $e) {
+        }
```
sergeyklay:
```php
$this->fail($e->getMessage());
```

src/Codeception/Command/MyCustomCommand.php
```diff
@@ -0,0 +1,84 @@
...
+ */
+namespace Codeception\Command;
+
+use Symfony\Component\Console\Command\Command;
+use Codeception\Lib\Interfaces\CustomCommands;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class MyCustomCommand extends Command implements CustomCommands
```
Naktibalda: Custom commands should not be placed in Codeception source tree, a better place for example is https://github.com/Codeception/Codeception/tree/2.1/tests/data/claypit

Naktibalda:
Your change broke some tests in cli suite, please fix them.

solutionDrive-Alt:
Thank you for your fast response, we will fix the open points as soon as possible

src/Codeception/Application.php
sd-tm: Hi Naktibalda,
i am outside of the try/catch() method. The variable $autoExit is mark as privat and only used in Application::run(). So i can't checked if it enabled and nobody else catch my exception.

DavertMik:
Thanks, that looks awesome! Well, let me review it

codecept
```diff
@@ -25,4 +25,5 @@ $app->add(new Codeception\Command\GenerateGroup('generate:groupobject'));
 $app->add(new Codeception\Command\GeneratePageObject('generate:pageobject'));
 $app->add(new Codeception\Command\GenerateStepObject('generate:stepobject'));
 $app->add(new Codeception\Command\GenerateEnvironment('generate:environment'));
+$app->registerCustomCommands();
```
DavertMik: when I was thinking of such feature I found that there is issue with loading time.
You are registering custom commands using current config from `Configuration::config()` but what if `-c` parameter passed and config path is different?

src/Codeception/Application.php
```diff
@@ -0,0 +1,62 @@
...
+     * Register commands from a config file
+     *
+     *  extensions:
+     *      commands:
+     *          - project\MyNewCommand
+     *
+     */
+    public function registerCustomCommands()
+    {
+        $config = Configuration::config();
```
DavertMik: Somehow you should get a value from `-c` option if custom config is specified. Otherwise that won't work in predictable way

DavertMik:
I really like the implementation. Please try to add ability to specify a custom config and resend this PR to master branch. It will be landed in Codeception 2.2 then.

codecept
sd-tm: Okay you are sure. I will be think about a solution.

sergeyklay:
@DavertMik :+1:

codecept
DavertMik: Probably you will need to parse args manually, the same thing I'm doing in Robo. Let me show you sample code: https://github.com/codegyre/robo/blob/master/src%2FRunner.php#L108

PR #2750, opened 2016-01-29
Gherkin Format Support

DavertMik:
For someone this is long awaited feature, for others an unexpected one. 
We bring Gherkin support to Codeception 2.2. This means that you can run your BDD-scenarios as tests.

In Codeception we were targeted developers audience so instead of Gherkin syntax we had PHP DSL. However, it is good to describe tests, and not for actual features. But what if we can mix flexibility and simplicity of PHP DSL for describing tests and Gherkin for features narrative. 

So if you expect that bringing Gherkin format to Codeception is just about writing tests in this kind of way:

```gherkin
Given I am on page "/"
When I click "Login"
And I fill field "Username" with "davert"
And I fill field "Password" with "123456"
And I click "Login"
Then I should see "Welcome, Davert"
```

Well, nope, this won't work in Codeception. At least by default. We expect that BDD will be used in its [right way](https://cucumber.io/blog/2015/03/24/single-source-of-truth) for describing a business rules, and not on testing UI.

So here goes the idea: you describe business specification in Gherkin format, then you implement steps using Codeception's `$I` (actor) object. So the example above should be reworked to something like that:

Inside a feature file:

```gherkin
Given I have account for "davert"
When I log in
Then I should see "Welcome, Davert" notification
```

Inside an actor file (AcceptanceTester) or a context file (see below):

```php
<?php
class AcceptanceTester extends \Codeception\Actor 
{
  use _generated\AcceptanceTesterActions;

  protected $userName;
  protected $password = '123456';
  /**
   * @Given I have account for :name
   */
  public function prepareUser($name)
  {
      $this->userName = $name;
      // let's assume user is already in database ;)
  }  

  /**
   * @When I log in
   */
  public function logIn() 
  {
    $I = $this;
    $I->amOnPage('/');
    $I->fillField('Username', $this->userName);
    $I->fillField('Password', $this->password); // standard password
    $I->click('Login', '#login-form'); // let's use UI-specific locators
  }

  /**
   * @Then I should see :text notification
   */
  public function seeOnPage($text)
  {
    // we just wrap standard action into a feature step
    $this->see($text, '#notification');
  }
}
?>
```
As always, step by step execution is shown when running with `--steps` option:
```
 Given I have account for "davert" 
 When I log in
  I am on page '/'
  I fill field 'Username', 'davert'
  I fill field 'Password', '123456'
  I click 'Login', '#login-form'
 Then I should see "Welcome, Davert" notification
  I see 'Welcome, Davert', '#notification'
```

With `@Given`, `@When` or `@Then` annotation any method of actor object can become a feature step. 
You can use either string with placeholders or regex for a step definition.

Here is another example. We can use tables to define multiline data. Supported in Codeception.

```gherkin
Feature: ToDo items
  In order to get my tasks done
  As a motivated person
  I need to be able to organize my todo list

  Scenario: show current todos
    Given there are todo items:
      | Task             | State   |
      | make Gherkin PR  | opened  |
      | update docs      | opened  |
      | create examples  | closed  |
    When I open my todos
    Then I should see 2 todo items
```

### Features

* [x] executing tests from *.features file 
* [x] defining feature steps in actor file
* [x] regex step definitions
* [x] placeholder step definitions
* [x] tables
* [x] gherkin tags => Codeception groups
* [x] multiple contexts
* [x] select context for a tag or role
* [x] list step definitions in console
* [x] generate snippets for undefined steps
* [ ] describe features in other languages
* [ ] gherkin examples
* [ ] ???

### Implementation Details

(for core team)

This PR includes lots of non-obvious changes. The overall Gherkin implementation requires "behat/gherkin" library and 2 files: `Codeception\Test\Gherkin` and its loader `Codeception\Test\Loader\Gherkin`. 

However, lot of work was done to improve console output so all test formats would look nicely. 
Also all improvements to output was also added to this branch (and not to 2.1 or master) for not to get into merge conflicts. 

Step, and MetaStep classes were changed to handle feature steps of Gherkin.

### Roadmap

This feature is about to land into 2.2. A new guide on Behavior-Driven Development with Codeception is coming by that time.

P.S. Appveyor tests may fail due to their technical issues


jadb:
Just saying thanks! :+1:

DavertMik:
Just about to get it merged today. Last chance to review )

codeception.yml
```diff
@@ -5,7 +5,7 @@ paths:
     support: tests/support
     envs: tests/envs
 settings:
-    shuffle: true
```
Naktibalda: You are trying to change the value of shuffle again. I got used to shuffle: true already.

composer.json
```diff
@@ -31,6 +31,7 @@
         "symfony/dom-crawler": ">=2.4|<3.1"
     },
     "require-dev": {
+        "behat/gherkin": "~4.4.0",
```
Naktibalda: why is it in require-dev?

src/Codeception/Subscriber/Console.php
```diff
@@ -452,115 +462,81 @@ public function printScenarioTrace(ScenarioDriven $failedTest)
...
-    protected function writeFinishedTest(\PHPUnit_Framework_Test $test)
+    protected function writelnFinishedTest(TestEvent $e, Message $result)
     {
+        $test = $e->getTest();
         if ($this->isDetailed($test)) {
             return;
         }
 
+        if ($this->output->isInteractive()) {
+            $this->output->write("\x0D");
```
Naktibalda: I had to lookup this one.
Why do you print \r here?

src/Codeception/Subscriber/Console.php
```diff
@@ -452,115 +462,81 @@ public function printScenarioTrace(ScenarioDriven $failedTest)
...
+        $atMessage->cut($this->width - 4 - strlen($title))->style('info');
+
+        $this->message($title)
+            ->prepend($prefix)
+            ->append($atMessage)
+            ->write();
     }
 
-    protected function writeFinishedTest(\PHPUnit_Framework_Test $test)
+    protected function writelnFinishedTest(TestEvent $e, Message $result)
```
Naktibalda: $e should be named $event.
The only place where $e is acceptable is catch (Exception $e)

src/Codeception/Test/Gherkin.php
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    public function getScenarioText($format = 'text')
+    {
+        return file_get_contents($this->getFileName());
+    }
+
+    public function getSourceCode()
+    {
+        // TODO: Implement getRawBody() method.
```
Naktibalda: TODO here

tests/cli/RunSkippedCept.php
```diff
@@ -3,6 +3,7 @@
 $I->wantTo('run skipped test');
 $I->amInPath('tests/data/sandbox');
 $I->executeCommand('run skipped SkipMeCept.php');
-$I->seeShellOutputMatches("~\(SkipMeCept\)[\s\.]*?Skipped~");
+$I->seeInShellOutput("S Skip it");
```
Naktibalda: Shouldn't it be "I skip it"?

tests/data/claypit/tests/_support/ScenarioGuy.php
```diff
@@ -20,10 +20,45 @@ class ScenarioGuy extends \Codeception\Actor
...
-    function seeCodeCoverageFilesArePresent()
+    public function seeCodeCoverageFilesArePresent()
     {
         $this->seeFileFound('c3.php');
         $this->seeFileFound('composer.json');
         $this->seeInThisFile('codeception/c3');
     }
+
+    /**
+     * @Given I have terminal opened
```
Naktibalda: Is it mandatory to have a Given line in the feature file?
This one is so useless that it would be better to skip it.

tests/unit/Codeception/Test/GherkinTest.php
```diff
@@ -0,0 +1,141 @@
...
+        $test->getMetadata()->setServices($this->getServices());
+        $test->test();
+        $this->assertEquals('aXc', self::$calls);
+    }
+
+    public function testMatchingPatterns()
+    {
+        $pattern = 'hello :name, are you from :place?';
+        $regex = $this->loader->makePlaceholderPattern($pattern);
+        $this->assertEquals('/hello "(.*?)", are you from "(.*?)"\?/', $regex);
```
Naktibalda: Why are the quotes added around placeholders?

composer.json
Naktibalda: Add behat/gherkin to suggest if you don't want to add it to require.

src/Codeception/Subscriber/Console.php
DavertMik: Let me explain. In interactive mode we can control the caret on a line. For a test we can print currently started test:

```
- CurrentTest
```
and when test finishes we replace that line with result + test name + time:

```
✓CurrentTest (0.1s) 
```
so we use \r character to print that line on top of previous one

tests/cli/RunSkippedCept.php
DavertMik: No, `S` is a status character, first in line

tests/data/claypit/tests/_support/ScenarioGuy.php
DavertMik: there is no real difference between `@Given`, `@When` and `@Then` but they should be present in annotation to become a gherkin step

tests/unit/Codeception/Test/GherkinTest.php
DavertMik: there are 2 step formats: regular string with placeholders and regex
so yes, we convert placeholders to a compatible regex.
In regex string we use quotes to highlight the variables.

src/Codeception/Test/Gherkin.php
DavertMik: removed the TODO, but I'm not planning to implement it for now =)

micaweb:
This is awesome Mr. Davert! I have like forever that 'Herkin language to Gherkin' gist bookmarked. Thank you so much!

PR #2751, opened 2016-01-30
AngularJS module

DavertMik:
* extends WebDriver module
* adds async waiter (taken from Protractor) after each action
* adds strict locator rule for finding elements by model
* added a test suite and basic test


src/Codeception/Module/AngularJS.php
```diff
@@ -0,0 +1,118 @@
+<?php
+namespace Codeception\Module;
+
+use Codeception\Step;
+use WebDriverBy;
```
DavertMik: should be `Facebook\Webdriver\WebDriverBy`

DavertMik:
@zbateson @Naktibalda can you take a brief look into it?

Naktibalda:
I can't comment about Javascript code and have nothing to say about PHP.

DavertMik:
> I can't comment about Javascript code

just don't care of it, it was copy-pasted from Protractor clientscripts.js.
the only thing I'm not satisfied here are the tests, as they are launching 3rd-party Angular application. I hope soon we get our own test site with Angular

DavertMik:
https://github.com/angular/protractor/blob/master/lib%2Fclientsidescripts.js#L50

Issue #2124, opened 2015-07-03
Problem with Guzzle on php 5.4


MAXakaWIZARD:
Codeception 2.1.x requires php >=5.4.0
But when running `codecept.phar` on php 5.4, I see such error:
`Call to undefined function GuzzleHttp\Handler\curl_reset() `
Function `curl_reset` was added in php 5.5, so it seems that we have guzzle >=6.0.0 in phar archive (maybe phar was built on php 5.5+). I think phar should be built on php 5.4, as per Codeception requirement.

atmoz:
+1

Got the same issue. Codeception 2.1.0 does not work with PHP 5.4.

DavertMik:
Ok, to clarify the issue: Codeception Phar is not working with PHP 5.4 
If you install it via Composer you will have it working

delboy1978uk:
i was about to say something until I saw the EOL is soon. I think I'll just drop testing for 5.4

Naktibalda:
I solved this issue by installing Guzzle 5.x with composer.

Glutexo:
The fact, you found a solution does not mean that the bug does not exist any more. Even though the EOL of PHP 5.4 is comming really soon, the Codeception website states that it is compatible with PHP 5.4, which it actually isn’t. At least not the PHAR package.

webfacer:
i have also on ovh server the 5.4 PHP version. Inpossiple to use the 6.x version of Guzzle. Please reopen this issue.

Naktibalda:
If it helps.

Naktibalda:
@DavertMik I believe that not providing a build for PHP 5.4 hurts our adoption on PHP 5.4.
It wouldn't be that hard to make 2 builds for every release.

For everyone else:

Here are instructions how to make your own build of Codeception with Guzzle 5.

git clone https://github.com/Codeception/Codeception/
cd Codeception
composer require guzzlehttp/guzzle 5.3.*
wget http://robo.li/robo.phar
php -d phar.readonly=0 robo.phar build:phar

Result: http://s000.tinyupload.com/index.php?file_id=05032114061555932630

Naktibalda:
Codeception 2.1.3 for PHP 5.4: http://s000.tinyupload.com/index.php?file_id=05304194827543980861

aniebiet:
Thanks @Naktibalda

agarzon:
Thank you @Naktibalda it works.

Naktibalda:
[...]

PR #2752, opened 2016-01-30
Generate separate phar file for PHP 5.4

Naktibalda:
PHP5.4 phar will be bundled with Guzzle 5.3 and Symfony 2.8
The main phar file will be bundled with the latest version of Guzzle 6 and Symfony 3.

Solves #2124

### DIFF
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -14,6 +14,9 @@ class RoboFile extends \Robo\Tasks
         $this->update();
         $this->buildDocs();
         $this->publishDocs();
+        $this->installDependenciesForPhp54();
+        $this->buildPhar54();
+        $this->installDependenciesForPhp55();
         $this->buildPhar();
         $this->publishPhar();
         $this->publishGit();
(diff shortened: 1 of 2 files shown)


RoboFile.php
```diff
@@ -142,13 +145,74 @@ public function testCli()
...
+        $this->installDependencies($dependencies);
+    }
+
+
+
+    private function installDependencies($dependencies)
+    {
+        $composerRequireCommand = str_replace(' install', ' require --no-update ', $this->taskComposerInstall()->getCommand());
+        foreach ($dependencies as $dependency => $version) {
+            $code = $this->taskExec($composerRequireCommand . $dependency . ' ' . $version )->printed(true)->run()->getExitCode();
```
DavertMik: sometimes Nitpick is just doing too much

DavertMik:
restarted appveyor build, should pass now

RoboFile.php
```diff
@@ -142,13 +145,74 @@ public function testCli()
...
+            'symfony/css-selector' => '2.8.*',
+            'symfony/dom-crawler' => '2.8.*',
+        ];
+
+        $this->installDependencies($dependencies);
+    }
+
+    private function installDependenciesForPhp55()
+    {
+        $dependencies = [
```
DavertMik: I'd like not to define current dependencies here but use data from `composer.json`.
In current case dependencies are overwritten by those settings from Robofile which is not good and makes releases unpredictable.

I'd recomment to copy current `composer.json`, build php 5.4 version, restore `composer.json` and continue;

Naktibalda:
I found a better way to install version specific dependencies without hardcoding them in Robofile.

DavertMik:
So I think I can merge it?

PR #2755, opened 2016-01-31
fix Bug in FTP Module

k-serenade:
1. Add the argument of _after ()
2. $this->ftp to NULL, after ftp_close.  
If you do not do this, then _closeConnection() for ftp_close is re-executed when is called.
3. Fixed a typo flase to false


DavertMik:
Thank you!

PR #2757, opened 2016-01-31
New feature adding Custom Commands to codeception

sd-tm:
Hello Codeception Team, 

the same [pull request](https://github.com/Codeception/Codeception/pull/2746), but now to the master branch.

cu


src/Codeception/Lib/Interfaces/CustomCommand.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Codeception\Lib\Interfaces;
```
DavertMik: I'd like to  move that to Codeception\CustomCommandInterface to make it simpler to use. 
Also `Lib\Interface` is more like internal stuff, and actually interfaces there are related to modules

DavertMik:
Something strange happened with Semaphore CI build, I will merge this PR without it.

sd-tm:
Some code ist comming from branch 2.1, because we have started from Branch 2.1.

That are not my Code: 
 - tests/unit/Codeception/Module/TestsForWeb.php
 - src/Codeception/Lib/InnerBrowser.php

docs/08-Customization.md
```diff
@@ -179,6 +179,25 @@ Passed configuration is accessible via `config` property: `$this->config['param'
...
+You can add your own commands to codeception.
+
+Your custom commands have to implement the interface Codeception\CustomCommandInterface,
+because there has to be a function to get the name of the command.
+
+You have to register your command in the file `codeception.yml` 
+
+```yaml
+extensions:
+    commands: [Codeception\Command\MyCustomCommand]
```
DavertMik: you use Codeception namespace here. I think your other example suites better:

```
commands: [project\MyNewCommand]
```

docs/08-Customization.md
```diff
@@ -179,6 +179,25 @@ Passed configuration is accessible via `config` property: `$this->config['param'
...
+```
+
+If you want to activate the Command globally, because you use more then one ```codeception.yml``` file, then
+you have to register your command in ```codeception.dist.yml``` in the root folder of your project.
+
+You can find an example here: [/tests/data/register_command/examples/MyCustomCommand.php]
```
DavertMik: such permalink won't work on our site. However, before it's merged to master we can't get such link working (
anyway, let's keep it as it is and update once we need it. Maybe we can have a gist for now?

DavertMik:
I'm ok with merging it. Anyone wants to review it? 
@sergeyklay @Naktibalda 
If there will be no objections I will merge it today with Gherkin branch too

docs/08-Customization.md
sd-tm: I think, it easy to get your access to our fork.

sd-tm:
:clap:  :+1:

docs/08-Customization.md
Naktibalda: https://gist.github.com/

src/Codeception/Application.php
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    /**
+     * Pre load Configuration, the config option is use.
+     *
+     * @param string $configfile Path to Configuration
+     *
+     * @throws ConfigurationException
+     */
+    protected function preloadConfigration($configfile)
```
Naktibalda: Mistype in the method name
$configfile should be named $configFile

src/Codeception/Application.php
```diff
@@ -0,0 +1,172 @@
...
+        return $inputDefinition;
+    }
+
+    /**
+     * Search for --config Option and if found will be loaded
+     *
+     * example:
+     * -c file.yml
+     * --config file.yml
+     * --config=file.yml
```
Naktibalda: Two issues here:
1. -c is usually used to provide a path to directory, not a path to file,
example: php codecept build -c frameworks-yii-basic/tests
http://codeception.com/docs/reference/Commands 
https://github.com/Codeception/Codeception/blob/2.1/.travis.yml#L118
2. -cfile.yml (without spaces) must be supported too, to be fully compatible with http://symfony.com/doc/current/components/console/console_arguments.html

sd-tm: What is the best solution? 
To take other character, remove it or change build shortcut?

docs/08-Customization.md
sd-tm: Like that: https://gist.github.com/sd-tm/3b8eea3c15795b99fc77 ?

src/Codeception/Application.php
DavertMik: I don't think that we need to support 2.
but I agree about 1.

src/Codeception/Application.php
```diff
@@ -0,0 +1,172 @@
...
+    {
+        if ($this->coreArguments !== null) {
+            return $this->coreArguments;
+        }
+
+        $argv = $_SERVER['argv'];
+        $argvWithoutConfig = array();
+
+        for ($i = 0; $i < count($argv); $i++) {
+            if (preg_match("/^(?:-c|--config(?:=(.*))?)$/", $argv[$i], $match)) {
```
DavertMik: I'm not sure I understand how this works. But I'd like to be sure it supports

```
-c src/AppBundle
-c src/AppBundle/codeception_extension.yml
--config src/AppBundle/codeception_extension.yml
--config src/AppBundle
```

src/Codeception/Application.php
Naktibalda: If we don't support 2, it could break someone's CI.

DavertMik: actually yes, you are right. I thought that will affect only custom commands, but actually it's affecting all the commands. So we need 100% Symfony Console compatibility for dealing with options

docs/08-Customization.md
DavertMik: nope, I meant you can publish /tests/data/register_command/examples/MyCustomCommand.php as gist and have it as permalink ) 
anyway, maybe some url shortener like bit.ly can help us to have a permalink which can be changed.

src/Codeception/Application.php
DavertMik: Just checked, those formats are supported so I'd like to have it merged

DavertMik:
Thanks. I think using `-cWithPath` or `-c=path` is a really minor issue and not documented anywhere, so I think we can have custom commands to be landed in master. Thanks for awesome work!

sergeyklay:
:+1:

sd-tm:
:lollipop: WOW!! I'am happy to can help this awesome project. Thank You!

PR #2766, opened 2016-02-03
Add Sequence per suite generation

johnatannvmd:
This feature is usefull for huge suites with multiple depend steps within.

### DIFF
--- a/src/Codeception/Module/Sequence.php
+++ b/src/Codeception/Module/Sequence.php
@@ -11,7 +11,8 @@ use Codeception\TestCase;
  * you can use generated unique names, that should not conflict.
  * When you create article on a site, for instance, you can assign it a unique name and then check it.
  *
- * This module has no actions, but introduces a function `sq` for generating unique sequences.
+ * This module has no actions, but introduces a function `sq` for generating unique sequences within test and
+ * `sqs` for generating unique sequences across suite.
  *
  * ### Usage
  *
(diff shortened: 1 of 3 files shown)


DavertMik:
Can you provide more detailed description. What problem are you solving with this? 
Thanks

johnatannvmd:
Simply, it generates sequence which values are the same during all Cest suite tests.

For example: I test email server: I generate unique emails and send it to the server. Then I just check is the email on the server in different ways.

```php
<?php
class EmailTest
{
    public function sendEmail(AcceptanceTester $I)
    {
        // Takes about 5 sec, depends on connection
        $I->sendEmail('email' . sqs('user') . '@mailserver.com', sqs('subject'), sqs('body'));
    }

    /**
     * @depends sendEmail
    */
    public function checkEmail(AcceptanceTester $I)
    {
        $I->seeInSubjectTo('email' . sqs('user') . '@mailserver.com', sqs('subject'));
        $I->seeInBodyTo('email' . sqs('user') . '@mailserver.com', sqs('body'));
    }

    /**
     * @depends sendEmail
    */
    public function removeEmail(AcceptanceTester $I)
    {
        $I->removeEmail('email' . sqs('user') . '@mailserver.com');
        $I->dontSeeEmailTo('email' . sqs('user') . '@mailserver.com');
    }
}
?>
```

or I test site's backoffice via WebDriver: I login as admin, create user, change his profile. And if I create a user on every step and remove him it would take too much time rather than I create a user at the very first step and then doing some actions with him before deleting in the end.

If Cest Classes support `_beforeSuite` and `_afterSuite`, then `sqs` function will become unneeded.

Naktibalda:
I didn't know that we have a Sequence module, but I had a use case for a sequence.

In my tests I needed to generate 2 globally unique ids per test.
So I ended up doing this:

Test1Cept.php
```
$paymentId1 = '10' . substr(time(), 2);
$paymentId2 = $paymentId1 + 1;
```
Test2Cept.php
```
$paymentId1 = '11' . substr(time(), 2);
$paymentId2 = $paymentId1 + 1;
```

Test3Cept.php
```
$paymentId = '12' . substr(time(), 2);
```

What I actually needed was a global counter (but using globals is bad, so I didn't even consider it).

Why is this module called Sequence if it is actually a unique id generator?

I could have made use of it if it had a setting which accepted a callback to set an initial value of the sequence, and a setting to use ++ instead of unique_id or a setting to set a generator function.

DavertMik:
> Why is this module called Sequence if it is actually a unique id generator?

good question :) Probably I was inspired by this https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#sequences

DavertMik:
> If Cest Classes support _beforeSuite and _afterSuite, then sqs function will become unneeded.

from the architectural point of view this is a bad idea. One Cest should not affect others by doing something before and after the whole suite. So yes, Helpers and Modules should be used for such cases. In other way you won't get a clue where does those preparations happen.

DavertMik:
Looks good to merge

PR #2770, opened 2016-02-05
console option are handled like cli standard

sd-tm:
Hi,
in this PR is the --config option treated as cli standard.
by
Tobi


DavertMik:
Just awesome. Thank you!

sd-tm:
:smiley:  Thank you, too!

PR #2777, opened 2016-02-08
Limit step output length to 1 line (to Gherkin)

Naktibalda:
Example of effect:
Command: `php codecept run tests/cli/RunCest.php:Failed --steps`
Output:
```
Cli Tests (28) ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Run test with failed scenario RunCest.php
Scenario --
 I am in path "tests\/data\/sandbox"
 I execute command "run scenario FailedCept --steps --no-exit"
 I see in shell output "Fail when file is not found at FailedCept.php\nScenario --\n I am in path \".\"\n I see file found \"games.zip\"\n FAIL"
 I expect to see scenario trace
 I see in shell output "Scenario Steps:\n\n 2. $I->seeFileFound(\"games.zip\") at tests\/scenario\/FailedCept.php:5\n 1. $I->amInPath(\".\") at tests\/scenario\/FailedCept.php:4\n"
 PASSED 

------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------


Time: 557 ms, Memory: 8.00Mb
```

Output in narrower window:
```
Cli Tests (28) -------------------------------------------------------------------------------------------
Run test with failed scenario RunCest.php
Scenario --
 I am in path "tests\/data\/sandbox"
 I execute command "run scenario FailedCept --steps --no-exit"
 I see in shell output "Fail when file is not found at FailedCept.php\nScenario --\n I am..."
 I expect to see scenario trace
 I see in shell output "Scenario Steps:\n\n 2. $I->seeFileFound(\"games.zip\") at tests\/..."
 PASSED 

----------------------------------------------------------------------------------------------------------
```

Even narrower:
```
Cli Tests (28) ----------------------------------
Run test with failed scenario RunCest.php
Scenario --
 I am in path "tests\/data\/sandbox"
 I execute command "run scenario..."
 I see in shell output "Fail whe..."
 I expect to see scenario trace
 I see in shell output "Scenario..."
 PASSED 

-------------------------------------------------
```

Test in even smaller window failed, because it affected shell output.

PR #2778, opened 2016-02-09
Improved steps output in 2.1 branch

Naktibalda:
Backporting some changes made in https://github.com/Codeception/Codeception/pull/2777 to 2.1

Don't trim external [ if the first or the argument is array;
Display "Closure" instead of "lambda function"
Eliminated "callable function" - a full content of array (or any other callable) is displayed, 
Objects in array are replaced with class name.

### DIFF
--- a/src/Codeception/Step.php
+++ b/src/Codeception/Step.php
@@ -98,11 +98,12 @@ abstract class Step
 
     protected function getArgumentsAsString(array $arguments)
     {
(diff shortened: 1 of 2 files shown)


sergeyklay:
yay

sergeyklay:
I used this hack https://github.com/phalcon/incubator/blob/master/.travis.yml#L78

Issue #1719, opened 2015-02-14
Local acceptance tests show 0% coverage


stephencroberts:
Hi, I'm working with codeception to test a Joomla! extension I'm developing. I've written a simple acceptance test (which was incredibly easy to do) and am running it locally with phpbrowser, and I've confirmed it's working as expected. However, I'm having issues where the code coverage report is showing that no code is being touched by the acceptance test. 

In codeception.yml I have this:

```
coverage:
    enabled: true
    include:
        - testapp/index.php
```

I'm including index.php in the coverage because it should always be executed for every request. I've included c3.php in index.php and I know it's being loaded. When I run `codecept run acceptance --coverage-html`, the report gets generated, but it looks like this:

![b0747277691f221b1a1616ea7e606657](https://cloud.githubusercontent.com/assets/5873529/6200957/7a138dee-b456-11e4-84d9-2a606a17167c.png)

I've confirmed xdebug is installed and remote_enabled is on. I've also tried setting the `remote: true` config option, but that generates the same result. Nothing obvious shows in apache logs (my site is running the traditional LAMP stack). Any ideas? It would be great to have a combined report of all code coverage between unit and acceptance tests. Thanks!

ThePeach:
I've had a similar problem (see #1621 and #1622) which was related to the unstable version of CodeCeption being installed. Could you please ensure you're running the stable version?

stephencroberts:
Yeah, I'm using 2.0.10.1 (the latest release). Any ideas what else I can try?

Ragazzo:
did you set correct `remote_config` right ?

stephencroberts:
Which `remote_config` are you referring to? I tried setting `remote: true` in codeception.yml. But if I understand it correctly, since I'm running it locally with phpbrowser, I don't need to connect to it as a remote. Locally is the only way to get a combined report of coverage for all types of tests. Correct?

Ragazzo:
Check guide for `[yii2-basic](https://github.com/yiisoft/yii2/blob/master/apps/basic/tests/README.md#remote-code-coverage)`

DavertMik:
@stephencroberts sorry, this kind of things are really hard to debug in discussion mode )
What I recommend to do is to put series of var_dumps into c3.php to see if it is actually gets executed in requests. Because it should.

stephencroberts:
[...]

PR #2781, opened 2016-02-10
Empty coverage simple workaround

neronmoon:
Sometimes coverage is empty because c3 creates coverage file later, than afterSuite executed.
Related to #1719

[Review on Reviewable](https://reviewable.io/reviews/codeception/codeception/2781)

PR #2785, opened 2016-02-11
Empty coverage simple workaround

neronmoon:
Sometimes coverage is empty because c3 creates coverage file later, than afterSuite executed.
Related to #1719

[Review on Reviewable](https://reviewable.io/reviews/codeception/codeception/2785)


neronmoon:
@DavertMik #2781 resended

DavertMik:
Thanks!

PR #2786, opened 2016-02-11
Wrong format in BrowserStack host address

javigomez:


DavertMik:
Oups. I will update docs myself as it should be done in WebDriver module itself

javigomez:
Oh! Ok, so if I see an issue next time, I have only to modify the docblock at the module: src/Codeception/Module/`module`.php... right? and then do I need to run some Robo command to generate docs? 

I saw https://github.com/Codeception/Codeception/commit/b38b519220ed5f877c081c7de8331ecb1dd607fa but not sure to understand the process.

DavertMik:
> and then do I need to run some Robo command to generate docs?

Yes, it's `robo build:docs`

javigomez:
nice! thanks :+1:

PR #2787, opened 2016-02-11
Used latest PHPUnit

sergeyklay:


Naktibalda:
What's the point? 
4.8 is the last minor version of 4.* series and your change does not allow 5.*

sergeyklay:
As you know current master does not allow 5.* anyway ;)

sergeyklay:
```
$ composer require --dev codeception/codeception:dev-master
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - phpunit/phpunit 4.8.20 requires php ~5.3.3|~5.4|~5.5|~5.6 -> your PHP version (7.0.1-1+deb.sury.org~trusty+2) or value of "config.platform.php" in composer.json does not satisfy that requirement.
```

Naktibalda:
run `composer update` first, 4.8.20 is the only version of PhpUnit which can't be installed on PHP 7,
4.8.19 and 4.8.21+ can - https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-4.8.md

PR #2788, opened 2016-02-11
Don't display php54 as a minor version of Codeception in /builds

Naktibalda:
http://codeception.com/builds

php54. is displayed above 2.1, this patch fixes that


DavertMik:
:+1:

PR #2789, opened 2016-02-11
REST: Added test to prove that HTTP_HOST header can be set

### LINKED ISSUE #1650: REST.php $I->haveHttpHeader not working as expected
I noticed that I can not set the Host header using:
```
$I->haveHttpHeader("Host", "www.expected.com")
```
When I do that, I see in my Apache log files that I'm getting the DNS name I used to make the request instead of the value I used e.g. "www.expected.com"
***However*** If I do the following:
```
$I->haveHttpHeader(" Host", "www.expected.com")
```
I can see the value of "www.expected.com" in my Apache logs. 

Is this a bug or a feature? Is this how I need to set header variables?

I'm using codeceptino 2.0.9 and php 5.5.10

Naktibalda:
Closes #1650 

My patch was accepted to symfony/browserkit on 27th of January and released in 3.0.2 version,
2.* branches will be fixed in the next release cycle.

### DIFF
--- a/tests/data/rest/index.php
+++ b/tests/data/rest/index.php
@@ -28,7 +28,11 @@ $GLOBALS['RESTmap']['GET'] = [
                 0
             ],
         ];
+    },
+    'http-host' => function() {
+        return 'host: "' . $_SERVER['HTTP_HOST'] . '"';
     }
+
 ];
 
 $GLOBALS['RESTmap']['POST'] = [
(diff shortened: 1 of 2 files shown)


Naktibalda:
My test passed on PHP 5.5+ and HHVM builds that are using Symfony 3.
I will keep this PR open until Symfony 2.7.10 and 2.8.3 are released. http://symfony.com/blog/category/releases

raistlin:
Both Symfony versions have been released.

PR #2792, opened 2016-02-12
Allow PHPUnit 5

Naktibalda:


DavertMik:
you can also simplify travis.yml and remove some conditionals in it, like installing symfony 2.7 and guzzle 5 for PHP 5.4, and so on

githubjeka:
:balloon: Awesome

PR #2793, opened 2016-02-12
Fix for codecept autoloader search paths

zinovyev:
I've tried to call codeception executable from the vendor/bin folder. But there was an error related to the search path of the autoload.php file. This fix should help codecept to find autoloader even if in the bin directory of another project

### DIFF
--- a/codecept
+++ b/codecept
@@ -4,7 +4,15 @@
  * Codeception CLI
  */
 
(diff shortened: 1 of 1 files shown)


zinovyev:
Sorry, my fault. Not a bug)

PR #2799, opened 2016-02-14
Making testing for phalcon 1.3.x usable

someson:
Due to differences of declaration for some function arguments in
different phalcon versions (1.3.x -> 2.x) was it unable to run tests on
version 1.3.x. The commit is an offer to make both (1.x and 2.x) version
usable for testing.


sergeyklay:
As you can see, we use Phalcon 2. This is why tests fail

DavertMik:
Phalcon maintainers says that won't support Phalcon < 2.1 (after the March 1st), so you should upgrade as well

someson:
Yeah, I probably schould. Thanks

PR #2802, opened 2016-02-15
Db module: Connect only if setting reconnect is true or there is no p…

### LINKED ISSUE #2801: Db module reconnects even if reconnect = false
After having many functional tests running I ended up with the error:

`An exception occured in driver: SQLSTATE[08004] [1040] Too many connections`

I am using symfony2 where I have a call to end the connection in _after method.

I made some digging and realised that the Db module calls connect with every run of a test, even if my setting reconnect is set to false. This is creating new connections with every test run and with the PDO disability to disconnect this creates many connections and results in failing my suite.

thanpa:
…revious connection - fixes #2801

### DIFF
--- a/src/Codeception/Module/Db.php
+++ b/src/Codeception/Module/Db.php
@@ -73,21 +73,21 @@ use Codeception\TestCase;
  *              reconnect: true
  *
  * ### SQL data dump
- * 
+ *
  *  * Comments are permitted.
  *  * The `dump.sql` may contain multiline statements.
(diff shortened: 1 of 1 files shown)


Naktibalda:
_initialize() method should only be called once per suite (before running tests).
Are you saying that it is executed before each test?

thanpa:
Hm looks like it's my bad, some debug code I left called it many times. sorry pls ignore the pull request.

Issue #2805, opened 2016-02-16
Queue module clearQueue function not working with AmazonSQS queue


mikitu:
```
1) Failed to perform actions and see result in WelcomeCept (tests/acceptance/WelcomeCept.php)
 [Error] Call to undefined method Codeception\Lib\Driver\AmazonSQS::debug()
Scenario Steps:
 1. $I->clearQueue("my-queue-name") at tests/acceptance/WelcomeCept.php:5
```
the bug is located in \Codeception\Lib\Driver\AmazonSQS::clearQueue line 100
`$this->debug("  - delete message: " . $msg['MessageId']);`

Naktibalda:
Could you fix it and make a pull request to 2.1 branch?

PR #2806, opened 2016-02-16
AmazonSQS module: fix call to undefined method Codeception\Lib\Driver…

mikitu:
…\AmazonSQS::debug()

fix issue: https://github.com/Codeception/Codeception/issues/2805

### DIFF
--- a/src/Codeception/Lib/Driver/AmazonSQS.php
+++ b/src/Codeception/Lib/Driver/AmazonSQS.php
@@ -8,133 +8,133 @@ use Aws\Common\Credentials\Credentials;
 
 class AmazonSQS implements Queue
 {
-  protected $queue;
+    protected $queue;
 
-  /**
-   * Connect to the queueing server. (AWS, Iron.io and Beanstalkd)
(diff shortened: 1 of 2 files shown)


src/Codeception/Lib/Driver/AmazonSQS.php
```diff
@@ -17,10 +17,14 @@ class AmazonSQS implements Queue
...
     public function openConnection($config)
     {
-        $this->queue = SqsClient::factory([
-            'credentials' => new Credentials($config['key'], $config['secret']),
+        $params = [
             'region' => $config['region']
-        ]);
+        ];
+        if (!empty($config['profile'])) {
+            $params['profile'] = $config['profile'];
```
Naktibalda: Please document the new parameter in Config section of Queue documentation https://github.com/Codeception/Codeception/blob/2.1/src/Codeception/Module/Queue.php#L39

DavertMik:
Looks like we can merge it. @Naktibalda, right?

PR #2808, opened 2016-02-16
DataFactory Module

DavertMik:
Added DataFactory module, which uses FactoryMuffin to generate data for tests. More docs to come


sergeyklay:
:+1:

PR #2821, opened 2016-02-20
Datafactory module

DavertMik:
Cleaner PR for DataFactory module

* Includes dependnecy on FactoryMuffin package
* Added Doctrine support
* Tests needed

### DIFF
--- a/src/Codeception/Module/DataFactory.php
+++ b/src/Codeception/Module/DataFactory.php
@@ -1,0 +1,246 @@
+<?php
+namespace Codeception\Module;
+
+use Codeception\Lib\Interfaces\DataMapper;
+use Codeception\Lib\Interfaces\DependsOnModule;
+use Codeception\Lib\Interfaces\ORM;
+use Codeception\TestInterface;
+use League\FactoryMuffin\FactoryMuffin;
+use League\FactoryMuffin\Stores\RepositoryStore;
+
+/**
+ * DataFactory allows you to easily generate and create test data using [**FactoryMuffin**](https://github.com/thephpleague/factory-muffin).
+ * DataFactory uses an ORM of your application to define, save and cleanup data. Thus, should be used with ORM or Framework modules.
+ *
+ * This module requires packages installed:
+ *
+ * ```json
+ * {
+ *  "league/factory-muffin": "^3.0",
+ *  "league/factory-muffin-faker": "^1.0"
+ * }
+ * ```
+ *
+ * Generation rules can be defined in a factories file. You will need to create `factories.php` (it is recommended to store it in `_support` dir)
+ * Follow [FactoryMuffin documentation](https://github.com/thephpleague/factory-muffin) to set valid rules.
+ * Random data provided by [Faker](https://github.com/fzaninotto/Faker) library.
+ *
+ * ```php
+ * use League\FactoryMuffin\Faker\Facade as Faker;
+ *
+ * $fm->define(User::class)->setDefinitions([
+ *  'name'   => Faker::name(),
+ *
+ *     // generate email
+ *    'email'  => Faker::email(),
+ *    'body'   => Faker::text(),
+ *
+ *    // generate a profile and return its Id
+ *    'profile_id' => 'factory|Profile'
+ *]);
+ * ```
+ *
+ * Configure this module to load factory definitions from a directory.
+ * You should also specify a module with an ORM as a dependency.
+ *
+ * ```yaml
+ * modules:
+ *     enabled:
+ *         - Yii2:
+ *             configFile: path/to/config.php
+ *         - DataFactory:
+ *             factories: tests/_support/factories
+ *             depends: Yii2
+ * ```
+ *
+ * (you can also use Laravel4, Laravel5, Phalcon).
+ *
+ * In this example factories are loaded from `tests/_support/factories` directory.
+gst * You should create this directory manually and create PHP files in it with factories definitions following [official documentation](https://github.com/thephpleague/factory-muffin#usage).
+ *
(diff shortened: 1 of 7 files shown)


src/Codeception/Module/DataFactory.php
```diff
@@ -0,0 +1,169 @@
...
+/**
+ * DataFactory allows you to easily generate and create test data using [**FactoryMuffin**](https://github.com/thephpleague/factory-muffin).
+ * DataFactory uses an ORM of your application to define, save and cleanup data. Thus, should be used with ORM or Frqmework modules.
+ *
+ * Generation rules can be defined in a factories file. You will need to create `factories.php` (it is recommended to store it in `_support` dir)
+ * Follow [FactoryMuffin documentation](https://github.com/thephpleague/factory-muffin) to set valid rules.
+ * Random data is set with [Faker](https://github.com/fzaninotto/Faker) library.
+ *
+ * ```php
+ * use League\FactoryMuffin\Faker\Facade as Faker;
```
Naktibalda: Faker support is provided by Factory Muffin Faker. If you want to enable faker support, then you need to add "league/factory-muffin-faker": "^1.0" too:

Naktibalda:
Would it be possible to make it work with Db module?

DavertMik:
> Would it be possible to make it work with Db module?

I think it's possible to make it, the API is pretty flexible. However, I'm not going to implement this in near future. It will require an extension to Muffin itself, which will pass `Db` module as storage.

src/Codeception/Module/DataFactory.php
```diff
@@ -0,0 +1,169 @@
...
+
+use Codeception\Lib\Interfaces\DataMapper;
+use Codeception\Lib\Interfaces\ORM;
+use Codeception\TestCase;
+use League\FactoryMuffin\FactoryMuffin;
+use League\FactoryMuffin\Stores\RepositoryStore;
+
+/**
+ * DataFactory allows you to easily generate and create test data using [**FactoryMuffin**](https://github.com/thephpleague/factory-muffin).
+ * DataFactory uses an ORM of your application to define, save and cleanup data. Thus, should be used with ORM or Frqmework modules.
```
Naktibalda: Comment says that Framework modules are supported.

src/Codeception/Module/DataFactory.php
```diff
@@ -0,0 +1,169 @@
...
+
+    public function _initialize()
+    {
+        if (!class_exists('League\FactoryMuffin\FactoryMuffin')) {
+            throw new \Exception('FactoryMuffin not installed. Please add `"league/factory-muffin": "~3.0"` to composer.json');
+        }
+
+        $store = null;
+        if ($this->orm instanceof DataMapper) { // for Doctrine
+            $store = new RepositoryStore($this->orm->_getEntityManager());
```
Naktibalda: But Doctrine is the only module implementing DataMapper interface.

Naktibalda:
> Would it be possible to make it work with Db module?
> 
> I think it's possible to make it, the API is pretty flexible. However, I'm not going to implement this in near future. It will require an extension to Muffin itself, which will pass Db module as storage.

I don't see how it could require an extension to Muffin.

It should be as simple as
```php
$store = new RepositoryStore($this->db, '_saveModel', , '_deleteModel', '_flushModel);
```

And this is how the code of Db::_saveModel($model) could look like:
```php
public function _saveModel($model)
{
  $this->haveInDatabase(get_class($model), (array)$module);
}
```

src/Codeception/Module/DataFactory.php
DavertMik: For now yes. I added DataMapperInterface just for Doctrine :(

DavertMik:
Wow. Good idea. But `$model` is not a class in case we deal with Db module. FactoryMuffin operates with ORM classes.

Naktibalda:
> But $model is not a class in case we deal with Db module. FactoryMuffin operates with ORM classes.

It is an object, because it was generated by FactoryMuffin, the example I gave transforms ORM class to format suitable for `haveInDatabase`.

DavertMik:
Ok, so maybe you can provide an additional commit into this branch with this integration?
I think you have access to it, so you can commit just into this PR

Naktibalda:
I am not familiar with ORMs and I don't know if there is a common way for retrieving table name and field values from model.

get_class($model) is simple, but I think that the problems start if it is namespaced.
What to do in that case? Drop namespaces? Call getTableName method instead?

DavertMik:
> I am not familiar with ORMs and I don't know if there is a common way for retrieving table name and field values from model.

It depends on particular ORM you are working with. In case of Db module we won't work with real models but with an object representing a table row. 

So probably in our case we need to replace muffin's generator so it would create not an object but an array which could be passed to `haveRecord` of database.

```
$user = $fm->generate('users'); // -> ['user', ['name' => 'asdsadsadasd', 'email' => 'asdas@asdasd']]
call_user_func_array([$db, 'haveRecord'], $user);
```

src/Codeception/Module/DataFactory.php
Naktibalda: Ah, ok, ModelStore is used by default and default methods match Eloquent methods, so Laravel should be working.

Naktibalda:
`generate` is a protected method and it takes an instance of the model as the first parameter: https://github.com/thephpleague/factory-muffin/blob/master/src/FactoryMuffin.php#L237

Maybe it wasn't such a good idea - making DataFactory work with DB module requires too many assumptions.

DavertMik:
In theory it is possible, but for sure we won't have it in current PR. 
Users can use Faker + Db module, which can be pretty much the same

PR #2825, opened 2016-02-21
Typo in "dontSeeXmlResponseMatchesXpath"

mangust404:
Assertion should be negative. Specified xpath shouldn't be in XML structure.


Naktibalda:
Thank you

PR #2827, opened 2016-02-21
Merge 2.1 to master

Naktibalda:


src/Codeception/Step.php
```diff
@@ -157,6 +157,7 @@ protected function stringifyArgument($argument)
             } else {
                 $argument = $this->getClassName($argument);
             }
+            return $this->getClassName($argument);
```
Naktibalda: This line should not be here

Issue #2810, opened 2016-02-18
Bad encoding in HTML report


kerimovaa:
I met problem with encoding test's name in HTML report . Tests' names on russian language (Windows 10). 
For example , a piece of test: 

1. `$I = new AcceptanceTester\MainPageSteps($scenario);
2. $I->am('Клиент PROFI.RU'); 
3. $I->wantToTest('PROFI-4:Проверка отправки email');
4. $I->lookForwardTo('Создал заказ и отправил свой email');

`   

Launch of test : 
`codecept run testname -- html report.html`

And a piece of html report:
`<td>
     <h3>Acceptance Tests</h3>
 </td>
</tr>   <tr class="scenarioRow scenarioSuccess">
    <td>
     <p class="scenarioSuccess" onclick="showHide('1', this)">[+] Test profi-4:�������� �������� email (PROFI-4_Cept) <span style="color: #34495e; font-size: 70%;">83.19s</span></p>
    </td>
   </tr>`

And screenshot : https://yadi.sk/i/6bBAWKsgp4GUG

pejaycz:
Same problem with czech language

![problem](https://cloud.githubusercontent.com/assets/3164102/13214839/24fba388-d952-11e5-99e5-d4fe5be70225.png)

Naktibalda:
@pejaycz: Can this ticket be closed now? #2831

pejaycz:
@Naktibalda I think you can close it. I am not experiencing any problem with encoding, so i think its fixed for @kerimovaa too.

kerimovaa:
Ok,thanks! Close.

Issue #120, opened 2013-01-23
utf-8 support problems


gennadiylitvinyuk:
Hi

I want to report some little problems.

If I make tests with utf-8 characters, for example with ü

I see "wu00fcrstchen" instead of "würstchen" in html-reports

The checks themself are ok, only the output has problems.

githubjeka:
## FIX for Windows

You need to use Lucida console fonts.

You can set code page before going to work with cmd.

```
$I->seeInTitle('Система учёта');
I see in title "╨б╨╕╤Б╤В╨╡╨╝╨░ ╤Г╤З╤С╤В╨"
```

set new code page 

```
chcp 65001
```

```
$I->seeInTitle('Система учёта');
I see in title "Система учёта"
```

PR #2784, opened 2016-02-10
Limit step output length to 1 line

Naktibalda:
Example of effect:
Command: `php codecept run tests/cli/RunCest.php:Failed --steps`
Output:
```
Cli Tests (28) ----------------------------------------------------------------------
Run test with failed scenario RunCest.php
Scenario --
 I am in path "tests\/data\/sandbox"
 I execute command "run scenario FailedCept --steps --no-exit"
 I see in shell output "Fail when file is not found at FailedCept.php\nScenario -..."
 I expect to see scenario trace
 I see in shell output "Scenario Steps:\n\n 2. $I->seeFileFound(\"games.zip\") at..."
 PASSED 

-------------------------------------------------------------------------------------

```

PR #2831, opened 2016-02-22
Fix encoding problems in PHP prior to 5.6 related to #120 and #2810

pejaycz:
Fix utf-8 encoding issue in PHP < 5.6 (mb_* functions default charset)

## Before
problem

## After
fix

### DIFF
--- a/src/Codeception/Actor.php
+++ b/src/Codeception/Actor.php
@@ -36,7 +36,7 @@ abstract class Actor
 
     public function wantTo($text)
     {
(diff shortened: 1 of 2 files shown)


Naktibalda:
This will need to be redone for master branch, because I added more mb_strlen and mb_substr calls in https://github.com/Codeception/Codeception/pull/2784/files  (but now I see a few places that needs replacing with mb_ equivalents in it).

A few failing tests that could reproduce these multibyte issues would be nice too.

PR #2844, opened 2016-02-25
[imp] Add scrollTo function to Webdriver Module

### LINKED ISSUE #2840: Webdriver module missing moveToElement
Long story short: since Selenium 2.48 there is a new addition causing a lot of trouble. Is the known error:

```
[UnknownServerException] Element is not clickable at point (XXX.XXXXXXXX, YY.YYYYYY). Other element would receive the click:
```

this happens when you have a floating element over the element that you want to click:

capture 2016-02-03 at 17 43 53

You can seen the whole issue described at: https://github.com/SeleniumHQ/selenium/issues/1202

I have tried with this suggestion (https://github.com/SeleniumHQ/selenium/issues/1202#issuecomment-159909722):

screen shot 2016-02-24 at 17 48 30

Adding this in my acceptance.suite.yml:

```yaml
            capabilities:
              elementScrollBehavior: '1'
```
But doesn't seem to do the trick.

I was now trying the advise at https://github.com/SeleniumHQ/selenium/issues/1202#issuecomment-173656799. And is to use:

screen shot 2016-02-24 at 17 47 02

And I have noticed that despite that the Facebook Webdriver implements the function:

https://github.com/facebook/php-webdriver/blob/community/lib/Interactions/WebDriverActions.php#L170-L186

The Codeception Webdriver module does not implement it.

I was wondering how to implement it. It should be something like adding this:

````php
    public function moveTo($selector)
    {
        $el = $this->findField($selector);
        $this->webDriver->getMouse()->moveToElement($el);
    }
````
at: https://github.com/Codeception/Codeception/blob/2.1/src/Codeception/Module/WebDriver.php

But I'm missing something, because I get: `FATAL ERROR. TESTS NOT FINISHED.
Call to undefined method RemoteMouse::moveToElement() `

Any advise on how to implement this?

Thanks ^_^

### PR DESCRIPTION
**Pull request by javigomez**
This pull closes #2840

This pull is work in progress. The code partially works but I need advise from you @DavertMik  and @Naktibalda on the following topics:

1.- Passing `offsets` to Facebook Webdriver doesn't do any action. Please see the following 5 seconds video:
https://www.youtube.com/watch?v=OlCn3W4cj_E

In the video you will notice that `scrolling to an element` works :+1: , however `scrolling to that element passing offsets` doesn't seems to work :-1: . It doesn't looks like Codeception issue but a Facebook webdriver issue, but I would need your help to confirm.

I'm running this test, you can use it as example:

```php
<?php

class scrollToCest
{
	public function scrollToConfigurationInWebdriverPage(AcceptanceTester $I, $scenario)
	{
		$I->amOnUrl('http://codeception.com/docs/modules/WebDriver');
		$I->scrollTo(['id' => 'configuration']);
		$I->wait(1);
		$I->scrollTo(['id' => 'configuration'], 0, 100);
		$I->wait(1);
		$I->scrollTo(['id' => 'configuration'], 10, 200);
		$I->wait(1);
		$I->scrollTo(['id' => 'configuration'], 20, 300);
		$I->wait(1);
		$I->scrollTo(['id' => 'configuration'], 30, 400);
		$I->wait(1);
		$I->scrollTo(['id' => 'configuration'], 40, 500);
		$I->wait(1);
		$I->scrollTo(['id' => 'configuration'], 50, 600);
	}
}
```

2.- I was planning to add the test of issue 1. in the Codeception tests folder: https://github.com/Codeception/Codeception/blob/2.1/tests/web/WebDriverTest.php#L616. But I would need your advise, because I'm not sure how to generate the dummy page to be tested.


DavertMik:
Thanks! 
To create a test page you can create a page at `tests/app/view/form/example*.php` where you can put a test code and probably JS script which to some action on mouse over.

This page will be accessible at `/form/example*` url.

Use `robo test:launch-server` to start a test app locally

javigomez:
I have added a test, but something is wrong, and I'm not sure what. 

Please have a look at this small video and give me your advise: https://www.youtube.com/watch?v=3qZcx4qsPA0

Thanks very much.

DavertMik:
@javigomez I took a brief investigation and looks like `scrollTo` makes only sense for mobile browsers, so its implementation at CodeceptJS and WebdriverIO is just about that. For desktop-based browsers seelenium scrolls page automatically, so in our case it's preferable to call method `moveCursorTo` (yes, same method exists in CodeceptJS http://codecept.io/helpers/WebDriverIO/#movecursorto

Also I pulled from your branch and resent your PR https://github.com/Codeception/Codeception/pull/2882

Issue #1090, opened 2014-06-13
How to retrieve the base URL used by PHPBrowser?


pwenzel:
I have a kind of special situation where I want to iterate over all the URLs in my site's XML sitemap and perform tests on each URL. 

To achieve this, I would like to replace the `BASE_URL` constant below in my sample code below with the real base URL defined in my YML config file. 

I am also using the `--env` parameter when running my tests, so the URL could change based on the environment. 

So, how do I access the base 'url' parameter used by PHPBrowser? 

```
<?php
function getSitemapUrls() {
    $urls = array();
    $x = new SimpleXMLElement(file_get_contents(BASE_URL.'/sitemap.xml'));
    foreach ($x->url as $n) {
        $urls[] = $n->loc->__toString();
    }

    return $urls;
}
```

Thanks!

PR #2846, opened 2016-02-25
Examples

DavertMik:
Implementation of Examples in Cest format and [Scenario Outlines](https://github.com/cucumber/cucumber/wiki/Scenario-Outlines) for Gherkin. 
Examples are DataProviders alternative for scenario driven tests. 

Examples can be passed as doctrine-style annotations, json-object or json-array.

```php
    /**
     * @example(path=".", file="scenario.suite.yml")
     * @example(path=".", file="dummy.suite.yml")
     * @example(path=".", file="unit.suite.yml")
     */
    public function filesExistsAnnotation(ScenarioGuy $I, Example $example)
    {
        $I->amInPath($example['path']);
        $I->seeFileFound($example['file']);
    }


    /**
     * @example { "path":".", "file":"scenario.suite.yml" }
     * @example { "path":".", "file":"dummy.suite.yml" }
     * @example { "path":".", "file":"unit.suite.yml" }
     */
    public function filesExistsByJson(ScenarioGuy $I, Example $example)
    {
        $I->amInPath($example['path']);
        $I->seeFileFound($example['file']);
    }

    /**
     * @example [".", "scenario.suite.yml"]
     * @example [".", "dummy.suite.yml"]
     * @example [".", "unit.suite.yml"]
     */
    public function filesExistsByArray(ScenarioGuy $I, Example $example)
    {
        $I->amInPath($example[0]);
        $I->seeFileFound($example[1]);
    }
```

See included file.

### DIFF
--- a/src/Codeception/Example.php
+++ b/src/Codeception/Example.php
@@ -1,0 +1,77 @@
+<?php
+namespace Codeception;
+
+class Example implements \ArrayAccess
+{
+    protected $data;
+
+    public function __construct($data)
+    {
+        $this->data = $data;
+    }
+
+    /**
+     * Whether a offset exists
+     * @link http://php.net/manual/en/arrayaccess.offsetexists.php
+     * @param mixed $offset <p>
+     * An offset to check for.
+     * </p>
+     * @return boolean true on success or false on failure.
+     * </p>
+     * <p>
+     * The return value will be casted to boolean if non-boolean was returned.
+     * @since 5.0.0
+     */
+    public function offsetExists($offset)
+    {
+        return isset($this->data[$offset]);
+    }
+
+    /**
+     * Offset to retrieve
+     * @link http://php.net/manual/en/arrayaccess.offsetget.php
+     * @param mixed $offset <p>
+     * The offset to retrieve.
+     * </p>
+     * @return mixed Can return all value types.
+     * @since 5.0.0
+     */
+    public function offsetGet($offset)
+    {
+        if (!$this->offsetExists($offset)) {
+            throw new \PHPUnit_Framework_AssertionFailedError("Example $offset doesn't exist");
+        };
+        return $this->data[$offset];
+    }
+
+    /**
+     * Offset to set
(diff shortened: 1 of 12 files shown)


sergeyklay:
:+1:

GaryJones:
Could you please give an example of what the example `.yml` files would look like please?

Simply:

```yml
path: ...
file: ...
```
?

DavertMik:
You don't need any files to use examples. 

```
@example(path=".", file="scenario.suite.yml")
```

is just a fragment of internal test suite. We pass 2 parameters: `path` and `file`. Then we use them in tests like `$example['path']`.

The same way you can pass absolutely any param:

```php
/**
@example(color="brown")
@example(color="black")
*/
function colors(\ColorsTester $I, \Codeception\Example $example)
{
    $I->seeColorInPalette($example['color']);
}
```

maff:
Is there a way to generate the examples via code  - e.g. `@example` referencing a method which will be called to generate example data (similar to the `@dataProvider` annotation for PHPUnit)?

akiessling:
I have the same question as @maff : can i provide the examples via code? I want to implement something similar like in #1090, but when running this in Jenkins with xunit output, i only get one testrun with all urls.

mikeuding:
Yes, via this extension: https://github.com/edno/codeception-dataprovider

GaryJones:
I just wanted to follow-up here with how I've been using this functionality. Here's a small part of my `RedirectsCest.php`:

screenshot 2016-11-02 16 47 41

The multiple functions are to allow specific groups of URLs to be tested, so that re-checking to see if a specific`@example` data now passes, doesn't mean running the whole Cest.

Clearly, the downside is, all of the functions are basically the same, and call a separate private method.

It's not ideal, but it works for me.

alexpts:
How I Can pass php constant to @example?

edno:
Can be done using [@dataprovider](https://github.com/Codeception/Codeception/pull/3737) annotation.

PR #2847, opened 2016-02-25
Reproduced and fixed issue #2841

### LINKED ISSUE #2841: Bug POST Submit is passing GET variable as well, causing problems
Since codecept 2.1.* about 40 of our tests are failing. These are all tests in wordpress that involve saving wordpress settings. 
##Symptoms: 
when we click submit after adjusting the settings, the form fails to submit and the settings are not saved. The rest of the test is relying on the settings being saved, so the tests all fail, oh noes.

##Problem:
 We have been able to track down the problem, and reproduce it in isolation.
The Problem is, when codecept submits a POST form, if there is a GET variable in the current URI, it adds the GET to the POST. This makes the GET available when it should not be.
In the wordpress settings code, at least the part we are using, the code is relying on the GET being unavailable.
Because the GET is available, the app starts running down some other pathway, and never actually resolves the POST request.

I have isolated this problem in this [repo here](https://github.com/cloudEdwards/codecept-bug).
It is a very simple html form with a post action to the same file.

Load the file, and add a get var to the URL, and reload it. It will print out the Get var.
Now submit the form, and the Get var is gone, and you see the Post var with what ever value you entered in the form.
Now run the test that does the same thing, and you will see that both the Post and the Get are available after the form submits.

##Solution:
I will try to use form submit as a work around, or try and make a PR. If someone can point me in the right direction to where I can find this in the codecept codebase that would help.

Naktibalda:
Fixes #2841 

Don't keep query string if new URI sets a path.
Don't keep anchor if new URI sets a path or a query string.

### DIFF
--- a/src/Codeception/Util/Uri.php
+++ b/src/Codeception/Util/Uri.php
@@ -42,16 +42,17 @@ class Uri
                 }
             }
             $base = $base->withPath($path);
-        }
-        if (isset($parts['scheme'])) {
-            $base = $base->withScheme($parts['scheme']);
+            $base = $base->withQuery('');
+            $base = $base->withFragment('');
         }
         if (isset($parts['query'])) {
             $base = $base->withQuery($parts['query']);
+            $base = $base->withFragment('');
         }
         if (isset($parts['fragment'])) {
             $base = $base->withFragment($parts['fragment']);
         }
+
         return (string) $base;
 
     }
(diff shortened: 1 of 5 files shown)


Naktibalda:
I noticed that another missing feature is support for schemaless urls.

DavertMik:
Great. Thank you!

cloudEdwards:
Hey, thanks Naktibalda. 
What is the usual timeline for stable releases?

DavertMik:
@cloudEdwards ~1 month
but you can switch set 2.1 branch in Composer to receive the latest stable updates

PR #2852, opened 2016-02-28
Replaced strlen with mb_strlen in Codeception\Step

Naktibalda:
Used mb_strlen for measuring total output length and sorting parameters by length in Codeception\Step.

But started by writing a failing test.


DavertMik:
:+1:

PR #2856, opened 2016-03-01
REST::execute() small fix in argument defaults

sasha-ch:
Add default value for $url in Module\REST::execute() accoarding to 
"Note that when using default arguments, any defaults should be on the right side of any non-default arguments; otherwise, things will not work as expected." 
http://php.net/manual/en/functions.arguments.php#functions.arguments.default

### DIFF
--- a/src/Codeception/Module/REST.php
+++ b/src/Codeception/Module/REST.php
@@ -440,7 +440,7 @@ EOF;
         $this->execute('UNLINK', $url);
     }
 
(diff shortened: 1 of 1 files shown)


Naktibalda:
I think that it would be more correct to remove a default value of $method.

sasha-ch:
@Naktibalda : no objections. 
Should I update PR?

Naktibalda:
Yes, please.

sasha-ch:
Done

sasha-ch:
Commit message updated.

Issue #2511, opened 2015-11-02
Webdriver does multiple implicit waits per one click() action


sustmi:
When you do eg.:

``` php
$I->click('Send');
```

`Webdriver` calls `findClickable()` method which progressively tries to match `Send` as:
- CSS or xpath
- narrow search in HTML content
- wide search in HTML content

If you have `config.WebDriver.wait` setting set to eg. 10 seconds, it waits 10 seconds several times. First when it tries to match `Send` as CSS or xpath, then when it does narrow search and so on.

That means even if the `Send` text is present on the page immediately, Codeception waits 20 seconds before it finds the text using wide search match.

This unacceptably slows down the tests.

javigomez:
This not solves your comment gives an alternative:

DavertMik:
Config `wait` behavior changed. See SmartWait http://codeception.com/docs/03-AcceptanceTests#SmartWait

PR #2865, opened 2016-03-02
docs: Clarify slowness of fuzzy selectors when using Selenium with implicit wait

sustmi:
I guess we close https://github.com/Codeception/Codeception/issues/2511 afterwards.


sustmi:
Should I update generated docs? I tried to run `robo build:docs` but it generated much more diffs than only in WebDriver. Do you update the generated docs in bulks?

DavertMik:
> Do you update the generated docs in bulks?

Yes, we do.

Thanks, we will update docs on release.

PR #2866, opened 2016-03-02
Laravel changes for Codeception 2.2

janhenkgerritsen:
- Removed the Laravel 4 module, because Laravel 4 depends on versions of Symfony components that will not be supported in Codeception 2.2.

- Refactored database methods for the Laravel5 and Lumen modules:
  - Removed the `createModel` method from the Laravel5 module, it was just an alias of the `haveModel` method.
  - Removed the `makeModel` method from the Laravel5 module, as it is not really needed for functional tests.
  - Renamed the `haveModel` method of the Laravel5 module to `have` to make it consistent with the API of the `DataFactory` module. 
  - Created a `haveMultiple` method to create more than one model per call, also for consistency with the `DataFactory` module.
  - The `haveRecord`, `seeRecord`, `dontSeeRecord` and `grabRecord` methods of the Laravel5 and Lumen modules now also accept class names of Eloquent models as the first parameter, instead of only database table names.


torkiljohnsen:
Ok. So this just broke my test suite, bigtime :) Some migration instructions would have been awesome. This should also be listed as breaking changes in the Codeception 2.2 release log. Right now I only see mentioned that the Laravel4 module was removed.

janhenkgerritsen:
@torkiljohnsen You are right, these changes should have been mentioned in the changelog. I have added the changes to the changelog with #3226.

Thanks for reporting the issue!

torkiljohnsen:
Should note that `grabRecord` used to return an object but now it returns an array if you use it with `table_name` string. Previously I did this:

```php
$item = $i->grabRecord('table_name', ['foo' => 'bar']);
$item->whatever;
```

This doesn't work anymore. Had to change `table_name` to `ModelName::class` in all `grabRecord` calls to migrate to 2.2.

janhenkgerritsen:
Ah you are right, that is due to the changes in commit https://github.com/Codeception/Codeception/commit/1cc05fce652d08de3ced51c6949d88c66fc20f1e.

In that commit I added a cast to array before returning the result in the internal `findRecord` method. To be honest I can't recall why I added this cast, it probably is a mistake.

torkiljohnsen:
I don't mind change, don't get me wrong, but it wasn't documented as a
breaking change so it threw us off guard here for a minute :)

cswgr:
@janhenkgerritsen 
>Removed the makeModel method from the Laravel5 module, as it is not really needed for functional tests.  

Apparently you forgot to remove the hint to the `makeModel` method in the [doc](https://codeception.com/docs/09-Data#ActiveRecord) (search for "Laravel5 module also provides").
Took me a while to get that it's been removed 🙁

Naktibalda:
@cswgr Could you update https://github.com/Codeception/Codeception/blob/2.6/docs/09-Data.md and raise a pull request?

cswgr:
Sure, have done it [here](https://github.com/Codeception/Codeception/pull/5468).

PR #2868, opened 2016-03-02
Set minimum version of Symfony components to 2.5.

janhenkgerritsen:


Naktibalda:
https://travis-ci.org/Codeception/Codeception/jobs/113108503
@janhenkgerritsen: Please review this build to see the error.

`Fatal error: Call to undefined method Symfony\Component\DomCrawler\Crawler::nodeName() in /home/travis/build/Codeception/Codeception/src/Codeception/Lib/InnerBrowser.php on line 497`

Also please test all console commands, like generate:cept, manually. Many commands don't have any unit tests.

janhenkgerritsen:
@Naktibalda I tested all the commands manually, against `composer install` and with lowest dependencies with `composer update --prefer-lowest`. I encountered no errors.

I also fixed the build error, you can see the fix in #2870.

The `Symfony\Component\DomCrawler\Crawler::nodeName()` method was introduced in Symfony 2.6 with this commit: https://github.com/symfony/dom-crawler/commit/82b5ca611e483de4dfc010b21dc63d5e7c8645c1. It is a simple getter that first checks for the count. 

However, this count will never be 0 in our case because if that is the case the call to `proceedGetValueFromField` in InnerBrowser's `proceedSeeInField` method would result in a call to `fail()`.

That is why I could simply replace the call to `nodeName()` with `getNode(0)->nodeName`.

With this fix we can keep 2.5 as the lowest version of the Symfony components in our composer.json.

PR #2881, opened 2016-03-08
Better layout of logic using switch

phpcodemonkey:
Previous version was confusing to follow, made more readable with outer switch statement.


DavertMik:
:+1: thanks
looks good to me. And more readable )

PR #2882, opened 2016-03-08
MoveCursorTo

DavertMik:
Based on @javigomez work at https://github.com/Codeception/Codeception/pull/2844

Squashing commits required.


javigomez:
Awesome! thanks very much ^_^

javigomez:
I have tested and still not working to me:

screen shot 2016-03-09 at 10 28 50

The moveToElement don't seem to do anything over Firefox

DavertMik:
I think it depends on window size, if it is too small element is not visible, or sort of. Locally my tests passed

DavertMik:
Looks like it depends on Selenium's verison. It passed for 2.48 but failed in 2.52.
And the strange thing that `click('Submit')` didn't actually perform click on a link, so it might be a bug in old selenium. I will try to experiment to make it work

DavertMik:
looks like `moveToElement` method is actually part of drag-and-drop sequence. So it's not built for your case...

DavertMik:
@javigomez here you go, it should work now! :)

src/Codeception/Module/WebDriver.php
```diff
@@ -2444,4 +2444,26 @@ protected function isPhantom()
...
+     * @param $selector
+     * @param int $offsetX
+     * @param int $offsetY
+     */
+    public function scrollTo($selector, $offsetX = null, $offsetY = null)
+    {
+        $el = $this->matchFirstOrFail($this->webDriver, $selector);
+        $x = $el->getLocation()->getX() + $offsetX;
+        $y = $el->getLocation()->getY() + $offsetY;
+        $this->webDriver->executeScript("window.scrollTo($x, $y)");
```
javigomez: brilliant!

javigomez:
works very nice:

screen shot 2016-03-11 at 13 02 41

PR #2883, opened 2016-03-09
REST::getRunningClient() made 'protected'

sasha-ch:
Visibility of REST::getRunningClient() changed to 'protected' to allow its using in module extends.
For example:
```php
    public function grabHttpHeaders()
    {
        return $this->getRunningClient()->getInternalResponse()->getHeaders();
    }
```
instead of:
`return $this->client->getInternalResponse()->getHeaders();`

### DIFF
--- a/src/Codeception/Module/REST.php
+++ b/src/Codeception/Module/REST.php
@@ -138,7 +138,7 @@ EOF;
         }
     }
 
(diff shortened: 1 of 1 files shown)


DavertMik:
ok, I see no problems with that

PR #2887, opened 2016-03-10
REST::seeResponseMatchesJsonType() doc block fix

sasha-ch:
REST::seeResponseMatchesJsonType() doc block fix accoarding to new method name and args.
Also need it in http://codeception.com/docs/modules/REST#seeResponseMatchesJsonType


DavertMik:
Thanks

PR #2889, opened 2016-03-11
Unified api for setting/delitting headers

DavertMik:
Requests from REST module are sent through `_reqest` method of InnerBrowser.
Headers managing also moved from REST to InnerBrowser.

Now InnerBrowser and REST can set and unset headers equally:

* haveHttpHeader - to set (was not available in framework modules)
* deleteHeader - to unset it (was not available in REST + framework modules)

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -47,6 +47,8 @@ class InnerBrowser extends Module implements Web, PageSourceSaver, ElementLocato
      */
     protected $forms = [];
 
+    public $headers = [];
+
     protected $defaultCookieParameters = ['expires' => null, 'path' => '/', 'domain' => '', 'secure' => false];
 
     protected $internalDomains = null;
@@ -64,6 +66,7 @@ class InnerBrowser extends Module implements Web, PageSourceSaver, ElementLocato
         $this->client = null;
         $this->crawler = null;
         $this->forms = [];
+        $this->headers = [];
     }
 
     public function _conflicts()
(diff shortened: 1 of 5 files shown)


tests/unit/Codeception/Module/PhpBrowserRestTest.php
```diff
@@ -51,7 +51,6 @@ public function testSendAbsoluteUrlGet()
     {
         $this->module->sendGET('http://127.0.0.1:8010/rest/user/');
         $this->module->seeResponseCodeIs(200);
-        $this->assertEquals('http://127.0.0.1:8010/rest/user/', $this->module->client->getHistory()->current()->getUri());
```
DavertMik: REST module is not updating a history list. Not sure if this a feature or bug :)

Naktibalda: There is no ->back() function in REST, so it is fine.

src/Codeception/Lib/InnerBrowser.php
```diff
@@ -136,10 +139,21 @@ public function _getResponseContent()
 
     protected function clientRequest($method, $uri, array $parameters = array(), array $files = array(), array $server = array(), $content = null, $changeHistory = true)
     {
+        $this->debugSection("Request Headers", $this->headers);
```
Naktibalda: It would probably be good to move this line to debugResponse() method.

DavertMik: why request headers should be in debug*Response*?

DavertMik:
@Naktibalda are you ok with merging it?

DavertMik:
and so I'm merging it...

PR #2907, opened 2016-03-16
Allow fetching of Current Environment & Allow parameters per environment

marcvanduivenvoorde:
## Get the current environment

- Allow for fetching the `current_environment` parameter from the moduleContainer from any module

`$environment = $this->moduleContainer->getCurrentEnvironment()`

## Get a parameter

- This would allow you to enter a parameter list in a codeception configuration file.

```
codeception.yml
tests/some.suite.yml
_envs/some-env.yml
```

```
parameters:
   foo: bar
```

- Every deeper level of config will overwrite a lower level parameter. New parameters will be merged into the parameter array.

- You would be able to fetch a parameter in any module you create by

`$foo = $this->moduleContainer->getParameter('foo')`

- The new `Parameter` module would allow you to use `getParameter` from your `..Tester..` classes which are injected into the `Cest` files by using `$i->getParameter('foo')`


src/Codeception/Lib/ModuleContainer.php
```diff
@@ -201,6 +201,42 @@ public function validateConflicts()
...
+    }
+
+    /**
+     * @param string $key
+     *
+     * @return string|array
+     */
+    public function getEnvironmentParameter($key)
+    {
+        if ( ! array_key_exists($this->config, 'parameters')) {
```
marcvanduivenvoorde: fixed

src/Codeception/Lib/ModuleContainer.php
```diff
@@ -201,6 +201,42 @@ public function validateConflicts()
...
+            throw new \RuntimeException('Parameters are not set.');
+        }
+
+        if ( ! array_key_exists($this->config['parameters'], $key)) {
```
marcvanduivenvoorde: fixed

src/Codeception/Lib/ModuleContainer.php
```diff
@@ -201,6 +201,42 @@ public function validateConflicts()
         }
     }
 
+    /**
+     * @return string
+     */
+    public function getCurrentEnvironment()
+    {
+        if (empty($this->config['current_environment'])) {
```
DemonTPx: What is the point of this check?

marcvanduivenvoorde: fair enough, it now throws an exception when the `current_environment` key is not available.

DavertMik:
Wow, this is simple yet powerful thing! 
Somehow it correlates to my recent updates to Codeception 2.2 http://codeception.com/03-05-2016/codeception-2.2.-upcoming-features.html#params but from environment's perspective. Thanks!

DavertMik:
However, I need to think is there a way to unify environment params and global prams...

marcvanduivenvoorde:
@DavertMik I just did a small xdebug session, whenever a parameters: list is inserted into a `some.suite.yml` file the list gets merged into the `config['parameters']` array. The list in the `codeception.yml` gets excluded for now.

Maybe the method `getEnvironmentParameter` should be renamed to `getParameter` which will fetch a parameter from the "global" scope. The the `getEnvironmentParameter` method should fetch the key from the `config['env'][{current_environment}]['parameters']` array.

marcvanduivenvoorde:
@DavertMik Could you have another look at this?

DavertMik:
yes, finally got back here.

src/Codeception/Module/Parameter.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace Codeception\Module;
+
+use Codeception\Module as CodeceptionModule;
+
+/**
+ * Class Parameter
+ */
+class Parameter extends CodeceptionModule
```
DavertMik: please don't... this doesn't look like a real module.
I think parameters should be taken in using Scenario->current

```
$scenario->current('paramName')
```

DavertMik:
Please resend this PR into master branch so we will have my and yours params implementations in one track. 
Sorry right now I don't have a vision how to make environment params to be implemented in the right way (too sleepy). Please look into my commit, maybe you will have some more ideas https://github.com/Codeception/Codeception/commit/33bc5958304971ea1ce2085576d3909548e39df3

But I will try to take a look more into it tomorrow

What I know for sure, that `Parameter` module is not a good idea.

marcvanduivenvoorde:
Fair enough. I have removed the `Parameter` module. 

I've added a `getParameter` method to `Scenario` and `Cest`. This will allow for a "global" access and bubbling down of a single named method to get a parameter. All parameters from the config files loaded will be available.

src/Codeception/TestCase/Cest.php
```diff
@@ -212,4 +213,16 @@ protected function executeHook($I, $hook)
         }
     }
 
+    /**
+     * @param string $key
+     * @return array|string
+     */
+    public function getParameter($key)
```
DavertMik: actually this is uselss. You are not accessing Cest class from inside your tests in any way

tests/unit/Codeception/Module/ParameterTest.php
```diff
@@ -0,0 +1,41 @@
+<?php
+namespace Codeception\tests\unit\Codeception\Module;
```
DavertMik: yeah, remove this file please

src/Codeception/Scenario.php
```diff
@@ -187,4 +187,12 @@ public function running()
         return true;
     }
 
+    /**
+     * @param string $key
+     * @return mixed
+     */
+    public function getParameter($key)
+    {
+        return $this->test->getParameter($key);
```
DavertMik: Nah... This won't work...

DavertMik:
Sorry, I don't see a way this can be merged. 
I know the overall architecture is complex and it is hard to figure out a way to make this feature work.
It just won't work this way, and especially it can't be merged into master.

However, Codeception 2.2 (which is in master now) is much more cleaner and easy to understand.
I will try to implement this feature by myself maybe even without touching ModuleContainer (this class is about modules, and not environments or additional params)

Nevertheless, thanks for your work.

PR #2908, opened 2016-03-16
Update WebDriver Doc

eXorus:
The link http://code.google.com/p/selenium/wiki/DesiredCapabilities doesn't exist anymore. I switch to the new one always from SeleniumHQ


DavertMik:
Awesome! Thanks

PR #2911, opened 2016-03-18
Update Page Object documentation outdated link

javigomez:
The link at (http://codeception.com/docs/06-ReusingTestCode#PageObjects):  

capture 2016-03-05 at 12 42 43

was not anymore valid. I'm updating it with the one that I guess is the right new location.


DavertMik:
Thanks!

PR #2912, opened 2016-03-18
Correction for non ascii encoding with mb_ function

omgnull:
While using utf-8 text as messages in tests output is not correct in some cases

```php
// example
public function testBlaBla($I)
{
    $I->wantTo('бла бла');
}

public function testBlaBlaBla($I)
{
    $I->wantTo('бла бла бла');
}
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,7 +80,6 @@ install:
   - '[[ "$TRAVIS_PHP_VERSION" == "5.4" ]] || composer install -d frameworks-zend-expressive --no-dev  --prefer-dist'
 
 before_script:
-  - '[[ "$TRAVIS_PHP_VERSION" == "hhvm" ]] || echo "mbstring.internal_encoding = UTF-8" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini'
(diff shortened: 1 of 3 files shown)


Naktibalda:
`if (function_exists('mb_substr')) {` checks are not necessary, ext-mbstring is required by composer.json of Codeception.

omgnull:
@Naktibalda should I remove this check too?
https://github.com/Codeception/Codeception/blob/2.1/src/Codeception/Lib/Console/Message.php#L111

Naktibalda:
Yes, it is already like that in master branch.

.travis.yml
```diff
@@ -80,6 +80,7 @@ install:
   - '[[ "$TRAVIS_PHP_VERSION" == "5.4" ]] || composer install -d frameworks-zend-expressive --no-dev  --prefer-dist'
 
 before_script:
+  - '[[ "$TRAVIS_PHP_VERSION" == "hhvm" ]] || echo "mbstring.internal_encoding = UTF-8" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini'
```
Naktibalda: I think that this line should be removed.

Naktibalda:
Related pull request hardcoded 'utf-8' charset in all mb_ function calls: https://github.com/Codeception/Codeception/pull/2831/files and https://github.com/Codeception/Codeception/pull/2852/files

Please do the same in your function calls.
If you think that it is a wrong approach, please let us know why.

PR #2915, opened 2016-03-22
Updated module Queue to work with pheanstalk library ~3.0

torus84:
Updated module Queue to work with pheanstalk library ~3.0; Codeception module Queue (adapter for Beanstalk)


Naktibalda:
Thank you

PR #2918, opened 2016-03-22
Test lowest dependencies for frameworks

Naktibalda:
In dependencies=lowest environment, install the lowest supported versions of dependencies for all framework sites except Symfony.

Install matching version of symfony/symfony for frameworks-symfony when SYMFONY=X variable is set (only 2.7 and 2.8 are supported).

Run dependencies=lowest on PHP 5.5 to test all frameworks (except Symfony).


DavertMik:
:+1:

PR #2920, opened 2016-03-23
Display correct path to output directory in the message

### LINKED ISSUE #2902: See complete response in '_output' directory
_output is hardcoded in https://github.com/Codeception/Codeception/blob/2.1/src/Codeception/PHPUnit/Constraint/Page.php#L25

```php
$debugMessage = new Message("[Content too long to display. See complete response in '_output' directory]");
```

It matches the default location but is very misleading when different directory is configured in codeception.yml

The fix could be as simple as using `realpath(codecept_output_dir())` instead of hardcoded directory name.

Naktibalda:
Fixes  #2902


Naktibalda:
FacebookTest failed as expected and printed "[Content too long to display. See complete response in '/home/travis/build/Codeception/Codeception/tests/log/' directory]".

Issue #2930, opened 2016-03-27
beta 2.2.0 html report


rebajz:
two bugs in html
1)
html output have all test everytime as success (in green background).
Command line say me:

```
FAILURES!
Tests: 5, Assertions: 68, Failures: 2.
```

html say me:

```
Codeception Results OK (49.5s)
Summary
Successful scenarios: 	5
Failed scenarios: 	0
Skipped scenarios: 	0
Incomplete scenarios: 	0
```

2) problem with / in url.
for example i see in html this:

```
I am on url "https:\/\/erpdev.ulozenka.cz\/"
I see in current url "\/user\/edit"
I see current url equals "\/"
```

Naktibalda:
2) is caused by Limit step output length to 1 line #2784 
I didn't realise that [stripcslashes](https://github.com/Codeception/Codeception/pull/2784/files#diff-bf39b9a6ccca147479241c668f020739L102#) is so important.

DavertMik:
> html output have all test everytime as success (in green background).

this may depend on test formats. Could you provide numbers of how many Cept/Cest/Test/PHPUnit_Framework_TestCase classes in your suite?

rebajz:
Right now I have two 00-otherCest.php and 01-crmCest.php (but it will be more in future).
If i run only this test (vendor\bin\codecept run tests/acceptance/00-otherCest.php --html
 ), i have same problem...

00-otherCest.php
class otherCest
{
    public function _before(AcceptanceTester $I)
    {
this is empty
    }
    public function _after(AcceptanceTester $I)
    {
this is empty
    }
    public function neprihlasen(AcceptanceTester $I)
    {
here i have passed test
    }
    public function prihlasen(AcceptanceTester $I)
    {
here i have passed test
    }
    public function logo(AcceptanceTester $I)
    {
here a have failed test
    }
}

from command line:
FAILURES!
Tests: 3, Assertions: 11, Failures: 1.
- HTML report generated in file://D:\cod-beta\tests/_output\report.html

from html:
Summary
Successful scenarios: 	3
Failed scenarios: 	0
Skipped scenarios: 	0
Incomplete scenarios: 	0

rebajz:
when i doing
vendor\bin\codecept g:scenarios acceptance
vendor\bin\codecept g:scenarios acceptance --format html

output files have same problem with /   as is written in first post.

DavertMik:
Thanks, looking into it

bvigneshk:
[...]

PR #2940, opened 2016-03-29
Don't json_encode strings in step output, escape \n,\r and \t

Naktibalda:


Naktibalda:
Fixes issue 1 in #2930

PR #2947, opened 2016-03-31
htmlentities on html report generation

liftingcortical:
When creating HTML report, if failure snippet contains html entities, the report HTML gets broken.
Added htmlentities to encode them when passing failures to the template.


DavertMik:
Thanks, I fixed this for Codeception 2.2 just in the same manner.

Issue #2934, opened 2016-03-28
setUpBeforeClass not working with @dataProvider


fabianbadoi:
The setupBeforeClass method is not called on tests that use the @dataProvider annotation.
Here's a minimal test case.

```
<?php

use Codeception\TestCase\Test;

class DataProviderTest extends Test
{
	protected static $setUpCalled = 0;

	public static function setUpBeforeClass()
	{
		self::$setUpCalled++;
	}
  
	public static function _beforeSuite()
	{
		self::$setUpCalled++;
	}

	/**
	 * @dataProvider getItems
	 */
	public function testSetup($i)
	{
		self::assertEquals(1, self::$setUpCalled);
	}

	public function getItems()
	{
		return [[1], [2]];
	}
}

```
```Output:
1) Test setup | #1 DataProviderTest.php
Failed asserting that 0 matches expected 1.
#1  /Users/fabi/Temp/codeption-bug copy/Tests/tests/unit/DataProviderTest.php:20
#2  DataProviderTest->testSetup
```

The problem can be solved with by replacing `$testClass = $test` with 
```
$testClass = $test instanceof \PHPUnit_Framework_TestSuite_DataProvider ?
                strstr($test->getName(), '::', true) :
                get_class($test);
```
In src/Codeception/Subscriber/BeforeAfterTest.php:26, but I suspect there are better ways.

PR #2949, opened 2016-03-31
Fix setupbeforeclass bug issue 2934

fabianbadoi:
Fix Issue #2934 (setUpBeforeClass not working with @dataProvider).

The method setUpBeforeClass, methods annotated with `@beforeClass`, and the _beforeSuite method did not get called if the test case only contained tests annotated with `@dataProvider`. 

This happened because such test cases were included in PHPUnit_Framework_TestSuite_DataProvider instances, which do not have hooks for setup or teardown 'before class'.

However, if a single test of the test case was not annotated, than the hooks would run, since at least one test did not get converted to PHPUnit_Framework_TestSuite_DataProvider.

I've provided a simple, though inelegant fix, and a test to guard against such bugs.

Kind regards,
Fabian

### DIFF
--- a/src/Codeception/Subscriber/BeforeAfterTest.php
+++ b/src/Codeception/Subscriber/BeforeAfterTest.php
@@ -23,6 +23,11 @@ class BeforeAfterTest implements EventSubscriberInterface
     {
         foreach ($e->getSuite()->tests() as $test) {
             /** @var $test \PHPUnit_Framework_Test  * */
+            if ($test instanceof \PHPUnit_Framework_TestSuite_DataProvider) {
+                $potentialTestClass = strstr($test->getName(), '::', true);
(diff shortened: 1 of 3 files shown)


fabianbadoi:
A more elegant solution would be to extend PHPUnit_Framework_TestSuite_DataProvide and add the required hoots to it. I'm prepared to implement this as well, if you wish it so.

DavertMik:
Thanks, finally got a time to look into it. 

> A more elegant solution would be to extend PHPUnit_Framework_TestSuite_DataProvide and add the required hoots to it. I'm prepared to implement this as well, if you wish it so.

I think not. Overriding too much PHPUnit classes have its cons, so I'd like to keep your current solution. Your case is pretty rare so I hope we can live with it.

fabianbadoi:
Cool, I'm glad we could fix this.

Thanks!

PR #2957, opened 2016-04-05
Fix non-init config value

Aib0:
When $config['settings']['lint'] is not initialized codeception throws PHP Notice, fixed it.


Naktibalda:
Thanks

PR #2959, opened 2016-04-05
[SOAP] Fix grabTextContentFrom() after the second request

eXorus:
**Test :**
```php
$I->sendRequest('ShowUser', '<user><name>notdavert</name></user>');
$I->grabTextContentFrom("//*[local-name() = 'login']");
$I->sendRequest('ShowUser', '<user><name>eXorus</name></user>');
$I->grabTextContentFrom("//*[local-name() = 'login']");
```
**Actual Result :**
notdavert
notdavert

**Result Expected with my fix**
notdavert
eXorus

The fix consist to always rewrite xmlStructure from xmlResponse and not only when it's null because if you call a second times the xmlStructure will be the xmlStructure of the fist call.

Bug was introduced in 2.1.7

### DIFF
--- a/src/Codeception/Module/SOAP.php
+++ b/src/Codeception/Module/SOAP.php
@@ -37,8 +37,8 @@ use Codeception\Util\XmlStructure;
  *
  * ## Public Properties
  *
- * * xmlRequest - last soap request (DOMDocument)
- * * xmlResponse - last soap response (DOMDocument)
+ * * xmlRequest - last SOAP request (DOMDocument)
+ * * xmlResponse - last SOAP response (DOMDocument)
  *
  */
 class SOAP extends CodeceptionModule implements DependsOnModule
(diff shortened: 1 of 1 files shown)


Naktibalda:
a) it's sendSoapRequest, not sendRequest (please update the docblock in your pull request);
b) no changes were made in 2.1.7,  what bug are you talking about?
c) I think that it would be better to unset $this->xmlStructure in sendSoapRequest method than to regenerate it on every assertion.

eXorus:
a) It's done
b) it was on 21 april 2015 not 2.1.7 when you created getXmlStructure()
c) it's done

And I added some more update to the docblock

eXorus:
I don't understand why the build failed on FacebookTest::testSeePostOnFacebookWithMessage ... I only change SOAP Module

Could you check this for me ?

Naktibalda:
Thanks for making these changes.
FacebookTest fails occasionally, don't worry about it

PR #2961, opened 2016-04-05
Remove container as public property on Symfony.

### LINKED ISSUE #2954: [BC] Container should not be a variable on Symfony2 module
Container variable should be refactored into a method call to $this->kernel->getContainer() to avoid trouble dealing with non-booted kernels and with tests with multiple requests, as each request boots its own kernel and creates a new container instance.

raistlin:
Fixes #2954

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 * [Phalcon] Merged `Phalcon1` and `Phalcon2` modules into one `Phalcon` due the fact that Phalcon Framework v1.3.x no longer supported at all
 * *Breaking* [Dbh] module removed
 * *Breaking* [Laravel4] module removed
+* *Breaking* [Symfony] Removed public property `container` you can use new `_getContainer` method
 
 
 #### 2.1.7
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -63,7 +63,6 @@ use Symfony\Component\Finder\Finder;
  *
  * * kernel - HttpKernel instance
  * * client - current Crawler instance
- * * container - dependency injection container instance
  *
  * ## Parts
  * 
@@ -91,11 +90,6 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
      */
     public $kernel;
 
-    /**
-     * @var \Symfony\Component\DependencyInjection\ContainerInterface
-     */
-    public $container;
-
     public $config = [
         'app_path' => 'app',
         'var_path' => 'app',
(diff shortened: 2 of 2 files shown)


src/Codeception/Module/Symfony.php
```diff
@@ -350,6 +346,14 @@ public function grabServiceFromContainer($service)
     }
 
     /**
+     * @return \Symfony\Component\DependencyInjection\ContainerInterface
+     */
+    public function getContainer()
```
DavertMik: should be `_getContainer` for not to appear in functional tests

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 * [Phalcon] Merged `Phalcon1` and `Phalcon2` modules into one `Phalcon` due the fact that Phalcon Framework v1.3.x no longer supported at all
 * *Breaking* [Dbh] module removed
 * *Breaking* [Laravel4] module removed
+* *Breaking* [Symfony] Removed public property `container` you can use new `getContainer` method
```
DavertMik: use new `_getContainer` method

src/Codeception/Module/Symfony.php
```diff
@@ -350,6 +346,14 @@ public function grabServiceFromContainer($service)
     }
 
     /**
+     * @return \Symfony\Component\DependencyInjection\ContainerInterface
+     */
+    public function _getContainer()
+    {
+        return $this->kernel->getcontainer();
```
DavertMik: getContainer

raistlin:
Agh! CHANGELOG fixed

src/Codeception/Lib/Connector/Symfony.php
```diff
@@ -10,11 +10,26 @@ class Symfony extends \Symfony\Component\HttpKernel\Client
...
      */
     public $persistentServices = [];
 
     /**
+     * Symfony constructor.
+     * @param \Codeception\Module\Symfony $module
+     */
+    public function __construct($module)
+    {
+        $this->module = $module;
```
DavertMik: I'd rather not have cyclic references here. Is that important for connector to be aware of module?

raistlin: If connector must create new kernels, connector must be able to set new kernel on symfony module.
It's not mandatory for this branch, but it's a step forward to allow this behaviour.

raistlin:
- Travis, Facebook failed.
- AppVeyor, I don't have any experience but seems that download on win64openssl failed ...

raistlin:
Potentially invalidated with #2968

Issue #2938, opened 2016-03-28
Symfony2 memory leaks and performance issues


vyarmolenko:
I wrote some functional tests for Symfony2 app with multiple requests in a test and encountered an issue with test failing due to lack of memory (3Gb's was not enough to complete the test). I rewrote my tests and still memory consumption was too big.
The cause of memory leak was the cloning of kernel on each request. it was introduced in https://github.com/Codeception/Codeception/pull/2042 as a solution to problem with service container introduced in https://github.com/Codeception/Codeception/pull/2491.
This issue may relate to the following issues:
https://github.com/Codeception/Codeception/issues/2203 - memory consumption
https://github.com/Codeception/Codeception/issues/2025 - first call to doctrine after client reboot
https://github.com/Codeception/Codeception/issues/2886 - similar memory consumption
https://github.com/Codeception/Codeception/issues/2853
https://github.com/Codeception/Codeception/issues/1804

Possible issues include:
- big memory consumption of tests with multiple requests
- long execution time/slow speed of tests with multiple requests

raistlin:
You have made me to spot some issues between Symfony2 module and Symfony2 connector.
First, Symfony2 module declares a public container variable. This variable is never used on the module, but user tests have access to this variable. Problem is that this container instance is never the container linked on the request running kernel. Doctrine works because it is a persisted service and shared among all kernels, but it's a very high trouble.

This container variable is totally invalidated just after first test are run, because Symfony2 connector boots a new kernel on every request, and booting a new kernel creates a new container instance. Just before booting a new kernel the connector save persisted services to inject them after kernel has been booted. As a side note, just because kernel is cloned here, kernel on Symfony2 module and Symfony2 connector will be out of sync, other big source of problems.

At conclusion: 

- Symfony2 module do not need the container anymore, if needed we should provide a getContainer method to return current container.
- Kernel on Symfony2 module and Symfony2 connector will be out of sync, connector should never clone or create a new kernel!

I will create issue to fix the first. The second will reopen #2042, but it's better to reopen and fix it the good way that having memleaks.

Many thanks for your research!

Issue #2853, opened 2016-02-29
Doctrine 2 and unit of work


VincentClair:
Hi,
I have a problem with unit of work of Doctrine2 / Symfony3 plugins.

Here is an example of a test that does not work:

``` php
$I->wantTo('remove "permission" action from ROLE_ADMIN');

$adminRoleId = $this->getRoleIdByKey($I, 'ROLE_ADMIN');
$permissionActionId = $this->getPermissionActionId($I, 'permission');

$I->sendDELETE('/roles/' . $adminRoleId . '/permissions/' . $permissionActionId);
$I->seeResponseCodeIs(400);
$I->seeResponseEquals('"permission-action-mandatory"');

// We could not verify that permission is not removed, because Symfony plugin of codeception does not create
// another real request. Requests are simulated with one Symfony instance, so the service of doctrine keep the
// same unit of work: objects are not refreshed between requests. If we made a modification on a object in one
// request, next requests get the same modified object.
$I->sendGET('/roles/' . $adminRoleId . '/permissions');
$I->seeResponseCodeIs(200);
$I->seeResponseContainsJson(['id' => $permissionActionId]);
```

I want to check that an entity could not be deleted. After I made the action and validated that DELETE is refused, I want to check that entity ever existed. But when I get the list, I don't found the NOT DELETED entity.

I suppose that this is is due to two calls to get entities with a relation between entities managed. It is the same unit of work of Doctrine, so a change of an entity is kept during all requests in one test. Is there a way to reset unit of work ?

DavertMik:
@raistlin ping

VincentClair:
Any news please ?

raistlin:
I don't see the problem, and I don't think that not creating a full request is the source of this issue.
Could you verify that if you get the list before this test is run, do you see the NOT DELETED entity?

VincentClair:
I will try to explain better what i feel where the problem comes from. If i well understand the actual process, here are the steps of a Codeception test:
- Initialize modules Doctrine...
- Begin a transaction (to no commit changes from test)
- Run the test. In my example:
    - method `getPermissionActionId` send a request $I->sendGET('/roles/' . $adminRoleId . '/permissions'); to get all permissions for a given role.
[...]

PR #2968, opened 2016-04-07
Update Symfony2 framework module and connector, fix client generation…

vyarmolenko:
…, update Doctrine EM retrieval. Improve memory usage. Unify service retrieval.

This PR provides solution for https://github.com/Codeception/Codeception/issues/2938, https://github.com/Codeception/Codeception/issues/2954.

This PR proposes the following model of Symfony2 framework module and client:
- module's `$container` is a read-only property. It returns active client's service container during test execution or self service container between tests when accessed externally.
- the module has `$persistentServices` array which contains services to persist between requests (kernel reboots) in a test.
- the module has `$permanentServices` array which contains services to persist between tests.
- persistence status of any service can be changed at any moment during test execution (public methods `persistService(...)` / `unpersistService(...)`).
- the app kernel is booted during module initialization in order to make framework features available before client creation.
- a client is created before each test. It is created based on current kernel, currently retrieved persistent services and a configuration option marking client's kernel rebootable between requests or not. Client's kernel is rebooted in its constructor to ensure clean environment for test. 
- client's persistent services are being updated before kernel shutdown and injected into newly initialized service container after kernel boot during test execution.
- client's kernel can be rebooted using `$I->rebootClientKernel()` during test execution.
- permanent services are updated after test execution.
- cached router is permanent by default if used.
- doctrine's EM service is permanent by default.

Test results of Codeception's Symfony2 framework test app (https://github.com/Codeception/symfony-demo):
symfony2_test_app

Test results of modified version of https://github.com/AlexStansfield/codeception-test - an application used to reproduce https://github.com/Codeception/Codeception/issues/2025:
issue2025_test_app
Modifications to this test app include:
- creation of a copy of EntityCest.php - Entinty1Cest.php to check several cests in a suite, 
- making some methods of EntityCest.php to perform several `sendPOST(...)` requests to check correct work and measure performance of a solution.

Performing tests with modified version of Codeception on my Symfony2 app showed memory consumption was around 190 Mb's (previously tests failed due to insufficient memory with a 3GB's configuration) and execution time became around 19 minutes (and even lower - 11 minutes - with `config['cached_router'] = true`)

This PR also includes:
- change retrieval of Doctrine's Entity Manager to make it compatible with new client generation logic and the model of persistent services and to avoid/solve https://github.com/Codeception/Codeception/issues/2025, https://github.com/Codeception/Codeception/issues/2853  
- unification of service retrieval in existing public methods of Symfony2 module.

### DIFF
--- a/src/Codeception/Module/Symfony2.php
+++ b/src/Codeception/Module/Symfony2.php
@@ -9,6 +9,7 @@ use Codeception\Lib\Connector\Symfony2 as Symfony2Connector;
 use Codeception\Lib\Interfaces\DoctrineProvider;
 use Codeception\Lib\Interfaces\PartedModule;
 use Symfony\Component\Finder\Finder;
+use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
  * This module uses Symfony2 Crawler and HttpKernel to emulate requests and test response.
@@ -30,8 +31,9 @@ use Symfony\Component\Finder\Finder;
  * * environment: 'local' - environment used for load kernel
  * * debug: true - turn on/off debug mode
  * * em_service: 'doctrine.orm.entity_manager' - use the stated EntityManager to pair with Doctrine Module.
- * * cache_router: 'false' - enable router caching between tests in order to [increase performance](http://lakion.com/blog/how-did-we-speed-up-sylius-behat-suite-with-blackfire) 
- * 
+ * * cache_router: 'false' - enable router caching between tests in order to [increase performance](http://lakion.com/blog/how-did-we-speed-up-sylius-behat-suite-with-blackfire)
+ * * rebootable_client: 'true' - reboot client's kernel before each request
+ *
  * ### Example (`functional.suite.yml`) - Symfony 2.x Directory Structure
  *
  * ```
@@ -48,7 +50,8 @@ use Symfony\Component\Finder\Finder;
  * * environment: 'local' - environment used for load kernel
  * * em_service: 'doctrine.orm.entity_manager' - use the stated EntityManager to pair with Doctrine Module.
  * * debug: true - turn on/off debug mode
- * * cache_router: 'false' - enable router caching between tests in order to [increase performance](http://lakion.com/blog/how-did-we-speed-up-sylius-behat-suite-with-blackfire) 
+ * * cache_router: 'false' - enable router caching between tests in order to [increase performance](http://lakion.com/blog/how-did-we-speed-up-sylius-behat-suite-with-blackfire)
+ * * rebootable_client: 'true' - reboot client's kernel before each request
  *
  * ### Example (`functional.suite.yml`) - Symfony 3 Directory Structure
  *
@@ -64,14 +67,13 @@ use Symfony\Component\Finder\Finder;
  *
  * * kernel - HttpKernel instance
  * * client - current Crawler instance
- * * container - dependency injection container instance
  *
  * ## Parts
- * 
- * * services - allows to use Symfony2 DIC only with WebDriver or PhpBrowser modules. 
- * 
+ *
+ * * services - allows to use Symfony2 DIC only with WebDriver or PhpBrowser modules.
+ *
  * Usage example:
- * 
+ *
  * ```yaml
  * class_name: AcceptanceTester
  * modules:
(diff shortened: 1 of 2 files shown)


raistlin:
You have removed my lasts commits, one that deprecated `grabServiceFromContainer` and introduces a new `grabService`.
Could you rebase please?

raistlin:
How your modifications behave with current situation:
- You make a request that grab some data from Doctrine2, update entities or even create new entities and call `persist` but you do not call to `flush`
- You make other request. Will your second request see modifications done on first one?

vyarmolenko:
There is no need to deprecate `grabServiceFromContainer` with proposed changes. The service is being retrieved from the appropriate container transparently.

We have persistent Entity Manager and Unit Of Work and in described situation all `persist`'ed entities become managed and preserved between requests. However as stated in https://github.com/doctrine/doctrine2/issues/5092 unflushed entities cannot be accessed in common way (eg something like`$entity = $em->find('AppBundle:Codeception', $name)` will return `null` for unflushed entity) but can be accessed using DQL.
So the answer is yes, the second request can see modifications done by first one but with some restrictions (and it is a Doctrine issue, not Codeception's). The behaviour is the same as within one request.

raistlin:
Rename of `grabServiceFromContainer` has been done to be consistent with other frameworks, that also have `grabService`.

vyarmolenko:
Updated, please review.

src/Codeception/Module/Symfony2.php
```diff
@@ -93,17 +99,18 @@ class Symfony2 extends Framework implements DoctrineProvider, PartedModule
     public $kernel;
 
     /**
-     * @var \Symfony\Component\DependencyInjection\ContainerInterface
+     * @var ContainerInterface
      */
-    public $container;
+    private $container;
```
raistlin: This var is not used, and is a breaking change.

src/Codeception/Module/Symfony2.php
```diff
@@ -143,41 +180,66 @@ public function _initialize()
...
         }
+        return $this->permanentServices[$this->config['em_service']];
+    }
+
+    /**
+     * Return active container.
+     *
+     * @return ContainerInterface
+     */
+    protected function getContainer()
```
raistlin: You can change visibility to `public` so users could still access container if they need.

src/Codeception/Module/Symfony2.php
```diff
@@ -143,41 +180,66 @@ public function _initialize()
             ini_set($xdebugMaxLevelKey, $maxNestingLevel);
         }
 
-        $this->bootKernel();
-        $this->container = $this->kernel->getContainer();
+        $this->kernel = new $this->kernelClass($this->config['environment'], $this->config['debug']);
+        $this->kernel->boot();
+
+        if ($this->config['cache_router'] === true) {
+            $this->persistService('router', true);
```
raistlin: Nice trick!

src/Codeception/Module/Symfony2.php
```diff
@@ -417,4 +513,14 @@ protected function getInternalDomains()
 
         return array_unique($internalDomains);
     }
+
+    /**
+     * Reboot client's kernel
+     */
```
raistlin: Could you improve a bit documentation of this method? Users should know why and when this should be called.

src/Codeception/Module/Symfony2.php
vyarmolenko: We can make `$container` a read-only property and retrieve it with magic method `__get($property)`

src/Codeception/Module/Symfony2.php
```diff
@@ -67,11 +70,11 @@
  * * container - dependency injection container instance
```
raistlin: One last change: As this property no longer exists as is, please remove it from documentation.

src/Codeception/Module/Symfony2.php
raistlin: Ok, you can do it to handle the BC on 2.1.x. We should remove this 'magic' in a future release, maybe 3.0, and have only `_getContainer`.

raistlin:
:+1: 
Status: Reviewed.

DavertMik:
Thanks, awesome work!

PR #2969, opened 2016-04-07
Improved Output | Additional Assertions

DavertMik:
### Overview

Changed the output format for console. 

white terminal:

Now each test have its class and human readable name displayed. This change is important for Cest files (now you see the exact file and a feature currently running) and the most important as this **changes output of unit tests**:

```
Codeception\Lib\ModuleContainerTest:testActionsExplicitlySetForNotInheritedModule
```
is changed to
```
ModuleContainerTest: Actions explicitly set for not inherited module
```
This change was done to make unit tests be named in more descriptive manner.

### Descriptor

Also refactored how test information from `Codeception\Test\Descriptor`. Methods of this class provide information about a test passed. 

* `getTestAsString` provides human readable test name. Used to be printed to console or HTML reports.
* `getTestFileName` returns a full name of a file with a test
* `getTestFullName` previous + if file contains several tests appends `:testName` of exact test. This name can be used to execute exact this test with `codecept run {testFullName}`
* `getTestSignature` a test name which can be used in `@depends`

Console Output now contains all those information about tests. Each failed test contains its full name so can be easily re-executed, each test in debug mode provides its signature, etc.

### Additional Changes

* In cases where PHPUnit produces diff output, Codeception enhances it with colors. 
* Added `ansi` and `no-ansi` mode to turn on/off special utf chars
* Also I accidentally added my previous commit for Asserts. More `assert*` methods added to `Asserts` module as well as into the trait itself.
* Added `expectException` method to `Asserts` module, which checks for exception enclosed in closure.


src/Codeception/Test/Gherkin.php
```diff
@@ -141,7 +140,7 @@ protected function makeContexts()
 
     public function toString()
     {
-        return $this->getFeature() . ' ' . $this->getFileName();
+        return $this->featureNode->getTitle() . ": " . $this->getFeature();
```
raistlin: Should be `': '`.

PR #2980, opened 2016-04-13
Module/Db: debugSection usage fix

sasha-ch:
In Module/Db.php:369 we pass to debugSection 3 params instead of 2 as in declaration.


Naktibalda:
Wouldn't it be better if we printed query and parameters in separate lines?

```php
$this->debugSection('Query', $query);
$this->debugSection('Parameters', $data);
```

sasha-ch:
Really I think it will be better if debugSection can print several separate parameters at a time, like:
```php
$this->debugSection('title', $param1, $param2, $param3, ...);
```
via `func_get_args()` 
with some formatting like new lines.

PR #2982, opened 2016-04-13
Merge 2.1

DavertMik:
Latest changes from 2.1


src/Codeception/Module/Asserts.php
```diff
@@ -263,35 +262,6 @@ public function assertFileNotExists($filename, $message = '')
     }
 
     /**
-     * @param $expected
```
Naktibalda: I don't like this change - you removed methods without warning and even without mentioning that in the changelog, but you left aliases that were marked as deprecated long time ago (assertGreaterThen and assertGreaterThenOrEqual ). This is a BC break if anyone is using them.

src/Codeception/Module/WebDriver.php
```diff
@@ -979,8 +983,10 @@ protected function proceedSeeInField(array $elements, $value)
                     break;
 
                 case 'textarea':
+                    // we include trimmed and real value of textarea for check
+                    $currentValues[] = $el->getText(); // trimmed value
```
Naktibalda: I think that `break;` is necessary here.

src/Codeception/Module/Asserts.php
DavertMik: yep, you are right. Scrutinzer has suggested that we had those methods written in wrong way

src/Codeception/Module/WebDriver.php
DavertMik: it's actually not.
$currentValues will contain both: '  text value   ' and 'text value' strings, so both checks will be correct:

```php
$I->seeInField('textarea', ' field value ');
$I->seeInField('textarea', 'field value');
```

PR #2992, opened 2016-04-15
Fixes bug generating helper classnames

raistlin:


src/Codeception/Lib/Generator/Helper.php
```diff
@@ -40,6 +40,6 @@ public function produce()
 
     public function getHelperName()
     {
-        return '\\' . $this->namespace . 'Helper\\' . $this->name;
+        return (empty($this->namespace) ? '' : '\\' . $this->namespace) . '\\Helper\\' . $this->name;
```
raistlin: Other possibility could have been
`return rtrim('\\' . $this->namespace, '\\') . '\\Helper\\' . $this->name;`

raistlin:
Will merge with 2.1 instead of master.

PR #2999, opened 2016-04-17
[ZF2] Create new Application for each request

### LINKED ISSUE #2964: ZF2 functional redirects
Currently redirects are not correctly processed in the ZF2 module when running functional tests, a `LogicException` is thrown.

It seems the headers are not being reset for the second request after the redirect, causing a second redirect to occur.

@Naktibalda has been informed about this via IRC and a link to sample code to reproduce the error can be found here: https://github.com/stevewest/codecept_zf2_redirect with full `-vvv` console output in the `readme.md`

Naktibalda:
A significant modification of ZF2 module in order to fix #2964

New Application is created for each request, all services from previous request are reused except Request and Response.
I am not very happy with this solution but this is the best I can do.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,7 +75,7 @@ install:
   - git clone -q -b 2.1 --recursive https://github.com/Naktibalda/codeception-zf1-tests frameworks-zf1
   - composer update -d frameworks-zf1 $composer_parameters
   # ZF2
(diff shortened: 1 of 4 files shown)


Naktibalda:
RedirectCept passes locally but fails on Travis, because it tries to use default Doctrine settings after redirect to /doctrine/.

I don't know if it is a sign of a bigger issue, so I will simply change the target of redirection to route which does not use Doctrine.

PR #3011, opened 2016-04-20
Fix detection of refresh timeout for max interval.

Frug:
Refresh headers always have a semicolon, but a space following it is optional.


Naktibalda:
Please make the same change in Guzzle.php

Issue #3026, opened 2016-04-26
Improve Redis module


marcverney:
The current [Redis module](http://codeception.com/docs/modules/Redis#.Vx8s5FaLSzc) has two limitations:

1. it doesn't provide any Actor method apart from `cleanup`. You have to get your hands dirty and code your own Helper
1. it uses the [sash/redis](https://github.com/sash/php-redis) driver, which is incomplete (doesn't provide interactions with Redis hashes, for instance) and has not been updated since 2012.

I'd like to propose a new version of the module which, at least, solves those two issues.

PR #3027, opened 2016-04-26
Replace Redis Module

marcverney:
This new version of the Redis module fixes issue #3026 by:

- adding Actor methods for the most common interactions with a Redis server
- using the ~~[Credis client class](https://github.com/colinmollenhour/credis)~~ [Predis library](https://github.com/nrk/predis) as a driver. This project is much more complete and active than the [current driver](https://github.com/sash/php-redis)

The module itself is tested (cf the `RedisTest` class)


Naktibalda:
Please add `colinmollenhour/credis` as dependency to composer.json instead of copying it to Codeception code base.

marcverney:
Should I fix the coding style issues reported by Nitpick-CI?

marcverney:
Errors in TravisCI tests made me realize the Credis library was not the best choice. I switched it to Predis. All Redis tests pass now.

src/Codeception/Module/Redis.php
```diff
@@ -3,87 +3,626 @@
...
+     *
+     * @param string $key   The key name
+     * @param mixed  $value Optional. If specified, also checks the key has this
+     * value. Booleans will be converted to 1 and 0 (even inside arrays)
+     */
+    public function dontSeeInRedis($key, $value = null)
+    {
+        $this->assertFalse(
+            $this->checkKeyExists($key, $value)
+        );
```
DavertMik: please add better failure message to this exception.
in current version user will see "failed asserting that true is false" which won't be really helpful

src/Codeception/Module/Redis.php
```diff
@@ -3,87 +3,626 @@
...
 class Redis extends CodeceptionModule
 {
+    /**
+     * {@inheritdoc}
+     *
+     * No default value is set for the database, as this module will delete
+     * every data in it. The user is required to explicitely set this parameter.
+     */
     protected $config = [
-        'cleanup' => true
```
DavertMik: I'd rather keep this setting. I think redis should be cleaned up before all tests by default.

src/Codeception/Module/Redis.php
```diff
@@ -3,87 +3,626 @@
...
+        'host'          => '127.0.0.1',
+        'port'          => 6379,
+        'cleanupBefore' => 'suite'
```
DavertMik: Do you really think that cleanup before suite is needed? Why not to allow cleaning only between tests?

src/Codeception/Module/Redis.php
```diff
@@ -3,87 +3,626 @@
...
+     * $I->seeInRedis('example:hash', ['riri' => true, 'fifi' => 'Dewey', 'loulou' => 2])
+     * ```
+     *
+     * @param string $key   The key name
+     * @param mixed  $value Optional. If specified, also checks the key has this
+     * value. Booleans will be converted to 1 and 0 (even inside arrays)
+     */
+    public function seeInRedis($key, $value = null)
+    {
+        $this->assertTrue(
```
DavertMik: same as above: please provide failure message

src/Codeception/Module/Redis.php
```diff
@@ -3,87 +3,626 @@
...
+     * @param string $key       The key
+     * @param mixed  $item      The item
+     * @param null   $itemValue Optional and only used for zsets and hashes. If
+     * specified, the method will also check that the $item has this value/score
+     *
+     * @return bool
+     */
+    public function seeRedisKeyContains($key, $item, $itemValue = null)
+    {
+        $this->assertSame(
```
DavertMik: failure message

DavertMik:
Thanks @marcverney this is really impressive :clap: It's the real Redis module!
I added some comments, and same did Scrutinizer, I think it knows something.

src/Codeception/Module/Redis.php
marcverney: Well, this allow someone to create a key in one test and interact with it in the next test(s), for instance. Codeception's base `Module` class provides those hooks, I'm only letting the user use them (I would do the same with there `_after` equivalent, without any problem). For the moment, I'll change the default value to `"test"` as it looks like it's the most common behavior in Codeception. If you prefer to only allow `"test"`, just let me know.

marcverney:
I'm not sure I completely agree with Scrutinizer's suggestions. I always put `break` statements even after a `throw`, a `return`, or in the last `case `block, because in the future someone may change the instructions in the block and forget to add the `break`. As for the code duplications, they are not, really. I could fix all of this, however, if this is important for the project. This is only personal preference. Just tell me.

src/Codeception/Module/Redis.php
```diff
@@ -42,7 +42,7 @@ class Redis extends CodeceptionModule
     protected $config = [
         'host'          => '127.0.0.1',
         'port'          => 6379,
-        'cleanupBefore' => 'suite'
+        'cleanupBefore' => 'test'
```
DavertMik: I'd replace this with `cleanup` options:

* cleanup: false - disable cleanup
* cleanup: true - cleanup before test
* cleanup: test - same as above
* cleanup: suite - cleanup suite.

DavertMik:
Thanks, the test especially is absolutely great!
The only thing I will change after merge is cleanup option. `cleanupBefore` to be removed (its ugly) and `cleanup` to reenabled again with following options:

* cleanup: false - disable cleanup
* cleanup: true - cleanup between tests
* cleanup: test - same as above
* cleanup: suite - cleanup before suite

PR #3029, opened 2016-04-26
Fixes #3028

### LINKED ISSUE #3028: Laravel4 Connector: false assumption of 'db' existence
Method `_after(TestCase $test)` in `codeception\src\Codeception\Module\Laravel4.php` contains following lines:
```php
if ($this->app['db'] && $this->cleanupDatabase()) {
    $this->app['db']->rollback();
}
```

My project does not use Database Service (Elastic Search in place) so I have disabled `DatabaseServiceProvider` in Laravel config. That's why it fails on `if($this->app['db'])` part. From debugging step by step I know that instead of calling `offsetExists` it immediately invokes `offsetGet` in `Illuminate\Container\Container`:
```php
public function offsetGet($key)
{
	return $this->make($key);
}
```

What I get is `ReflectionException` - "Class db does not exist" from `Container::build()`

Same thing happens inside `function _before(TestCase $test)`.

I'm not sure what triggers this error: this module, Laravel, PHP engine or quantum physics.

Namek:
I did not rebase on top of master branch because there is no Laravel4 module anymore.

### DIFF
--- a/src/Codeception/Lib/Connector/Laravel4.php
+++ b/src/Codeception/Lib/Connector/Laravel4.php
@@ -71,7 +71,7 @@ class Laravel4 extends Client
         // Store a reference to the database object
         // so the database connection can be reused during tests
         $oldDb = null;
-        if ($this->app['db'] && $this->app['db']->connection()) {
+        if (isset($this->app['db']) && $this->app['db']->connection()) {
             $oldDb = $this->app['db'];
         }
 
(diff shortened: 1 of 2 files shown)


janhenkgerritsen:
I created my own PRs to fix the issue, it adds some additional checks and also applies the fix for the Laravel 5 module.

The PR for Codeception 2.1 is #3049. The PR for Codeception 2.2 (master) is #3048.

Thanks for reporting and providing the solution to fix the issue.

PR #3032, opened 2016-04-26
Improved debug output of DB module

Naktibalda:
Replaces #2980


javigomez:
tested, works:

screen shot 2016-05-05 at 12 34 31

Issue #3033, opened 2016-04-26
Count for tests being executed when using groups


gdscei:
When using groups, the tests total that is being showed by suite, is the total number of tests, not the amount of tests in the group(s). 

To reproduce:
- Create two tests in the same suite
- Assign a group to one of them
- Run codecept with the -g option, specifying the group
- In console '[suite name] Tests (2)' is shown

Naktibalda:
Test filter has the same issue:
```
$ php codecept run tests/unit/Codeception/Module/DbTest.php:SeeInDatabase
Codeception PHP Testing Framework v2.1.9
Powered by PHPUnit 5.3.2 by Sebastian Bergmann and contributors.

Unit Tests (9) -------------------------------------------------
DbTest::testHaveAndSeeInDatabase                          Ok
DbTest::testSeeInDatabase                                 Ok
DbTest::testDontSeeInDatabase                             Ok
DbTest::testDontSeeInDatabaseWithEmptyTable               Ok
----------------------------------------------------------------
```

PR #3034, opened 2016-04-26
Fix the count of tests printed in console when tests are filtered

gdscei:
Fix for #3033 
I'm not sure this would be the best solution to the issue, maybe a refactor of the Loader would be better, but this seems more like something that can join easily into 2.2.


DavertMik:
Thanks. I don't see why not to merge it )
Finally this small ugly bug is fixed!

PR #3035, opened 2016-04-27
[FIX] pass the zend request the files

devartbvba:
While using codeception for testing a ZF2 project, we found out that files attached to a REST call were not received in Zend's action methods. After investigation, we could see that get and post parameters were attached to the Zend request object, but files were not. This fixes this issue.

### DIFF
--- a/src/Codeception/Lib/Connector/ZF2.php
+++ b/src/Codeception/Lib/Connector/ZF2.php
@@ -76,6 +76,7 @@ class ZF2 extends Client
 
         $zendRequest->setQuery(new Parameters($query));
         $zendRequest->setPost(new Parameters($post));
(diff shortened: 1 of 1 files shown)


Naktibalda:
Thanks.

To make behaviour consistent with other modules please change your line of code to 
`$zendRequest->setFiles(new Parameters($this->remapFiles($request->getFiles())));`
and add  `use Shared\PhpSuperGlobalsConverter;` at the top of the class (it is a trait).

Then expected format of $files array will be a format of $_FILES.

Naktibalda:
Actually it isn't necessary, other modules remap from structure used by ZF2 to $_FILES, we don't need to do that here.

Issue #2836, opened 2016-02-23
Friend killed after Suite


eXorus:
Hello,

When I use haveFriend/does in a test case of Cest File the new window is closing at the end of the suite.

My conf is restart: true so I'm expected to kill all the windows after each test case.

```
    public function testFriend1(AcceptanceTester $I)
    {
        $toto = $I->haveFriend('toto', $I);
        $toto->does(function($I) {
            $I->amOnUrl('http://localhost/toto-friend');
        });

        $I->amOnUrl('http://localhost/toto-nofriend');
    }

    public function testFriend2(AcceptanceTester $I)
    {
        $titi = $I->haveFriend('titi', $I);
        $titi->does(function($I) {
            $I->amOnUrl('http://localhost/titi-friend');
        });
        $I->amOnUrl('http://localhost/titi-nofriend');
    }
```
### Actual behaviour (order by closed windows) :

At the end of the test case testFriend1
1. toto-nofriend

At the end of the test case testFriend2
2. titi-nofriend

At the end of the test suite
3. toto-friend
4. titi-friend
### Expected behaviour (order by closed windows) :

At the end of the test case testFriend1
1. toto-nofriend
2. toto-friend

At the end of the test case testFriend2
3. titi-nofriend
4. titi-friend

DavertMik:
Sorry for your friend... :)

eXorus:
So what I suspect is that friend is killed when Actor is killed because this is Actor that have friends.

Unfortunately, all actors injected in the test case are killed at the end of the test suite. I'm not sure it's good for performance.

PR #3039, opened 2016-04-27
Friends - MultiSession : Replace close() by quit() in WebDriver and quit all sessions after test

eXorus:
Hello,

I'm using Selenium Grid with Codeception Friends and Codeception doesn't kill correctly the browser so I'm stuck with a lot of node used for nothing :

```
public function testFriend1(AcceptanceTester $I)
{
   $I->amOnUrl('http://localhost/toto-nofriend');

   $toto = $I->haveFriend('toto', $I);
   $toto->does(function($I) {
      $I->amOnUrl('http://localhost/toto-friend');
   });
}
```

with Selenium Hub (2.53) and 2 Firefox nodes
"http://localhost/toto-nofriend" will be on node1
"http://localhost/toto-friend" will be on node2

**At the end of the test with close()**
node1 will be free slot
node2 will be always used (with state deleted but still used)

**At the end of the test with quit()**
node1 will be free slot
node2 will be free slot

### DIFF
--- a/src/Codeception/Lib/Actor/Shared/Friend.php
+++ b/src/Codeception/Lib/Actor/Shared/Friend.php
@@ -23,15 +23,8 @@ trait Friend
         if (!isset($this->friends[$name])) {
             $actor = $actorClass === null ? $this : new $actorClass($this->getScenario());
             $this->friends[$name] = new LibFriend($name, $actor, $this->getScenario()->current('modules'));
+            $this->getScenario()->friends[$name] = $this->friends[$name];
         }
         return $this->friends[$name];
-    }
-
-    public function leaveFriends()
-    {
-        foreach ($this->friends as $friend) {
-            $friend->leave();
-        }
-        $this->friends = [];
     }
 }
--- a/src/Codeception/Lib/Actor/Shared/Friend.php
+++ b/src/Codeception/Lib/Actor/Shared/Friend.php
@@ -26,4 +26,12 @@ trait Friend
         }
         return $this->friends[$name];
     }
+
+    public function leaveFriends()
+    {
+        foreach ($this->friends as $friend) {
+            $friend->leave();
+        }
+        $this->friends = [];
+    }
 }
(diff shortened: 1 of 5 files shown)


eXorus:
Fix #2836

DavertMik:
Somehow WebDriver tests fail in this build. Any ideas what is wrong? Please check the Semaphore CI, webdriver tests are running there

eXorus:
It's working well with Selenium but doesn't work for Selenium Grid.

I'm working on it.

eXorus:
@DavertMik I need to access to array friends in the Actor class from _after() in Module/WebDriver.php, do you know if it's possible ?

I have only TestCase variable...

eXorus:
It' think I'm finished. It's my first big contribution to codeception so I'm not sure to do the best things.

I don't like so much to add friends array in Scenario but I  don't know all the friends created during a test in WebDriver and I get only scenario so ...

src/Codeception/Lib/Actor/Shared/Friend.php
```diff
@@ -23,6 +23,7 @@ public function haveFriend($name, $actorClass = null)
         if (!isset($this->friends[$name])) {
             $actor = $actorClass === null ? $this : new $actorClass($this->getScenario());
             $this->friends[$name] = new LibFriend($name, $actor, $this->getScenario()->current('modules'));
+            $this->getScenario()->friends[$name] = $this->friends[$name];
```
DavertMik: I'd suggest to save current friends into module and not to scenario. Scenario is a general purpose object available in all test suites, even if there are no modules with friends in it.

I'd like you to save friend instances into the WebDriver modules itself. You can get it through
```
$this->getScenario()->current('modules') // check that module implements Interfaces\MultiSession
```
and set friends into it.

DavertMik:
I like the implementation but I'd ask not to add additional properties to Scenario. I'd suggest to use WebDriver module for it.

eXorus:
I did it in another PR with doc, interfaces, friends in webdriver ...

PR #3040, opened 2016-04-27
Support LIKE operator in Db Module

teeencho:
Adding the ability to generate where clause with LIKE operator usage ['field LIKE' => '%value%']

### DIFF
--- a/src/Codeception/Lib/Driver/Db.php
+++ b/src/Codeception/Lib/Driver/Db.php
@@ -164,6 +164,9 @@ class Db
             if ($v === null) {
                 $params[] = $this->getQuotedName($k) . " IS NULL ";
                 unset($criteria[$k]);
+            } elseif (strpos(strtolower($k), ' like') > 0) {
+                $k = str_replace(' like', '', strtolower($k));
+                $params[] = $this->getQuotedName($k) . " LIKE ? ";
             } else {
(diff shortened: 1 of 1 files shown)


Naktibalda:
What's your use case?
In tests you should know exactly what data is present in database before test and what record should be added by the application you are testing.

teeencho:
@Naktibalda Hi, we have a table where we store tasks to be executed in the background. The field is in JSON format. I couldn't possibly know the exact parameter values, since some are generated by my test and some are tokens generated by an external payment service.

Naktibalda:
Looks valid.

@DavertMik are you happy with the syntax?

DavertMik:
I like the syntax but docs should be updated to mention this. 
Could you update Db module docblock? (.md docs are generated form it)

Naktibalda:
I will update the docblock.

Naktibalda:
@DavertMik I am confused.
How did you add Example to http://codeception.com/docs/modules/Db#seeInDatabase ?
seeInDatabase has no docblock https://github.com/Codeception/Codeception/blob/2.1/src/Codeception/Module/Db.php#L325

DavertMik:
Ok, looks like I added this from `DbInterface` from https://github.com/Codeception/Codeception/blob/2.1/src/Codeception/Lib/Interfaces/Db.php

Issue #3044, opened 2016-04-28
There is no way to pass backupGlobals variable to PHPUnit


mkeasling:
I want to set the backupGlobals variable to false, but there seems to be no way to do that.  I'd like to be able to set that in the YAML file for a suite, and have it be passed to PHPUnit

PR #3045, opened 2016-04-28
Provide method for setting backupGlobals

mkeasling:
This address issue #3044 

Pass the backup_globals variable, if provided in the .yml file, to the test suite.

### DIFF
--- a/src/Codeception/SuiteManager.php
+++ b/src/Codeception/SuiteManager.php
@@ -143,7 +143,7 @@ class SuiteManager
         }
         $suite->setName($name);
         if (isset($this->settings['backup_globals'])) {
-            $suite->setBackupGlobals(boolval($this->settings['backup_globals']));
+            $suite->setBackupGlobals((bool) $this->settings['backup_globals']);
         }
         $suite->setModules($this->moduleContainer->all());
         return $suite;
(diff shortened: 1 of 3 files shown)


DavertMik:
Ok, I got your point but I'd like to improve the implementation:

Please add `backup_globals` parameter to the configuration https://github.com/Codeception/Codeception/blob/2.1/src%2FCodeception%2FConfiguration.php#L88

So you won't need to check for `isset` as you always receive default values from the global config.

also could you add this parameter to reference? https://github.com/Codeception/Codeception/blob/2.1/docs%2Freference%2FConfiguration.md

mkeasling:
Great feedback; thank you!

I've addressed your requests; please let me know if there's anything else I can do to improve this.

src/Codeception/SuiteManager.php
```diff
@@ -142,6 +142,9 @@ protected function createSuite($name)
             $name = $this->settings['namespace'] . ".$name";
         }
         $suite->setName($name);
+        if (isset($this->settings['backup_globals'])) {
+            $suite->setBackupGlobals(boolval($this->settings['backup_globals']));
```
Naktibalda: boolval method was added in PHP 5.5, please use (bool) cast instead.

DavertMik:
Thanks!

Issue #2038, opened 2015-06-11
Self-update issue


sergeyklay:
```sh
$ sudo codecept self-update
Codeception version 2.0.14

Checking for a new version...

A newer version is available: 2.1.0-beta

Do you want to update? y

Updating...

/usr/local/bin/codecept has been updated.

$ sudo codecept --version
Codeception version 2.0.14
```

sergeyklay:
```
$ ls -l /usr/local/bin/codecept
-rwxr-xr-x 1 root root 1292559 Jun 11 19:43 /usr/local/bin/codecept
```

Naktibalda:
There is an inconsistency between what self-update checks and what it downloads.
It checks what is the latest tag in github, but downloads http://codeception.com/codecept.phar

Naktibalda:
Have you fixed it?

DavertMik:
I think we don't have beta versions now. 
And I think we won't have betas in phar.

Naktibalda:
It's better to prepare for beta versions if you are going to publish 2.2.0-beta or 3.0.0-beta.

DavertMik:
I think they can be downloaded via Composer only. Why not? Phar should be for stable environments.

DavertMik:
You can't do self-revert after self-update :)

Naktibalda:
http://codeception.com/builds

DavertMik:
Yes, I know we had beta-versions and you can download older versions. But I don't see a reason to provide packages for beta versions, as majority of our users (ok, I didn't count them actually) can download it with Composer

Naktibalda:
> $ php codecept.phar self-update
> Codeception version 2.1.4
> 
> Checking for a new version...
> 
> A newer version is available: 2.2.0-RC
> 
> Do you want to update? y
> 
> Updating...
> 
> codecept.phar has been updated.
> 
> $ php codecept.phar self-update
> Codeception version 2.1.8
> 
> Checking for a new version...
> 
> A newer version is available: 2.2.0-RC

PR #3050, opened 2016-04-29
Fix self-update command to update only to stable versions

MAXakaWIZARD:
This PR fixes the issue when `self-update` updates `2.1.7` to `2.2.0-RC2` instead of `2.1.8`.


Naktibalda:
Thanks, it looks good.

Fixes #2038

MAXakaWIZARD:
@Naktibalda Should we merge this also to `master`? Or you regularly merge branch `2.1` to `master`?

PR #3064, opened 2016-04-30
Check installed mongo extension to use.

Orgun109uk:
Updated MongoDb driver to check if using the legacy driver or not.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,8 +30,7 @@
         "symfony/browser-kit": ">=2.7 <3.1",
         "symfony/css-selector": ">=2.7 <3.1",
         "symfony/dom-crawler": ">=2.7 <3.1",
-        "behat/gherkin": "~4.4.0",
-        "mongodb/mongodb": "^1.0"
+        "behat/gherkin": "~4.4.0"
     },
     "require-dev": {
         "monolog/monolog": "~1.8",
@@ -49,7 +48,8 @@
         "flow/jsonpath": "For using JSONPath in REST module",
         "phpseclib/phpseclib": "for SFTP option in FTP Module",
         "league/factory-muffin": "For DataFactory module",
-        "league/factory-muffin-faker": "For Faker support in DataFactory module"
+        "league/factory-muffin-faker": "For Faker support in DataFactory module",
+        "mongodb/mongodb": "For the MongoDB extension and the MongoDb module"
     },
 
     "autoload":{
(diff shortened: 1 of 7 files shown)


src/Codeception/Lib/Driver/MongoDb.php
```diff
@@ -59,12 +126,7 @@ public function __construct($dsn, $user, $password)
             ];
         }
 
-        try {
-            $this->client = new \MongoClient($dsn, $options);
-            $this->dbh    = $this->client->selectDB($this->dbName);
-        } catch (\MongoConnectionException $e) {
-            throw new ModuleException($this, sprintf('Failed to open Mongo connection: %s', $e->getMessage()));
-        }
+        $this->{$this->legacy ? 'setupMongo' : 'setupMongoDB'}($dsn, $options);
```
DavertMik: what a strange language construct

DavertMik:
Ok, actually I could merge it but could you fix Travis build? Looks like some dependencies are not installed

src/Codeception/Lib/Driver/MongoDb.php
Orgun109uk: Yeah, an old trick i picked up some time ago.

Orgun109uk:
Have made the mongodb library as a suggestion instead of a requirement.

DavertMik:
You can add `mongodb` to require-dev so it was installed on Travis. 
From what I see, MongoDb tests are skipped. Could you make mongodb package installed and all mongo tests passed on Travis?

Orgun109uk:
Yep, can do. This will require installing the mongodb extension via pecl, if your happy for me to do this?

Orgun109uk:
@DavertMik - I have finally managed to get all builds building successfully, with the exception for "semaphoreci" which appears will require the mongodb extension to be enabled within its settings.

composer.json
```diff
@@ -40,7 +40,9 @@
         "pda/pheanstalk": "~3.0",
         "flow/jsonpath": "~0.2",
         "league/factory-muffin": "^3.0",
-        "league/factory-muffin-faker": "^1.0"
+        "league/factory-muffin-faker": "^1.0",
+        "ext-mongodb": "*",
```
DavertMik: nah, codeception by itself does not require mongodb. I will remove this constraint

DavertMik:
Thanks! I think we can go with it. Really great stuff

PR #3066, opened 2016-04-30
[PhpBrowser] added `handler` and `middleware` config options

DavertMik:


Naktibalda:
It isn't Guzzle, it is Guzzle6!

Changelog line should be
`* [PhpBrowser] added `handler` and `middleware` config options to customize *Guzzle6* handlers and middleware`

PhpBrowser doc block should say that these options have no effect on PHP 5.4 and only have effect on PHP 5.5 if Codeception is installed with composer (not available in phar file).

DavertMik:
Ok, right :+1:

Naktibalda:
Guzzle5 has handler option too: http://docs.guzzlephp.org/en/5.3/clients.html#creating-a-client
> Configures the RingPHP handler used to transfer the HTTP requests of a client. 
RingPHP has a middleware option too.
http://ringphp.readthedocs.io/en/latest/spec.html#middleware

But there is not much point to implement support if we are going to drop support for Guzzle 5 in 3.0

PR #3068, opened 2016-05-02
Friends - MultiSession : Replace close() by quit() in WebDriver and quit all sessions after test

eXorus:
Hello,

I'm using Selenium Grid with Codeception Friends and Codeception doesn't kill correctly the browser so I'm stuck with a lot of node used for nothing :

```
public function testFriend1(AcceptanceTester $I)
{
   $I->amOnUrl('http://localhost/toto-nofriend');

   $toto = $I->haveFriend('toto', $I);
   $toto->does(function($I) {
      $I->amOnUrl('http://localhost/toto-friend');
   });
}
```

with Selenium Hub (2.53) and 2 Firefox nodes
"http://localhost/toto-nofriend" will be on node1
"http://localhost/toto-friend" will be on node2

**At the end of the test with close()**
node1 will be free slot
node2 will be always used (with state deleted but still used)

**At the end of the test with quit()**
node1 will be free slot
node2 will be free slot

#3039 #2836

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -2325,16 +2325,13 @@ class WebDriver extends CodeceptionModule implements
                 }
 
                 throw new ElementNotFound(json_encode($value), "Option inside $field matched by name or value");
-                break;
             case "textarea":
                 $el->sendKeys($value);
                 return;
-                break;
             case "div": //allows for content editable divs
                 $el->sendKeys(WebDriverKeys::END);
                 $el->sendKeys($value);
                 return;
-                break;
             //Text, Checkbox, Radio
             case "input":
                 $type = $el->getAttribute('type');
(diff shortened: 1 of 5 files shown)


eXorus:
Fix 11 issues from scrutinizer. I'm done with this PR.

eXorus:
I found an issue when test failed (element not visible for example). The PR is not ready yet

eXorus:
I'm done with this PR, it's working well for me on local selenium and selenium grid @DavertMik @Naktibalda 

Actually Codeception kill the browsers of the friends at the end of the PHP script (when all objects are destructs).
This PR saved all the sessions opened during a run (friends or not) and kill them at the end of the test case or suite according to the configuration.

DavertMik:
Thank you very much!
This is awesome.

src/Codeception/Module/WebDriver.php
```diff
@@ -394,10 +392,21 @@ protected function formatLogEntries(array $logEntries)
     public function _afterSuite()
     {
         // this is just to make sure webDriver is cleared after suite
-        if (isset($this->webDriver)) {
-            $this->webDriver->quit();
+        $this->cleanWebDriver();
+    }
+
+    public function cleanWebDriver()
```
DavertMik: I will make it protected

Issue #2986, opened 2016-04-14
getMock will be deprecated in PhpUnit 5.4


Naktibalda:
https://github.com/sebastianbergmann/phpunit/commit/5b6283a80cdb98dd482a86eee75f8e6a8f598cd8#diff-511c90242d5589e941582006e0485184R1445
```
PHPUnit_Framework_TestCase::getMock() is deprecated, use PHPUnit_Framework_TestCase::getMockBuilder() instead
```

Codeception\Util\Stub::doGenerateMock must be updated.

DavertMik:
Looks like we have serious issue here. Did this ever work?

```php
        $testCase = self::extractTestCaseFromArgs($args);
        $class = $testCase instanceof \PHPUnit_Framework_TestCase ? $testCase : new \PHPUnit_Framework_MockObject_Generator;
        $methodName = $isAbstract ? 'getMockForAbstractClass' : 'getMock';

        $mock = call_user_func_array([$class, $methodName], $args);
```

If a testcase is passed we are using it for mocj generation. However, `getMock` and `getMockForAbstractClass` methods in `PHPUnit_Framework_TestCase` are **protected**. How this was expected to be working? 

I blame myself I ever merged PR which added mocking capabilities to `Codeception\Util\Stub` :no_good: It was just wrong by name and design and now we have consequences of it :(

DavertMik:
Nevermind. `getMock` becomes protected in PHPUnit 5.4

DavertMik:
Ok, here is my patch https://github.com/Codeception/Codeception/pull/3093

PR #3080, opened 2016-05-04
Set upper boundary of supported PHPUnit versions to <5.4

Naktibalda:
PHPUnit 5.4 is likely to break Codeception\Util\Stub, so I'm setting this restriction to prevent automatic upgrade to PHPUnit 5.4

I will revisit this issue in June when PHPUnit 5.4 is released.


Naktibalda:
#2986

DavertMik:
This should not affect users that are not using `Codeception\Util\Stub`. So I'm against setting upper boundary. What we can do is to throw an exception with a hint to downgrade to PHPUnit v5.3 in class. And update the changelog

Naktibalda:
It won't do much harm if the users will not be able to upgrade to Phpunit 5.4 for a week. But if someone really want it, they could use the penultimate version of Codeception which has no restriction.

There is a risk and I think that it would be safer to put a restriction. But if you disagree, then we can do nothing and see how it works out.

DavertMik:
Oh, what a mess there in `Codeception\Util\Stub` :(

DavertMik:
The most nice change I see in PHPUnit 5.4 is this method https://github.com/sebastianbergmann/phpunit/blob/master/src%2FFramework%2FTestCase.php#L1407

Naktibalda:
So we actually needed it.

DavertMik:
Yes :(
Actually the only change which really affected us was not in PHPUnit itself but in PHPUnit Code overage :(

PR #3105, opened 2016-05-13
PSR-2 compliance

MAXakaWIZARD:
This PR introduces some efforts on making code PSR2-compliant. Some rules can not be followed at the moment, so I've added custom ruleset.


MAXakaWIZARD:
@DavertMik @Naktibalda Please review

Issue #3097, opened 2016-05-12
Symfony route tests fail with additional query params


davidcochrum:
`seeCurrentRouteIs()` and `seeInCurrentRoute()` will only match routes which generated URLs match exactly. In our project, we're adding additional query string parameters outside of any defined placeholders on occasion which cause these tests to fail (unnecessarily IMHO).

For instance (where route named `foo` resolves to `/foo` with no placeholders) this test passes:
``` php
$I->amOnPage('/foo');
$I->seeInCurrentRoute('foo');
```
But this test fails:
``` php
$I->amOnPage('/foo?bar');
$I->seeInCurrentRoute('foo');
```

Technically I am still on the route and the Symfony router will choose the correct controller defined by the `foo` route, but the tests themselves doesn't seem to make the same match.

raistlin:
Could you test branch 2.1?

raistlin:
I have been testing with a new application on Symfony 2.8.5, but when I define a route as ```/my/route/{p1}?param={p2}``` or only ```/my/route/{p1}?param``` or even ```/my/route/?param```, Symfony route matcher is unable to match any route, even written the url on the browser, a 404 exception is always returned.

I couldn't find how to enable this kind of route definitions should work at Symfony2 documentation.
So I understand that this kind of feature is not supported.

PR #3106, opened 2016-05-13
[Symfony] Reworked seeCurrentRouteIs

raistlin:
Now it will properly check route name and route parameters, avoiding query parameters.

### DIFF
--- a/src/Codeception/Module/Symfony2.php
+++ b/src/Codeception/Module/Symfony2.php
@@ -358,8 +358,17 @@ class Symfony2 extends Framework implements DoctrineProvider, PartedModule
         if (!$router->getRouteCollection()->get($routeName)) {
             $this->fail(sprintf('Route with name "%s" does not exists.', $routeName));
         }
-        $url = $router->generate($routeName, $params);
-        $this->seeCurrentUrlEquals($url);
+
+        try {
+            $match = $router->match(explode('?', $this->grabFromCurrentUrl())[0]);
+        } catch (\Symfony\Component\Routing\Exception\ResourceNotFoundException $e) {
+            $this->fail(sprintf('The "%s" url does not match with any route', $routeName));
+        }
+        $expected = array_merge(array('_route' => $routeName), $params);
+        $intersection = array_intersect_assoc($expected, $match);
+
+        $this->assertEquals($expected, $intersection);
+
     }
 
     /**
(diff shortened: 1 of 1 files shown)


raistlin:
Fixes #3097

src/Codeception/Module/Symfony2.php
```diff
@@ -358,8 +358,17 @@ public function seeCurrentRouteIs($routeName, array $params = [])
         if (!$router->getRouteCollection()->get($routeName)) {
             $this->fail(sprintf('Route with name "%s" does not exists.', $routeName));
         }
-        $url = $router->generate($routeName, $params);
-        $this->seeCurrentUrlEquals($url);
+
+        try {
+            $match = $router->match(explode('?', $this->grabFromCurrentUrl())[0]);
```
davidcochrum: Is this overly simplistic if the route is defined with placeholders within query params?

src/Codeception/Module/Symfony2.php
```diff
@@ -382,7 +391,7 @@ public function seeInCurrentRoute($routeName)
         }
 
         try {
-            $matchedRouteName = $router->match($this->grabFromCurrentUrl())['_route'];
+            $matchedRouteName = $router->match(explode('?', $this->grabFromCurrentUrl())[0])['_route'];
         } catch (\Exception\ResourceNotFoundException $e) {
```
davidcochrum: The correct exception should be the same as Line 364: `\Symfony\Component\Routing\Exception\ResourceNotFoundException`

src/Codeception/Module/Symfony2.php
```diff
@@ -358,8 +358,17 @@ public function seeCurrentRouteIs($routeName, array $params = [])
         if (!$router->getRouteCollection()->get($routeName)) {
             $this->fail(sprintf('Route with name "%s" does not exists.', $routeName));
         }
-        $url = $router->generate($routeName, $params);
-        $this->seeCurrentUrlEquals($url);
+
+        try {
+            $match = $router->match(explode('?', $this->grabFromCurrentUrl())[0]);
+        } catch (\Symfony\Component\Routing\Exception\ResourceNotFoundException $e) {
+            $this->fail(sprintf('The "%s" url does not match with any route', $routeName));
```
davidcochrum: Both these methods end up outputting an incorrect message because the URL should be placed in the string instead of the route name. Or the message should be reworded.

raistlin: You are right, I'll fix error messages.

src/Codeception/Module/Symfony2.php
raistlin: Not really, above we need ALL the parameters to do the comparison, at this method we only need '_route' because we will ignore parameters.

src/Codeception/Module/Symfony2.php
raistlin: We don't support parameter definition on route query parameters.
I did some tests and Symfony route matched doesn't seem to support it either.
So our solution is good enough.

PR #3109, opened 2016-05-16
Add Robo command for checking code style

MAXakaWIZARD:
`php robo.phar check:code-style`
`php robo.phar check:code-style -r diff`

### DIFF
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -763,4 +763,21 @@ class RoboFile extends \Robo\Tasks
             ->run();
     }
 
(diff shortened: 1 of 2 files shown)


DavertMik:
Thank you!

Naktibalda:
This change didn't age well.

`phpcs` has never been a part of CI process, so this code style wasn't enforced.
Robofile.php was deleted in 4.0
and I will remove dependency on squizlabs/php_codesniffer in 5.0.

PR #3120, opened 2016-05-22
InnerBrowser::see matches UTF-8 text case-insensitively

### LINKED ISSUE #3114: A PhpBrowser test 'see' is case sensitive in some cases
A method 'see' from PhpBrowser module should be case insensitive http://codeception.com/docs/modules/PhpBrowser#see

But this is not entirely true for strings which contains utf multibyte chars, like š,č, etc.

Source:
`<h1>Šak</h1>`
Tests:
`$I->see('Šak')` success
`$I->see('šak')` fail

Naktibalda:
Fixes #3114


elcuro:
Thanks @Naktibalda

Issue #3118, opened 2016-05-20
Cannot read from/write to MySQL BIT fields


davidcochrum:
We use `BIT(1)` columns in a number of tables for booleans. When attempting to write to one of those fields (using `haveInDatabase()`):
```
[PDOException] SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'active' at row 1
```

When attempting to verify a record exists matching the boolean value (using `seeInDatabase()`):
```
No matching records found for criteria {"active":0} in table ...
```

Obviously this is a bit of a conundrum and I don't know of any sweeping, reliable solution.

DavertMik:
> Obviously this is a bit of a conundrum and I don't know of any sweeping, reliable solution.

Yes, Codeception is not an ORM so we are not trying to support all RBMS 100%. It's not a core feature. However, I encourage you to contribute. There is a Db module and Mysql Driver in `Codeception\Lib\Driver\Mysql` for it. Please check it out, maybe you will get an idea how to fix it.

Send a pull request then.

PR #3121, opened 2016-05-23
Allow INT Parameter SQL Binding

davidcochrum:
Solves Issue #3118

### DIFF
--- a/src/Codeception/Lib/Driver/Db.php
+++ b/src/Codeception/Lib/Driver/Db.php
@@ -226,7 +226,7 @@ class Db
             $i++;
             if (is_bool($value)) {
                 $type = \PDO::PARAM_BOOL;
-            } elseif (preg_match('/^[-+]?\d+$/', $value)) {
+            } elseif (is_int($value)) {
                 $type = \PDO::PARAM_INT;
             } else {
                 $type = \PDO::PARAM_STR;
--- a/src/Codeception/Lib/Driver/Db.php
+++ b/src/Codeception/Lib/Driver/Db.php
@@ -224,7 +224,13 @@ class Db
         $i = 0;
         foreach ($params as $value) {
             $i++;
-            $type = preg_match('/^[-+]?\d+$/', $value) ? \PDO::PARAM_INT : \PDO::PARAM_STR;
+            if (is_bool($value)) {
+                $type = \PDO::PARAM_BOOL;
+            } elseif (preg_match('/^[-+]?\d+$/', $value)) {
+                $type = \PDO::PARAM_INT;
+            } else {
+                $type = \PDO::PARAM_STR;
+            }
             $sth->bindValue($i, $value, $type);
         }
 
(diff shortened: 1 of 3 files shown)


src/Codeception/Lib/Driver/Db.php
```diff
@@ -221,7 +221,14 @@ public function executeQuery($query, array $params)
             throw new \Exception("Query '$query' can't be prepared.");
         }
 
-        $sth->execute($params);
+        $i = 0;
+        foreach ($params as $value) {
+            $i++;
+            $type = preg_match('/^[-+]?\d+$/', $value) ? \PDO::PARAM_INT : \PDO::PARAM_STR;
```
raistlin: @davidcochrum How does this line interact with floats or doubles?

davidcochrum: Leaves them as strings which should never be an issue. If you [refer to the issue I opened](https://github.com/Codeception/Codeception/issues/3118), this is really only enabling read/write access to `BIT(1)` fields which give a "data too long" error when passing a string parameter.

davidcochrum: Determining whether the column and/or value represent booleans is near impossible whereas an integer check is entirely feasible and allows `0`/`1` to be passed for boolean values. Can implement logic here to enable passing `true`/`false` values as well.

src/Codeception/Lib/Driver/Db.php
```diff
@@ -221,7 +221,20 @@ public function executeQuery($query, array $params)
             throw new \Exception("Query '$query' can't be prepared.");
         }
 
-        $sth->execute($params);
+        $i = 0;
+        foreach ($params as $value) {
+            $i++;
+            if (is_bool($value)) {
+                $type = \PDO::PARAM_BOOL;
```
davidcochrum: Added a condition for boolean as well. This *only* handles literal booleans because that is all that can be safely assumed.

src/Codeception/Lib/Driver/Db.php
```diff
@@ -221,7 +221,20 @@ public function executeQuery($query, array $params)
...
+            } elseif (preg_match('/^[-+]?\d+$/', $value)) {
+                $type = \PDO::PARAM_INT;
```
Naktibalda: I wonder if this could cause an issue if someone is trying to store a number in text column.
Probably is_int check would be more appropriate.

Naktibalda:
Does this patch solve your issue with BIT type?

Could you add a test to https://github.com/Codeception/Codeception/blob/2.1/tests/unit/Codeception/Lib/Driver/MysqlTest.php ?
New table can be added to https://github.com/Codeception/Codeception/blob/2.1/tests/data/dumps/mysql.sql

src/Codeception/Lib/Driver/Db.php
davidcochrum: Writing an int into a text column gets converted to text regardless

Naktibalda: It isn't an issue for mysql, but other systems like Sybase/MSSQL are more sensitive.

Naktibalda:
Thanks

Issue #2607, opened 2015-12-04
Lumen module not loading Service providers


NeokamiCode:
Hi, 

I am trying to load a validation service provider that extends the validator facade but none of the extensions are found when the tests are run. The bootstrap/app.php is registering the provider and the logic works just fine out of Codeception.

Any ideas? 
Thanks!

janhenkgerritsen:
Can you provide some code samples of what you are doing? And what versions of Lumen and Codeception are you using?

NeokamiCode:
Lumen is version 5.1.6 and Codeception 2.1.4
So under app/Providers there is this class
```php
<?php

namespace App\Providers;

use Illuminate\Support\ServiceProvider;

class ValidationServiceProvider extends ServiceProvider {

	public function register() {}

	public function boot() {
		\Validator::extend('alpha_dash_spaces', function($attribute, $value)
		{
			return (bool) preg_match( '/^[a-z0-9 .\-]+$/i', $value );
		});

		\Validator::extend('is_array_of_strings', function($attribute, $values, $parameters)
		{
			$checkStrings = function($carry, $string){
				return $carry && is_string($string) !== false;
			};
			if(!is_array($values)) return false;
			return  array_reduce($values,$checkStrings,true);
		});
	}
}
```

The problem is that the provider seems to not to be registered so the extension methods are not available and a BadMethodCallException is thrown while running the test. 

On the app.php file under app/bootstrap there is this line to register the provider:
```php 
$app->register('App\Providers\ValidationServiceProvider');
```

NeokamiCode:
The following happened: 
It threw the exception when I was running the test like that
```shell 
vendor/bin/codecept run tests/api/ProvidersCest.php:test_it_configures_provider
```
But I don't have the described problem if I run it without specifying a test function. So just
```shell 
vendor/bin/codecept run tests/api/ProvidersCest.php
```

janhenkgerritsen:
I cannot replicate your issue. But how do you use your validator? I tested it with the following code after adding your validator definition to the `AppServiceProvider::register()` method:

```
$data = ['test' => 'abcde%'];
$rules = ['test' => 'alpha_dash_spaces'];
$validator = \Validator::make($data, $rules);
$validator->fails(); // or $validator->passes()
```

Everything worked as expected when doing it like this.

PR #3124, opened 2016-05-24
Fix container issue in Lumen Module

kt81:
In Lumen module, initializeLumen() method has been called twice on that starts. (_initialize, _before)
That causes a problem that the IoC container does not store any singleton instances which created via Facade.
Facade::clearResolvedInstance() should be called every before bootApplication().

I guess #2607 is same problem.

### DIFF
--- a/src/Codeception/Module/Lumen.php
+++ b/src/Codeception/Module/Lumen.php
@@ -116,22 +116,14 @@ class Lumen extends Framework implements ActiveRecord
     }
 
     /**
-     * After step hook.
-     *
-     * @param \Codeception\Step $step
-     */
-    public function _afterStep(Step $step)
-    {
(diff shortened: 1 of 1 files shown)


DavertMik:
@janhenkgerritsen could you look into it?

janhenkgerritsen:
I will look into it, I still need to bring the structure of the Lumen module a bit more in line with the structure of the Laravel modules. Also, there were some significant breaking changes with the release of Lumen 5.2, so the module probably needs some work for that as well.

I am very busy right now, and I am also going away on holiday soon, so it will probably take a while before I can work on the Lumen module. For now I will merge this PR since it looks good and does not break any tests. I can rework it later if necessary.

kt81:
Thank you!! 😂

PR #3013, opened 2016-04-21
Fixed namespace issue in the generated page object class

eugene-manuilov:
I have fixed issue described in the #3012 ticket.

PR #3126, opened 2016-05-24
Fixes page object generation with namespaces

### LINKED ISSUE #3012: Namespace is missed for PageObject constructor type hinting
Generate page object command creates a class with constructor function which has type hinting for its single argument. Unfortunately it doesn't take into consideration project namespace and just uses actor's name.

Naktibalda:
Replaces #3013, fixes #3012


DavertMik:
Thanks

PR #3130, opened 2016-05-26
Fix format with Double-Width Characters

kt81:
When the test case named with double-width characters, test result's format is broken.

after this fix:

### DIFF
--- a/src/Codeception/Lib/Console/Message.php
+++ b/src/Codeception/Lib/Console/Message.php
@@ -31,7 +31,7 @@ class Message
 
     public function width($length, $char = ' ')
     {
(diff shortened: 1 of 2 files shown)


Naktibalda:
Are you sure that you don't need to update getLength method too?

raistlin:
A better solution would be update `getLength` method and use `$this->getLength()` on `width` method.

So message length is only calculated on the same place.

kt81:
Thank you for your comments. 😭 

I changed that to use getLength() method.
getLength() is used only by Console::writeFinishedTest for now.
In most cases we may not want that the "Length" contains tags. So getLength will do strip_tags in default. 
(apply('strip_tags') has Side Effect.)

Issue #3128, opened 2016-05-25
Db Module doesn't work on big dumps


dima-stefantsov:
I am on windows, using wamp. I actually get system PHP CLI process exception when I try to run "codecept run" with a big dump.sql

When I comment out this line of code of Db Module:
`$sql = preg_replace('%/\*(?!!\d+)(?:(?!\*/).)*\*/%s', "", $sql);`
, it works like it should, no problem.

String of 11MB is too big or something.
Feels like this line is not needed at all. Please fix.

Naktibalda:
This line removes comments from SQL.

dima-stefantsov:
I know :)
But looks like it's not needed. Is it?

Naktibalda:
It may be needed if dump file contains sql statements in multi-line comments.
And this code exists since 1.0.1 so I am not very keen to delete it.

Could you run Codeception with -vvv flag and copy error output including exception message and complete stacktrace?

dima-stefantsov:
Sql in multiline comments... I see. Then it must be implemented in some way.

There are no exceptions and traces. My PHP CLI (php.exe) simply dies with default Windows "application stopped working" if that php code line is present.

It works even with this line, if instead of 11MB dump I pass a short one, few KB.

Actually I usually do restores with
`shell_exec("7z e -so __{$local_db}.7z | mysql -u {$local_user} {$local_pass} {$local_db}");`

1. Is there a way to maybe override "what Db module does to restore dump"? I'd just pass my shell_exec command in there.
2. Can you reproduce fatal php interpreter error if you pass big/huge string into preg_replace? 12MB+

Naktibalda:
I don't use PHP on Windows.
Had no problem with running that preg_replace on a longer string here: https://3v4l.org/j0WGa#tabs

DavertMik:
> Is there a way to maybe override "what Db module does to restore dump"? I'd just pass my shell_exec command in there.


Yes, you can and should do that. Db module can't perform better than mysql istself. So what you need to do is to create a [new helper](http://codeception.com/docs/06-ReusingTestCode#Helpers) `Helper\Db` and implement its logic in `_before` method:

```
function _before()
{
shell_exec("7z e -so __{$local_db}.7z | mysql -u {$local_user} {$local_pass} {$local_db}");
}
```

Also you will need to disable loading dumps in Db module by setting cleanup and populate to false

```
cleanup: false
populate: false
```

[...]

PR #3138, opened 2016-05-30
#3128 Fixed stack overflow too comples regex

dima-stefantsov:
Simple lazy/non-greedy match-all instead of complex non-capturing negative lookahead.

Still matches both single line and multiline comments. Just like it did before.
Matching single-line comments should not be required, but excluding them would make regex more complex.


Naktibalda:
Thank you, it looks good to me.

DavertMik:
The only problem I don't know how to mention it in changelog. In other things, looks good to me too

PR #3140, opened 2016-05-30
--steps will not disable code coverage

### LINKED ISSUE #2620: Adding --steps parameter removes the code coverage output
I have a build script that calls
`bin/codecept run --coverage --coverage-html --steps`
does not create `tests/_output/coverage`
`bin//codecept run --coverage --coverage-html`
produces `tests/_output/coverage`

The first call I wanted so that the developers will have a nice command line output and HTML report.

Naktibalda:
Fixes #2620


DavertMik:
:laughing: :laughing: if only I could remember why it was written so )

PR #3142, opened 2016-05-30
Merge21

DavertMik:
Latest updates from 2.1 branch


Naktibalda:
It looks that `src/Codeception/Lib/TestLoader.php` was added by mistake.

DavertMik:
Yes, PHPStorm sucked up on merging... :(

PR #3150, opened 2016-05-31
Added missing support for LIKE condition to SqlSrv driver

Naktibalda:
SqlSrv didn't have it because it overrides generateWhereClause method.

Also I deleted misleading comment from ZF1 module.

I made these changes after reviewing  #3142


DavertMik:
Thanks

PR #3159, opened 2016-06-02
Docker parallel testing guide

DavertMik:


docs/12-ParallelExecution.md
```diff
@@ -40,7 +35,87 @@ The default working directory in the container is `/project`.
     
     docker run -v ${PWD}:/project codeception/codeception run
 
-For local testing of the Codeception repository with Docker and `docker-copmose`, please refer to the [testing documentation](../tests/README.md). 
+To prepare application and tests to be executed inside containers you will need to use [Docker Compose](https://docs.docker.com/compose/) to run multiple containers and connect them together. 
+
+Define all required services in `docker-compose.yml` file. Make sure to follow Docker philisophy: 1 service = 1 container. So each process should be defined as its own service. Those services can use official Docker images pulled from DockerHub. Directories with code and tests should be mounted using `volume` directive. And exposed ports should be explicitly set using `ports` directive.
+
```
schmunk42: There are two directives in `docker-compose.yml`, `ports` and `expose` - the latter is also available in `Dockerfile` and is only for communication between containers.
While the former is merely for connecting the container to the outside world or in this case for debugging.
I.e. you only need to define `ports` for the database service if you want to access the database for debugging. It's usually `expose`d by default so the PHP service finds your db.

PR #3162, opened 2016-06-03
Avoid the last version of code coverage

pablofmorales:
The last version of php-code-coverage introduce the use of namepsace and broke the code coverage generation in codecept

https://github.com/sebastianbergmann/php-code-coverage/blob/4.0.0/src/CodeCoverage.php#L25


DavertMik:
Thanks for hotfix. I hope we can have it supported too

DavertMik:
What a mess :( 

Package `phpunit/codecoverage` have namespace `SebastianBergmann\CodeCoverage`

PR #3172, opened 2016-06-04
Check the return value of application->handle()

hinablue:
In the Phalcon Micro Application, when you use the middleware in `before`, and it will be return the boolean (true or false) to make the application process pass through or stop.

When I use `return false` to stop the application, the Lib of Phalcon will got the boolean in LINE 115,

```
// Here will got the boolean when middleware `before` return false.
$response = $application->handle();
```

And the Library will be broken. So I add the check and IF NOT `ResponseInterface`, get the response from `$application->response` will be ok.

### DIFF
--- a/src/Codeception/Lib/Connector/Phalcon.php
+++ b/src/Codeception/Lib/Connector/Phalcon.php
@@ -11,7 +11,6 @@ use Phalcon\Mvc\Application;
 use Phalcon\Mvc\Micro as MicroApplication;
 use Phalcon\Http\Request;
 use Phalcon\Http\RequestInterface;
-use Phalcon\Http\Response as PhResponse;
 use Phalcon\Http\ResponseInterface;
 use Phalcon\Session\AdapterInterface as SessionInterface;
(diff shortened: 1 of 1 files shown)


src/Codeception/Lib/Connector/Phalcon.php
```diff
@@ -11,6 +11,8 @@
 use Phalcon\Mvc\Micro as MicroApplication;
 use Phalcon\Http\Request;
 use Phalcon\Http\RequestInterface;
+use Phalcon\Http\Response as PhResponse;
```
sergeyklay: Could you please remove this line

sergeyklay:
@hinablue

hinablue:
@sergeyklay

OK, I removed.

I found some issue from Phalcon2 which the `$application->response` will also be `NULL`, so I add this line for setting the new response ( for testing ). But I cannot reproduce the error, remove it is fine.

If I can reproduce the error, I'll send the PR :-)

sergeyklay:
Thanks

PR #3174, opened 2016-06-04
Curl extension is required

Naktibalda:
I has always been required: https://github.com/Codeception/Codeception/blob/2.1.10/src/Codeception/Command/Run.php#L380
I just made it explicit in composer.json


DavertMik:
Thanks.

PR #3175, opened 2016-06-05
Added a grabNumRecords to Db module

tocsick:
You might want to compare how many records are in a table before and after some operations. The table might not be empty to begin with.

For example:
```
$count = $I->grabNumRecords( 'book' );
$I->click( 'Save' );
$I->see( 'Field is required' );
$I->seeNumRecords( 'book', $count );
```


DavertMik:
Thanks :+1:

PR #3183, opened 2016-06-06
Added missing deleteHeader method to REST module

### LINKED ISSUE #3161: deleteHeader not supported by REST module
The documentation of the REST module suggests using the "deleteHeader" function to undo the "haveHttpHeader" function: http://codeception.com/docs/modules/REST#haveHttpHeader. The deleteHeader function is however currently only implemented in the PhpBrowser module, which uses setHeader, which is ignored by the REST module. 

Using deleteHeader after using haveHttpHeader does not seem to have any effect. 

code sample:

```
$I = new ApiTester($scenario);
$I->wantTo("do two things");
$I->amBearerAuthenticated(Fixtures::get('apikey'));
$I->haveHttpHeader('Content-Type', 'application/json');
$I->sendPOST('/action/',
    [
        'name' => 'action 1'
    ]
);
$I->seeResponseCodeIs(200);

// Now do a sendDELETE, which should not have application/json in its header:
$I->deleteHeader('Content-Type');
$I->sendDELETE("/action/1");
$I->seeResponseCodeIs(204);
```

api.suite.yml:
```
class_name: ApiTester
modules:
    enabled:
        - REST:
            url: http://localhost/api/v2/
            depends: PhpBrowser
            part: Json
        - PhpBrowser
```

Removing `- PhpBrowser` from the last line results in an error that function deleteHeader() does not exist.

Naktibalda:
Fixes #3161

### DIFF
--- a/src/Codeception/Module/REST.php
+++ b/src/Codeception/Module/REST.php
@@ -168,6 +168,28 @@ EOF;
     }
 
     /**
+     * Deletes the header with the passed name.  Subsequent requests
+     * will not have the deleted header in its request.
+     *
+     * Example:
+     * ```php
+     * <?php
+     * $I->haveHttpHeader('X-Requested-With', 'Codeception');
+     * $I->sendGET('test-headers.php');
+     * // ...
+     * $I->deleteHeader('X-Requested-With');
+     * $I->sendPOST('some-other-page.php');
+     * ?>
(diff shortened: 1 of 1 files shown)


DavertMik:
Yes, thank you. This was accidentally missing

PR #3188, opened 2016-06-07
Fix in the Logger extension.

gnuget:
Using the logger ext in the version 2.2  codeception fails with the following error:

`PHP Fatal error:  Call to undefined method path/to/an/specific/class::getFileName() in /vendor/codeception/codeception/ext/Logger.php on line 82`

Checking the code it seems to since the 2.2 version the `Codeception\TestCase` was replaced with `Codeception\TestInterface` and  some methods were moved to the MetaData Class. 

So the Logger extension fails when it tries to use the `getFileName` method because it is still looking in the Test class instead of the Metadata class. 

In this PR I just fixed that.

### DIFF
--- a/ext/Logger.php
+++ b/ext/Logger.php
@@ -79,7 +79,7 @@ class Logger extends Extension
 
     public function beforeTest(TestEvent $e)
     {
-        $this->logger = new \Monolog\Logger($e->getTest()->getFileName());
+        $this->logger = new \Monolog\Logger($e->getTest()->getMetadata()->getFilename());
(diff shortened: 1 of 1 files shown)


DavertMik:
Thanks for pointing to this bug. There is a much cleaner way to get information about test. It is `Codeception\Test\Descriptor` class. We need it because Codeception 2.2 has different test formats to support. 

Here is my alternative fix:

https://github.com/Codeception/Codeception/pull/3190

gnuget:
Ok, Thanks for your review to my PR,  glad to help to report bugs! :-)

PR #3189, opened 2016-06-07
Location header only redirects when status code is 3xx

### LINKED ISSUE #3171: On POST creation, 201 status with a Location header should not fail
In the case of a POST interaction, `InnerBrowser` is aggressively evaluating a redirection response header, and evaluating a response as a failure if the browser is not redirected.

In this particular case, a _Location_ header doesn't indicate a redirection, but instead indicates the location of the created record (defined in boring detail in [RFC2616, 10.2.2](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.2.2)).

Would there be other issues I'm not seeing by also checking for an HTTP status code of 201 and not interpreting the Location header as a redirect? Even better than that, check for no status code or a [`3xx`](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.3) code and then evaluate the purpose of the header?

As a side note, `2.2.1` sees the issue, but `2.1.10` does not.

Naktibalda:
Fixes #3171

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -1480,7 +1480,8 @@ class InnerBrowser extends Module implements Web, PageSourceSaver, ElementLocato
     protected function redirectIfNecessary($result, $maxRedirects, $redirectCount)
     {
         $locationHeader = $this->client->getInternalResponse()->getHeader('Location');
-        if ($locationHeader) {
+        $statusCode = $this->getResponseStatusCode();
+        if ($locationHeader && $statusCode >= 300 && $statusCode < 400) {
             if ($redirectCount == $maxRedirects) {
                 throw new \LogicException(sprintf(
                     'The maximum number (%d) of redirections was reached.',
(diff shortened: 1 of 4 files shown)


DavertMik:
With tests! Nice

stipsan:
We just upgraded to 2.2 and our entire testing suite failed due to our REST api returning a location header on every 2xx request.
Updated to latest 2.2 few hours ago but it's missing this fix, @DavertMik any chance there's a new 2.2 release out soon?

torkiljohnsen:
Our team also have this issue :) This needs to be merged into 2.2 ASAP. REST APIs will return a `Location` header when a resource is created (and a `201 created` is returned), but right now this results in a redirect. Yikes.

Also: CodeCeption 2.2 Laravel5 module has [a lot of breaking changes](https://github.com/Codeception/Codeception/pull/2866) that are not listed in [the changelog](http://codeception.com/changelog).

sureshHARDIYA:
I am facing the same issue. Need Fix ASAP with this thing. Embarrassing  !!!

Naktibalda:
I would like to use this opportunity to invite you guys to test the release candidate next time.

Issue #3191, opened 2016-06-08
bad encoding in command test title


itigoppo:
```
$I->wantTo('マルチバイト対応wantTo');

Codeception PHP Testing Framework v2.2.1
Powered by PHPUnit 5.4.2 by Sebastian Bergmann and contributors.

Cli Tests (1) 
✔ MbStringCept: Ã?ルチバイト対応wantto (0.12s)
```

wantTo displays wrong character encoding in console

Naktibalda:
Fixed by #3192

PR #3192, opened 2016-06-08
#3191 change ucfirst for multibyte character encodings

itigoppo:


Naktibalda:
1. Please make request to 2.1 branch instead.
2. There is a block of code exactly like this one in https://github.com/Codeception/Codeception/blob/2.1.10/src/Codeception/Subscriber/Console.php#L527 it would be good to make it a function somewhere.

Naktibalda:
Actually, this PR is good enough.
It can't go to 2.1, because `Codeception/Test/Cept.php` does not exist in 2.1

PR #3203, opened 2016-06-10
gherkin languages support

dizzy7:


DavertMik:
I've been waiting for such PR but current implementation can't be accepted. 
Would you mind to improve it so it could actually work for all distributions?

src/Codeception/Test/Loader/Gherkin.php
```diff
@@ -58,7 +45,8 @@ public function __construct($settings = [])
         if (!class_exists('Behat\Gherkin\Keywords\ArrayKeywords')) {
             throw new TestParseException('Feature file can only be parsed with Behat\Gherkin library. Please install `behat/gherkin` with Composer');
         }
-        $keywords = new GherkinKeywords(['en' => static::$defaultKeywords]);
+        $i18n = require __DIR__ . '/../../../../../../behat/gherkin/i18n.php';
```
DavertMik: this **hardcoded** path is really really bad. For instance it will work ONLY if behat and codeception both are in `vendor` dir of Codeception, but in Codeception repo, Codeception itself is in `src` fir, while behat is in `vendor/behat`.

Here you should get a path to Composer vendor dir and provide path from it

DavertMik:
Btw, how would I set the language to use? I think there should be something like in config:

```
gherkin:
    language: en
```

but how to pass it to Gherkin parser?

dizzy7:
I agree. Unfortunately I haven't found a better way to do it than using reflection.

About choosing a language - there is a standart way in gherkin to point a language in the first line of each particular file; for example, it's "#language: ru" for russian. This allows to write tests in several languages, and I can't see why to include it into the configuration.

DavertMik:
> About choosing a language - there is a standart way in gherkin to point a language in the first line of each particular file; for example, it's "#language: ru" for russian. This allows to write tests in several languages, and I can't see why to include it into the configuration.

Oh, thanks, I didn't know that. Looks good then :+1:

DavertMik:
Thanks. Looks good now

PR #3205, opened 2016-06-10
remove same name methods of parent class

dbwhddn10:
removed methods are already existed in parent class


Naktibalda:
Thanks, but it has to stay this way.

This is the only way to change visibility of trait methods which is compatible with HHVM.

PR #3209, opened 2016-06-11
fix search input field by label

dizzy7:


Naktibalda:
What issue does it solve?

dizzy7:
Original xpath without normalize-space() can find 
`<label>Lorem</label>`
but can not find
```
<label>
   Lorem
<label>
```

PR #3210, opened 2016-06-12
Deleted TestLoader, because @DavertMik merged it in from 2.1 by accident

Naktibalda:
#3142


DavertMik:
Not the first time :( 

PHPStorm is getting mad when I'm merging 2.1=>2.2 and it throws errors when I ask it to delete files

Issue #3119, opened 2016-05-21
[imp] Gherkin:snippets list snippets of one specific .feature file 


javigomez:
Right now we run:

```
tests/vendor/bin/codecept g:snippets acceptance
```

...to get a list of the steps implementing the methods.

It would be a great improvement if we could do instead:

```
tests/vendor/bin/codecept g:snippets tests/my.feature
```

...in order to get a list of were are the methods at the `.feature` file (not all the acceptance suite).

edno:
I've pushed the PR #3212 that implements this request.

PR #3216, opened 2016-06-13
Add optional parameter test to command gherkin:snippets

edno:
Implementation of #3119 for creating gherkin snippets for a specific feature. The new command line is similar to the syntax of `run`:
```
codecept gherkin:snippets suite [test]
```
Example:
```
codecept gherkin:snippets acceptance admin.feature
```

---
This change is [Reviewable](https://reviewable.io/reviews/codeception/codeception/3216)

### DIFF
--- a/src/Codeception/Lib/Generator/GherkinSnippets.php
+++ b/src/Codeception/Lib/Generator/GherkinSnippets.php
@@ -25,7 +25,7 @@ EOF;
     public function __construct($settings, $test = null)
     {
         $loader = new Gherkin($settings);
-        if (is_null($test)) {
+        if (empty($test)) {
             $pattern = $loader->getPattern();
         } else {
             $pattern = $test;
--- a/src/Codeception/Lib/Generator/GherkinSnippets.php
+++ b/src/Codeception/Lib/Generator/GherkinSnippets.php
@@ -25,7 +25,7 @@ EOF;
     public function __construct($settings, $test = null)
     {
         $loader = new Gherkin($settings);
-        if (empty($test) || is_null($test)) {
+        if (is_null($test)) {
             $pattern = $loader->getPattern();
         } else {
             $pattern = $test;
--- a/src/Codeception/Lib/Generator/GherkinSnippets.php
+++ b/src/Codeception/Lib/Generator/GherkinSnippets.php
@@ -16,22 +16,27 @@ class GherkinSnippets
      {
         throw new \Codeception\Exception\Incomplete("Step `{{text}}` is not defined");
      }
-     
+
 EOF;
 
     protected $snippets = [];
     protected $processed = [];
 
-    public function __construct($settings)
+    public function __construct($settings, $test = null)
     {
         $loader = new Gherkin($settings);
+        if (empty($test) || is_null($test)) {
+            $pattern = $loader->getPattern();
+        } else {
+            $pattern = $test;
+        }
 
         $finder = Finder::create()
             ->files()
             ->sortByName()
             ->in($settings['path'])
             ->followLinks()
-            ->name($loader->getPattern());
+            ->name($pattern);
 
         foreach ($finder as $file) {
             $pathname = str_replace("//", "/", $file->getPathname());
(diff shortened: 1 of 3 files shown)


src/Codeception/Lib/Generator/GherkinSnippets.php
```diff
@@ -16,22 +16,27 @@ public function {{methodName}}({{params}})
...
 EOF;
 
     protected $snippets = [];
     protected $processed = [];
 
-    public function __construct($settings)
+    public function __construct($settings, $test = null)
     {
         $loader = new Gherkin($settings);
+        if (empty($test) || is_null($test)) {
```
Naktibalda: null is empty, so you could make this line shorter by removing unnecessary condition

edno: my bad. old habits.

src/Codeception/Lib/Generator/GherkinSnippets.php
```diff
@@ -16,22 +16,27 @@ public function {{methodName}}({{params}})
...
 EOF;
 
     protected $snippets = [];
     protected $processed = [];
 
-    public function __construct($settings)
+    public function __construct($settings, $test = null)
     {
         $loader = new Gherkin($settings);
+        if (is_null($test)) {
```
sergeyklay: `var_dump(is_null(false));` // false

edno: Complete set https://www.virendrachandak.com/techtalk/php-isset-vs-empty-vs-is_null/
@sergeyklay Better to revert my commit?

sergeyklay: what about empty($test)

edno: Apparently Symfony Console handles correctly optional param if not set. If not set then `$test` will be null when passed to `GherkinSnippets` constructor.
This is confirmed by the tests that are passed so far.

sergeyklay: So empty(null) === empty(false) === empty('') === empty([])

src/Codeception/Lib/Generator/GherkinSnippets.php
```diff
@@ -16,22 +16,27 @@ public function {{methodName}}({{params}})
...
     protected $processed = [];
 
-    public function __construct($settings)
+    public function __construct($settings, $test = null)
     {
         $loader = new Gherkin($settings);
+        if (empty($test)) {
+            $pattern = $loader->getPattern();
+        } else {
+            $pattern = $test;
```
DavertMik: ok, so practically anything can be passed here. Cest/Test/Cept would be loaded as well

src/Codeception/Lib/Generator/GherkinSnippets.php
```diff
@@ -16,22 +16,27 @@ public function {{methodName}}({{params}})
...
         $finder = Finder::create()
             ->files()
             ->sortByName()
             ->in($settings['path'])
             ->followLinks()
-            ->name($loader->getPattern());
+            ->name($pattern);
 
         foreach ($finder as $file) {
             $pathname = str_replace("//", "/", $file->getPathname());
```
DavertMik: I think that each found file should be checked against `$loader->getPattern()` to ensure that we are working only with `.feature` files here

edno: Good point 👍

DavertMik: Ok, probably this can be it. But you can point this command to a folder with different tests formats in it. So I'd suggest to not throw the exception but to show the number of processed files

edno: I've created a new (local) branch https://github.com/edno/Codeception/tree/pr-3216 that contains the code refactored into 3 commits:
- New param test
- Extend param test for supporting folder
- Additional tests
@DavertMik if ok with you, I'll close this PR and create a new one into the branch 2.2

DavertMik:
> if ok with you, I'll close this PR and create a new one into the branch 2.2

Great, please do. Or you can update current one.

edno:
Code refactored in PR #3280

PR #3219, opened 2016-06-14
Bugfix with duplicate instances in the container

dizzy7:
In some cases duplicates were present in the container, for example "Codeception\Modules\Symfony2" and "\Codeception\Modules\Symfony2"


DavertMik:
Thanks

Issue #3112, opened 2016-05-18
Group events fire twice


jstaudenmaier:
### My Use Case

Using groups to add data to the Db for certain tests (ex: Not all tests need to have two companies so only add the second company for the group of tests that need it)

Created the Group, added it to the `codeception.yml` file, added the appropriate `@group` annotation to the tests in question.  

**Problem**: 
The `_before` event of the group is firing twice, leading to SQL Integrity constraint errors because duplicate data is getting entered.
### Root Cause Analysis

The `test.before.groupname` event is getting fired twice because the group is listed twice in the Scenario->groups[] array.

The group gets added once to the Scenario during the Cest `preload()` process in the `Parser::attachMetadata()` function. 

Then it gets added again in the SuiteManager::addToSuite() function (which, incidentally is the same function that calls preload()) 

There are no guards on the $groups[] array in the scenario to prevent the same group from being added twice so the group ends up in the array twice and when Listener::fire() iterates through the Scenario's groups, it get two items for the group and therefore fires the group's events twice.
### Suggested Solution

Rather than re-engineering the system to avoid double-adds, or adding complicated logic to avoid adding the same group to Scenario->$groups[] twice, a simple solution is to wrap the return of Scenario->getGroups() to only return unique values, like so:

`public function getGroups()
    {
       return array_unique($this->groups);
    }
`

jstaudenmaier:
Just in case you were wondering, this still occurs in Codeception 2.2.  It's just migrated from the Scenario object to the Test\Metadata object.

Naktibalda:
Could you make a pull request?

jstaudenmaier:
I could! :-)

https://github.com/Codeception/Codeception/pull/3221

Naktibalda:
This issue was fixed by #3221 in 2016.

PR #3221, opened 2016-06-14
Fix for issue #3112 Group events fire twice.

jstaudenmaier:


DavertMik:
Great! Tests look cool. Thank you

PR #3223, opened 2016-06-15
bugfix: Symfony tests always fail when the profiler is disabled

dizzy7:


DavertMik:
@raistlin is everything ok here?

dizzy7:
Code using this method (debugResponse, seeEmailIsSent) takes into account the possibility of lack of service profiler.
In my case (258 gherkin tests) everything works properly. Besides, disabling profile service in symfony config reduce memory consumption and speedup tests (7.02 minutes, 198MB -> 5.79 minutes, 158.00MB)

raistlin:
Seems ok.

DavertMik:
Thanks

PR #3224, opened 2016-06-15
Fix loading of parameters from `.env.*` files

smotesko:
The documentation [states](http://codeception.com/docs/06-ModulesAndHelpers#Dynamic-Configuration-With-Params) that we can load params from files named `.env` or `.env.something` but the latter doesn't work because the regex only matches filenames ending with `.env` or `.ini`

### DIFF
--- a/src/Codeception/Configuration.php
+++ b/src/Codeception/Configuration.php
@@ -646,7 +646,7 @@ class Configuration
             }
 
             // .env and ini files
-            if (preg_match('~(\.ini|\.env)$~', $paramStorage)) {
+            if (preg_match('~(\.ini$|\.env(\.|$))~', $paramStorage)) {
                 $params = parse_ini_file($paramsFile);
                 static::$params = array_merge(self::$params, $params);
                 continue;


src/Codeception/Configuration.php
```diff
@@ -646,7 +646,7 @@ private static function prepareParams($settings)
             }
 
             // .env and ini files
-            if (preg_match('~(\.ini|\.env)$~', $paramStorage)) {
+            if (preg_match('~(\.ini$|\.env)~', $paramStorage)) {
```
stipsan: why limit .ini files?

smotesko: @stipsan I see two reasons to "limit" .ini files:
1. They were "limited" before, so I don't change anything in that regard.
2. I've never seen a config file named `.ini.something`.

examples of file names that would match:
```
prod.ini
testing.ini
.env
.env.testing
```

stipsan: ok that makes sense regarding `.ini`.
I'm no regex wizard but I noticed a problem with your updated pattern though: https://regex101.com/r/hO5rU9/2

It's now matching stuff like `.envy` or anything with `.env` in it.

smotesko: @stipsan Yes, I'm aware of it. Don't think if would be a problem because that's the last condition in the chain of possible options (after the `.yml`). See, it doesn't match any files in the project root. It matches files that are explicitly specified in `codeception.yml`, so if user want to specify file named `.envy` there's no reason to deny it in this case, I think.

stipsan: I see, that's good. I think that's a developer friendly approach 👍

I don't know if this functionality is covered in the testing suite already, as I've only recently started to participate here, but I think adding a test that catch the edge case you're fixing in this PR would prevent future regressions.

Basically just a test that makes sure `.env.something` is loaded correctly :)

smotesko: @stipsan 👌 I've looked into covering this with test and I see it requires mocking the PHP functions. It can be done, but needs some thought to not make a mess. I'll have to do it later.

stipsan: that sounds complicated indeed.
Perhaps @DavertMik or @Naktibalda know more about this?

DavertMik: Thanks @stipsan for noticing it. 

I think next pattern should be used:

```
(\.ini$|\.env(\.|$))
```

DavertMik:
Please update the regex

src/Codeception/Configuration.php
smotesko: @DavertMik looks great, thanks. I've changed it.

smotesko:
@DavertMik changed.

DavertMik:
Thanks!

PR #3228, opened 2016-06-16
REST module improvements

DavertMik:
* added HttpCode util class to handle Http codes in more verbose way
* changed error description for seeResponseCode failures
* REST to use seeResponseCode for connector module
* REST stop/startFollowingRedirects methods to be part of JSON and XML parts


src/Codeception/Util/HttpCode.php
```diff
@@ -0,0 +1,162 @@
...
+    const CONFLICT = 409;
+    const GONE = 410;
+    const LENGTH_REQUIRED = 411;
+    const PRECONDITION_FAILED = 412;
+    const REQUEST_ENTITY_TOO_LARGE = 413;
+    const REQUEST_URI_TOO_LONG = 414;
+    const UNSUPPORTED_MEDIA_TYPE = 415;
+    const REQUESTED_RANGE_NOT_SATISFIABLE = 416;
+    const EXPECTATION_FAILED = 417;
+    const I_AM_A_TEAPOT = 418;                                               // RFC2324
```
stipsan: 😂 awesome easter egg, but since 418 is in the HTCPCP and not HTTP/1 I think it's kinda inaccurate to have it in a class that is named `HttpCode`.

DavertMik: I don't have my own opinion on this. I just copy pasted this code :)
Actually teapot is also availble in a similar constant list of Symfony https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/Response.php#L61 so I don't see why not to include it. Anyway, no ona have died of +1 constant

PR #3238, opened 2016-06-18
Search field by label without "for"

dizzy7:
Search input fields in constructions like
```
<label>
   <b>Some label</b>
   <input name="name">
</label>
```

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -817,11 +817,13 @@ class InnerBrowser extends Module implements Web, PageSourceSaver, ElementLocato
         }
 
         // by label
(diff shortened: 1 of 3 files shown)


DavertMik:
Is it even a valid markup? I'm not sure it is widely used or it is recommended somewhere to do markup like this...

raistlin:
A little comment I found related to this topic: http://stackoverflow.com/a/24596641/3036692

dizzy7:
Unfortunately its valid markup [W3C](https://www.w3.org/wiki/HTML/Elements/label#Example_B)
This markup is sometimes used in third-party projects (Sonata Admin in my case), which are not easy to change, but still need to be tested.

PR #3239, opened 2016-06-18
Fix for "File name too long" error in failed tests

dizzy7:
Fix for "File name too long" error in failed gherkin tests with too long description


Naktibalda:
This change conflicts with #3231 which fixes the illegal character issue in a different way.
But limiting the length of filename is a good idea.
Could you make that change in Module/WebDriver.php too?

PR #3244, opened 2016-06-19
[REST] Support simple key-value format for file uploads

Naktibalda:
When REST module was used with PhpBrowser(or more specifically with Guzzle), file upload could be done in 2 formats: https://github.com/Codeception/Codeception/blob/2.2.1/tests/unit/Codeception/Module/PhpBrowserRestTest.php#L205

```php
        $files = [
            'file' => [
                'name' => 'file.txt',
                'type' => 'text/plain',
                'size' => 9,
                'tmp_name' => $tmpFileName,
            ]
        ];
```
and
```php
        $files = [
            'file' => $tmpFileName,
        ];
```

The magic happened in in the Guzzle (if I remember correctly).
When REST module was used with framework modules, the user had to provide a complete $_FILES array (the first type).

I added support for the second type of file upload for all frameworks, but I only tested it with ZF2,
I will test it with ZF1 and Zend-Expressive tomorrow.
https://github.com/Naktibalda/codeception-rest-testcase/commit/024d3870d60ee901cbbb064d12569cea0edd70c3


DavertMik:
Thank you!

VincentClair:
Before, with version 2.1.8, we could pass a `Symfony\Component\HttpFoundation\File\UploadedFile` as value. Something like:
```
$I->sendPOST('/ui/options/images',
    ['name'  => 'logo'],
    ['image' => new UploadedFile($filePath, $filename, $mimeType, filesize($filePath))]
);
```

Since version 2.2.2, it seems to be broken. I get an error that the value of key image is invalid.
I changed to the indexed array and it works.
Is it normal ?

Naktibalda:
@VincentClair Thanks for report, I will change the code to allow objects.

Could you raise it as an issue?

VincentClair:
Here it is #3298 :)

src/Codeception/Module/REST.php
```diff
@@ -527,6 +529,67 @@ protected function encodeApplicationJson($method, $parameters)
...
+                    'size' => filesize($value),
+                    'type' => $this->getFileType($value),
+                    'error' => 0,
+                ];
+                continue;
+            } elseif (is_array($value)) {
+                if (isset($value['tmp_name'])) {
+                    $this->checkFileBeforeUpload($value['tmp_name']);
+                    if (!isset($value['name'])) {
+                        $value['name'] = basename($value);
```
haveyaseen: If `$value` is an array, why call `basename($value)` on it? This leads to the error `basename() expects parameter 1 to be string, array given`.

PR #3248, opened 2016-06-20
selectOption - select option from different selector

eXorus:
Hello,

Module :  WebDriver
Version : 2.2.1
Browser : Firefox

```html
<select id="select1">
<option value="select1_value1">Value1</option>
<option value="select1_value2">Value2</option>
</select>

<select id="select2">
<option value="select2_value1">Value1</option>
</select>
```

```php
$I->selectOption('#select2', 'Value2');
```

**Actual behaviour**
The value "Value2" in field "#select1" is selected

**Expected behaviour**
Failed to select the value "Value2" because it is not in "#select2" field.

**Fix**
I only replace xPath "//option" to ".//option". Like this the findElement will only search in the childs and not everywhere

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -1101,7 +1101,7 @@ class WebDriver extends CodeceptionModule implements
         // partially matching
         foreach ($option as $opt) {
             try {
-                $optElement = $el->findElement(WebDriverBy::xpath('//option [contains (., "' . $opt . '")]'));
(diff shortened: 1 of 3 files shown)


raistlin:
Please add some tests.

raistlin:
It's ok for me.

DavertMik:
> Like this the findElement will only search in the childs and not everywhere

Wow, this should increase overall performance as well. Thank you

Issue #3251, opened 2016-06-20
Get symfony container parameter


samusenkoiv:
I'd be great to be able to grab symfony container parameters. Something like $I->grabContainerParameter($key)

raistlin:
Why do you need grab a parameter while testing?

What's your use case?

samusenkoiv:
e.g. make sure a service is affected by some of attributes.

But since you asked me the question I started having doubts whether this is necessary...
What do you think?

raistlin:
Just because parameters are on container for application use doesn't mean that you as tester need to have access to them.

From my point of view, you as a tester, only need to get the services (class instances) and test they work properly as they are configured on your app.

If you need to test the behavior of your class when some parameter has some specific value, you are not doing a functional or acceptance test: you need a unit test suite.
And unit test suite do not need the framework configuration. You simply instantiate your class to be tested and you inject the 'parameters' at the constructor or set them via setters.

I think that exposing ```getParameter``` method is a bad idea, because some users will start to test behaviors that should be tested on unit suites instead than on functional or acceptance ones.

samusenkoiv:
I think you're right.

iquito:
I think there are valid use cases which need parameter values - in my case for dynamic routes, where route prefixes can be defined in Symfony parameters. I can only do my functional tests if I know what URL to call, and the prefix for that URL is contained in the Symfony configuration of the application. Dynamic routes are generally a valid use case for parameter value access, and probably there are other use cases, especially when testing packages/bundles which are highly configurable by parameter values.

raistlin:
I'm strongly against exposing ```getParameter``` because of my reasons two comments above.

@Codeception/committers What are your thoughts?

janhenkgerritsen:
In the Laravel module I expose the Laravel Application object through a `getApplication` method. This application object is also Laravel's IoC container. This provides a lot of flexibility for the users, which can be helpful in uncommon test scenarios.

idchlife:
I'm testing whether my UserListener service in Symfony2 is doing something when I update some entity. So fields that are should be checked when I update entity are listed in the parameters in config.yml file.
What do you call this test? Integrational? Behaviour? I'm not sure this is unit test but I'm writing one for this case right now!

Actually, to be more clear, I write UserListener that prevents ordinary, not admin user to change users important fields. And I wan't to test if it's working. Create user, call from console, get error 'you are not allowed changing value from console if not root and from web if not admin'.

[...]

PR #3252, opened 2016-06-20
Symfony module grab container param

samusenkoiv:
Add ability to grab container parameters

https://github.com/Codeception/Codeception/issues/3251


samusenkoiv:
I should've branched from 2.2 I think

PR #3253, opened 2016-06-20
Add ability to grab Symfony container parameter

samusenkoiv:
Add ability to grab container parameters

https://github.com/Codeception/Codeception/issues/3251


raistlin:
The code is ok, but I don't understand why do you need to access a parameter on container...

raistlin:
Applied "Needs clarification" until someone explain a good use case of this feature.

Issue #3250, opened 2016-06-20
SOAP and WebDriver


eXorus:
Hello,

Why SOAP doesn't work with WebDriver like REST do it ?

This is working for REST : 

``` yml
class_name: AcceptanceTester
modules:
    enabled:
        - WebDriver:
            browser: firefox
            url: http://localhost/myapp
            host: localhost
            port: 4446
        - REST:
            url: http://localhost/myapp
            depends: PhpBrowser
        - \Helper\Acceptance
```

This is NOT working for SOAP :

``` yml
class_name: AcceptanceTester
modules:
    enabled:
        - WebDriver:
            browser: firefox
            url: http://localhost/myapp
            host: localhost
            port: 4446
        - SOAP:
            url: http://localhost/myapp
            endpoint: locahost
            depends: PhpBrowser
        - \Helper\Acceptance
```

Result will be
  [Codeception\Exception\ModuleConfigException]  
  PhpBrowser module is not configured!                              

  Options: url are required  
  Please, update the configuration and set all the required fields

Is there any other solution ? it nice and very fast to call API at the beginning of acceptance test to construct some data.

PR #3254, opened 2016-06-20
[SOAP] Add Parts to be more flexible like REST module

eXorus:
This PR fix the issue #3250

### DIFF
--- a/src/Codeception/Module/SOAP.php
+++ b/src/Codeception/Module/SOAP.php
@@ -3,7 +3,6 @@ namespace Codeception\Module;
 
 use Codeception\Lib\Interfaces\ConflictsWithModule;
 use Codeception\Lib\Interfaces\DependsOnModule;
-use Codeception\Lib\Interfaces\PartedModule;
 use Codeception\Module as CodeceptionModule;
 use Codeception\TestInterface;
 use Codeception\Exception\ModuleException;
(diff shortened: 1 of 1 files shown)


DavertMik:
It's actually not about parts but about SOAP module setting PhpBrowser config as REST module does. I think it can be implemented without adding PartedInterface and `@part` to annotation.

Currently you can overcome this issue with this config:

```yaml
class_name: AcceptanceTester
modules:
    enabled:
        - WebDriver:
            browser: firefox
            url: http://localhost/myapp
            host: localhost
            port: 4446
        - SOAP:
            url: http://localhost/myapp
            endpoint: locahost
            depends: PhpBrowser
        - \Helper\Acceptance
    config:
        PhpBrowser:
            url: http://localhost      
```

src/Codeception/Module/SOAP.php
```diff
@@ -118,12 +123,22 @@ public function _conflicts()
...
+
+    public function _inject(InnerBrowser $connection)
+    {
+        $this->connectionModule = $connection;
+        if ($this->connectionModule instanceof Framework) {
             $this->isFunctional = true;
         }
+        if ($this->connectionModule instanceof PhpBrowser) {
+            if (!$this->connectionModule->_getConfig('url')) {
+                $this->connectionModule->_setConfig(['url' => $this->config['url']]);
```
DavertMik: please leave only this, changing parts are not needed

eXorus:
Hello @DavertMik 

Without Part it's not working, I have :

```
[Codeception\Exception\ModuleConflictException]                                                      
  WebDriver module conflicts with SOAP                                                                 
  --                                                                                                   
  This usually happens when you enable two modules with the same actions but with different backends.  
  For instance, you can't run PhpBrowser, WebDriver, Laravel5 modules in one suite,                    
  as they implement similar methods but use different drivers to execute them.                         
  You can load a part of module (like: ORM) to avoid conflict.                                         
  Conflicts with method: grabAttributeFrom
```

PR #3255, opened 2016-06-20
Apply parameters in codeception config files

LeRondPoint:
The current code only allows for parameters to be used inside suite config files.  This PR allows for parameters to be used inside codeception.yml and codeception.dist.yml files as well.


DavertMik:
Thank you!

dardrone:
Where is the best place to see the documentation or implementation of this feature?

thinkspill:
I'm not sure if this covers the specific new features from this thread, but I think this is the section that would/does: 

http://codeception.com/docs/06-ModulesAndHelpers#Dynamic-Configuration-With-Params

Please correct me otherwise.

PR #3271, opened 2016-06-24
added module dependencies exceptions for Redis and Datafactory

DavertMik:


Naktibalda:
Related issue:
https://github.com/Codeception/Codeception/blob/2.2/tests/unit/Codeception/Module/RedisTest.php
should be markedAsSkipped if Predis\Client is not installed or failed to connect.

Naktibalda:
Thanks

DavertMik:
THanks for reviewing

PR #3283, opened 2016-06-27
Moved FB tests to separate suite

DavertMik:
And added to separate CI server

### DIFF
--- a/docs/modules/Facebook.md
+++ b/docs/modules/Facebook.md
@@ -10,9 +10,11 @@ To use this module with Composer you need <em>"facebook/php-sdk4": "5.*"</em> pa
 
 ## Status
 
-* Maintainer: **tiger-seo**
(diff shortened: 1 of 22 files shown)


docs/modules/Facebook.md
```diff
@@ -10,9 +10,11 @@ To use this module with Composer you need <em>"facebook/php-sdk4": "5.*"</em> pa
 
 ## Status
 
-* Maintainer: **tiger-seo**
+[ ![Facebook Status for Codeception/Codeception](https://codeship.com/projects/e4bc90d0-1ed5-0134-566c-1ed679ae6c9d/status?branch=2.2)](https://codeship.com/projects/160201)
+
 * Stability: **beta**
```
Naktibalda: How about downgrading it to **unstable**?

DavertMik: to me beta == unstable.

Also the instability of most of our tests are not related to module itself but to Facebook servers and to Facebook data.

Issue #3302, opened 2016-07-01
2.2.2 DataFactory loads factories multiply times


samusenkoiv:
I use Codeception with Symfony, as it is described here http://codeception.com/for/symfony.
Today I updated to 2.2.2 version and got such error

> [League\FactoryMuffin\Exceptions\DefinitionAlreadyDefinedException]
  The model definition 'AppBundle\Entity\MyEntity' has already been defined. 

I do not experience the same with 2.2.1. So I had to rollback. 
What could be the reason?

DavertMik:
There were some updates with DataFactory... Probably one of them made this regression.

samusenkoiv:
how can I see the diff?

jstewmc:
Looks like the issue is caused by line 166 in `Codeception\Module\DataFactory.php`. The `loadFactories()` method is called on both lines using the same path. I'd bet the author meant to delete line 166 after adding the path check on line 167. 

I commented out line 166, and so far, it works fine.

![problem](https://cloud.githubusercontent.com/assets/1192893/16589961/c7ac85da-4292-11e6-9be5-be3bb80dc9c3.PNG)
![solution](https://cloud.githubusercontent.com/assets/1192893/16589968/cd8ce008-4292-11e6-82b2-2387787f308b.PNG)

DavertMik:
Thanks for patch!

PR #3314, opened 2016-07-06
Data factory - fix loading factories twice

samusenkoiv:
Fixes this issue  https://github.com/Codeception/Codeception/issues/3302


DavertMik:
Thanks for fixing this regression. Looks like I failed to see it on code review :|

PR #3315, opened 2016-07-06
Fix checking "host" attribute in Yii2 Module / getInternalDomains()

TomaszKane:
Accessing $rule->host fails because \yii\rest\UrlRule not have "host" attribute.


samdark:
Merged. Thank you!

PR #3321, opened 2016-07-07
[doc] PhpBrowser doesn't support session

glensc:
cloned from https://github.com/Codeception/codeception.github.com/pull/108

doc here is incorrect:
http://codeception.com/docs/06-ReusingTestCode#Session-Snapshot

at least for current stable version (2.2.2)

    [RuntimeException] Call to undefined method AcceptanceTester::loadSessionSnapshot

confirmed by [StackOverFlow post](http://stackoverflow.com/q/33273378/2314626)

### DIFF
--- a/docs/06-ReusingTestCode.md
+++ b/docs/06-ReusingTestCode.md
@@ -131,7 +131,7 @@ Let's improve code of our `login` method by making it executed only once for the
 
 ```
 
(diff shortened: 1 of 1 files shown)


Naktibalda:
Do you mean that "PhpBrowser doesn't support session **restoration**"?

DavertMik:
Probably there is some problem with cloning and restoring Guzzle objects and sessions. I didn't have chance to try it recently so probably it doesn't work as expected. Maybe we will fix that in next releases but now let's update docs

glensc:
and if you add the functionality, update doc with what version is required to avoid the confusion

Issue #2936, opened 2016-03-28
Apigility support


artmnv:
Hello! In our work we are using Codeception with [Apigility](https://apigility.org/). It is Zend Framework 2 based application for building APIs. The default ZF2 Module doesn't work with Apigility because of some internal differences. We are implemented this module based on ZF2 module. Does you interested in Apigility support? Can we send PR without tests?

Thank you!

Naktibalda:
@artmnv Are you going to maintain it? (Fix bugs, ensure compatibility with new versions of Apigility and dependencies).

Creating some tests would be a good start.

artmnv:
@Naktibalda Thank you for your reply! I can't promise that I'll fix bugs very fast, but I want to try. I can't find tests for ZF2 Module. Are there any?

Maybe we don't need a new module. Can we do some settings to ZF2 module? Or maybe we can do it in REST module. There are three main differences between Apigility and ZF2 modules:

1. You need to json_encode() request content. `$zendRequest->setContent(json_encode($post));`
2. Add headers:
`$zendHeaders->addHeaders(array('Content-Type' => 'application/json'));
$zendHeaders->addHeaders(array('Accept' => 'application/json'));`
Yes we can add this header in every test or cest but I think it is not the best solution in this situation. It's not a setting which can be changed. It's not a coincidence, it's a rule. Requests may be json or xml only.
3. I didn't check but I think Apigility recreates response object so you need to get it again `$zendResponse = $this->application->getResponse();` after application run `$this->application->run();` This is the easiest difference. I think we can move this get in ZF2 module without any problems.

What do you think? What is the best way of doing this? Thank you!

Naktibalda:
First point is not necessary, because REST module automatically encodes request if Content-Type header is set to 'application/json'
Third point doesn't make any difference.

Second point can be achieved with helper methods:
```php
public function sendJsonPost($url, $params = [], $files = [])
{
    $restModule = $this->getModule('REST');
    $restModule->haveHttpHeader('Accept', 'application/json');
    $restModule->haveHttpHeader('Content-Type', 'application/json');
    $restModule->sendPost($url, $params, $files);
}
```

jstewmc:
I'm not sure if I'm supposed to comment on a closed issue or create a new one, but I can attest that @artmnv's third point is an issue. 

[...]

PR #3325, opened 2016-07-08
Update ZF2 module

jstewmc:
Update the `doRequest()` method in the ZF2 module's connector to grab the application's response _after_ the application has run. 

Before this fix, the `doRequest()` method grabbed the response _before_ the application was run. This resulted in an empty response when other ZF libraries like Api Agility replaced the application's response object with their own.

Fixes issue #2936.


Naktibalda:
Thanks

PR #3327, opened 2016-07-10
Support different line endings in dump

sustmi:
This adds support for all common end-of-line sequences.
Effectively it fixes problems with COPY command in SQL dump on Windows.

Splitting only by LF caused problems on Windows when SQL dump contained Windows line endings.
For example, `PostgreSql` driver expects lines to not contain any of end-of-line characters here: https://github.com/Codeception/Codeception/blob/2.2/src/Codeception/Lib/Driver/PostgreSql.php#L90
If a line ends with CR+LF then the `$sql` argument contains a string ending with CR and therefore it does never call `pg_end_copy` and `pg_close`.


Naktibalda:
Thanks

PR #3338, opened 2016-07-14
I->see improvements

dizzy7:
For example i have template:
```
<div>
        Comment on 
        <a href="{{ generatePostHref(entity) ~ commentHref }}">
            review about
            {{  entity.place.mainTypeName(entity.place)) }}
            {{ entity.place.name -}}
        </a>
</div>
```
Actually in browser i see "Comment on review about cafe Starbucks",
Selenium test $I->see('Comment on review about cafe Starbucks') is successful, but PhpBrowser expects something like "Comment on\n           review about            \ncafe             \nStarbucks" and failed.

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -461,13 +461,13 @@ class InnerBrowser extends Module implements Web, PageSourceSaver, ElementLocato
 
     public function seeCheckboxIsChecked($checkbox)
     {
-        $checkboxes = $this->getCrawler()->filter($checkbox);
+        $checkboxes = $this->getFieldsByLabelOrCss($checkbox);
         $this->assertDomContains($checkboxes->filter('input[checked=checked]'), 'checkbox');
     }
 
     public function dontSeeCheckboxIsChecked($checkbox)
     {
-        $checkboxes = $this->getCrawler()->filter($checkbox);
+        $checkboxes = $this->getFieldsByLabelOrCss($checkbox);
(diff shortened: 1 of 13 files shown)


Naktibalda:
This is not a bugfix, but a feature, and potentially a breaking one - so I think that it isn't a good idea to make it in 2.1 branch. It would be better to make it in 2.2.

Please add some tests to https://github.com/Codeception/Codeception/blob/2.2/tests/unit/Codeception/Module/TestsForWeb.php for see() and dontSee() with and without newlines.

Also please add some tests for see($text, $selector), because see with selector uses assertDomContains instead of assertPageContains.

src/Codeception/Lib/InnerBrowser.php
```diff
@@ -1626,4 +1626,18 @@ protected function setCookiesFromOptions()
             }
         }
     }
+
+    /**
+     * @return string
+     */
+    protected function getNormalizedResponceContent()
```
Naktibalda: Response, not Responce

Naktibalda:
@samdark Could you check why does the lowest dependency build [fail](https://travis-ci.org/Codeception/Codeception/jobs/144742379#L808) in 2.1 branch?

`php codecept build -c frameworks-yii-basic/tests`
```
PHP Fatal error:  Class 'Codeception\Specify\Config' not found in /home/travis/build/Codeception/Codeception/frameworks-yii-basic/tests/codeception/_bootstrap.php on line 23
```

The previous build 16 days ago was successful.

Issue #3340, opened 2016-07-15
amLoggedAs causes error in Lumen


a300600st:
```
[PHPUnit_Framework_Exception] call_user_func_array() expects parameter 1 to be a valid callback,
    class 'Illuminate\Auth\RequestGuard' does not have a method 'driver'
```
When I try to use amLoggedAs. Changing the function from
```
public function amLoggedAs($user, $driver = null)
{
    if ($user instanceof Authenticatable) {
        $this->app['auth']->driver($driver)->setUser($user);
    } else {
        $this->app['auth']->driver($driver)->attempt($user);
    }
}
```
to
```
public function amLoggedAs($user, $driver = null)
{
    if ($user instanceof Authenticatable) {
        $this->app['auth']->guard($driver)->setUser($user);
    } else {
        $this->app['auth']->guard($driver)->attempt($user);
    }
}
```
fixes the error but I don't have enough experience to know the full ramifications of this change.

Naktibalda:
It looks consistent with https://github.com/Codeception/Codeception/commit/192e3339fa259d53b0e7caa7a1afe5b2767b55fb change in Laravel5 module.

@a300600st Could you copy that change from Laravel5 to Lumen and make a pull request?

@janhenkgerritsen Do you support Lumen module?

a300600st:
I messed up the first pull request but I think the second one is correct.

janhenkgerritsen:
Thanks for the pull request! 

The Lumen module probably needs some more work for version 5.2, because there were some major changes between Lumen 5.1 and Lumen 5.2. I haven't had the time to do this though, hopefully soon.

a300600st:
Good to know. We just switched to codeception with lumen 5.2 so if I find other problems I'll see what I can do to help fix them.

janhenkgerritsen:
@a300600st I recently finished a large refactoring of the Lumen module, but it probably needs a bit more testing in real applications before I merge it into the Codeception 2.2 branch. Since I know you are using the Lumen module, would it be possible for you to test out the refactored module and see if you run into  any issues?

If so, take a look at [this comment](https://github.com/Codeception/Codeception/issues/3494#issuecomment-246591781) I made on another issue for more details.

a300600st:
@janhenkgerritsen We've since switched to using Laravel (Lumen was an experiment to see what speed improvements we'd see) so I no longer have a code base to test it with. Sorry about that.

janhenkgerritsen:
[...]

PR #3341, opened 2016-07-15
Update Lumen amLoggedAs to be compatible with 5.2

a300600st:
Addresses #3340


a300600st:
Oops I messed this up.

PR #3343, opened 2016-07-16
Yii module improvements

DavertMik:
* added `init` part to make module work for acceptance and unit tests
* server vars are now set by module from a config
* `entryScript` and `entryConfig` values added

TODO:

- [x] add methods to work with routes
- [x] add `grabService` method
- [x] add fixtures methods
- [x] add email assertion method
- [x] add quick login method


RoboFile.php
```diff
@@ -815,7 +815,7 @@ public function publishBase($branch = null, $tag = null)
 
         $this->taskComposerUpdate()->run();
         $this->taskGitStack()
-            ->add('composer*')
```
Naktibalda: This is not a yii improvement.

Naktibalda:
@DavertMik Could you use more meaningful commit messages than "Improved" and "improvements"?

DavertMik:
> Could you use more meaningful commit messages than "Improved" and "improvements"?

I wrote the wider description in Pull Request. Writing this inside a terminal was not that convenient, actually

RoboFile.php
DavertMik: yes, I mentioned it. I will try to cherry pick that and move to another PR

DavertMik:
Once I finish all issues I can rebase and add this description as commit message

src/Codeception/Module/Yii2.php
```diff
@@ -89,22 +150,137 @@ public function _after(\Codeception\TestInterface $test)
...
+    {
+        return call_user_func_array('array_merge',
+            array_map(function($fixturesStore) {
+                return $fixturesStore->getFixtures();
+            })
+        );
+    }
+
+    /**
+     * Gets a fixtures by name.
```
samdark: fixtures → fixture

src/Codeception/Module/Yii2.php
```diff
@@ -442,6 +442,62 @@ public function grabComponent($component)
...
+    /**
+     * Checks that no email was sent
+     */
+    public function dontSeeEmailIsSent()
+    {
+        $this->seeEmailIsSent(0);
+    }
+
+    /**
+     * Returns array fo all sent email messages.
```
samdark: fo → of

src/Codeception/Lib/Connector/Yii2/Logger.php
```diff
@@ -0,0 +1,26 @@
+<?php
+namespace Codeception\Lib\Connector\Yii2;
+
+use Codeception\Util\Debug;
+
+class Logger extends \yii\log\Logger
+{
+    public function init()
+    {
```
samdark: I'd add comment that empty method is intentional i.e.:

```
// prevent shutdown function registration
```

src/Codeception/Lib/Connector/Yii2/Logger.php
```diff
@@ -0,0 +1,26 @@
...
+    public function log($message, $level, $category = 'application')
+    {
+        if (!in_array($level, [
+            \yii\log\Logger::LEVEL_INFO,
+            \yii\log\Logger::LEVEL_WARNING,
+            \yii\log\Logger::LEVEL_ERROR,
+        ])) {
+            return;
+        }
+        if (strpos($category, 'yii\db\Command')===0) {
```
samdark: Could be done via config (`except`) property:

http://www.yiiframework.com/doc-2.0/guide-runtime-logging.html#message-filtering

PR #3346, opened 2016-07-19
Using valid example of expectException()

igorsantos07:
Also, using `Exception::class` is generic enough for any sort of Exception implementation, and lighter than actually instantiating an exception just to get its class.

### DIFF
--- a/docs/05-UnitTests.md
+++ b/docs/05-UnitTests.md
@@ -336,7 +336,7 @@ To check your code for exception you can use `expectException` method from `Asse
 
 ```php
 <?php
-$t->expectException(new Exception, {
+$t->expectException(Exception::class, function() {
    throw new Exception; 
 });
 ```


DavertMik:
I think both examples are equivalent. Depending on what you imported as `Exception`

igorsantos07:
The biggest point here is actually the said "callback". That syntax with sole curly braces is invalid :)

Naktibalda:
`Exception::class` does not work on PHP 5.4

igorsantos07:
@Naktibalda so what about using `expectException('Exception', $callback)` instead? I still feel it's bad so _suggest_ people to instantiate a class for nothing but getting its name. What if the needed exception has some required arguments? Someone may get helpless thinking it needs an actual instance of the exception for that.

DavertMik:
>  I still feel it's bad so suggest people to instantiate a class for nothing but getting its name.

If you want to test that exception with the exact code and exact message, you should instantiate it. At least this is how it works now. An example demonstrates that this is possible to pass exception instance. Reference for this module provides all this variations: http://codeception.com/docs/modules/Asserts#expectException

But I don't mind merging this. Sure it doesn't work for PHP 5.4 but it is only a guide, so anyone using this method can see its reference and choose the way they like to use it.

igorsantos07:
Whoa! I didn't even think about looking for the actual class reference. As the guide gave no further insights, I thought that was the "canonical" way of doing stuff. That's why I got worried with the weird syntax. And no, I didn't know it was possible to compare the actual code and message that way :)

Can I send another PR including links for the referenced classes and methods, when appropriate? This way people will know there's further information down the rabbit hole, if needed.

PR #3350, opened 2016-07-20
I->see improvements

dizzy7:
replaces #3338

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -1469,7 +1469,7 @@ class InnerBrowser extends Module implements Web, PageSourceSaver, ElementLocato
     {
         $constraint = new PageConstraint($needle, $this->_getCurrentUri());
         $this->assertThat(
-            html_entity_decode(strip_tags($this->_getResponseContent()), ENT_QUOTES),
+            $this->getNormalizedResponseContent(),
             $constraint,
             $message
         );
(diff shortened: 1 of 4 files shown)


Naktibalda:
In the previous ticket I asked for some tests to ensure backwards compatibility -
see() assertions containing newlines and multiple spaces should still pass.

```php
$I->see("Comment on\n review about \ncafe \nStarbucks");
$I->see('Comment on review about cafe Starbucks') 
```
```php
$I->see('Some text with        formatting on separate lines');
$I->see('Some text with formatting on separate lines');
```

DavertMik:
I don't know how to make this works without breaking BC. @Naktibalda are we ok with this?

DavertMik:
There is an edge case for this. What if I want to test the content of `pre` tag. I should probably treat it as it is without the normalization.

samdark:
@DavertMik is absolutely right about the fact that it's both BC breaking and we're losing an ability to check any formatting-sensitive output. It could be useful but as a separate method.

samdark:
No idea how to name it though.

DavertMik:
If we still think on the implementation probably should be 2 behaviors: 

* current
* normalized

Current should be left as it is, and if someone needs normalized check they can pass data in special way. I can propose this syntax:

```php
$I->see(['Some text with formatting on separate lines']);
```

Wrapping a string in array makes assertion to be performed on normalized string. What to you think? 
Any other idea? 

Also as it is important to leave current behavior - please revert your change to Page Constraint. If you need to use it for normalized output - just extend this class.

samdark:
I don't like this syntax. It's rather implicit. I'm either for separate method like `$I->seeWithoutSpaces()` or a meaningful wrapper:

```php
$I->see(nospaces('Some text with formatting on separate lines'));
$I->see(new NoSpaces('Some text with formatting on separate lines'));
```

Naktibalda:
A simple way to maintain backwards compatibility is to normalize $needle too.
see() will pass if spaces and line breaks are removed from both texts.

On the other hand, see() backwards compativility was broken just 9 month ago in 2.1.3 by https://github.com/Codeception/Codeception/commit/474b0959fb935392b567b5f9d06d350b390d8d8a
so breaking it again isn't such a big deal.

DavertMik:
Closed in favor of #3439

PR #3355, opened 2016-07-21
FIX: PHP Fatal error:  Call to undefined function Codeception\Module\…

kusnir:
FIX: PHP Fatal error:  Call to undefined function Codeception\Module\mb_strcut() in phar:///usr/bin/codecept/src/Codeception/Module/WebDriver.php on line 370


Naktibalda:
It isn't necessary to namespace functions.
http://php.net/manual/en/language.namespaces.fallback.php

> For functions and constants, PHP will fall back to global functions or constants if a namespaced function or constant does not exist.


It looks like multibyte string extension is missing in your PHP.

kusnir:
jop, sorry, you are right. I thought that it's already installed on my server so that the PHP take it as not language construct.

PR #3365, opened 2016-07-23
fixed time calculation. Fixes #3331

### LINKED ISSUE #3331: Timing in 2.2.2 seems to be wrong
The seconds taken on each test seem to be wrong. Codeception seems to think that 0.7 < 0.14, so I think perhaps sometimes 0.7 should be 0.07.

Here is an example:

*Before* I had some methods in my Cest's `_before` which were unnecessary. The tests were performing like this prior to the cleanup (names have been removed):

```bash
SuiteName Tests (5) --------------------------------------------------------------------------
✔ MyCest: Test 1 (0.17s)
✔ MyCest: Test 2 (0.14s)
✔ MyCest: Test 3 (0.13s)
✔ MyCest: Test 4 (0.13s)
✔ MyCest: Test 5 (0.13s)
---------------------------------------------------------------------------------------

Time: 945 ms, Memory: 20.00MB
```

*After* I cleaned up the `_before` method, the results look like this:

```bash
SuiteName Tests (5) --------------------------------------------------------------------------
✔ MyCest: Test 1 (0.10s)
✔ MyCest: Test 2 (0.7s)
✔ MyCest: Test 3 (0.7s)
✔ MyCest: Test 4 (0.7s)
✔ MyCest: Test 5 (0.7s)
---------------------------------------------------------------------------------------

Time: 644 ms, Memory: 20.00MB
```

The times vary of course, but on average a 300 ms improvement I would say.

The times for the individual tests are all wrong though. One would think that the tests had gotten faster, but no, they have gotten much slower! :) Codeception seems to think that 0.13 < 0.7, which is very weird to me ;)

Possibly related to #3262.

DavertMik:


src/Codeception/Subscriber/Console.php
```diff
@@ -610,11 +610,8 @@ protected function writeTimeInformation(TestEvent $event)
     {
         $time = $event->getTime();
         if ($time) {
-            $seconds = (int) ($milliseconds = (int) ($time * 100)) / 100;
-            $time = ($seconds % 60) . '.' . $milliseconds;
-
             $this
-                ->message($time)
+                ->message(number_format(round($time, 2),2))
```
Naktibalda: I think that merging should be forbidden if Nitpick founds any issues.

PR #3366, opened 2016-07-23
Specify package requirements for module

DavertMik:
Each module implementing `RequiresPackage` can now specify its external dependencies by implementing `_requires` method. If required class can't be found and exception is thrown with proposal to add corresponding package to composer.json.

For instance, PHPBrowser module has its requirements specified
```php
    public function _requires()
    {
        return ['GuzzleHttp\Client' => '"guzzlehttp/guzzle": ">=4.1.4 <7.0"'];
    }
```
If `codeception/base` is installed you will get the the exception which suggests to install '"guzzlehttp/guzzle": ">=4.1.4 <7.0"'


Naktibalda:
I like this change

PR #3369, opened 2016-07-24
Maintenance (CodeStyle fix, Release Script updates, ...)

DavertMik:
* fixed codestyle
* updated codeception/base release script
* updated changelog
* version bump


package/stub.php
```diff
@@ -6,4 +6,4 @@
 
 require_once 'phar://codecept.phar/codecept';
 
-__HALT_COMPILER();
+__halt_compiler();
```
DavertMik: Shit, why does this change happens? This breaks phar compilation!

PR #3381, opened 2016-07-27
Yii2 `@webroot` fix

DavertMik:
Fixes creating of `@webroot` directory on testing.


src/Codeception/Lib/Connector/Yii2.php
```diff
@@ -252,6 +251,6 @@ protected function persistDb()
 
     private function mockAssetManager()
     {
-        $this->app->set('assetManager', Stub::make('yii\web\AssetManager'));
+        $this->app->set('assetManager', new Maybe());
```
dmitry-kulikov: Alternative fix
```
$this->app->set('assetManager', Stub::make('yii\web\AssetManager', ['bundles' => false]));
```
then assetManager will consider all bundles as dummy and they won't be published. However developers can directly use
```
Yii::$app->assetManager->publish($path, $options);
```
and requested file or directory will be published and published incorrectly.
I can't say how widely direct call of `publish` is used, I never used it.

DavertMik: I think you approach is better because Maybe class is pretty flexible, however once anywhere an application `AssetManager`'s class or interface will be checked a test will fail.

DavertMik:
Closed in favor of #3387

Issue #2509, opened 2015-11-01
REST amDigestAuthenticated doesn't work with framework modules


Naktibalda:
REST amDigestAuthenticated method doesn't work with framework modules.

https://github.com/Codeception/Codeception/blob/6f30eeb1032cff54ffcf99a87b01d8d6c839450b/src/Codeception/Module/REST.php#L263

``` php
public function amDigestAuthenticated($username, $password)
{
    $this->client->setAuth($username, $password, CURLAUTH_DIGEST);
}
```

Only Guzzle connectors have `setAuth` method.

doctorx:
```diff
Index: src/Codeception/Lib/Connector/Guzzle.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- src/Codeception/Lib/Connector/Guzzle.php	(revision 2e51a84355f69b8b091abdac497a21a2ebc5621c)
+++ src/Codeception/Lib/Connector/Guzzle.php	(revision f057e640435466d64a929faa69a4a9cafb8a4da5)
@@ -79,13 +79,18 @@
         unset($this->requestOptions['headers'][$name]);
     }
 
-    public function setAuth($username, $password)
+    /**
+     * @param string $username
+     * @param string $password
+     * @param string $type  Default: 'basic'
+     */
+    public function setAuth($username, $password, $type='basic')
     {
         if (!$username) {
             unset($this->requestOptions['auth']);
             return;
         }
-        $this->requestOptions['auth'] = [$username, $password];
+        $this->requestOptions['auth'] = [$username, $password, $type];
     }
 
     /**
Index: src/Codeception/Lib/Connector/Guzzle6.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- src/Codeception/Lib/Connector/Guzzle6.php	(revision 2e51a84355f69b8b091abdac497a21a2ebc5621c)
+++ src/Codeception/Lib/Connector/Guzzle6.php	(revision f057e640435466d64a929faa69a4a9cafb8a4da5)
@@ -82,13 +82,18 @@
         unset($this->requestOptions['headers'][$name]);
     }
 
-    public function setAuth($username, $password)
+    /**
+     * @param string $username
+     * @param string $password
+     * @param string $type  Default: 'basic'
+     */
+    public function setAuth($username, $password, $type='basic')
     {
[...]

PR #3383, opened 2016-07-28
Fix REST.php amDigestAuthenticated and Guzzle (#2509)

doctorx:
Fix method setAuth in Guzzle Connector


Naktibalda:
@doctorx Please make pull request to 2.2 branch instead
and fix formatting of $type argument.

PR #3394, opened 2016-08-01
exactly callback receives $params

antonmarin:


DavertMik:
I think it could better to add an example where `$params` are actually used. How do you think?

PR #3396, opened 2016-08-01
[SOAP] Add flexibility with the config SOAPAction

### LINKED ISSUE #2201: Having trouble with SOAP request and Microsofts namespace http://tempuri.org
If you are using the SOAP module with sendSOAPRequest it would be nice if it knew how to handle a SOAP request to a .Net WebService. The action needs to look like `SOAPAction: http://tempuri.org/action` in the Request Header but in the tag it should look like `<action xmlns="http://tempuri.org">`. I'm thinking that there so many .Net webservices it would help people a lot. I think the only alternative is to use the PHP SOAP client directly.

eXorus:
Hello,

I'm having trouble with Tomcat API (CXF Framework) that doesn't accept SOAPAction in HTTP Header because SOAPAction is deprecated from SOAP 1.2.
It will also fix #2201 maybe

So actual user will continue without config SOAPAction and the SOAPAction header will be send with the action value
Other people like me that want to do SOAP 1.2 will put '' in SOAPAction config.

### DIFF
--- a/src/Codeception/Module/SOAP.php
+++ b/src/Codeception/Module/SOAP.php
@@ -36,6 +36,7 @@ use Codeception\Lib\Interfaces\API;
  * ## Configuration
  *
  * * endpoint *required* - soap wsdl endpoint
+ * * SOAPAction - replace header SOAPAction (Set to '' to SOAP 1.2)
  *
  * ## Public Properties
  *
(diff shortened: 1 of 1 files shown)


Naktibalda:
@DavertMik It does no harm, so it can be merged, right?

DavertMik:
Yes, thanks! :)

PR #3402, opened 2016-08-03
added link to internal tests manual

DavertMik:


DavertMik:
Should something else be added here?

P.S. Don't look into wrecker CI, I'm experimenting with it to move HHVM builds to it

PR #3412, opened 2016-08-05
Fix DataFactory::haveMultiple argument order

buffcode:
Correct the order of the FactoryMuffin::seed call which is `$times, $name, ...` instead of `$name, $times, ...`.

1. [DataFactory::haveMultiple call](https://github.com/Codeception/Codeception/blob/master/src/Codeception/Module/DataFactory.php#L245)
2. [FactoryMuffin::seed function](https://github.com/thephpleague/factory-muffin/blob/master/src/FactoryMuffin.php#L82)


Naktibalda:

buffcode:
see #3413

PR #3413, opened 2016-08-05
Fix DataFactory::haveMultiple argument order

buffcode:
Correct the order of the FactoryMuffin::seed call which is `$times, $name, ...` instead of `$name, $times, ...`.

1. [DataFactory::haveMultiple call](https://github.com/Codeception/Codeception/blob/2.2/src/Codeception/Module/DataFactory.php#L254)
2. [FactoryMuffin::seed function](https://github.com/thephpleague/factory-muffin/tree/v3.0.1/src/FactoryMuffin.php#L82)


Naktibalda:
Great, thank you.

PR #3428, opened 2016-08-10
Refactor ModuleContainer

janhenkgerritsen:


src/Codeception/Lib/ModuleContainer.php
```diff
@@ -246,18 +429,79 @@ protected function getModuleConfig($module)
...
+
+        $conflicts = $this->normalizeConflictSpecification($module->_conflicts());
+        if ($otherModule instanceof $conflicts) {
+            throw new ModuleConflictException($module, $otherModule);
+        }
+    }
+
+    /**
+     * Normalize the return value of ConflictsWithModule::_conflicts() to a class name.
+     * This is necessary because it can return a module name instead of the name of a class or interface.
```
DavertMik: Does it? I think only class name or interface was supposed to. I don't think we have conflicts in core other than interface conflicts

DavertMik:
Thank you. This is so clean now!

src/Codeception/Lib/ModuleContainer.php
janhenkgerritsen: I added it because of this test: https://github.com/Codeception/Codeception/blob/2.2/tests/unit/Codeception/Lib/ModuleContainerTest.php#L168

PR #3429, opened 2016-08-10
Laravel5 fix issue with uploaded files

janhenkgerritsen:
See #3417.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+#### 2.2.4
+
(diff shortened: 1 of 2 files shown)


janhenkgerritsen:
The `Illuminate\Http\UploadedFile` class was added in Laravel 5.2.15, so the implementation in this PR will cause a fatal error if it is used with an earlier version of Laravel. This problem was fixed in https://github.com/Codeception/Codeception/pull/3436/commits/c764ccfccbfb370f74fbfd0c55a4cb28e94eec4b.

PR #3430, opened 2016-08-10
FIX 3427 Support of standard mongodump/mongorestore tools to populate…

GSokol:
… MongoDB database


DavertMik:
Yes, this is effective approach but how general it is? Can this be executed on Windows? I see there is external dependencies to tools like mongotstore and tgz. I'd rather recommend to release this a standalone module; If someone needs better dump loader they will find it on Addons page of a site and install it.

GSokol:
@DavertMik mongorestore is common tool, provided with mongodb (like mongo tool), so if mongo tool is already used, mongorestore tool could be used too. About using tar -- yes, that's *nix feature, but there is a check, so if that's it, the behavior will be like in case of empty script.

Naktibalda:
mongorestore is supported on Travis.
I see no reason why `loadFromMongoDump` could have issues with Windows.

However I think that it would be a lot more in line with Codeception way of database testing, if loading was implemented in [MongoDb::loadDump](https://github.com/Codeception/Codeception/blob/2.2.3/src/Codeception/Module/MongoDb.php#L146) and [MongoDb Driver](https://github.com/Codeception/Codeception/blob/2.2/src/Codeception/Lib/Driver/MongoDb.php).

A new option should be added to MongoDb module to choose the method of loading (default to the old way).

Also it would be good if you could add some tests for this kind of loading to [MongoDbTest](https://github.com/Codeception/Codeception/blob/2.2.3/tests/unit/Codeception/Module/MongoDbTest.php)

GSokol:
@Naktibalda done!

docs/modules/MongoDb.md
```diff
@@ -30,6 +31,8 @@ Check out the driver if you get problems loading dumps and cleaning databases.
 * dsn *required* - MongoDb DSN with the db name specified at the end of the host after slash
 * user *required* - user to access database
 * password *required* - password
+* dump_type *required* - type of dump.
```
Naktibalda: It must be optional

docs/modules/MongoDb.md
```diff
@@ -30,6 +31,8 @@ Check out the driver if you get problems loading dumps and cleaning databases.
...
+  One of MongoDb::DUMP_TYPE_JS ('js'), MongoDb::DUMP_TYPE_MONGODUMP ('mongodump') or MongoDb::DUMP_TYPE_MONGODUMP_TAR_GZ ('tar-gz'). Default value is MongoDb::DUMP_TYPE_JS.
```
Naktibalda: Constants can't be used in configuration, so it is better just to list the string values here, the last value should probably be a 'mongodump-tar-gz'

docs/modules/MongoDb.md
```diff
@@ -8,6 +8,7 @@ To have your database properly cleaned you should configure it to access the dat
 
 In order to have your database populated with data you need a valid js file with data (of the same style which can be fed up to mongo binary)
 File can be generated by RockMongo export command
+You can also use directoy, generated by ```mongodump``` tool or it's ```.tar.gz``` archive (not available for Windows systems), generated by ```tar -czf <archve_file_name>.tar.gz <path_to dump directory>```.
```
Naktibalda: Actually you done have to update docs, they are generated from the code comments automatically on release.

src/Codeception/Module/MongoDb.php
```diff
@@ -39,13 +39,20 @@
  * * dsn *required* - MongoDb DSN with the db name specified at the end of the host after slash
  * * user *required* - user to access database
  * * password *required* - password
+ * * dump_type *required* - type of dump
+ *   (one of MongoDb::DUMP_TYPE_JS, MongoDb::DUMP_TYPE_MONGODUMP, MongoDb::DUMP_TYPE_MONGODUMP_TAR_GZ,
```
Naktibalda: Constants can't be used in configuration, so it is better just to list the string values here, the last value should probably be a 'mongodump-tar-gz'

Naktibalda:
Please fix everything what Nitpick complained about

src/Codeception/Module/MongoDb.php
GSokol: @Naktibalda So how to let the one to use defined constants, not "magic strings"?

GSokol:
@Naktibalda done

GSokol:
@DavertMik @Naktibalda rebuild of this pull request: https://github.com/Codeception/Codeception/pull/3437

Issue #3433, opened 2016-08-11
output of `codecept_relative_path()` incorrect when `codeception.yml` not in project root


loren-osborn:
#### What are you trying to achieve?

I'm trying to read the filenames displayed in the stack trace when a test throws an exception, even when `codeception.yml` is `include:`ed from a subdirectory relative to the primary `codeception.yml`.
#### What do you get instead?

If the file path is not a subdirectory of the included directory, the beginning of the file path will be omitted. This filename is processed through `codecept_relative_path()` which does not account for this edge case.
### Details

``` bash
...
Time: 27.8 seconds, Memory: 29.50MB

There was 1 failure:

---------
1) CheckoutTestingScenarioCept: Log in as regular user
 Test  tests/acceptance/CheckoutTestingScenarioCept.php
 Step  Fail "Web server threw Exception"
 Fail  Web server threw Exception

Scenario Steps:

 6. $I->fail("Web server threw Exception") at xtensions/CatchExceptionOutput.php:155
 5. $I->click({"css":"#login-form button[type=submit]"}) at tests/acceptance/_steps/UserSteps.php:18
 4. $I->fillField("#LoginForm_password","*******") at tests/acceptance/_steps/UserSteps.php:17
 3. $I->fillField("#LoginForm_username","*******") at tests/acceptance/_steps/UserSteps.php:16
 2. $I->canSeeElement("#login-form") at tests/acceptance/_steps/UserSteps.php:15
 1. $I->amOnPage("/site/login") at tests/acceptance/_steps/UserSteps.php:14

...
```

In this case, I expect line 6. of the stack trace to read:

``` bash
 6. $I->fail("Web server threw Exception") at ../_custom/Extensions/CatchExceptionOutput.php:155
```

...or possibly have all paths relative to `proj/` instead of `proj/tests/frontend/`.
- Codeception version: 2.2.3
- PHP Version: PHP 5.5.9-1ubuntu4.17 (cli)
- Operating System: Ubuntu 14.04.4 LTS x86_64 (within VirtualBox/Vagrant/Windows 10)
- Installation type: Composer via Makefile
- Suite configuration:

`proj/codeception.yml`:

``` yml
include:
  - tests/frontend
  - tests/admin
actor: Tester
paths:
    tests: tests
    log: tests/_output
    data: tests/_data
    support: tests/_support
    envs: tests/_envs
settings:
    bootstrap: _bootstrap.php
    colors: true
    memory_limit: 512M
extensions:
[...]

PR #3434, opened 2016-08-11
Fixed bad base path edge case

loren-osborn:
This is a fix to the issue I brought up in #3433 . 

While the code is a bit rough around the edges, it appears to conform to project guidelines, and resolves the issue. Any suggestions to make the code more elegant are welcome.

### DIFF
--- a/src/Codeception/Configuration.php
+++ b/src/Codeception/Configuration.php
@@ -500,6 +500,126 @@ class Configuration
     }
 
     /**
+     * Returns path to a given directory relative to $projDir.
+     * @param string $path
+     * @param string $projDir
+     * @param string $dirSep
+     * @return string
+     */
+    public static function getRelativeDir($path, $projDir, $dirSep = DIRECTORY_SEPARATOR)
+    {
+        // ensure $projDir ends with a trailing $dirSep
+        $projDir = preg_replace('/'.preg_quote($dirSep, '/').'*$/', $dirSep, $projDir);
+        // if $path is a within $projDir
+        if (self::fsCaseStrCmp(substr($path, 0, strlen($projDir)), $projDir, $dirSep) == 0) {
+            // simply chop it off the front
+            return substr($path, strlen($projDir));
+        }
+        // Identify any absoluteness prefix (like '/' in Unix or "C:\\" in Windows)
+        $pathAbsPrefix = self::getPathAbsolutenessPrefix($path, $dirSep);
+        $projDirAbsPrefix = self::getPathAbsolutenessPrefix($projDir, $dirSep);
+        $sameAbsoluteness = (self::fsCaseStrCmp($pathAbsPrefix['wholePrefix'], $projDirAbsPrefix['wholePrefix'], $dirSep) == 0);
+        if (!$sameAbsoluteness) {
+            // if the $projDir and $path aren't relative to the same
+            // thing, we can't make a relative path.
+
+            // if we're relative to the same device ...
+            if (strlen($pathAbsPrefix['devicePrefix']) &&
+                (self::fsCaseStrCmp($pathAbsPrefix['devicePrefix'], $projDirAbsPrefix['devicePrefix'], $dirSep) == 0)
+            ) {
+                // ... shave that off
+                return substr($path, strlen($pathAbsPrefix['devicePrefix']));
+            }
+            // Return the input unaltered
+            return $path;
+        }
+        // peel off optional absoluteness prefixes and convert
+        // $path and $projDir to an subdirectory path array
+        $relPathParts = array_filter(explode($dirSep, substr($path, strlen($pathAbsPrefix['wholePrefix']))), 'strlen');
+        $relProjDirParts = array_filter(explode($dirSep, substr($projDir, strlen($projDirAbsPrefix['wholePrefix']))), 'strlen');
+        // While there are any, peel off any common parent directories
+        // from the beginning of the $projDir and $path
+        while ((count($relPathParts) > 0) && (count($relProjDirParts) > 0) &&
+            (self::fsCaseStrCmp($relPathParts[0], $relProjDirParts[0], $dirSep) == 0)
+        ) {
+            array_shift($relPathParts);
+            array_shift($relProjDirParts);
+        }
+        if (count($relProjDirParts) > 0) {
+            // prefix $relPath with '..' for all remaining unmatched $projDir
+            // subdirectories
+            $relPathParts = array_merge(array_fill(0, count($relProjDirParts), '..'), $relPathParts);
+        }
+        // only append a trailing seperator if one is already present
+        $trailingSep = preg_match('/'.preg_quote($dirSep, '/').'$/', $path) ? $dirSep : '';
+        // convert array of dir paths back into a string path
+        return implode($dirSep, $relPathParts).$trailingSep;
+    }
+
+    /**
+     * Are we in a Windows style filesystem?
+     *
+     * @param string $dirSep
+     * @return bool
+     */
+    private static function isWindows($dirSep = DIRECTORY_SEPARATOR)
+    {
+        return ($dirSep == '\\');
+    }
+
+    /**
+     * FileSystem Case String Compare
+     * compare two strings with the filesystem's case-sensitiveness
+     *
+     * @param string $str1
+     * @param string $str2
+     * @param string $dirSep
+     * @return int -1 / 0 / 1 for < / = / > respectively
+     */
+    private static function fsCaseStrCmp($str1, $str2, $dirSep = DIRECTORY_SEPARATOR)
+    {
+        $cmpFn = self::isWindows($dirSep) ? 'strcasecmp' : 'strcmp';
+        return $cmpFn($str1, $str2);
+    }
+
+    /**
+     * What part of this path (leftmost 0-3 characters) what
+     * it is absolute relative to:
+     *
+     * On Unix:
+     *     This is simply '/' for an absolute path or
+     *     '' for a relative path
+     *
+     * On Windows this is more complicated:
+     *     If the first two characters are a letter followed
+     *         by a ':', this indicates that the path is
+     *         on a specific device.
+     *     With or without a device specified, a path MAY
+     *         start with a '\\' to indicate an absolute path
+     *         on the device or '' to indicate a path relative
+     *         to the device's CWD
+     *
+     * @param string $path
+     * @param string $dirSep
+     * @return string
+     */
+    private static function getPathAbsolutenessPrefix($path, $dirSep = DIRECTORY_SEPARATOR)
+    {
+        $devLetterPrefixPattern = '';
+        if (self::isWindows($dirSep)) {
+            $devLetterPrefixPattern = '([A-Za-z]:|)';
+        }
+        $matches = [];
+        if (!preg_match('/^'.$devLetterPrefixPattern.preg_quote($dirSep, '/').'?/', $path, $matches)) {
+            // This should match, even if it matches 0 characters
+            throw new ConfigurationException("INTERNAL ERROR: This must be a regex problem.");
+        }
+        return [
+            'wholePrefix'  => $matches[0], // The optional device letter followed by the optional $dirSep
+            'devicePrefix' => self::isWindows($dirSep) ? $matches[1] : ''];
+    }
+
+    /**
      * Returns path to tests directory
      *
      * @return string
--- a/tests/unit/Codeception/ConfigurationTest.php
+++ b/tests/unit/Codeception/ConfigurationTest.php
@@ -67,4 +67,83 @@ class ConfigurationTest extends \PHPUnit_Framework_TestCase
         $commandsConfig = $defaultConfig['extensions'];
         $this->assertArrayHasKey('commands', $commandsConfig);
     }
+
+    /**
+     * data provider for testGetRelativeDir
+     * 
+     * @return array(array(strings))
+     */
+    public function getRelativeDirTestData()
+    {
+        return [
+            // Unix style paths:
+                // projectDir() with & without trailing directory seperator: actual subdir
+                    ['/my/proj/path/some/file/in/my/proj.txt',           '/my/proj/path/',               '/',  'some/file/in/my/proj.txt'],
+                    ['/my/proj/path/some/file/in/my/proj.txt',           '/my/proj/path',                '/',  'some/file/in/my/proj.txt'],
+                    ['/my/proj/pathsome/file/in/my/proj.txt',            '/my/proj/path',                '/',  '../pathsome/file/in/my/proj.txt'],
+                // Case sensitive:
+                    ['/my/proj/Path/some/file/in/my/proj.txt',           '/my/proj/path/',               '/',  '../Path/some/file/in/my/proj.txt'],
+                    ['/my/Proj/path/some/file/in/my/proj.txt',           '/my/proj/path',                '/',  '../../Proj/path/some/file/in/my/proj.txt'],
+                    ['My/proj/path/some/file/in/my/proj.txt',            'my/proj/path/foo/bar',         '/',  '../../../../../My/proj/path/some/file/in/my/proj.txt'],
+                    ['/my/proj/path/some/file/in/my/proj.txt',           '/my/proj/Path/foobar/',        '/',  '../../path/some/file/in/my/proj.txt'],
+                    ['/my/PROJ/path/some/dir/in/my/proj/',               '/my/proj/path/foobar/',        '/',  '../../../PROJ/path/some/dir/in/my/proj/'],
+                // Absolute $path, Relative projectDir()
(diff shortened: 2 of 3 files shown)


src/Codeception/Configuration.php
```diff
@@ -500,6 +500,58 @@ public static function projectDir()
...
+                $sameAbsoluteness =
+                    (($pathPrefix[$rootPathIndex] == DIRECTORY_SEPARATOR) ==
+                     ($path[$rootPathIndex]       == DIRECTORY_SEPARATOR));
+                $rootPathIndex += ($pathPrefix[$rootPathIndex] == DIRECTORY_SEPARATOR) ? 1 : 0;
+            }
+            if ($sameAbsoluteness) {
+                $relPath = substr($path, $rootPathIndex);
+                $pathPrefix = substr($pathPrefix, $rootPathIndex);
+                $basePathArr = explode(DIRECTORY_SEPARATOR, $pathPrefix);
+                while (
```
loren-osborn: ... because of overly complex while condition. Please make code simplification suggestions. (@Nitpick-CI, Not you personally, I know you're just an automated tool.)

src/Codeception/Configuration.php
```diff
@@ -500,6 +500,58 @@ public static function projectDir()
...
     /**
+     * Returns path to a given directory relative to your project.
+     * @param string $path
+     * @return string
+     */
+    public static function getRelativeDir($path)
+    {
+        $relPath = $path;
+        if (substr($path, 0, strlen(self::projectDir())) == self::projectDir()) {
+            $relPath = substr($path, strlen(self::projectDir()));
```
DavertMik: don't do `else`, just return it here if you already matched it

src/Codeception/Configuration.php
```diff
@@ -500,6 +500,58 @@ public static function projectDir()
...
+        } else {
+            $pathPrefix = self::projectDir();
+            $sameAbsoluteness = true;
+            $rootPathIndex = 0;
+            if (DIRECTORY_SEPARATOR == '\\') {
+                $hasDriveSpecifier = preg_match('/^[A-Za-z]:/', $pathPrefix);
```
DavertMik: please add some comments here on why regex is using and what is matched

src/Codeception/Configuration.php
```diff
@@ -500,6 +500,58 @@ public static function projectDir()
...
+                $sameAbsoluteness = (preg_match('/^[A-Za-z]:/', $pathPrefix) == preg_match('/^[A-Za-z]:/', $path));
+                $rootPathIndex = $hasDriveSpecifier ? 2 : 0;
+                if ($sameAbsoluteness && $hasDriveSpecifier && preg_match('/^[A-Za-z]:/', $pathPrefix)) {
+                    $sameAbsoluteness = ($pathPrefix[0] == $path[0]);
+                }
+            }
+            if ($sameAbsoluteness) {
```
DavertMik: add some comments here too... I can't understand the logic :|

src/Codeception/Configuration.php
```diff
@@ -500,6 +500,58 @@ public static function projectDir()
...
+                $sameAbsoluteness =
+                    (($pathPrefix[$rootPathIndex] == DIRECTORY_SEPARATOR) ==
+                     ($path[$rootPathIndex]       == DIRECTORY_SEPARATOR));
+                $rootPathIndex += ($pathPrefix[$rootPathIndex] == DIRECTORY_SEPARATOR) ? 1 : 0;
+            }
+            if ($sameAbsoluteness) {
```
DavertMik: comments! what are you trying to achieve here too

src/Codeception/Configuration.php
```diff
@@ -500,6 +500,58 @@ public static function projectDir()
...
+                    $relPath = substr($relPath, strlen($basePathArr[0])+1);
+                    array_shift($basePathArr);
+                }
+                while (count($basePathArr)) {
+                    $dirName = array_pop($basePathArr);
+                    if ($dirName != '') {
+                        $relPath = '..' . DIRECTORY_SEPARATOR . $relPath;
+                    }
+                }
+            } else {
```
DavertMik: don't do `else`

src/Codeception/Configuration.php
loren-osborn: I know there are a couple regexes... This one matches Windows style drive letters.

src/Codeception/Configuration.php
loren-osborn: I mentioned when I submitted the PR, that the code was pretty rough. Basically you can only make a relative path of one path against the other if they both have the same "absoluteness":

* On *nix this is easy:
    * A path is absolute if it starts with a slash (a Boolean)
    * Two paths have the same absoluteness if both or neither start with a slash
* On Windows this is less clear:
    * a path can be relative to the current or a specified device
    * it can also be absolute or relative to the CWD of that device
    * If a drive is specified, they both have to specify the same drive
    * If no drive is specified, they both need to *not* specify a drive
    * whether or not they specify a drive letter the first non-drive letter character (after the `:`) need to both or neither start with a backslash.

If the paths don't have the same absoluteness, you can't make a relative path because they're *not* relative to one another... 

I need to clarify the code to capture that more clearly, but does that clarify the logic to you at least?

src/Codeception/Configuration.php
loren-osborn: Instead end the if and the function with distinct `return`s?

src/Codeception/Configuration.php
loren-osborn: I will add comments:

In advance of that, I'm peeling off all common parts of the path from left of both paths, and for all unmatched directory remaining pretending a `..` directory to the result.

loren-osborn:
Thanks for the feedback... I will revise the code as soon as I'm able.

DavertMik:
Thanks, will be waiting for update!

loren-osborn:
@DavertMik, Thanks... I will update this PR as soon as I can find time to update it. I didn't see a way to PM you, but wanted to know if you could take a peek at an issue I think I found: https://github.com/Codeception/Codeception/compare/2.2...loren-osborn:losborn_configModsLooksLikeABug  It shouldn't be hard to write a unit test to catch this issue... (If a enabled module in the module `.yml` file has options, the Codeception should [incorrectly] try to enable a module with the name of the module's first option, instead of its own name.)

loren-osborn:
@DavertMik, let me know if you found this much easier to follow. I prefer descriptive variable and method names to comments, but I included good comments too.

DavertMik:
@loren-osborn looks etter now, however I still have doubts. I think code still can be simplified but I'm not sure how :no_mouth:

Could you add tests to it? They should go to `tests/unit/ConfigurationTest.php`. 

@Naktibalda @sergeyklay what your thoughts of it?

sergeyklay:
+1 tests

loren-osborn:
I'm certainly down with tests... It's more a matter of knowing what kind of tests, and where to find them. I just found the right place for unit tests for this particular issue, and the issue I found with `Codeception\Configuration::modules()`:

* This file: `Codeception/tests/unit/Codeception/ConfigurationTest.php` for unit tests.

Where else do you suggest I put tests for these two issues. I currently don't have a setup where I can run the tests (might I suggest a `Vagrantfile` with a Puppet, Chef or Ansible configuration) so I'll have to rely on Travis and the other CI services.

Naktibalda:
I found `getRelativeDir()` very hard to read.

I think that it would be much easier to unit test it, if project dir was passed in as a parameter:
`public static function getRelativeDir($path, $projectDir)`

```php
function codecept_relative_path($path)
{
    return \Codeception\Configuration::getRelativeDir($path, \Codeception\Configuration::projectDir());
}
```

All you need to run ConfigurationTest is PHP,
but if you want to run all tests, try to use Docker as documented in https://github.com/Codeception/Codeception/blob/2.2/tests/README.md

loren-osborn:
@Naktibalda, passing the project directory as an argument is a great idea. I was trying to keep the `codecept_relative_path()` API, but as you pointed out there is no need for that.

As far as a dev environment, I currently have limited storage to keep multiple VMs around, and my current VM is missing a PECL module that composer requires for Codeception. (I don't want to alter one project's dev environment VM for development of another.) I'll see what I can put together if I have the time, but I won't make any promises with my looming deadlines.

loren-osborn:
I think I took what @Naktibalda said to heart. I rewrote this so I can inject both `projectDir()` *and* `DIRECTORY_SEPARATOR` so I will now be able to test the Windows path handling without being on a Windown system. I also rewrote the `$path` and `$projDir`handling for clairity, so both are either being manipulated as strings *or* as arrays, not one of each. I still need to write the actual tests, but that should be straightforward now.

I'm at dinner now, but hope to

loren-osborn:
Status update: Wrote some tests, found some edge casses I missed. I'll address before I push my code.

loren-osborn:
Enabled test cases... disabled code changes... test suite *should FAIL*

loren-osborn:
@DavertMik, @sergeyklay, @Naktibalda, I think the `semaphoreci` build needs to be restarted...

Once I get a signoff from one of you, I can get started on a git rebase squash of this code. (Of course, I'm still open to suggestions. It's far from the most readable thing I've written.)

loren-osborn:
All changed squashed into single commit.  Original unsquashed commits are still available here: https://github.com/loren-osborn/Codeception/tree/losborn_badBasePathFix_unsquashed

Thanks for all of your help. Do any of you want any additional revisions before merging this? (i.e. It may be a good idea to build the output array returned by `ConfigurationTest::getRelativeDirTestData()` to be generated as a permutation of several test conditions, but I'm unsure if that would be more or less readable. Any feedback is appreciated.)

src/Codeception/Configuration.php
```diff
@@ -500,6 +500,126 @@ public static function projectDir()
...
+     * Returns path to a given directory relative to $projDir.
+     * @param string $path
+     * @param string $projDir
+     * @param string $dirSep
+     * @return string
+     */
+    public static function getRelativeDir($path, $projDir, $dirSep = DIRECTORY_SEPARATOR)
+    {
+        // ensure $projDir ends with a trailing $dirSep
+        $projDir = preg_replace('/'.preg_quote($dirSep, '/').'*$/', $dirSep, $projDir);
```
Naktibalda: No an issue, but mu preferred way is $projDir = rtrim($projDir, $dirSep) . $dirSep;

src/Codeception/Configuration.php
```diff
@@ -500,6 +500,126 @@ public static function projectDir()
...
+            ) {
+                // ... shave that off
+                return substr($path, strlen($pathAbsPrefix['devicePrefix']));
+            }
+            // Return the input unaltered
+            return $path;
+        }
+        // peel off optional absoluteness prefixes and convert
+        // $path and $projDir to an subdirectory path array
+        $relPathParts = array_filter(explode($dirSep, substr($path, strlen($pathAbsPrefix['wholePrefix']))), 'strlen');
```
Naktibalda: 'strlen' is not necessary, array_filter removes empty strings when callback is not provided.

tests/unit/Codeception/ConfigurationTest.php
```diff
@@ -67,4 +67,83 @@ public function testDefaultCustomCommandConfig()
...
+                    ['my\\proj\\path\\some\\file\\in\\my\\proj.txt',     '\\my\\proj\\path\\',           '\\',  'my\\proj\\path\\some\\file\\in\\my\\proj.txt'],
+                // No device letter... absoluteness match
+                    ['my\\proj\\path\\some\\file\\in\\my\\proj.txt',     'my\\proj\\path\\foo\\bar',     '\\',  '..\\..\\some\\file\\in\\my\\proj.txt'],
+                    ['\\my\\proj\\path\\some\\file\\in\\my\\proj.txt',   '\\my\\proj\\path\\foobar\\',   '\\',  '..\\some\\file\\in\\my\\proj.txt'],
+                    ['\\my\\proj\\path\\some\\dir\\in\\my\\proj\\',      '\\my\\proj\\path\\foobar\\',   '\\',  '..\\some\\dir\\in\\my\\proj\\'],
+                // Device letter (both)... path absoluteness mismatch
+                    ['C:\\my\\proj\\path\\some\\file\\in\\my\\proj.txt', 'C:my\\proj\\path\\',           '\\',  '\\my\\proj\\path\\some\\file\\in\\my\\proj.txt'],
+                    ['d:my\\proj\\path\\some\\file\\in\\my\\proj.txt',   'd:\\my\\proj\\path\\',         '\\',  'my\\proj\\path\\some\\file\\in\\my\\proj.txt'],
```
Naktibalda: Is `d:my` a valid path on Windows?

tests/unit/Codeception/ConfigurationTest.php
```diff
@@ -67,4 +67,83 @@ public function testDefaultCustomCommandConfig()
...
+     * data provider for testGetRelativeDir
+     * 
+     * @return array(array(strings))
+     */
+    public function getRelativeDirTestData()
+    {
+        return [
+            // Unix style paths:
+                // projectDir() with & without trailing directory seperator: actual subdir
+                    ['/my/proj/path/some/file/in/my/proj.txt',           '/my/proj/path/',               '/',  'some/file/in/my/proj.txt'],
```
Naktibalda: The size of this data provider is impressive, so it must be good.

tests/unit/Codeception/ConfigurationTest.php
loren-osborn: To my knowledge it is... It certainly was in the MSDOS days... Each drive letter keeps its own CWD

tests/unit/Codeception/ConfigurationTest.php
loren-osborn: I believe you were being sarcastic, but I disagree (so actually agreeing)... I think shorter is clearer... Just not sure how to pair this down better

loren-osborn: ... Each device has its own CWD on a per-shell basis, I believe

src/Codeception/Configuration.php
loren-osborn: Agree... Was trying to make the code more clear... I probably failed. ;-)

src/Codeception/Configuration.php
loren-osborn: Yes... Probably clearer your way.

src/Codeception/Configuration.php
loren-osborn: I take back my previous comment... Without a callback, array_filter is supposed to remove any element that evaluates to FALSE... I believe this includes some non-empty strings, including `"0"`... Looks like I should add some dies named `0` to my data provider test cases.

Naktibalda: then using strlen is fine.

loren-osborn:
@DavertMik, @sergeyklay, @Naktibalda, Status update:

* I have confirmed that `D:Foo` is a perfectly valid Windows path (referring to file or directory `Foo` in this shell's current working directory for device `D:`)
* While I'm open to some tweaking, I think this patch is basically done and ready to merge.
* Welcome additions include:
    * Additional test cases with `0` as one of the directories in paths
    * Easier to read generation of test cases instead of an confusing exhaustive list
* I can get to these when I have time, but consider them low priority.

Naktibalda:
Your PR looks good to me, I am waiting for somebody else to review the code before merging it.

loren-osborn:
Thanks

src/Codeception/Configuration.php
DavertMik: Yeah, don't use regex here. rtrim is much clearer

tests/unit/Codeception/ConfigurationTest.php
DavertMik: Yes, the list is impressive. And that's really good

DavertMik:
Thanks. I think you did a great job and this PR looks almost perfect to me. Especially tests look great.
However, now I have a feeling that `isWindows`, `getPathAbsolutenessPrefix`, and other added methods may not belong to Configuration class. They are not about Configuration, right? That's why I'd suggest to move the path resolving logic to a utility class, probably `Codeception\Util\PathResolver` and make Configuration only call them to get relative path.

Sorry for this additional request but that's how we can maintain the clean and sane codebase.

loren-osborn:
This seems like a very reasonable request , though I'm leaning toward `Codeception\Util\PathHandler` unless the other already exists. I don't expect address this this week, but will try to get to it shortly. @DavertMik, if you can suggest a more readable way to generate the given test cases, I would be grateful.

DavertMik:
Looks good. Please proceed

Naktibalda:
@loren-osborn you are about to miss a release.

DavertMik:
Thanks @loren-osborn 
We will extract class by ourselves and your fix is about to land in next release

loren-osborn:
@DavertMik, @Naktibalda thanks for picking up the slack on this. My last comment in this PR was actually in the few days between when I was laid off from my last employer and when my daughter was born, so you can imagine I was a bit distracted. I ended up being out of work about 6 months and didn't realize you had actually merged this. Thank you for picking up the slack in my absence.

PR #3437, opened 2016-08-11
3427 support of standard mongodump/mongorestore tools to populate mongo db database v2

GSokol:


GSokol:
Rebuild of https://github.com/Codeception/Codeception/pull/3430 pull request.

src/Codeception/Lib/Driver/MongoDb.php
```diff
@@ -167,6 +169,71 @@ public function load($dumpFile)
...
+                escapeshellarg($dumpFile)
+            );
+        }
+        shell_exec($cmd);
+    }
+
+    public function loadFromTarGzMongoDump($dumpFile)
+    {
+        list($host, $port) = $this->getHostPort();
+        $getDirCmd = 'tar -tf ' . escapeshellarg($dumpFile)
```
DavertMik: please use `sprintf` here as well

src/Codeception/Lib/Driver/MongoDb.php
```diff
@@ -167,6 +169,71 @@ public function load($dumpFile)
...
+                escapeshellarg($dumpFile),
+                $host,
+                $port,
+                $this->user,
+                $this->password,
+                $dirName,
+                $dirName
+            );
+        } else {
+            $cmd = sprintf(
```
DavertMik: I think you can avoid using `else` by making this by default value of `$cmd` and appending `--username` and `--password` in case they are set.

src/Codeception/Module/MongoDb.php
```diff
@@ -114,6 +106,70 @@ public function _initialize()
...
+                throw new ModuleConfigException(
+                    __CLASS__,
+                    "File with dump doesn't exist.\n
+                    Please, check path for dump file: " . $this->config['dump']
+                );
+            }
+            $this->dumpFile = Configuration::projectDir() . $this->config['dump'];
+            $this->isDumpFileEmpty = false;
+
+            switch ($this->config['dump_type']) {
```
DavertMik: this switch-case seems to be too complicated. I'd suggest to transform it to using single `if` (try to avoid `else`) and moving complex logic to other methods, like `validateJSDump`, `validateMongoDump`, etc.

src/Codeception/Module/MongoDb.php
```diff
@@ -114,6 +106,70 @@ public function _initialize()
...
+                        if ($entity !== '..' && $entity !== '.') {
+                            $this->isDumpFileEmpty = false;
+                            break;
+                        }
+                    }
+                    $dumpDir->close();
+                    break;
+                case self::DUMP_TYPE_MONGODUMP_TAR_GZ:
+                    if (strtoupper(substr(PHP_OS, 0, 3)) === 'WIN') {
+                        $this->isDumpFileEmpty = true;
```
DavertMik: shouldn't an error be thrown that tgz is not supported on Windows?

src/Codeception/Module/MongoDb.php
GSokol: Do you suggest using something like chain pattern?

src/Codeception/Module/MongoDb.php
```diff
@@ -148,13 +204,46 @@ protected function loadDump()
         if ($this->isDumpFileEmpty) {
             return;
         }
+        switch ($this->config['dump_type']) {
```
DavertMik: I think you can wrap this into try-catch and not use `loadDump*` methods here. As they are one liner + catching exceptions

src/Codeception/Module/MongoDb.php
```diff
@@ -148,13 +204,46 @@ protected function loadDump()
...
         } catch (\Exception $e) {
             throw new ModuleException(__CLASS__, $e->getMessage());
         }
     }
 
+    private function loadDumpFromDirectory()
+    {
+        try {
+            $this->driver->loadFromMongoDump($this->dumpFile);
+        } catch (\Exception $e) {
```
DavertMik: this part looks like duplication, move it to `_before`

src/Codeception/Module/MongoDb.php
DavertMik: Not sure, personally I'd try to avoid `switch`. So yes, I'd check inside those methods if we deal with the right data type and `return` if not. But anyway, switch still can be ok in this case

GSokol:
@DavertMik Done!

src/Codeception/Lib/Driver/MongoDb.php
```diff
@@ -167,6 +169,56 @@ public function load($dumpFile)
         shell_exec($cmd);
     }
 
+    public function loadFromMongoDump($dumpFile)
+    {
+        list($host, $port) = $this->getHostPort();
+        if ($this->user && $this->password) {
+            $cmd = sprintf(
+                'mongorestore --host %s --port %s --username %s --password %s %s',
```
DavertMik: you can add 

```
$this->user && $this->password ? printf("--username %s --password %s ", $this->user, $this->password) : ''
```

as you do it on line 215 and remove this `if/else`

src/Codeception/Lib/Driver/MongoDb.php
```diff
@@ -167,6 +169,56 @@ public function load($dumpFile)
...
+                'Archive MUST contain single directory with db dump'
+            );
+        }
+        $dirName = trim(shell_exec($getDirCmd));
+        $cmd = sprintf(
+            "tar -xzf %s && mongorestore --host %s --port %s%s %s && rm -r %s",
+            escapeshellarg($dumpFile),
+            $host,
+            $port,
+            $this->user && $this->password ? printf("--username %s --password %s ", $this->user, $this->password) : '',
```
DavertMik: I think you can use the same trick on line 177

DavertMik:
Just one more change and I'm good to merge :)

GSokol:
@DavertMik Done!

DavertMik:
:+1: :+1: :+1:

PR #3439, opened 2016-08-12
[InnerBrowser] normalize page source and see() parameter

Naktibalda:
Replaces #3350 

I made a small improvement to make it work the way I wanted - this assertion passed before change made by @dizzy7 and it will pass again after my change.
```php
$this->module->see("Test Link \n\n\n    Test");
```


Naktibalda:
I haven't noticed that the build was broken before,
I fixed it now.

DavertMik:
Thanks

PR #3445, opened 2016-08-15
Fix deprecation notices

### LINKED ISSUE #3424: E_USER_DEPRECATED errors bypass error_level setting
It appears that Codeception's `Subscriber\ErrorHandler` doesn't respect the `error_level` configuration setting for `E_USER_DEPRECATION` errors.

# Background

After a recent `composer update`, one of our third-party libraries started triggering `E_USER_DEPRECATED` errors. These errors appeared as notifications after our tests. 

I tried setting Codeception's `error_level` to `E_ALL & ~E_STRICT & ~E_DEPRECATED & ~E_USER_DEPRECATED` (note the addition of `E_USER_DEPRECATED`). However, the notifications continued to appear.

config

notifications

I added `var_dump(error_reporting())` to my test suite's `_beforeSuite()` and `_afterSuite()` methods. Both methods printed `6143`, the correct value for `E_ALL & ~E_STRICT & ~E_DEPRECATED & ~E_USER_DEPRECATED`.

Error reporting appears to be correctly set from start to finish.

# Issue

The issue appears to stem from the `Subscriber\ErrorHandler`. 

As I understand it, setting PHP's error level via [`error_reporting()`](http://php.net/manual/en/function.error-reporting.php) will set the error levels for PHP's _native_ error handler.  However, when you register a _custom_ error handler via the [`set_error_handler()`](http://php.net/manual/en/function.set-error-handler.php) function, the _native_ error handlers will be bypassed for all the _custom_ handler's error types. If you don't specify error types, a _custom_ error handler defaults to `E_ALL | E_STRICT`. 

On line 36 of the `ErrorHandler` object, we correctly set the error reporting level to my configuration setting, `E_ALL & ~E_STRICT & ~E_DEPRECATED & ~E_USER_DEPRECATED`. However, on line 41, we register an error handler without specifying the error types. 

handle

Now, in the `errorHandler()` method, we do check to be sure the error being handled is one we've set with `error_reporting()`, however, this occurs after the check for `E_USER_DEPRECATED` errors. 

errorhandler

Even though I've excluded `E_USER_DEPRECATED` errors from Codeception's `error_level` reporting, they are being reported. That's not great.

# Solutions

In my mind, there are two solutions:

## Specify the handler's error types

We could specify the custom error handler's error types when we set it. This would remove the need to check the error types in the `errorHandler()` method.

solution1

## Move the E_USER_DEPRECATED check

We could move the E_USER_DEPRECATED check in the `errorHandler` method below the error types check.

solution2

Either solution fixes the issue.

success

# Pull request

The first solution seems more correct. We shouldn't handle events only to not handle them, unless we want to suppress PHP's native error handling. The second solution is definitely a much smaller change. 

I'm happy to make a pull request for either one. I wanted to check with the authors/contributors first, because there is plenty going on in here that I don't understand haha.

**Which solution do you recommend?**

jstewmc:
Before, if the Symfony Deprecations Helper was not installed, Codeception would display each deprecation notice even if deprecation errors were ignored in Codeception's configuration.

Now, it should respect the configuration settings.

Fixes #3424.

### DIFF
--- a/src/Codeception/Subscriber/ErrorHandler.php
+++ b/src/Codeception/Subscriber/ErrorHandler.php
@@ -43,14 +43,14 @@ class ErrorHandler implements EventSubscriberInterface
 
     public function errorHandler($errno, $errstr, $errfile, $errline, $context)
     {
-        if (!(error_reporting() & $errno)) {
-            // This error code is not included in error_reporting
-            return false;
-        }
-
         if (E_USER_DEPRECATED === $errno) {
             $this->handleDeprecationError($errno, $errstr, $errfile, $errline, $context);
             return;
+        }
+
+        if (!(error_reporting() & $errno)) {
+            // This error code is not included in error_reporting
+            return false;
         }
 
         if (strpos($errstr, 'Cannot modify header information') !== false) {
--- a/src/Codeception/Subscriber/ErrorHandler.php
+++ b/src/Codeception/Subscriber/ErrorHandler.php
@@ -115,8 +115,8 @@ class ErrorHandler implements EventSubscriberInterface
             call_user_func($this->oldHandler, $type, $message, $file, $line, $context);
             return;
         }
-        
-        // if the error is included in the configured error reporting 
+
+        // if the error is included in the configured error reporting
         if (error_reporting() & $type) {
             Notification::deprecate("$message", "$file:$line");
         }
--- a/src/Codeception/Subscriber/ErrorHandler.php
+++ b/src/Codeception/Subscriber/ErrorHandler.php
@@ -115,6 +115,10 @@ class ErrorHandler implements EventSubscriberInterface
             call_user_func($this->oldHandler, $type, $message, $file, $line, $context);
             return;
         }
-        Notification::deprecate("$message", "$file:$line");
+        
+        // if the error is included in the configured error reporting 
+        if (error_reporting() & $type) {
+            Notification::deprecate("$message", "$file:$line");
+        }
     }
 }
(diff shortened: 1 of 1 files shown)


Naktibalda:
Thanks, it looks simple and good to me.

Naktibalda:
It isn't good.

The reason why error_reporting is ignored is because Symfony recommends using `@trigger_error` for depreciation messages - so they don't affect live app, but show up in tests.

jstewmc:
Shoot, I thought that would work haha. If Symfony suggests triggering deprecation notices outside of the normal error reporting system, I can see the reason we'd ignore `error_reporting()` levels and why my pull request wouldn't work.

I'd just like to find a solution that fixes it for people who don't use Symfony but at least doesn't break it for people who do. 

How about this one? Basically, if the deprecation type is in Codeception's configured error levels, print the notification. Otherwise, do nothing with deprecation errors.

Naktibalda:
Oh, your whole problem are these deprecation messages?
Before the handler was implemented, deprecation messages were literally killing my tests.

I don't think that this change will be accepted.

jstewmc:
Oh, haha, yes. I have dozens of deprecation notices printed after every test, because of an update to a third-party library I use. 

I turned off deprecation notices in the `error_level` setting of my functional.suite.yml config file. However, the notices are still printed to the screen when I run my tests.

I realize my first fix boned Symfony. That's why I reverted it and tried a different approach.

I thought my latest attempt (i.e., only printing deprecation notices if the suite's `error_level` allows it) would be a nice middle ground. It should allow Symfony to keep working, and it should respect my `error_level` configuration setting. 

It's up to you guys, but having a configuration setting that doesn't work for all errors seems less than optimal haha.

Either way, thank you for your help! And, thanks to the Codeception team for a great library!

PR #3446, opened 2016-08-15
[REST] requests are added to browser history

### LINKED ISSUE #3197: [Regression] grabFromCurrentUrl() doesn't work with REST in v2.2
After upgrading to v2.2, `grabFromCurrentUrl()` does no longer return the url requested using `sendGET()` but instead errors after printing `[LogicException] The page history is empty.`. This change is not mentioned in the release notes.

#### Minimal example
The following example works v2.1 but fails in v2.2
**acceptance.suite.yml:**
```
class_name: AcceptanceTester
modules:
    enabled:
        - REST:
            depends: PhpBrowser
            url: http://localhost:8080
        - PhpBrowser
```

**acceptance/grabUrlCept.cept:**
```
$I->wantTo('I want to retrieve the current URL');
$I->sendGET('/');
$I->grabFromCurrentUrl();
```

Naktibalda:
Fixes #3197

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -119,7 +119,7 @@ class InnerBrowser extends Module implements Web, PageSourceSaver, ElementLocato
         array $server = [],
         $content = null
     ) {
-        $this->clientRequest($method, $uri, $parameters, $files, $server, $content, false);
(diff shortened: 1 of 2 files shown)


Naktibalda:
I checked if sendAjaxRequest methods were affected by this change, but they weren't.

DavertMik:
Good. Thank you!

PR #3451, opened 2016-08-18
Fixed problems with default messages of asserts

B4rb4ross4:
I created a helper and extended \Codeception\Module\REST. When using

`$I->assertInternalType`

I recognized that the default message in Util/Shared/Asserts.php was missing. I found the same problem in some Asserts in Util/Asserts.php

This pull request adds this default messages and also corrects some default values for the messages which were set to null. I you look into PhpUnit itself the default message is ALWAYS an empty string.

### DIFF
--- a/src/Codeception/Module/Asserts.php
+++ b/src/Codeception/Module/Asserts.php
@@ -248,7 +248,7 @@ class Asserts extends CodeceptionModule
      * @param $actual
      * @param $description
      */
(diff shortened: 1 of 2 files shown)


B4rb4ross4:
// @DavertMik 
This would be pretty urgent, because otherwise i would have to add a
> , ''
to several hundred assertions.  And that would be pretty annoying for such a minor issue

PR #3453, opened 2016-08-18
Basic strict coverage support for Cest-format tests. Fixes #3409

### LINKED ISSUE #3409: Strict code coverage support
It doesn't appear that Codeception supports using `@covers` or `@codeCoverageIgnore` annotations to control code coverage like [PHPUnit does](https://phpunit.de/manual/current/en/code-coverage-analysis.html#code-coverage-analysis.specifying-covered-methods). Personally I like to use this feature because it helps me separate code that was incidentally executed during a test from code that was actually tested. Is this kind of enhancement possible for Codeception?

If this feature already exists in some form, then the enhancement I'm requesting is documentation of the feature.

TehJoE:
Fixes #3409

### DIFF
--- a/src/Codeception/Test/Feature/CodeCoverage.php
+++ b/src/Codeception/Test/Feature/CodeCoverage.php
@@ -26,8 +26,20 @@ trait CodeCoverage
             return;
         }
 
+        if (method_exists($this, 'getLinesToBeCovered')) {
+            $linesToBeCovered = $this->getLinesToBeCovered();
+        } else {
+            $linesToBeCovered = [];
+        }
+
+        if (method_exists($this, 'getLinesToBeUsed')) {
+            $linesToBeUsed = $this->getLinesToBeUsed();
+        } else {
+            $linesToBeUsed = [];
+        }
+
         try {
-            $codeCoverage->stop(true);
+            $codeCoverage->stop(true, $linesToBeCovered, $linesToBeUsed);
         } catch (\PHP_CodeCoverage_Exception $cce) {
             if ($status === \Codeception\Test\Test::STATUS_OK) {
                 $this->getTestResultObject()->addError($this, $cce, $time);
--- a/src/Codeception/Test/Feature/CodeCoverage.php
+++ b/src/Codeception/Test/Feature/CodeCoverage.php
@@ -2,6 +2,7 @@
 namespace Codeception\Test\Feature;
 
 use Codeception\Test\Descriptor;
+use Codeception\Test\Interfaces\StrictCoverage;
 
 trait CodeCoverage
 {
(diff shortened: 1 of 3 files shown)


src/Codeception/Test/Feature/CodeCoverage.php
```diff
@@ -26,8 +26,20 @@ public function codeCoverageEnd($status, $time)
             return;
         }
 
+        if (method_exists($this, 'getLinesToBeCovered')) {
```
DavertMik: could you use create an interface for that. Sonething like `StrictCoverageInterface` and make Cest implementing that. This way you can simplify code by checking this interface only.

DavertMik:
This looks really useful. I didn't have opportunity to look and review, but I have one minor suggestion (see above). If you will implement that I will merge this PR. 

Thank you!

src/Codeception/Test/Feature/CodeCoverage.php
Naktibalda: Wouldn't it be better to add these methods to Cept and Gherkin formats?

DavertMik: Not sure how this should work with Gherkin...

DavertMik:
On second thought I have some doubts about this PR. Isn't `getLinesToBeCovered` expect a class and method that should be executed inside a test? Something that is defined via `@covers` annotation in PHPUnit? In current implementation you just pass a Cest class and a test method for it.

Naktibalda:
@DavertMik I would like to point out your own earlier implementation: https://github.com/Codeception/Codeception/commit/16f1f26df9d62ec1eec8e351b82e18c5c7550c05#diff-dfd87d7762b7147eeacf44d56dd67af2L26 

`testClassInstance` only exists in Cest format.

DavertMik:
Ok, then, taken as a proof :)
So we need an interface and this PR can be merged.

TehJoE:
Sorry about the delay on this, guys. I added the interface as requested. Let me know if it needs anything else.

src/Codeception/Test/Feature/CodeCoverage.php
TehJoE: It may be possible to get something like this working in Cept/Gherkin, but the PHPUnit helper my implementation uses is looking for the `@covers` and `@uses` annotations on a test method (and on the test class, where it also checks for `@coversDefaultClass` and maybe some other stuff I'm not aware of) which pretty much rules out any test format that doesn't use a class-method structure. If I become familiar enough with the Cept or Gherkin formats I may try to extend this implementation to them as well if it hasn't been done already by then.

Naktibalda: I thought that this change does something bigger than just supporting `@covers` and `@uses` when I wrote that comment, but I was wrong.
So, because Cept and Gherkin formats don't support `@covers` and `@uses` annotations, we don't need to add anything to format files, right?

TehJoE: That's correct. As far as I'm aware, there's nowhere to actually put the
annotations in tests of those formats.

An alternative implementation of the methods in this interface would still
allow a test writer to control code coverage, it would just have to be
configured differently by the test writer. Hooking ´$I->wantToTest()´ or a
similar "comment" method and parsing out what one might put in a `@covers`
might work for Cept, for example.

In addition to `@covers` and ´@uses´, this change also enables
`@codeCoverageIgnore` since it's controlled using the same mechanism.

Naktibalda: Cept files have `pseudo-annotations` http://codeception.com/docs/07-AdvancedUsage#Groups . I think that additional annotations could be implemented in the same way.

DavertMik:
Thank you for a good work. Merging!

PR #3455, opened 2016-08-19
Display only distinct deprecation messages

Naktibalda:
Related to #3424

Before:
```
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
DEPRECATION: Usage of Zend\ServiceManager\ServiceManager::getServiceLocator is deprecated since v3.0.0; please use the container passed to the factory instead /home/.../codeception-zf2-tests/vendor/zendframework/zend-servicemanager/src/ServiceManager.php:164
DEPRECATION: Usage of Zend\ServiceManager\ServiceManager::getServiceLocator is deprecated since v3.0.0; please use the container passed to the factory instead /home/.../codeception-zf2-tests/vendor/zendframework/zend-servicemanager/src/ServiceManager.php:164
DEPRECATION: Usage of Zend\ServiceManager\ServiceManager::getServiceLocator is deprecated since v3.0.0; please use the container passed to the factory instead /home/.../codeception-zf2-tests/vendor/zendframework/zend-servicemanager/src/ServiceManager.php:164


Time: 2.27 seconds, Memory: 20.00MB
```

After:
```
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
DEPRECATION: Usage of Zend\ServiceManager\ServiceManager::getServiceLocator is deprecated since v3.0.0; please use the container passed to the factory instead /home/.../codeception-zf2-tests/vendor/zendframework/zend-servicemanager/src/ServiceManager.php:164


Time: 2.54 seconds, Memory: 20.00MB
```

### DIFF
--- a/src/Codeception/Subscriber/Console.php
+++ b/src/Codeception/Subscriber/Console.php
@@ -322,8 +322,11 @@ class Console implements EventSubscriberInterface
     public function afterSuite(SuiteEvent $e)
     {
         $this->message()->width($this->width, '-')->writeln();
-        $deprecationMessages = Notification::all();
-        foreach (array_unique($deprecationMessages) as $message) {
+        $messages = Notification::all();
+        foreach (array_count_values($messages) as $message => $count) {
+            if ($count > 1) {
+                $message = $count . 'x ' . $message;
+            }
             $this->output->notification($message);
         }
     }
--- a/src/Codeception/Subscriber/Console.php
+++ b/src/Codeception/Subscriber/Console.php
@@ -323,7 +323,7 @@ class Console implements EventSubscriberInterface
     {
         $this->message()->width($this->width, '-')->writeln();
         $deprecationMessages = Notification::all();
-        foreach ($deprecationMessages as $message) {
+        foreach (array_unique($deprecationMessages) as $message) {
             $this->output->notification($message);
         }
     }


Naktibalda:
For comparison, symfony/phpunit-bridge aggregates deprecation messages and produces this output:
```
Unsilenced deprecation notices (3)

Usage of Zend\ServiceManager\ServiceManager::getServiceLocator is deprecated since v3.0.0; please use the container passed to the factory instead: 3x
    3x in Application::run from Codeception

```

Naktibalda:
I've looked at the code of phpunit-bridge error handler and I can  confirm that the only reason to use it with Codeception is the command line flag which hides deprecation messages.

Aggregated output was intented to display the line of test code which calls deprecated method (directly or indirectly) as demonstrated in this picture:


phpunit-bridge achieves that by finding the first class in stack trace which is not ReflectionMethod and does not begin with 'PhpUnit_'.
It works for PhpUnit, but not for Codeception, because the first class passing this criteria is [Codeception\Application::run](https://github.com/Codeception/Codeception/blob/2.2/codecept#L33)

```
Usage of Zend\ServiceManager\ServiceManager::getServiceLocator is deprecated since v3.0.0; please use the container passed to the factory instead: 3x
    3x in Application::run from Codeception
```

DavertMik:
Do I understand it correctly that we can drop off phpunit-bridge if we handle and display deprecation errors correctly?

Can you improve output to provide this `3x` into message?

Naktibalda:
phpunit-bridge has a useful feature that it can stop execution and display stack trace of matching deprecation message `SYMFONY_DEPRECATIONS_HELPER=/message text/ codecept run`.

I will improve the output.

Naktibalda:
```
3x DEPRECATION: Usage of Zend\ServiceManager\ServiceManager::getServiceLocator is deprecated since v3.0.0; please use the container passed to the factory instead /.../codeception-zf2-tests/vendor/zendframework/zend-servicemanager/src/ServiceManager.php:164
```

DavertMik:
Thank you!

PR #3460, opened 2016-08-22
Handle deprecation messages according to error_level setting

### LINKED ISSUE #3424: E_USER_DEPRECATED errors bypass error_level setting

Naktibalda:
Fixes #3424

### DIFF
--- a/src/Codeception/Subscriber/ErrorHandler.php
+++ b/src/Codeception/Subscriber/ErrorHandler.php
@@ -48,14 +48,14 @@ class ErrorHandler implements EventSubscriberInterface
 
     public function errorHandler($errno, $errstr, $errfile, $errline, $context)
     {
-        if (!($this->errorLevel & $errno)) {
-            // This error code is not included in error_reporting
-            return false;
-        }
-
         if (E_USER_DEPRECATED === $errno) {
             $this->handleDeprecationError($errno, $errstr, $errfile, $errline, $context);
             return;
+        }
+
+        if (!(error_reporting() & $errno)) {
+            // This error code is not included in error_reporting
+            return false;
         }
 
         if (strpos($errstr, 'Cannot modify header information') !== false) {
@@ -116,6 +116,9 @@ class ErrorHandler implements EventSubscriberInterface
 
     private function handleDeprecationError($type, $message, $file, $line, $context)
     {
+        if (!($this->errorLevel & $type)) {
+            return;
+        }
         if ($this->deprecationsInstalled && $this->oldHandler) {
             call_user_func($this->oldHandler, $type, $message, $file, $line, $context);
             return;
--- a/src/Codeception/Subscriber/ErrorHandler.php
+++ b/src/Codeception/Subscriber/ErrorHandler.php
@@ -25,15 +25,20 @@ class ErrorHandler implements EventSubscriberInterface
     /**
      * @var int stores bitmask for errors
      */
-    private $errorLevel = 'E_ALL & ~E_STRICT & ~E_DEPRECATED';
+    private $errorLevel;
+
+    public function __construct()
+    {
+        $this->errorLevel = E_ALL & ~E_STRICT & ~E_DEPRECATED;
+    }
 
     public function handle(SuiteEvent $e)
     {
         $settings = $e->getSettings();
         if ($settings['error_level']) {
-            $this->errorLevel = $settings['error_level'];
+            $this->errorLevel = eval("return {$settings['error_level']};");
         }
-        error_reporting(eval("return {$this->errorLevel};"));
+        error_reporting($this->errorLevel);
         // We must register shutdown function before deprecation error handler to restore previous error handler
         // and silence DeprecationErrorHandler yelling about 'THE ERROR HANDLER HAS CHANGED!'
         register_shutdown_function([$this, 'shutdownHandler']);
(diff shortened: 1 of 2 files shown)


Naktibalda:
It appears that we don't want to respect error_level for all levels, it has to be limited to E_USER_DEPRECATED.

Naktibalda:
The test is flaky, I didn't expect that.

tests/unit/Codeception/Subscriber/ErrorHandlerTest.php
```diff
@@ -0,0 +1,23 @@
...
+
+    public function testDeprecationMessagesRespectErrorLevelSetting()
+    {
+        $errorHandler = new ErrorHandler();
+
+        $suiteEvent = new SuiteEvent(new Suite(), null, ['error_level' => 'E_ERROR']);
+        $errorHandler->handle($suiteEvent);
+
+        $messagesBeforeRun = count(Notification::all());
+        $errorHandler->errorHandler(E_USER_DEPRECATED, 'deprecated message', __FILE__, __LINE__, []);
```
Naktibalda: How can it add messages that happened in other tests at this line of code?

tests/unit/Codeception/Subscriber/ErrorHandlerTest.php
```diff
@@ -0,0 +1,23 @@
...
+
+    public function testDeprecationMessagesRespectErrorLevelSetting()
+    {
+        $errorHandler = new ErrorHandler();
+
+        $suiteEvent = new SuiteEvent(new Suite(), null, ['error_level' => 'E_ERROR']);
+        $errorHandler->handle($suiteEvent);
+
+        $messagesBeforeRun = Notification::all();
+        $errorHandler->errorHandler(E_USER_DEPRECATED, 'deprecated message', __FILE__, __LINE__, []);
```
Naktibalda: How can it add messages that happened in other tests at this line of code?

Naktibalda:
```
 Test  tests/unit/Codeception/Subscriber/ErrorHandlerTest.php:testDeprecationMessagesRespectErrorLevelSetting

Deprecation message was added to notifications

Failed asserting that Array &0 () is identical to Array &0 (

    0 => 'WARNING: Environments are not configured tests/data/claypit/tests/skipped/NoEnvironmentCept.php'

    1 => 'WARNING: Environments are not configured tests/data/claypit/tests/skipped/NoEnvironmentCept.php'

).

#1  /home/travis/build/Codeception/Codeception/tests/unit/Codeception/Subscriber/ErrorHandlerTest.php:21

**Comment by Naktibalda** [comment]
Oh, I haven't noticed that Notification::all() clears the old messages.

File: tests/unit/Codeception/Subscriber/ErrorHandlerTest.php [hunk 3]
```diff
@@ -0,0 +1,23 @@
+<?php
+
+use Codeception\Event\SuiteEvent;
+use Codeception\Lib\Notification;
+use Codeception\Subscriber\ErrorHandler;
+use Codeception\Suite;
+
+class ErrorHandlerTest extends \PHPUnit_Framework_TestCase
```
DavertMik: wow, that's cool

src/Codeception/Subscriber/ErrorHandler.php
```diff
@@ -25,15 +25,20 @@ class ErrorHandler implements EventSubscriberInterface
...
+    {
+        $this->errorLevel = E_ALL & ~E_STRICT & ~E_DEPRECATED;
+    }
 
     public function handle(SuiteEvent $e)
     {
         $settings = $e->getSettings();
         if ($settings['error_level']) {
-            $this->errorLevel = $settings['error_level'];
+            $this->errorLevel = eval("return {$settings['error_level']};");
```
DavertMik: eval 0.o
what if eval is blocked?

PR #3464, opened 2016-08-23
Allow PhpUnit 5.* in composer.json

Naktibalda:


DavertMik:
Let's cross fingers and hope that this will work for us :)
Looks like there won't be any breaking changes but who really knows

PR #3467, opened 2016-08-24
Use Doctrine\ORM\EntityManagerInterface to check for valid em instance

akbwm:
Use \Doctrine\ORM\EntityManagerInterface instead of
\Doctrine\ORM\EntityManager to check for valid entity manager in
Doctrine2 module.


DavertMik:
@raistlin looks good to me. Please merge if it is ok

Naktibalda:
ZF2 Doctrine tests passed, so it must be good

DavertMik:
Thanks

PR #3473, opened 2016-08-30
Fixes mongorestore command syntax and adds --quiet option to config

PhoenixShin:
I was unable to load my tar.gz mongo dump and I found that there were some things missing from the files that would allow me to load my dumps. For example not using the `validateDump()` function even though it's there and the mongorestore syntax not including what database to restore to. 

While I was at it, I added the ability to `--quiet` the restore or else if you are using `cleanup` and `populate` options you will be getting the load message every time.


DavertMik:
Thank you for improvements

PR #3474, opened 2016-08-30
Updated Memcache::seeInMemcached

sergeyklay:
* Updated `Memcache::seeInMemcached` to check if the key exists alone or with the desired value
* Added `Memcache::haveInMemcached`
* Fixed `Memcache::dontSeeInMemcached`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+#### 2.2.5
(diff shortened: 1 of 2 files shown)


DavertMik:
I will need detailed description to put into changelog. Or you can update changelog by yourself :)

sergeyklay:
@DavertMik done

sergeyklay:
@DavertMik Latest change fixes the issue with message

> Failed asserting that false is not equal to false.

because memcached **can** return false

DavertMik:
Thanks

PR #3476, opened 2016-09-01
Fix support for nested arrays in the file array

torkiljohnsen:
With the latest patch to this file, I accidentally removed support the files array being able to contain nested arrays of files. This fixes that blooper.


torkiljohnsen:
This is related to the Laravel module, so ping @janhenkgerritsen

janhenkgerritsen:
Thanks!

PR #3480, opened 2016-09-02
[REST] Improved output of failed JsonType assertions

### LINKED ISSUE #2858: seeResponseMatchesJsonType error message is confusing
Lets say i pass this array to seeResponseMatchesJsonType.
```
$I->seeResponseMatchesJsonType([
	'sessions' => [
		[
			'id_session_FAIL' => 'string', //force error
			'session_name' => 'string',
			'course_name' => 'string',
			'course_code' => 'string',
			'course_description' => 'string'
		],
	],
	'success'=>'boolean'
]);
```
In this example im forcing it to fail, but the error message i get is confusing.
```
Failed asserting that 'Key `id_session_FAIL` doesn't exist in {.............} is true.
```
It almost sounds like im trying to make sure the key doesn't exist and it failed ( basically found that key ).
After some digging around i found this that this method uses another method `matches` which also uses another method - `typeComparison`
`typeComparison` can return true, or some error message. One of the error messages is this:
```
return "Key `$key` doesn't exist in " . json_encode($data);
```
That message sounds ok. Sure, it failed because it didn't find some key.
But after that, this error message is used in a assertTrue in the `seeResponseMatchesJsonType` method
```
$matched = (new JsonType($jsonArray))->matches($jsonType); //Error message is returned here
$this->assertTrue($matched, $matched);
```
Which spits out the confusing message that i pointed out earlier. 
I saw issue #2544, which might cover this, as it might rework this method in the process.

Naktibalda:
Fixes #2858

### DIFF
--- a/src/Codeception/Module/REST.php
+++ b/src/Codeception/Module/REST.php
@@ -5,6 +5,7 @@ use Codeception\Exception\ModuleException;
 use Codeception\Lib\Interfaces\ConflictsWithModule;
 use Codeception\Module as CodeceptionModule;
 use Codeception\PHPUnit\Constraint\JsonContains;
+use Codeception\PHPUnit\Constraint\JsonType as JsonTypeConstraint;
 use Codeception\TestInterface;
 use Codeception\Lib\Interfaces\API;
 use Codeception\Lib\Framework;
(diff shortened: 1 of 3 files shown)


DavertMik:
Looks good. Thank you

Issue #3481, opened 2016-09-04
Running tests by name includes too many methods


oehlschl:
#### What are you trying to achieve?

The 'Running Tests' documentation [here](http://codeception.com/docs/02-GettingStarted#Running-Tests) states that 'You can execute one test from a test class (for Cest or Test formats)' via `php codecept run tests/acceptance/SignInCest.php:anonymousLogin`.

If I run this command in a file SignInCest.php containing the method 'anonymousLogin', I expect that only this single test will be run, regardless of the names of other test methods. 

#### What do you get instead?

However, since the method name above is passed to PHPUnit's --filter option, it actually functions as a pattern, matching all method names in which it is a substring. For example, the command above would run all of 'anonymousLogin', 'anonymousLoginFromHomepage', and 'nonanonymousLogin' (trivial examples, assuming all of these methods exist in SignInCest.php), when the expectation is that only a single test would be run.

Similarly, in a test class 'MyClass' containing test methods 'new_user' and 'guest_new_user', there is no way to run just the 'new_user' test; `codecept run MyClass:new_user` would run both.

Additionally, there is no way to way to prevent this behavior without changing the source. Test methods could be uniquely identified with a pattern like '/:testMethod$/', as Codeception test signatures are of the format 'ClassName:TestMethod', but ':' is already used to separate the path from the filter [here](https://github.com/Codeception/Codeception/blob/e31097051056e26d2d70e23b4df57a0f6f7df1b0/src/Codeception/Command/Run.php#L347).

Fixing this would admittedly break functionality for users expecting the test method name passed to `codecept run` to function like a pattern, but it would also make running tests by name work as expected and described.

oehlschl:
I would propose either returning `':' . $filter . '$'` from \Codeception\Command\Run::matchFilteredTestName() so that method names are exactly matched, or updating the explode in the same method to instead use the first ':' char only to separate the path and filter strings, allowing the filter to contain additional ':' chars; in the latter case, regex patterns (including exact matches) could actually be used.

Ex:
- `codecept run MyClass::exactMatch$`
- `codecept run MyClass::prefixMatch`
- `codecept run MyClass:containsMatch`
- `codecept run MyClass:suffixMatch$`
- `codecept run :/:.*arbitraryPatternWithDelimiters/`

Naktibalda:
I prefer updating documentation and leaving behaviour unchanged.

oehlschl:
[...]

Issue #2736, opened 2016-01-25
[FR] Filter Cept tests by name


Naktibalda:
Situation: I have 30 Cept files in a directory and I want to run 5 of them sharing the same prefix.
I don't want to run all tests every time because functional tests are slow.

Current options:
a) run each test manually
b) implement a deeper hierarchy and run a subdirectory.

It would be nice to be able to run  `codecept run tests/functional/module/ :create` and get only matching tests executed.

oehlschl:
FYI, I think this can be closed since #3482 was merged.

You should be able to run all tests containing "create" in dir tests/functional/module/ via `codecept run tests/functional/module/:create`, or `codecept run tests/functional/module/:^create` for tests beginning with "create".

PR #3482, opened 2016-09-04
bugfix issue 3481: fix pattern matching in run test by name

oehlschl:
Bugfix for #3481 

Changes are low risk in that the current majority case behavior remains the same; for test names of the format dir/ClassName:testName, nothing has changed. However, this change allows the use of additional ':' characters, which can be used to more accurately filter tests by name.

This change should also cover enhancement #2736

### DIFF
--- a/docs/02-GettingStarted.md
+++ b/docs/02-GettingStarted.md
@@ -223,13 +223,13 @@ Tests can be started with the `run` command.
 php codecept run
 ```
 
-With the first argument you can run tests from one suite.
+With the first argument you can run all tests from one suite.
 
 ```bash
 php codecept run acceptance
 ```
 
-To run exactly one test, add a second argument. Provide a local path to the test, from the suite directory.
+To limit tests run to a single class, add a second argument. Provide a local path to the test class, from the suite directory.
 
 ```bash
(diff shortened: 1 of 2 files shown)


oehlschl:
Per the discussion in #3481, I updated the "Getting Started" documentation to include pattern matching. Realizing that it was more complicated to explain why the patterns needed to begin with ':' to indicate the test method (since PHPUnit matches against the fully-qualified method name), I added support for the standard ^ char as start-of-string. Accordingly, functioning patterns are as follows, and are entirely intuitive:

- :^prefix
- :contains
- :suffix$
- :^exactMatch$

DavertMik:
I really love this regex filtering! :+1: 

Looks like there left no examples of non-regex filtering. However, in most cases non-strict filtering with `:` is more than enough. Can you bring back `run tests/acceptance/SignInCest.php:anonymousLogin` example? You can say that `:` will work like a grep, if you want to explain it better.

But yes, you should state that using regex is possible, and leave the `:^login` example.

oehlschl:
Yeah, no problem. I'll re-add the anonymousLogin example and repost shortly.

DavertMik:
Thank you!

PR #3490, opened 2016-09-06
Save all screenshots and page sources + show their names in log message

kusnir:
When you run test with "can" asserts and some of them fails then you will have screenshot and page source only from last one.
Try this test:

```
<?php 
$I = new NoGuy($scenario);

$code = "document.body.innerHTML = '<div style=\"text-align: center; padding-top: 50px;\"><h1>TEST %d</h1></div>';";
$I->wantTo('Test creating multiple screenshots');
$I->amOnPage('/');

$I->executeJS(sprintf($code, 1));
usleep(500000);
$I->canSee('TEST 2');

$I->executeJS(sprintf($code, 2));
usleep(500000);
$I->canSee('TEST 3');

$I->executeJS(sprintf($code, 3));
usleep(500000);
$I->canSee('TEST 4');
```

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -374,9 +374,18 @@ class WebDriver extends CodeceptionModule implements
         $this->debugWebDriverLogs();
         $filename = preg_replace('~\W~', '.', Descriptor::getTestSignature($test));
         $outputDir = codecept_output_dir();
-        $this->_saveScreenshot($outputDir . mb_strcut($filename, 0, 245, 'utf-8') . '.fail.png');
-        $this->_savePageSource($outputDir . mb_strcut($filename, 0, 244, 'utf-8') . '.fail.html');
-        $this->debug("Screenshot and page source were saved into '$outputDir' dir");
+        $filenamePng = mb_strcut($filename, 0, 245, 'utf-8') . '.fail.png';
+        $filenameHtml = mb_strcut($filename, 0, 244, 'utf-8') . '.fail.html';
+        for ($i = 1; $i < 1000; $i++) {
+            if ((! file_exists($outputDir . $filenamePng)) && (! file_exists($outputDir . $filenameHtml))) {
+                break;
+            }
+            $filenamePng = mb_strcut($filename . '.' . $i, 0, 245, 'utf-8') . '.fail.png';
+            $filenameHtml = mb_strcut($filename . '.' . $i, 0, 244, 'utf-8') . '.fail.html';
+        }
+        $this->_saveScreenshot($outputDir . $filenamePng);
+        $this->_savePageSource($outputDir . $filenameHtml);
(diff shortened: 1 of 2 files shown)


src/Codeception/Module/WebDriver.php
```diff
@@ -374,9 +374,18 @@ public function _failed(TestInterface $test, $fail)
         $this->debugWebDriverLogs();
         $filename = preg_replace('~\W~', '.', Descriptor::getTestSignature($test));
         $outputDir = codecept_output_dir();
-        $this->_saveScreenshot($outputDir . mb_strcut($filename, 0, 245, 'utf-8') . '.fail.png');
-        $this->_savePageSource($outputDir . mb_strcut($filename, 0, 244, 'utf-8') . '.fail.html');
-        $this->debug("Screenshot and page source were saved into '$outputDir' dir");
+        $filenamePng = mb_strcut($filename, 0, 245, 'utf-8') . '.fail.png';
+        $filenameHtml = mb_strcut($filename, 0, 244, 'utf-8') . '.fail.html';
+        for ($i = 1; $i < 1000; $i++) {
```
DavertMik: what is 1000?

src/Codeception/Module/WebDriver.php
```diff
@@ -374,9 +374,18 @@ public function _failed(TestInterface $test, $fail)
         $this->debugWebDriverLogs();
         $filename = preg_replace('~\W~', '.', Descriptor::getTestSignature($test));
         $outputDir = codecept_output_dir();
-        $this->_saveScreenshot($outputDir . mb_strcut($filename, 0, 245, 'utf-8') . '.fail.png');
-        $this->_savePageSource($outputDir . mb_strcut($filename, 0, 244, 'utf-8') . '.fail.html');
-        $this->debug("Screenshot and page source were saved into '$outputDir' dir");
+        $filenamePng = mb_strcut($filename, 0, 245, 'utf-8') . '.fail.png';
```
DavertMik: what is 245????

DavertMik:
Just curious if this patch works for you as you expect? 
I think it's not. The `_failed` method is execucted only when a test completely fails. `canSee` assertions won't trigger it, so screenshots are never saved for them. Probably the feature you want can't be implemented that easily

kusnir:
For 2.2 (7fa26a7e4bab45d74a406685f14c05994c9b98d2) with WebDriver it works for me. So the test generates 3 PNG and 3 HTML files to my _output directory (and also failed file).

Very simple: after failing "can" assert the ConditionalAssertionFailed exception is throw which run:
-> addFailure
-> fire(Events::TEST_FAIL, new FailEvent($test, $time, $e))
-> dispatch($event . '.' . $group, $eventType);
-> failed(FailEvent $e)
-> $module->_failed($e->getTest(), $e->getFail()); = WebDriver:_failed(TestInterface $test, $fail)

... the test continues so the method WebDriver:_failed is 3 times called in this test case.

Naktibalda:
Wait, why did you revert your change?

Please make a new pull request for Windows colors change.
Use feature branches.

PR #3492, opened 2016-09-07
Update WebDriver Doc - Wrong Regex

eXorus:
Hello,

According to the documentation 

```php
$I->amOnUrl('http://localhost/user/42/');
$link = $I->grabFromCurrentUrl();
$user_id = $I->grabFromCurrentUrl('~$/user/(\d+)/~');
```

Result will be:
link = /user/42/
user_id = 42

That's not true, result will be:
 Step  Grab from current url "~$/user/(\d+)/~"
 Fail  Couldn't match ~$/user/(\d+)/~ in /user/42/

So I fixed the doc with a new regex '~/user/(\d+)/~'

### DIFF
--- a/docs/modules/WebDriver.md
+++ b/docs/modules/WebDriver.md
@@ -761,7 +761,7 @@ If no parameters are provided, the full URI is returned.
 
 ``` php
 <?php
-$user_id = $I->grabFromCurrentUrl('~/user/(\d+)/~');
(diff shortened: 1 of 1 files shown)


Naktibalda:
Use ^ instead of $ to match the beginning of string.

eXorus:
Done

DavertMik:
Thank you

PR #3499, opened 2016-09-08
expected type for $param set for pressKey webdriver action.

iBotPeaches:
screen shot 2016-09-08 at 5 08 53 pm

Minor, but my IDE complains that its expecting some random type `Can`, because a description was written on a doc, without a type. This adds the expected string or array.


Naktibalda:
thanks

PR #3504, opened 2016-09-09
Update 09-Data.md

pavelkovar:
Changed link to HTML, since it is not displayed correctly.

### DIFF
--- a/docs/09-Data.md
+++ b/docs/09-Data.md
@@ -22,7 +22,7 @@ modules:
 
 
 <div class="alert alert-notice">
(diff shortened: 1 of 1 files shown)


Naktibalda:
Please change it to `[module params](http://codeception.com/docs/06-ModulesAndHelpers#Dynamic-Configuration-With-Params)`

pavelkovar:
I tried to use markdown syntax first, but it doesn't work in github viewer because link is located in _div_ element.

Naktibalda:
OK, thanks

Issue #2062, opened 2015-06-18
v2.0.9 selectOption() throws InvalidArgumentException: Unreachable field ""


daison12006013:
So I am using a chrome extension to get the exact xpath, I was able to use ``grabTextFrom()`` function to show the lists of options in the select tag. When trying to use ``grabValueFrom()`` it returns a ``NULL``.

Here's my codeception code
```
$el = "//div[@class='formBlock']/div[@class='row form-group'][2]/div[@class='col-sm-4'][1]/div[@class='form-group']/select[@class='citySelection form-control input-sm']";
//dd($this->I->grabTextFrom($el));
$city_name = $this->cities[array_rand($this->cities)];
$city = City::where('name', '=', $city_name)->first();
$this->I->selectOption($el, $city->id);
````

Here's the output fail file.
``http://pastebin.com/raw.php?i=VNc7Z6Y8``

When using the ``grabTextForm($el)``:
``http://pastebin.com/raw.php?i=M8Xnd1q9``

Thrown InvalidArgumentException from CLI:
``http://pastebin.com/raw.php?i=FT8FeRDq``

This is an acceptance test, this is the config in acceptance.suit.yml:
````
class_name: AcceptanceTester
modules:
    enabled:
        - PhpBrowser
        - AcceptanceHelper
        - Laravel4
        - Db
    config:
        PhpBrowser:
            url: '<I didn't include here, for security purposes>'
        Laravel4:
            cleanup: false
````

Using the input tag works well, while this select tag, really not working, I also tried creating a simple html just to test, but still 2.0.9 was failing to select a value.

Fyi,
I'm using codeception 2.0.9 for my old Laravel4 Framework Project.

zbateson:
As it is nothing's jumping out at me specifically... could you send the simple test you created instead?  That would probably be much easier to see what's going on and if it's a bug, create a test case out of :wink:

daison12006013:
Hi @zbateson, I did blame to your commit in the file src/Codeception/Lib/InnerBrowser.php at line 693, as per checking you're using ``$fieldName = $this->getSubmissionFormFieldName($field->attr('name'));``

While checking function ``getSubmissionFormFieldName()``, it seems it removes the last 2 characters which it is ``[]``

As you can see, most of the fields that i'm using they're inputs with array.

As per checking the ``fillField()`` function, it doesn't call this ``getSubmissionFormFieldName()`` function, that's why ``fillField()`` works well.

[...]

PR #3505, opened 2016-09-10
Add another test case for <select> names with square brackets

Suseika:
I'm having an issue similar to this one: https://github.com/Codeception/Codeception/issues/2062

I'm getting `[InvalidArgumentException] Unreachable field "product"` exception in my tests, and the test cases added in https://github.com/Codeception/Codeception/pull/2077 don't seem to be covering my case.

I couldn't run the tests for this project locally to check if the issue is reproduced with this test, so I'm sending this pull request to see what CI will show.

### DIFF
--- a/tests/data/app/view/form/names-sq-brackets.php
+++ b/tests/data/app/view/form/names-sq-brackets.php
@@ -35,7 +35,7 @@
             <option value="4">Select 4</option>
             <option value="5">Select 5</option>
         </select>
-        <select name="product[stock_data][is_in_stock]">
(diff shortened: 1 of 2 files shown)


Suseika:
Looks like the problem is on my side. Feel free to close this PR if you don't need it.

Suseika:
Looks like the problem is that my `<select>` was outside any form. If only the error message was more specific, or there was a `@throws` in phpdoc for the generated `AcceptanceTester::selectOption()`, it could have saved me a lot of debugging time : (

PR #3506, opened 2016-09-10
fix for then your pSQL sequences are not standard format

alexjeen:
This fixed a problem when your sequences are not a standard format (ie. table_id_seq).

### DIFF
--- a/src/Codeception/Lib/Driver/PostgreSql.php
+++ b/src/Codeception/Lib/Driver/PostgreSql.php
@@ -123,16 +123,38 @@ class PostgreSql extends Db
         }
     }
 
+	/**
+	* Get the last inserted ID of table
+	*/
     public function lastInsertId($table)
     {
-        /**
+        /*
          * We make an assumption that the sequence name for this table
          * is based on how postgres names sequences for SERIAL columns
          */
-        $sequenceName = $this->getQuotedName($table . '_id_seq');
-        return $this->getDbh()->lastInsertId($sequenceName);
+                
+        $sequenceName = $this->getQuotedName($table.'_id_seq');
+        $lastSequence = null;
+
+		try {
+	        $lastSequence = $this->getDbh()->lastInsertId($sequenceName);
+		} catch (\PDOException $e) {
+			// in this case, the sequence name might be combined with the primary key name
+		}
+		
+		// here we check if for instance, it's something like table_primary_key_seq instead of table_id_seq
+		if(!$lastSequence) {
+			$primaryKeys = $this->getPrimaryKey($table);
+			$pkName = array_shift($primaryKeys);
+	        $lastSequence = $this->getDbh()->lastInsertId($this->getQuotedName($table.'_'.$pkName.'_seq'));
+		}	
+        
+        return $lastSequence;
     }
 
+	/**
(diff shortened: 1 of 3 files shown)


Naktibalda:
1. Please convert tabs to spaces and delete that unnecessary space between `[]` and `=` in `$primaryKey [] = $column['attname'];`
2. Could you add a test for non-standard sequence to https://github.com/Codeception/Codeception/blob/2.2/tests/unit/Codeception/Lib/Driver/PostgresTest.php ?
3. Please explain the reason to change query in `getPrimaryKey` method. The original query was copied from https://wiki.postgresql.org/wiki/Retrieve_primary_key_columns and yours isn't. (Add test for table not supported by original query if it isn't covered by the test for non-standard sequence).

alexjeen:
Hi there,

1. I've let my sublime handle the PSR requirement (not in the first case because it often messed up the CodeCeption docs). I hope it's good now.
2. I've added the test but have yet to figure out how to run it (spent an hour trying to get it to work)
3. Your query was right, my query was based on "one schema" so I've reverted it again.

src/Codeception/Lib/Driver/PostgreSql.php
```diff
@@ -42,12 +48,12 @@ public function load($sql)
                 $this->searchPath = $match[0];
             }
 
-            $query .= "\n" . rtrim($sqlLine);
+            $query .= "\n".rtrim($sqlLine);
```
Naktibalda: Please put spaces around concatenation operator. This isn't specified by PSR-2, but it is common convention followed by many coding standards and elsewhere in Codeception.

src/Codeception/Lib/Driver/PostgreSql.php
```diff
@@ -162,9 +196,8 @@ public function getPrimaryKey($tableName)
                 AND    i.indisprimary';
             $stmt = $this->executeQuery($query, [$tableName]);
             $columns = $stmt->fetchAll(\PDO::FETCH_ASSOC);
-
             foreach ($columns as $column) {
-                $primaryKey []= $column['attname'];
+                $primaryKey [] = $column['attname'];
```
Naktibalda: Please delete one space around [], you pick the side.

Naktibalda:
To run the tests you need a running postgresql daemon, schema named `codeception_test`
and passwordless user named `postgres` (or you can change credentials in PostgresTest.php file)

Run the tests using command
`php codecept run tests/unit/Codeception/Lib/Driver/PostgresTest.php -vv`

alexjeen:
Alright, I've commited and pushed. All 10 tests pass.

tests/unit/Codeception/Lib/Driver/PostgresTest.php
```diff
@@ -124,6 +124,12 @@ public function testGetEmptyArrayIfTableHasNoPrimaryKey()
         $this->assertEquals([], $this->postgres->getPrimaryKey('no_pk'));
     }
 
+    public function testLastInsertIdReturnsSequenceValueWhenNonStandardSequenceNameIsUsed()
```
Naktibalda: I replaced your test with a test which actually tests the change in lastInsertId.

PR #3507, opened 2016-09-10
Escape tags in debug output

### LINKED ISSUE #3495: Console formatter processes excess tags from debug responses
I use REST module and receive xml http responses. Some of them contain <error> tag. And with debug module enabled I see red formatted text in the middle of the http response without error tag itself. This is confusing.
I know tags with such formatting are used to improve readability of console output. But can you output responses in «raw» format?

Codeception version 2.2.4

Naktibalda:
I found `Symfony\Component\Console\Formatter\OutputFormatter::escape` and added it to all relevant places that I could find:
* codecept_debug
* diff output
* assertion messages and stack traces
* step parameters

Fixes #3495

### DIFF
--- a/src/Codeception/Lib/Console/Colorizer.php
+++ b/src/Codeception/Lib/Console/Colorizer.php
@@ -1,9 +1,8 @@
 <?php
 namespace Codeception\Lib\Console;
 
-/**
- * Colorizer
- **/
+use Symfony\Component\Console\Formatter\OutputFormatter;
+
 class Colorizer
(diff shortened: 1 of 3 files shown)


DavertMik:
Looks good! :+1:

Naktibalda:
Today I tested and escaped tags in the output of failed step in MetaStep (by using Step Object).
I think that's everything.

DavertMik:
Cool, then I'm merging

Issue #3513, opened 2016-09-14
Codception ?


adrian-enspired:
Just wondering whether [this spelling is intentional](https://github.com/Codeception/Codeception/blob/2.2/tests/unit/Codeception/ApplicationTest.php#L3).

* Codeception version: 2.2.3

oehlschl:
Looks unintentional. Just posted a fix for `Codception` misspellings at #3515

PR #3515, opened 2016-09-14
bugfix 3513: fix codception misspellings

oehlschl:
Bugfix for #3513 -- misspellings in doc and namespace.

Neither was referenced anywhere else, so changes are non-breaking.

Asserted that unit test passed after changes via `codecept run tests/unit/Codeception/ApplicationTest.php`


Naktibalda:
Thanks

PR #3518, opened 2016-09-15
Fix bug in Coverage Filter related to relative filepaths

sbacic:
This fixes an edge-case bug with regular filepaths (not asterixed wildcards). 

And here is the related issue describing the bug in more detail: https://github.com/Codeception/Codeception/issues/3519

### DIFF
--- a/src/Codeception/Coverage/Filter.php
+++ b/src/Codeception/Coverage/Filter.php
@@ -73,7 +73,7 @@ class Filter
         if (isset($coverage['whitelist']['include'])) {
             foreach ($coverage['whitelist']['include'] as $fileOrDir) {
                 $finder = strpos($fileOrDir, '*') === false
-                    ? [$fileOrDir]
+                    ? [Configuration::projectDir() . DIRECTORY_SEPARATOR . $fileOrDir]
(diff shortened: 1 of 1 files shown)


src/Codeception/Coverage/Filter.php
```diff
@@ -73,7 +73,7 @@ public function whiteList($config)
         if (isset($coverage['whitelist']['include'])) {
             foreach ($coverage['whitelist']['include'] as $fileOrDir) {
                 $finder = strpos($fileOrDir, '*') === false
-                    ? [$fileOrDir]
+                    ? [Configuration::projectDir() . DIRECTORY_SEPARATOR . $fileOrDir]
```
Naktibalda: I think that this code creates an incorrect path when codeception.yml is included. #3433 
@loren-osborn what do you think about it?

sbacic:
Good news! I've managed to reproduce the bug in a clean environment. Here's the repo: https://github.com/sbacic/cctest

It seems that using chdir() in setUpBeforeClass is causing issues with Codeception's ability to match relative paths in codecept.yml.

Naktibalda:
I tested this fix and it works correctly, modified code is consistent with matchWildcardPattern method.

PR #3522, opened 2016-09-16
fixes #3520 nested transactions in yii2 module

### LINKED ISSUE #3520: Nested transactions in Yii2 unit tests

I use Yii2 Codeception module and ActiveFixtures for unit tests. It's required before each test begin a transaction in the database and rollback it after the test.

I use option **cleanup** to wrapping each test within transaction.

#### What do you get instead?

My DB is MySQL and it does not supports ALTER TABLE within transactions. But the ActiveFixture model clears all data in table including a sequence.

If my code has no nested transactions it's ok, but if testing code has nested transaction I saw PDOException like this:

```bash
There were 2 errors:

---------
1) UserManagerTest: Create admin
 Test  codeception/unit/UserManagerTest.php:testCreateAdmin
                                                                                                                                                                       
  [yii\db\Exception] SQLSTATE[42000]: Syntax error or access violation: 1305 SAVEPOINT LEVEL1 does not exist
The SQL being executed was: ROLLBACK TO SAVEPOINT LEVEL1  
                                                                                                                                                                       
#1  /Users/max/www/comitka/web/vendor/yiisoft/yii2/db/Command.php:855
#2  /Users/max/www/comitka/web/vendor/yiisoft/yii2/db/Schema.php:423
#3  /Users/max/www/comitka/web/vendor/yiisoft/yii2/db/Transaction.php:189
#4  /Users/max/www/comitka/web/app/modules/user/UserModule.php:98
#5  /Users/max/www/comitka/web/app/modules/user/UserModule.php:54
#6  /Users/max/www/comitka/web/app/tests/codeception/unit/UserManagerTest.php:55
#7  UserManagerTest->testCreateAdmin

---------
2) UserManagerTest: Create user
 Test  codeception/unit/UserManagerTest.php:testCreateUser
                                                                                                                                                                   
  [yii\db\Exception] SQLSTATE[42000]: Syntax error or access violation: 1305 SAVEPOINT LEVEL1 does not exist
The SQL being executed was: RELEASE SAVEPOINT LEVEL1  
                                                                                                                                                                   
#1  /Users/max/www/comitka/web/vendor/yiisoft/yii2/db/Command.php:855
#2  /Users/max/www/comitka/web/vendor/yiisoft/yii2/db/Schema.php:414
#3  /Users/max/www/comitka/web/vendor/yiisoft/yii2/db/Transaction.php:159
#4  /Users/max/www/comitka/web/app/modules/user/UserModule.php:428
#5  /Users/max/www/comitka/web/app/tests/codeception/unit/UserManagerTest.php:290
#6  UserManagerTest->testCreateUser
```

```php
<?php
namespace models;

use Codeception\Test\Unit;
use tests\codeception\fixtures\UserFixture;
use UnitTester;
use user\models\User;

/**
 * Test user manager
 */
class UserManagerTest extends Unit
{
    /**
     * @var UnitTester
     */
    protected $tester;

    /**
     * @inheritdoc
     */
    public function setUp()
    {
        parent::setUp();
        $this->getModule('Yii2')->haveFixtures([
            'users' => UserFixture::className(),
        ]);
    }

    /**
     * Test model validation and save
     *
     * @return User
     */
    public function testMe()
    {
        // assertions here
    }
}
```

### Details

* Codeception version: dev-master
* PHP Version: 5.6.15
* Operating System: OS X El Capitan
* Installation type: Composer
* Suite configuration: 

```yml
class_name: UnitTester
modules:
    enabled:
        - Yii2:
            configFile: 'codeception/config/unit.php'
            part: init
            cleanup: true
        - Asserts
        - \Helper\Unit
```

kalyabin:
I think it solves a problem discussed in #3520

MySQL does not supports ALTER TABLE within transactions. But the ActiveFixture model clears all data in table including a sequence before test.

If use cleanup option each unit test with nested transactions throws PDOException.

### DIFF
--- a/src/Codeception/Module/Yii2.php
+++ b/src/Codeception/Module/Yii2.php
@@ -10,6 +10,7 @@ use Codeception\Lib\Interfaces\ActiveRecord;
 use Codeception\Lib\Interfaces\PartedModule;
 use Codeception\Lib\Notification;
 use Codeception\TestInterface;
+use Codeception\Test\Unit as UnitTest;
 use Yii;
 use yii\db\ActiveRecordInterface;
 
@@ -142,7 +143,7 @@ class Yii2 extends Framework implements ActiveRecord, PartedModule
         $this->app = $this->client->getApplication();
 
         // load fixtures before db transaction
-        if (method_exists($test, 'fixtures') && ($fixtures = $test->fixtures()) !== false) {
+        if ($test instanceof UnitTest && method_exists($test, 'fixtures') && ($fixtures = $test->fixtures()) !== false) {
             $this->haveFixtures($fixtures);
         }
 
--- a/src/Codeception/Module/Yii2.php
+++ b/src/Codeception/Module/Yii2.php
@@ -141,6 +141,11 @@ class Yii2 extends Framework implements ActiveRecord, PartedModule
         $this->client->configFile = Configuration::projectDir() . $this->config['configFile'];
         $this->app = $this->client->getApplication();
 
+        // load fixtures before db transaction
+        if (method_exists($test, 'fixtures') && ($fixtures = $test->fixtures()) !== false) {
+            $this->haveFixtures($fixtures);
+        }
+
         if ($this->config['cleanup'] && $this->app->has('db')) {
             $this->transaction = $this->app->db->beginTransaction();
         }


Naktibalda:
@samdark Please merge if you approve it.

src/Codeception/Module/Yii2.php
```diff
@@ -141,6 +141,11 @@ public function _before(TestInterface $test)
         $this->client->configFile = Configuration::projectDir() . $this->config['configFile'];
         $this->app = $this->client->getApplication();
 
+        // load fixtures before db transaction
+        if (method_exists($test, 'fixtures') && ($fixtures = $test->fixtures()) !== false) {
```
DavertMik: Should only work for `Codeception\Test\Unit` instances.

DavertMik:
Thanks! 

Looks like I didn't think of using old `fixtures()` methods, as I think `haveFixtures` could be a cleaner way to load fixtures. But yeah, I understand `fixtures` method is a nice and habitual. So why not to use it. However, it should not be used with Cests or other test cases where methods can accidentally be tests by themsleves

kalyabin:
I properly understand that my code has collision with yii2-codeception extension? 

I see that FixtureTrait also uses this method with yii2-codeception extension. May be to use method with other name, fixtureConfig for example? And add a check by instance type, of course.

src/Codeception/Module/Yii2.php
kalyabin: fixed at https://github.com/Codeception/Codeception/pull/3522/commits/04cd17647c64e2920e3b597e9ca411513a8f9aa0

samdark:
Overall `yii2-codeception` is to be deprecated. @kalyabin could these changes cause some scenarios to fail now?

kalyabin:
@samdark these changes could not be cause for any failures, if do not use FixtureTrait with Unit tests. Fixtures will be loaded twice in one Unit test if use it with FixtureTrait.

DavertMik:
Ok, I will partially accept the PR. 

I think this use case can also be useful for Cest files, but in this case special methods should stard with underscore. So I will rename fixtures to `_fixtures`, to mark it to be a special method.
I will also update a documentation for that

githubjeka:
`Codeception version 2.2.6`
Yii module enabled:
My code:
```php
public function create() :Node
    {
        $tr = \Yii::$app->db->beginTransaction();
        // ... 
        $tr->commit();

        return $nodeModel;
    }
```

Functional test fail
```
Test  ..\backend/tests\functional\NodeCrudCest.php:tryToCreateNodeTest
  [yii\db\Exception] SQLSTATE[42000]: Syntax error or access violation: 1305 SAVEPOINT LEVEL1 does not exist
The SQL being executed was: RELEASE SAVEPOINT LEVEL1
```

for this code test passed:
```php
public function create() :Node
    {
        // $tr = \Yii::$app->db->beginTransaction();
        // ... 
        // $tr->commit();

        return $nodeModel;
    }
```

PR #3524, opened 2016-09-16
Using Doctrine2 module with ZF2 as partly loaded results into an error

chris1312:
Configuration:

```
class_name: AcceptanceTester
modules:
    enabled: [PhpBrowserHelper, \Helper\Acceptance, Asserts, Doctrine2, DbHelper, RestHelper]
    config:
        PhpBrowserHelper:
            url: 'http://localhost/'
        Doctrine2:
            depends: ZF2
```

Error:
```
Call to a member function grabServiceFromContainer() on a non-object in 
/var/www/localhost/vendor/codeception/codeception/src/Codeception/Module/ZF2.php:157
```

### DIFF
--- a/src/Codeception/Module/ZF2.php
+++ b/src/Codeception/Module/ZF2.php
@@ -135,6 +135,11 @@ class ZF2 extends Framework implements DoctrineProvider, PartedModule
 
     public function _getEntityManager()
     {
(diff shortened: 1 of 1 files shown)


Naktibalda:
Thanks

Issue #3531, opened 2016-09-20
[WebDriver] Improve stability of Codeception when Selenium Server is gone


eXorus:
#### What are you trying to achieve?

I'm trying to improve the stability of Codeception when the Selenium Server is gone or the Selenium Hub lost the connection with the node.

I would like to always have WebDriver Exception catched by Codeception and continue to the next test like we already have when PHP Exception is throw.

```
1) WelcomeCest: Try to test3
 Test  tests/acceptance/WelcomeCest.php:tryToTest3

  [RuntimeException] Call to undefined method AcceptanceTester::methodXXX
```
#### What do you get instead?

Some exception are throws and the shell is returned without information. So build is failed in jenkins, no failed file generated, no test result updated when xml is asked and the others tests are not runned.

``` bash
[Facebook\WebDriver\Exception\WebDriverCurlException]                                      
  Curl error thrown for http GET to /session/3d3f9b91-49da-4523-b3bb-f939562b5f58/log/types  
  Failed to connect to localhost port 4446: Connection refused

[Facebook\WebDriver\Exception\UnknownServerException]                                            
  Error forwarding the new session Error forwarding the request 172.17.0.4:5555 failed to respond

[Facebook\WebDriver\Exception\UnknownServerException]                                           
  Session [f5dd0664-2c1b-4d20-b88e-e6d6c013b040] was terminated due to FORWARDING_TO_NODE_FAILED
```

> Provide console output if related. Use `-vvv` mode for more details.

``` bash
➜  ./vendor/bin/codecept run acceptance -vvv
Codeception PHP Testing Framework v2.2.4
Powered by PHPUnit 5.4.8 by Sebastian Bergmann and contributors.

Acceptance Tests (4) 
Modules: WebDriver, \Helper\Acceptance

WelcomeCest: Try to test
Signature: WelcomeCest:tryToTest
Test: tests/acceptance/WelcomeCest.php:tryToTest
Scenario --
 I am on url "http://codeception.com"
  [Host] http://codeception.com
 I wait 5
 I see "Elegant and Efficient Testing for PHP"
 I wait 5
 I click "Quick Start"
 I wait 5
 PASSED 

WelcomeCest: Try to test2
Signature: WelcomeCest:tryToTest2
Test: tests/acceptance/WelcomeCest.php:tryToTest2
Scenario --
 I am on url "http://codeception.com"
  [Host] http://codeception.com
 I wait 5
 I see "Elegant and Efficient Testing for PHP"


[...]

PR #3534, opened 2016-09-21
Improve the stability of WebDriver when Selenium server is gone

eXorus:
Fix issue #3531 

**Step to reproduce :**
- Launch Selenium Server
- Launch Codeception (Acceptance Test)
- Stop Selenium Server

Each time I stopped server  during the $I->wait() and before $I->acceptPopup()
This is all the the fix I need to do to be able to have an Codeception that never exit and that finish all the tests and give the feedback about how many errors ....

Can solved this issue:

- Failed to connect to localhost port 4446: Connection refused
- Error forwarding the request 172.17.0.4:5555 failed to respond
- Session xxx was terminated due to FORWARDING_TO_NODE_FAILED

	------------------------
	Exception Found
	WebDriver.php:xxx
	WebDriver.php:xxx

	Fix x
	method that throw exception => description what it does before fix
	description what it does after fix

	------------------------
	WebDriverCurlException
	WebDriver.php:389
	WebDriver.php:374

	Fix 1
	debugWebDriverLogs => catch only UnknownServerException for IE
	so now catch all exception

	------------------------
	WebDriverCurlException
	WebDriver.php:513
	WebDriver.php:377

	Fix 2
	_saveScreenshot => no catch, bad design
	so now I'm using Exception to trap it

	------------------------
	WebDriverCurlException
	WebDriver.php:537
	WebDriver.php:378

	Fix 3
	_savePageSource => no catch, bad design
	so now I'm using Exception to trap it

	------------------------
	WebDriverCurlException
	WebDriver.php:439
	WebDriver.php:364

	Fix 4
	cleanWebDriver => catch only UnknownServerException
	so now I'm using Exception to trap it

	------------------------
	ConnectionException
	WebDriver.php:308
	WebDriver.php:324

	Fix 5
	_initialize => catch only WebDriverCurlException to throw ConnectionException
	so now I'm replacing to catch all Exception, move the catch at the end and remove new exception

The last fix is not enough good for me:
If I throw ConnectionException, the script exit => bad
If I only print debug line => the test is runned and failed because WebDriver is null ([PHPUnit_Framework_Exception] Undefined property: Codeception\Module\WebDriver::$webDriver)

The best way was to mark the test has skipped but I didn't succeed to call this function in WebDriver module.

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -508,6 +508,7 @@ class WebDriver extends CodeceptionModule implements
     public function _saveScreenshot($filename)
     {
         try {
+            if ($this->webDriver === null) throw new Exception("WebDriver::_saveScreenshot method has been called when webDriver is not set");
             $this->webDriver->takeScreenshot($filename);
         } catch (\Exception $e) {
             $this->debug("Unable to retrieve screenshot from Selenium : ".$e->getMessage());
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -233,32 +233,32 @@ use Symfony\Component\DomCrawler\Crawler;
  *
  */
 class WebDriver extends CodeceptionModule implements
-WebInterface,
-RemoteInterface,
-MultiSessionInterface,
-SessionSnapshot,
-ScreenshotSaver,
-PageSourceSaver,
-ElementLocator,
-ConflictsWithModule,
-RequiresPackage
+    WebInterface,
+    RemoteInterface,
+    MultiSessionInterface,
+    SessionSnapshot,
+    ScreenshotSaver,
+    PageSourceSaver,
+    ElementLocator,
+    ConflictsWithModule,
+    RequiresPackage
 {
     protected $requiredFields = ['browser', 'url'];
     protected $config = [
-    'host'               => '127.0.0.1',
-    'port'               => '4444',
-    'restart'            => false,
-    'wait'               => 0,
-    'clear_cookies'      => true,
-    'window_size'        => false,
-    'capabilities'       => [],
-    'connection_timeout' => null,
-    'request_timeout'    => null,
-    'http_proxy'         => null,
-    'http_proxy_port'    => null,
-    'ssl_proxy'          => null,
-    'ssl_proxy_port'     => null,
-    'debug_log_entries'  => 15,
+        'host'               => '127.0.0.1',
+        'port'               => '4444',
+        'restart'            => false,
+        'wait'               => 0,
+        'clear_cookies'      => true,
+        'window_size'        => false,
+        'capabilities'       => [],
+        'connection_timeout' => null,
+        'request_timeout'    => null,
+        'http_proxy'         => null,
+        'http_proxy_port'    => null,
+        'ssl_proxy'          => null,
+        'ssl_proxy_port'     => null,
+        'debug_log_entries'  => 15,
     ];
 
     protected $wd_host;
@@ -302,7 +302,7 @@ RequiresPackage
                 $this->requestTimeoutInMs,
                 $this->httpProxy,
                 $this->httpProxyPort
-                );
+            );
             $this->sessions[] = $this->_backupSession();
             $this->webDriver->manage()->timeouts()->implicitlyWait($this->config['wait']);
             $this->initialWindowSize();
@@ -328,7 +328,7 @@ RequiresPackage
         $test->getMetadata()->setCurrent([
             'browser' => $this->config['browser'],
             'capabilities' => $this->config['capabilities']
-            ]);
+        ]);
     }
 
     protected function loadFirefoxProfile()
@@ -342,7 +342,7 @@ RequiresPackage
             throw new ModuleConfigException(
                 __CLASS__,
                 "Firefox profile does not exist under given path " . $firefox_profile
-                );
+            );
         }
         // Set firefox profile as capability
         $this->capabilities['firefox_profile'] = file_get_contents($firefox_profile);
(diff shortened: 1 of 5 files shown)


src/Codeception/Module/WebDriver.php
```diff
@@ -304,13 +304,11 @@ public function _initialize()
                 $this->httpProxyPort
             );
             $this->sessions[] = $this->_backupSession();
-        } catch (WebDriverCurlException $e) {
-            throw new ConnectionException(
-                $e->getMessage() . "\n \nPlease make sure that Selenium Server or PhantomJS is running."
-            );
+            $this->webDriver->manage()->timeouts()->implicitlyWait($this->config['wait']);
+            $this->initialWindowSize();
```
Naktibalda: Could you explain these 2 lines?

src/Codeception/Module/WebDriver.php
```diff
@@ -511,11 +507,13 @@ public function _getCurrentUri()
 
     public function _saveScreenshot($filename)
     {
-        if ($this->webDriver !== null) {
+        try {
+            if ($this->webDriver === null) {
+                throw new \Exception("WebDriver::_saveScreenshot method has been called when webDriver is not set");
```
Naktibalda: I am not sure if it is a very good idea to throw exception and catch it immediately,
wouldn't it be better to print some debug output and return?

Naktibalda:
@eXorus I have added spaces around concatenation operator, please pull before making more changes.

src/Codeception/Module/WebDriver.php
eXorus: These 2 lines was already existing but I put them in the try to catch exception from Selenium.

src/Codeception/Module/WebDriver.php
eXorus: I agree with you but I didn't know how to do that. I don't want to repeat the error.

```php
try {
if ($this->webDriver === null) {
throw new \Exception("WebDriver::_saveScreenshot method has been called when webDriver is not set");
}
$this->webDriver->takeScreenshot($filename);
 } catch (\Exception $e) {
 $this->debug("Unable to retrieve screenshot from Selenium : ".$e->getMessage());
}
```

or I can do that if you prefer
```php
if ($this->webDriver === null) {
$this->debug("WebDriver::_saveScreenshot method has been called when webDriver is not set");
}

try {
$this->webDriver->takeScreenshot($filename);
 } catch (\Exception $e) {
 $this->debug("Unable to retrieve screenshot from Selenium : ".$e->getMessage());
}
```

eXorus:
And could you help me for the skipped test in initialize() ? because for the moment this PR can't be merged.

src/Codeception/Module/WebDriver.php
Naktibalda:
```php
if ($this->webDriver === null) {
    $this->debug("WebDriver::_saveScreenshot method has been called when webDriver is not set");
    return;
}
```

src/Codeception/Module/WebDriver.php
Naktibalda: oh, it is fine then.

Naktibalda:
I updated `_before` method, please test.

eXorus:
It's working well:
The first one failed: [Facebook\WebDriver\Exception\UnknownServerException] cannot forward the request Connect to 172.17.0.6:5555 [/172.17.0.6] failed: Connection refused
And the others are skipped.
Tests: 101, Assertions: 597, Errors: 2, Skipped: 91.

Tests are OK for me

Naktibalda:
I think that it would be better to allow _initialize fail hard. Because it usually means that Selenium is not running.

This PR must be approved by @davert.

eXorus:
I understand it's a choice to do. 
Fail hard or fail for the first and skip the others tests.

DavertMik:
Thanks. 
I made a few tries and I think this can't be done in current way. 
If selenium is not running - all tests are skipped, no errors. This means that the overall result is OK, and no failures in exit code. It's good to skip other tests but at least one test should fail to indicate problem

eXorus:
Sorry I'm not able to do it.
I don't know how to failed $test in the method _before() in the module WebDriver.

Any help ?

Naktibalda:
I think that we have to introduce new methods `setFail` and `getFail` to 
[Codeception\Test\Metadata](https://github.com/Codeception/Codeception/blob/0e268a10e2e5c7d640a70ddf7dc7d558f2d3b0c1/src/Codeception/Test/Metadata.php), similar to `setSkip` and `setIncomplete`.

A handling of failure should be added to [IgnoreIfMetadataBlocked feature](https://github.com/Codeception/Codeception/blob/0e268a10e2e5c7d640a70ddf7dc7d558f2d3b0c1/src/Codeception/Test/Feature/IgnoreIfMetadataBlocked.php#L20) and [Unit format](https://github.com/Codeception/Codeception/blob/146b58e6d53008dbe289ab59ae9dd7cb80415c65/src/Codeception/Test/Unit.php#L34).

eXorus:
lol, I just begin to do that before you read ... I will continue

Naktibalda:
@DavertMik it is good now, right?

DavertMik:
There are some internal changes. I see what you are up to but probably I have a better solution. https://github.com/Codeception/Codeception/pull/3598 This will allow us to throw exceptions in _before if it is needed. 

Please revert your changes to metadata behavior and merge with my branch (or just copy-paste my code, actually). Then WebDriver._before will be able just to throw exception in case a connection can't be established. `$this->fail` would be ok

DavertMik:
Also it should be noted and not forgotten. This breaks AngularJS module. Execute `codecept run angular` to see output

DavertMik:
Thanks! That was a long story but finally it is ready to be merged :+1:

eXorus:
Thanks for your help @DavertMik and @Naktibalda 
Each day I learn more about Codeception :)

PR #3345, opened 2016-07-19
Don't fail test validation when exec function is disabled

Naktibalda:
Fixes #3324 

I haven't tested this change - I made it using GitHub editor.

PR #3536, opened 2016-09-21
Don't fail test validation when exec function is disabled (#3345)

### LINKED ISSUE #3324: Fatal warning in Parser::validate
codecept version 2.2.2
execute codecept.phar run command prompt:
[PHPUnit_Framework_Exception]  Undefined variable: output

mklemarczyk:
Fixes #3324
It is much more elegant solution than @exec. If you want warn user about that validation was not performed. You should do this on entry point, but not each time when this function is being executed.

### DIFF
--- a/src/Codeception/Lib/Parser.php
+++ b/src/Codeception/Lib/Parser.php
@@ -131,7 +131,7 @@ class Parser
         if (empty($config['settings']['lint'])) { // lint disabled in config
             return;
         }
-        if(function_exists("exec")) {
+        if (function_exists("exec")) {
             exec("php -l ".escapeshellarg($file)." 2>&1", $output, $code);
         }
         if ($code !== 0) {
--- a/src/Codeception/Lib/Parser.php
+++ b/src/Codeception/Lib/Parser.php
@@ -131,7 +131,9 @@ class Parser
         if (empty($config['settings']['lint'])) { // lint disabled in config
             return;
         }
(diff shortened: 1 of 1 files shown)


src/Codeception/Lib/Parser.php
```diff
@@ -131,7 +131,9 @@ public static function validate($file)
         if (empty($config['settings']['lint'])) { // lint disabled in config
             return;
         }
-        exec("php -l ".escapeshellarg($file)." 2>&1", $output, $code);
+        if(function_exists("exec")) {
```
mklemarczyk: Fixed in 5df1255

Naktibalda:
I am moving discussion here from closed issue https://github.com/Codeception/Codeception/pull/3345#issuecomment-248819173

@mklemarczyk wrote that "The yii2 framework in version 2.0.9 (current stable) require Codeception 2.1.x version for test execution. I do not have any possibility to change that."

@mklemarczyk could you provide some evidence about that?
yii2 framework does not require Codeception directly or indirectly: https://github.com/yiisoft/yii2/blob/2.0.9/composer.json#L63

I created yii-app-basic project with yii 2.0.9 and successfully installed Codeception 2.2.4.

Naktibalda:
@mklemarczyk Thanks for idea to use function_exists, 

but I am closing this pull request, because 2.1 branch is dead for a long time, please upgrade to 2.2.

Issue #3472, opened 2016-08-29
Group Files not working with a non-empty data provider (regression since #3311)


motin:
This is just strange. Most of our unit tests were suddenly not executed after the upgrade from 2.2.2 to 2.2.3. Tracked this down that the test method that has a dataprovider (that provides some data) are ignored when using a group argument.

This 
```
<?php

class FooTest extends \Codeception\Test\Unit
{

    /**
     * @var \DbAgnosticCodeGuy
     */
    protected $guy;

    protected function _before()
    {
    }

    protected function _after()
    {
    }

    // tests


    public function dataProviderFoo()
    {
        return [
        ];
    }

    /**
     * @group coverage:full
     * @dataProvider dataProviderFoo
     */
    public function testMe($foo = 'bar')
    {
        $this->assertEquals('foo', $foo);
    }

}
```

... but this one does not work (the testMe is ignored). The only difference is that the dataprovider provides some data. 
```
<?php

class FooTest extends \Codeception\Test\Unit
{

    /**
     * @var \DbAgnosticCodeGuy
     */
    protected $guy;

    protected function _before()
    {
    }

    protected function _after()
    {
    }

    // tests


    public function dataProviderFoo()
    {
        return [
            [],
        ];
    }

    /**
     * @group coverage:full
     * @dataProvider dataProviderFoo
     */
    public function testMe($foo = 'bar')
    {
        $this->assertEquals('foo', $foo);
    }

}
```

Git bisecting showed that this behavior started after #3311 was merged.

Naktibalda:
@eXorus you caused it, can you fix it?

eXorus:
@Naktibalda of course I will try to fix it today. Sorry for that

Naktibalda:
Fixed

PR #3538, opened 2016-09-22
Fix issue #3472

eXorus:
Fix issue #3472


Naktibalda:
Thanks for the quick fix.
However I have no idea what is the effect of your change.

Could you implement a test for this issue in https://github.com/Codeception/Codeception/blob/2.2/tests/cli/RunCest.php ?

More information in https://github.com/Codeception/Codeception/tree/2.2/tests#cli

Issue #3263, opened 2016-06-22
Laravel5 Undefined index: HTTP_HOST issue


mohamed-atef:
I'm facing an exception of "[ErrorException] Undefined index: HTTP_HOST" while running API test with codeception 2.2 with Laravel module.
Everything was working just fine before but after upgrading to the new edition this exception has been thrown.
thanks,
Mohamed.

Naktibalda:
Don't use $_SERVER in your code, retrieve headers from request object.
https://laravel.com/api/5.2/Illuminate/Http/Request.html#method_header

RomaGilyov:
Have the same issue, didn't use $_SERVER in my code, getting all headers from the Request object only

janhenkgerritsen:
Can you provide your Codeception and suite configuration files?

And what version of Laravel are you using?

Do you get this error for all tests or only in specific tests?

Please provide additional details to make it a bit easier for us to find out what the problem could be.

mohamed-atef:
Hello,
[api.suite.yml.zip](https://github.com/Codeception/Codeception/files/329751/api.suite.yml.zip)

- I'm using Laravel Framework version 5.2.39.
- Yes, I'm getting this error for all tests, not just specific one.

thanks

janhenkgerritsen:
You should remove the `url` configuration option for the REST module from your api suite config file. If your tests do depend on an URL for the application you should set it through the `url` configuration option for the Laravel5 module.

I don't know why this does not give problems in earlier versions of Codeception and starts giving errors for Codeception 2.2 though. Any idea @Naktibalda or @DavertMik?

mohamed-atef:
I have removed the "URL" configuration option and hardcoded the URL inside the test itself to avoid this conflict but unfortunately got the same error

janhenkgerritsen:
I created a clean install of the latest versions of Laravel 5 and Codeception, you can view the code at https://github.com/janhenkgerritsen/codeception-issue-3263.

I added an API suite with a sample test `CheckErrorCept`. The API suite configuration looks like this:

```
class_name: ApiTester
modules:
    enabled:
        - \Helper\Api
        - Asserts
        - Filesystem
        - Laravel5:
            environment_file: .env.test
            url: http://test-api.dev
        - REST:
            depends: Laravel5
```

Like this, everything works as expected. But if you move the `url` configuration option from the Laravel5 section to the REST section you will get the error you mention. So to reproduce your error you should have the following in the API suite configuration file:

```
[...]

PR #3539, opened 2016-09-22
Fix issue with REST module resetting server parameters

janhenkgerritsen:
If the REST module is used with the Laravel5 module and the REST module is configured after the Laravel5 module in the suite configuration file, this method will overwrite the server variables set by the Laravel5 module, in this case HTTP_HOST.

My guess is that this method call should have been removed when the `DependsOnModule` functionality for the REST module was implemented, however, I am not really familiar with the REST module, so please review this change before merging.

The issue fixed with this PR is the root cause for the problem in #3263.

### DIFF
--- a/src/Codeception/Module/REST.php
+++ b/src/Codeception/Module/REST.php
@@ -108,9 +108,6 @@ EOF;
         $this->params = [];
         $this->response = "";
         $this->connectionModule->headers = [];
-        if ($this->client) {
-            $this->client->setServerParameters([]);
-        }
     }
 
     public function _conflicts()


Naktibalda:
Your change reverts #1232 and could cause issues for modules that do not recreate client in _before:
Phalcon and PhpBrowser.

janhenkgerritsen:
What about keeping the `HTTP_HOST` variable if it is set? Any possible issues with that? That does fix my issue with the Laravel 5 module.

DavertMik:
Maybe we should update PHPBrowser and Phalcon to do that?

Naktibalda:
Sorry, it was a false alarm, InnerBrowser unsets $this->client in _after, so there is no reason to resetServerParameters in _before.

PR #3311, opened 2016-07-05
Group Files not working with @example

eXorus:
Fix #3278 

acceptance/welcomeCest.php :
```php
class welcomeCest
{
    public function testCase1(AcceptanceTester $I)
    {
        $I->amOnUrl('https://www.google.fr');
    }
    /**
    * @group grouptc2
    */
    public function testCase2(AcceptanceTester $I)
    {
        $I->amOnUrl('https://www.google.fr');
    }

    /**
    * @group grouptc3
    * @example (url="https://www.google.fr")
    * @example (url="https://www.bing.com")
    * @example (url="https://duckduckgo.com")
    */
    public function testCase3(AcceptanceTester $I, $scenario, \Codeception\Example $example)
    {
        $I->amOnUrl($example['url']);
    }
    /**
    * @example (url="https://www.google.fr")
    * @example (url="https://www.bing.com")
    * @example (url="https://duckduckgo.com")
    */
    public function testCase4(AcceptanceTester $I, $scenario, \Codeception\Example $example)
    {
        $I->amOnUrl($example['url']);
    }
}
```

acceptance.suite.yml:
```yml
groups:
    grouptc4: tests/_data/grouptc4
```
_data/grouptc4:
```
tests/acceptance/welcomeCest.php
```

**Result before the fix:**
run => tc1, tc2, tc3, tc4
run -g grouptc2 => tc2
run -g grouptc3 => tc3
run -g grouptc4 =>  tc1, tc2

**Result after the fix:**
run => tc1, tc2, tc3, tc4
run -g grouptc2 => tc2
run -g grouptc3 => tc3
run -g grouptc4 =>  tc1, tc2, tc3, tc4

PR #3541, opened 2016-09-22
Test running test with data provider by group

Naktibalda:
Test for PR  #3538


Naktibalda:
The test failed because the fix is not merged yet.

eXorus:
Thanks i did the same test yesterday but I didn't push it because I was working on others tests to cover more like #3311

PR #3546, opened 2016-09-25
[WebDriver] Improved exception message thrown when click('name') does not match any element

### LINKED ISSUE #3528: WebDriver throws misleading exception for missing element
For the following steps:

```php
$I->amOnUrl("https://www.google.com");
$I->click("Link Text");
```

The following failure is seen (as expected):

```
Step  Click "Link Text"
Fail  Link or Button or CSS or XPath element with 'Link Text' was not found.
```

However this works as expected because `Codeception\Util\Locator::isCSS("Link Text")` returns true, so the call to `$this->match` in `WebDriver::click` is returning an empty list rather than a MalformedLocatorException.

So if the following steps are used instead:

```php
$I->amOnUrl("https://www.google.com");
$I->click("Link 2");
```

Since "Link 2" does not match the same `isCSS` check, `match` throws the exception.

This is equivalent of #2408 for WebDriver, and was not properly fixed by #2960/#2966.

I am aware that `$I->click(["link" => "Link 2"]);` will work, but this is about fixing the exception message.

Naktibalda:
New message matches message thrown by InnerBrowser and looks like this:
```php
$I->click('Link 2');
```
> 'Link 2' is invalid CSS and XPath selector and Link or Button element with 'name=Link 2' was not found.

Fixes #3528

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -734,7 +734,11 @@ class WebDriver extends CodeceptionModule implements
         }
         $el = $this->findClickable($page, $link);
         if (!$el) {
-            $els = $this->match($page, $link);
+            try {
+                $els = $this->match($page, $link);
+            } catch (MalformedLocatorException $e) {
+                throw new ElementNotFound("name=$link", "'$link' is invalid CSS and XPath selector and Link or Button");
+            }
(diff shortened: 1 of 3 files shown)


DavertMik:
Thanks

PR #3553, opened 2016-09-27
Detect cmd/console width on Windows

emninet:
Hello
I run codeception on Windows
I noticed that on Windows the debug may cause a print error when large method names are used in test suites.

In the code the default width used in method detectWidth is set to 60
As the terminal width is not detected on Windows, this width remains set to 60
This causes the maxLength of the message displayed in debug to be limited to 60 characters
As soon as your method name goes above a certain size the message length grows and may go over 60
And you may have an error occuring in the Step class in method getArgumentsAsString provoked by a division by zero on line 117

I dont know if there is a way to capture the cli width on Windows with PHP. I am not sure of that.
But giving a few more characters to the defaut width value on Windows allowed me to run all tests in debug mode without facing any printing errors

### DIFF
--- a/src/Codeception/Subscriber/Console.php
+++ b/src/Codeception/Subscriber/Console.php
@@ -521,12 +521,12 @@ class Console implements EventSubscriberInterface
     {
         $this->width = 60;
         if (!$this->isWin()
-            && (php_sapi_name() == "cli")
+            && (php_sapi_name() === "cli")
             && (getenv('TERM'))
             && (getenv('TERM') != 'unknown')
         ) {
             $this->width = (int) (`command -v tput >> /dev/null 2>&1 && tput cols`) - 2;
-        } elseif ($this->isWin() && (php_sapi_name() == "cli")) {
+        } elseif ($this->isWin() && (php_sapi_name() === "cli")) {
             exec('mode con', $output);
             preg_match('/^ +.* +(\d+)$/', $output[4], $matches);
             if (!empty($matches[1])) {
--- a/src/Codeception/Subscriber/Console.php
+++ b/src/Codeception/Subscriber/Console.php
@@ -529,7 +529,7 @@ class Console implements EventSubscriberInterface
         } else if ($this->isWin() && (php_sapi_name() == "cli")) {
             exec('mode con', $output);
             preg_match('/^ +.* +(\d+)$/', $output[4], $matches);
-            $this->width = (int) $matches[1] - 2;
+            $this->width = (int) $matches[1];
         }
 
         return $this->width;
--- a/src/Codeception/Subscriber/Console.php
+++ b/src/Codeception/Subscriber/Console.php
@@ -527,7 +527,7 @@ class Console implements EventSubscriberInterface
         ) {
             $this->width = (int) (`command -v tput >> /dev/null 2>&1 && tput cols`) - 2;
         } else if ($this->isWin() && (php_sapi_name() == "cli")) {
-           $this->width = 200; 
+            $this->width = 200; 
         }
 
         return $this->width;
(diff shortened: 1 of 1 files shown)


emninet:
Hi. I dont think hardcoding the value is a good idea either.
It's the quick fix i found to have codeception run, and had no idea to how to get the actual width of the terminal.
i will try working on your suggestion

src/Codeception/Subscriber/Console.php
```diff
@@ -526,6 +526,8 @@ protected function detectWidth()
             && (getenv('TERM') != 'unknown')
         ) {
             $this->width = (int) (`command -v tput >> /dev/null 2>&1 && tput cols`) - 2;
+        } else if ($this->isWin() && (php_sapi_name() == "cli")) {
```
DavertMik: Console always executed in CLI mode, so last check is not required

emninet: I just retrieved the same check used in the above if

src/Codeception/Subscriber/Console.php
```diff
@@ -527,7 +527,9 @@ protected function detectWidth()
         ) {
             $this->width = (int) (`command -v tput >> /dev/null 2>&1 && tput cols`) - 2;
         } else if ($this->isWin() && (php_sapi_name() == "cli")) {
-            $this->width = 200;
+            exec('mode con', $output);
+            preg_match('/^ +.* +(\d+)$/', $output[4], $matches);
+            $this->width = (int) $matches[1] - 2;
```
samdark: Why `-2`?

emninet: I was not sure if adding this -2 was required
I saw that for linux / mac calculation added this -2
If you tell me the meaning of this -2 maybe i could check if it is required to arbitrary reduce the detected width on Windows or not

DavertMik: Maybe it was supposed to left some padding on the screen? 
I can't say why for sure in Linux version there was -2 but if it is there I found it is nicer so have some padding on the screen. Maybe it, maybe not. 

Just try by yourself how it looks better on Windows.

src/Codeception/Subscriber/Console.php
sergeyklay: `else if` => `elseif`

src/Codeception/Subscriber/Console.php
```diff
@@ -526,6 +526,10 @@ protected function detectWidth()
             && (getenv('TERM') != 'unknown')
         ) {
             $this->width = (int) (`command -v tput >> /dev/null 2>&1 && tput cols`) - 2;
+        } else if ($this->isWin() && (php_sapi_name() == "cli")) {
+            exec('mode con', $output);
+            preg_match('/^ +.* +(\d+)$/', $output[4], $matches);
+            $this->width = (int) $matches[1];
```
sergeyklay:
```php
if (!empty($matches[1])) {
    $this->width = (int) $matches[1];
}
```

src/Codeception/Subscriber/Console.php
```diff
@@ -526,8 +526,13 @@ protected function detectWidth()
             && (getenv('TERM') != 'unknown')
         ) {
             $this->width = (int) (`command -v tput >> /dev/null 2>&1 && tput cols`) - 2;
+        } elseif ($this->isWin() && (php_sapi_name() == "cli")) {
```
samdark: `===` could be used instead of `==`

emninet: yes, true
should have been used in the above if linux test too, why wasn't it used?

samdark: No idea. You can fix that as well.

samdark:
Looks good to me. @sergeyklay please re-review.

sergeyklay:
LGTM

samdark:
@DavertMik could be merged.

DavertMik:
Thanks!

PR #3556, opened 2016-09-28
Extracted getRelativeDir method to Codeception\Util\PathResolver

Naktibalda:


DavertMik:
Thanks!

loren-osborn:
Thank you for picking up the slack when I dropped the ball on this.

PR #3560, opened 2016-09-29
Add support for different selectors in WebDriver::submitForm

imjoehaines:
Hi

We've been updating some acceptance tests to use the `submitForm` method, but have hit an issue where some of our buttons don't have a `name` attribute so `submitForm` can't find them. We've been getting around this by adding names to the buttons that were missing them, but it would be nice to be able to find the button through different means so that we don't have to change our code just to write a test.

This PR adds support for strict selectors (e.g. `['id' => 'foo']`) and instances of `WebDriverBy` to be passed to `WebDriver::submitForm` as the `$button` parameter, so that the buttons can be found by more methods than just `name`.

For example, with the following HTML:

``` html
<input type="submit" id="submit" name="submitButton" value="Submit" />
```

`$button` could be any one of the following:
  - `'submitButton'`
  - `['id' => 'submit']`
  - `WebDriverBy::cssSelector('input[name="submitButton"]')`

Or any other valid strict locator or `WebDriverBy`.

The existing behaviour should be unchanged so this won't cause any backwards incompatibility issues.

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -1742,6 +1742,24 @@ class WebDriver extends CodeceptionModule implements
      *     ]
      * ]);
      * ```
+     *
+     * The `$button` parameter can be either a string, an array or an instance
+     * of Facebook\WebDriver\WebDriverBy. When it is a string, the
+     * button will be found by its "name" attribute. If $button is an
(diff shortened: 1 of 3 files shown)


Naktibalda:

imjoehaines:
Hi @Naktibalda, updated :)

DavertMik:
Looks very good. Thank you!

PR #3565, opened 2016-09-30
Corrected call_user_func arguments in Yii2

iRipVanWinkle:
This code
``` php
// inside Cest file or Codeception\TestCase\Unit
public function _fixtures()
{
    return ['posts' => PostsFixture::className()]
}
```
threw exception: "call_user_func() expects parameter 1 to be a valid callback, no array or string given"

and I fixed it up


DavertMik:
Thank you!

PR #3532, opened 2016-09-20
No more conflict between SOAP and REST

eXorus:
Fix #3512

**User of Codeception  <= v2.2.3**
No conflict between SOAP and REST
it will take seeResponseCodeIs of SOAP or seeResponseCodeIs of REST depends on the order of modules in yml

**User of Codeception v2.2.4**
Conflict between SOAP and REST
SOAP and REST can't be together so nobody are using both.

**User of Codeception > v2.2.4**
No conflict between SOAP and REST
it will take seeResponseCodeIs of SOAP or seeResponseCodeIs of REST depends on the order of modules in yml
seeSoapResponseCodeIs will always be the response of SOAP.

Tips : Put the REST module after the SOAP module to be sure to have:
seeResponseCodeIs => REST
seeSoapResponseCodeIs => SOAP

Related to  #3104

PR #3566, opened 2016-09-30
SOAP conflict with REST - come back

eXorus:
Fix bad merged of #3532


DavertMik:
Oh, yes, sorry. I did that intentinally because REST and SOAP actually are APIs so probably the interface should be there. But I forgot that REST has declared its conflict soo. 

Sorry for tht

PR #3569, opened 2016-10-02
[InnerBrowser] click improvements

Naktibalda:
Support clicking button in a link.
Click on the first clickable item when clickBySelector is used.

I expect WebDriver to fail some tests, but I can't run it locally now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 * [Laravel5] Add `database_migrations_path` configuration option. See #3628. By @janhenkgerritsen
 * [Laravel5][Lumen] Fixed issue that caused the `have` and `haveMultiple` methods not being available when using the ORM part of the modules. See #3587. By @janhenkgerritsen
 * [APC] Added `Codeception\Module\Apc` module to interact with the Alternative PHP Cache (APC) using either APCu or APC extension. By @sergeyklay
+* [InnerBrowser] Fixed clicking on a button inside the link
+* [InnerBrowser] Click on the first clickable item when clickBySelector is used
+* Removed tags from see/dontSee output and friends output
 
 #### 2.2.5
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,9 @@
 * [Laravel5] Add `database_migrations_path` configuration option. See #3628. By @janhenkgerritsen
 * [Laravel5][Lumen] Fixed issue that caused the `have` and `haveMultiple` methods not being available when using the ORM part of the modules. See #3587. By @janhenkgerritsen
 * [APC] Added `Codeception\Module\Apc` module to interact with the Alternative PHP Cache (APC) using either APCu or APC extension. By @sergeyklay
-* [InnerBrowser] Fixed clicking on a button inside the link
-* [InnerBrowser] Click on the first clickable item when clickBySelector is used
+* [PhpBrowser][Frameworks] Fixed clicking on a button inside the link
+* [PhpBrowser][Frameworks] Click on the first clickable item when clickBySelector is used
+* [PhpBrowser][Frameworks] Anchor is no longer sent to server
 * Removed tags from see/dontSee output and friends output
 
 #### 2.2.5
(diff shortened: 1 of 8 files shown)


Naktibalda:
I don't know how to make WebDriver click only on links and buttons, so I moved `ClickSelectsClickableElement` tests from TestsForWeb to PhpBrowserTest.

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 #### 2.2.6
 
 * [Laravel5][Lumen] Fixed issue that caused the `have` and `haveMultiple` methods not being available when using the ORM part of the modules. See #3587. By @janhenkgerritsen
+* [InnerBrowser] Fixed clicking on a button inside the link
```
DavertMik: Next time please mention this as [PhpBrowser][Frameworks], because InnerBrowser is too inner for others to know about it :)

src/Codeception/Lib/InnerBrowser.php
```diff
@@ -379,24 +376,48 @@ protected function clickByLocator($link)
...
+        $buttonValue = $node->getAttribute('value');
+
+        if ($buttonName !== '' && $buttonValue !== null) {
+            $formParams = [$buttonName => $buttonValue];
+        }
+
+        while ($node->parentNode !== null) {
+            $node = $node->parentNode;
+            if ($node->tagName === 'a') {
+                $this->amOnPage($node->getAttribute('href'));
```
DavertMik: Just wondering should this behavior be applied to any element? If it is a part of a link clicking on it should move to corresponding page. How do you think?

Also what if href=='#'? This will produce a failure.

tests/data/app/view/form/button_in_link.php
```diff
@@ -0,0 +1,10 @@
+<html>
+<body>
+<form action="/form/button" method="POST">
+    <button type="submit" name="btn0">Submit</button>
+</form>
+<a href="/info"><input type="submit" value="More Info"></a>
+
+<a href="/info"><span><input type="submit" value="Span Info"></span></a>
```
DavertMik: from the point of valid HTML markup I'd say this is a dirty hack :/
not sure how many people actually use it (but they shouldn't)

DavertMik:
This is a good fix for actually edge case. I don't think <a><button></a> should ever be used... If someone needs a button in link they can style their links as button. On other side, from according to HTML specification everything inside <a> should be clickable as link unless handled by JS. 

This behavior can be widened, what do you think?

src/Codeception/Lib/InnerBrowser.php
Naktibalda: I will test that, but this is not a regression, earlier code had no special handling for anchors.

tests/data/app/view/form/button_in_link.php
Naktibalda: It is a valid XHTML 1.0 Strict!

Naktibalda:
> On other side, from according to HTML specification everything inside **should be** clickable as link unless handled by JS.

This is exactly what this change is about.

DavertMik:
@Naktibalda current behavior is only about buttons. What about any other divs or spans inside <a>? Should we allow them to be clicked too, how do you think?

src/Codeception/Lib/InnerBrowser.php
Naktibalda: Before clicking on button submitted the first form on page.
BrowserKit handles clicking href="#" just fine, I added some tests.

Naktibalda:
Clicking divs and spans is out of scope for this pull request.

Naktibalda:
The tests uncovered an existing issue in FrameworksTest.
InnerBrowser has to strip anchor before sending a request, anchor is never sent to server by a real browser.

DavertMik:
Ok, then. Please fix the anchor thing and we can merge it.

Naktibalda:
To behave more like a real browser InnerBrowser should not execute GET (?) requests if Url starts with # and it should trim anchor before passing request to connector.

Naktibalda:
Fixed

DavertMik:
@Naktibalda I am about to merge PR but there is one conflict. Could you update it? Thanks

Issue #3573, opened 2016-10-03
Autoload problem with Codeception 2.0+ and composer /w phpdi(zf2)


Mattin:
Hello,

I updated fully functional codeception from 1.8 to 2.2 today and within our application based on composer and php-di using autogenerated classmap I experienced strange fatal error:

``` bash
root@1ede0d5ee450:/var/www/html# ./vendor/bin/codecept --version
PHP Fatal error:  Cannot redeclare codecept_debug() (previously declared in /var/www/html/vendor/codeception/codeception/autoload.php:74) in /var/www/html/vendor/codeception/codeception/autoload.php on line 77
PHP Stack trace:
PHP   1. {main}() /var/www/html/vendor/codeception/codeception/codecept:0
PHP   2. require_once() /var/www/html/vendor/codeception/codeception/codecept:7
PHP   3. class_exists(*uninitialized*) /var/www/html/vendor/codeception/codeception/autoload.php:20
PHP   4. spl_autoload_call(*uninitialized*) /var/www/html/vendor/codeception/codeception/autoload.php:20
PHP   5. Composer\Autoload\ClassLoader->loadClass($class = *uninitialized*) /var/www/html/vendor/codeception/codeception/autoload.php:20
PHP   6. Composer\Autoload\includeFile($file = *uninitialized*) /var/www/html/vendor/composer/ClassLoader.php:301

Fatal error: Cannot redeclare codecept_debug() (previously declared in /var/www/html/vendor/codeception/codeception/autoload.php:74) in /var/www/html/vendor/codeception/codeception/autoload.php on line 77

Call Stack:
    0.0384     361240   1. {main}() /var/www/html/vendor/codeception/codeception/codecept:0
    0.0834     386424   2. require_once('/var/www/html/vendor/codeception/codeception/autoload.php') /var/www/html/vendor/codeception/codeception/codecept:7
    4.9170    4620632   3. class_exists(???) /var/www/html/vendor/codeception/codeception/autoload.php:20
    4.9170    4620672   4. spl_autoload_call(???) /var/www/html/vendor/codeception/codeception/autoload.php:20
    4.9170    4620792   5. Composer\Autoload\ClassLoader->loadClass(???) /var/www/html/vendor/codeception/codeception/autoload.php:20
    4.9170    4620792   6. Composer\Autoload\includeFile(???) /var/www/html/vendor/composer/ClassLoader.php:301
```

[...]

PR #3575, opened 2016-10-03
Fixed redeclaring functions in autoload

Mattin:
Issue: #3573 

_Note: Fixed bad (previously created from master) pull request #3574_

### DIFF
--- a/autoload.php
+++ b/autoload.php
@@ -62,32 +62,45 @@ if (!function_exists('json_last_error_msg')) {
 }
 
 // function not autoloaded in PHP, thus its a good place for them
-function codecept_debug($data)
-{
-    \Codeception\Util\Debug::debug($data);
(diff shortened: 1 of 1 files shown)


autoload.php
```diff
@@ -62,32 +62,45 @@ function json_last_error_msg()
...
 }
 
-function codecept_relative_path($path)
-{
-    return \Codeception\Util\PathResolver::getRelativeDir($path, \Codeception\Configuration::projectDir(), DIRECTORY_SEPARATOR);
+if (!function_exists('codecept_relative_path')) {
+    function codecept_relative_path($path)
+    {
+        return \Codeception\Util\PathResolver::getRelativeDir($path, \Codeception\Configuration::projectDir(),
+            DIRECTORY_SEPARATOR);
```
Mattin: That was previously there :), but I fixed it ;).

PR #3579, opened 2016-10-04
[Yii2] Check if Yii::$app is set in _after before using it

Naktibalda:
Fixes http://stackoverflow.com/questions/39810706/using-fixtures-in-functional-test-for-yii2-application


samdark:
Looks good to me.

Issue #3581, opened 2016-10-05
InnerBrowser Unknown modifier


nathanphan:
#### What are you trying to achieve?
I am using click like this:

`$I->click('Create Hotfix', 'li.active');`

I see this in log

` preg_match(): Unknown modifier '/' in file /var/www/devops/protected/vendor/codeception/codeception/src/Codeception/Lib/InnerBrowser.php at line 202`

My functional.suite.yml is

`modules:
    enabled:
        # add framework module here
        - Yii1:
            appPath: 'public/test.php'
            url: 'http://mysite.local/test.php'
`

When I use var_dump before line 202 in InnerBrowser.php for the $pattern variable, it gives me this

`/^http\\://mysite\\.local$/u`

Which is wrong. It should be 

`/^http:\/\/mysite\.local$/u`

Please help

amashigeseiji:
I think the cause of this problem is `\Codeception\Module\Yii1::getDomainRegex:227`.

`InnerBrowser::isInternalDomain` calls `\Codeception\Module\Yii1::getInternalDomains`.

line 248 of `\Codeception\Module\Yii1`(`getInternalDomain`) is:
```
        $domains = [$this->getDomainRegex(Yii::app()->request->getHostInfo())];
```

and line 227 of `\Codeception\Module\Yii1`(`getDomainRegex`) is:
```
        if (preg_match('#https?://(.*?)/#', $template, $matches)) {
```

`Yii::app()->request->getHostInfo()` returns `http://mysite.local`, not `http://mysite.local/`,
so `preg_match` on line 227 (`\Codeception\Module\Yii1::getDomainRegex)` does not operate as expected.
Regex pattern of `preg_match` on line 227 expects argument with trailing slash like `http://mysite.local/`.

I think [this commit](https://github.com/amashigeseiji/Codeception/commit/ff8804ce487a5e98daa59042ffc6455426aaecb2?diff=unifie://github.com/amashigeseiji/Codeception/commit/ff8804ce487a5e98daa59042ffc6455426aaecb2?diff=unified) will solve your problem.

nathanphan:
I got your point but when it comes to line 202 in InnerBrowser.php 

the $pattern and $domain cannot be matched.

amashigeseiji:
yes, I see.
This is a bug of `Yii1::getInternalDomain` that returns invalid regex pattern.
[...]

PR #3582, opened 2016-10-05
fix to get right domain host, not the whole url

nathanphan:
problem: Cannot compare

/^http:\/\/mysite\.local$/u with mysite.local in InnerBrowser.php line 202

It happens for Yii1 module only

Please merge it. I need this changes so My test will passed :)

### DIFF
--- a/src/Codeception/Module/Yii1.php
+++ b/src/Codeception/Module/Yii1.php
@@ -1,262 +1,262 @@
 <?php
-    namespace Codeception\Module;
+namespace Codeception\Module;
 
-    use Codeception\Lib\Framework;
-    use Codeception\Exception\ModuleConfigException;
-    use Codeception\Lib\Interfaces\PartedModule;
-    use Codeception\TestInterface;
-    use Codeception\Lib\Connector\Yii1 as Yii1Connector;
-    use Codeception\Util\ReflectionHelper;
-    use Yii;
+use Codeception\Lib\Framework;
+use Codeception\Exception\ModuleConfigException;
+use Codeception\Lib\Interfaces\PartedModule;
+use Codeception\TestInterface;
+use Codeception\Lib\Connector\Yii1 as Yii1Connector;
+use Codeception\Util\ReflectionHelper;
+use Yii;
+
+/**
+ * This module provides integration with [Yii Framework 1.1](http://www.yiiframework.com/doc/guide/).
+ *
+ * The following configurations are available for this module:
+ *
+ *  * `appPath` - full path to the application, include index.php</li>
+ *  * `url` - full url to the index.php entry script</li>
+ *
+ * In your index.php you must return an array with correct configuration for the application:
+ *
+ * For the simple created yii application index.php will be like this:
+ *
+ * ```php
+ * <?php
+ * // change the following paths if necessary
+ * $yii=dirname(__FILE__).'/../yii/framework/yii.php';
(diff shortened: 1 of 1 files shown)


plutov:
I vote for this PR will be merged!

Naktibalda:
@nathanphan Please replace tabs with spaces, delete empty lines and indent new code correctly.

nathanphan:
@Naktibalda can you check that Travis CI build failed

Fatal error: Allowed memory size of 1073741824 bytes exhausted (tried to allocate 134217728 bytes) in phar:///home/travis/.phpenv/versions/5.4.37/bin/composer/src/Composer/DependencyResolver/Solver.php on line 220

amashigeseiji:
If `$template` is defined like `'http://<hoge:\w+>.hoge.com/'`, `parse_url` returns false.

nathanphan:
yeah it will be false

I just wonder why people config their appPath like that.

amashigeseiji:
This PR does not work if the case `$I->amOnPage('http://mysite.local/')` because it will be treated as external domain. It is not expected.
This is the same problem `$I->click('some link')` which href is `http://mysite.local/`.

I think it should be this.

```diff
--- a/src/Codeception/Module/Yii1.php
+++ b/src/Codeception/Module/Yii1.php
@@ -224,8 +224,8 @@ public function _after(TestInterface $test)
      */
     private function getDomainRegex($template, $parameters = [])
     {
-        if (preg_match('#https?://(.*?)/#', $template, $matches)) {
-            $template = $matches[1];
+        if ($host = parse_url($template, PHP_URL_HOST)) {
+            $template = $host;
         }
         if (strpos($template, '<') !== false) {
             $template = str_replace(['<', '>'], '#', $template);
```

This problem is mainly invalid match pattern.
I pointed [here](https://github.com/Codeception/Codeception/issues/3581#issuecomment-251861095), the problem is `Yii::app()->request->getHostInfo()` returns `http://mysite.local` (without trailing slash), but `Yii1::getDomainRegex` expects string with trailing slash([code](https://github.com/Codeception/Codeception/blob/2.2/src/Codeception/Module/Yii1.php#L227)).
Regex pattern `'#https?://(.*?)/#'` does not match `http://mysite.local` because trailing slash does not exist.
This is the reason `Yii1::getDomainRegex` returns invalid pattern, I think.

another solution is:
```diff
--- a/src/Codeception/Module/Yii1.php
+++ b/src/Codeception/Module/Yii1.php
@@ -224,8 +224,8 @@ public function _after(TestInterface $test)
      */
     private function getDomainRegex($template, $parameters = [])
     {
-        if (preg_match('#https?://(.*?)/#', $template, $matches)) {
+        if (preg_match('@^(?:https?://)?([^/]+)@i', $template, $match) {
             $template = $match[1];
         }
         if (strpos($template, '<') !== false) {
             $template = str_replace(['<', '>'], '#', $template);
```

I misunderstood that `$rule->template` on [here](https://github.com/Codeception/Codeception/blob/2.2/src/Codeception/Module/Yii1.php#L254) is like `'http://<subdomain:\w+>.mysite.local/<action:\w+>'`,
but it is `'http://<subdomain>.mysite.local/<action>'`.
So `parse_url` is behave as expected, and `parse_url` looks more cool for me than regex.

Naktibalda:
@amashigeseiji Could you make a PR? Your solution looks more correct.

Naktibalda:
Replaced by #3597

PR #3589, opened 2016-10-07
decompose array compare logic, so that it can be used independently f…

tiger-seo:
decompose array compare logic, so that it can be used independently from json


src/Codeception/Util/ArrayContainsComparator.php
```diff
@@ -0,0 +1,149 @@
+<?php
+namespace Codeception\Util;
```
tiger-seo: fixed

DavertMik:
Thanks!

PR #3591, opened 2016-10-09
[Yii1] Escape domain regex

konstin:
When using Yii1 together with the REST module, the following situation occurs in `InnerBrowser.php:202` when using `http:yii-codeception.test` as url:
```php
preg_match('/^http\://yii\-codeception\.test$/u', 'yii-codeception.test')
```

The call to `preg_match` fails as it treats the slash behind the `http\:` as the end of the regex and sees the following slash as (invalid) regex option. This pull request fixes this.

### DIFF
--- a/src/Codeception/Module/Yii1.php
+++ b/src/Codeception/Module/Yii1.php
@@ -234,6 +234,7 @@ class Yii1 extends Framework implements PartedModule
         foreach ($parameters as $name => $value) {
             $template = str_replace("#$name#", $value, $template);
         }
+        $template = str_replace('/', '\\/', $template);
         return '/^' . $template . '$/u';
     }
 


konstin:
appveyor seems to be failing. Unfortunately I can't see why. Any suggestions on this?

src/Codeception/Module/Yii1.php
```diff
@@ -234,6 +234,7 @@ private function getDomainRegex($template, $parameters = [])
         foreach ($parameters as $name => $value) {
```
Naktibalda: forward slashes are already escaped by a previous line.

Naktibalda:
I think that your issue will be fixed by https://github.com/Codeception/Codeception/pull/3582/files#diff-41cc717334ae24027609d71ab27a7bddR233

konstin:
This will indeed be fixed by #3582, thanks.

PR #3593, opened 2016-10-10
Turn off colors in Windows console by default

kusnir:


kusnir:
It is useful for using Codeception in console mode in Windows environment.

kusnir:
OK. Anyway there should by way to set default value so I changed it to use setting from config.

DavertMik:
This fix applies only for interactive console. All other commands especially `run` won't be affected by this change. @kusnir is that what you are actually want to achieve?

samdark:
I think it should be applied to all console output.

kusnir:
I want just FIX output in interactive console (which I run in console without colors support) because only there you can not change default settings by configuration.

I checked it and I recommend:
- remove adding 'colors' settings from Codeception/Command/Bootstrap.php - for using auto default value
- to Codeception/Codecept.php and Codeception/Configuration.php add mentioned detection of colors support

samdark:
Removing isn't a good idea but making it autodetect by default is good.

kusnir:
remove adding 'colors' ... = remove from generated codeception.yml file, so the colors settings will not be in config file by default

samdark:
Yes. Still, ability to configure it explicitly should remain.

enumag:
I'm not quite sure why but somehow colors work for me on windows. So they should not be auto-disabled.

PR #3594, opened 2016-10-10
Save all screenshots and page sources + show their names in log message

kusnir:
See https://github.com/Codeception/Codeception/pull/3490

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -374,9 +374,18 @@ class WebDriver extends CodeceptionModule implements
         $this->debugWebDriverLogs();
         $filename = preg_replace('~\W~', '.', Descriptor::getTestSignature($test));
         $outputDir = codecept_output_dir();
-        $this->_saveScreenshot($outputDir . mb_strcut($filename, 0, 245, 'utf-8') . '.fail.png');
-        $this->_savePageSource($outputDir . mb_strcut($filename, 0, 244, 'utf-8') . '.fail.html');
-        $this->debug("Screenshot and page source were saved into '$outputDir' dir");
+        $filenamePng = mb_strcut($filename, 0, 245, 'utf-8') . '.fail.png';
+        $filenameHtml = mb_strcut($filename, 0, 244, 'utf-8') . '.fail.html';
+        for ($i = 1; $i < 1000; $i++) {
+            if ((! file_exists($outputDir . $filenamePng)) && (! file_exists($outputDir . $filenameHtml))) {
+                break;
+            }
+            $filenamePng = mb_strcut($filename . '.' . $i, 0, 245, 'utf-8') . '.fail.png';
+            $filenameHtml = mb_strcut($filename . '.' . $i, 0, 244, 'utf-8') . '.fail.html';
+        }
+        $this->_saveScreenshot($outputDir . $filenamePng);
+        $this->_savePageSource($outputDir . $filenameHtml);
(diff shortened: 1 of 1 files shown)


src/Codeception/Module/WebDriver.php
```diff
@@ -374,9 +374,18 @@ public function _failed(TestInterface $test, $fail)
...
-        $this->_saveScreenshot($outputDir . mb_strcut($filename, 0, 245, 'utf-8') . '.fail.png');
-        $this->_savePageSource($outputDir . mb_strcut($filename, 0, 244, 'utf-8') . '.fail.html');
-        $this->debug("Screenshot and page source were saved into '$outputDir' dir");
+        $filenamePng = mb_strcut($filename, 0, 245, 'utf-8') . '.fail.png';
+        $filenameHtml = mb_strcut($filename, 0, 244, 'utf-8') . '.fail.html';
+        for ($i = 1; $i < 1000; $i++) {
+            if ((! file_exists($outputDir . $filenamePng)) && (! file_exists($outputDir . $filenameHtml))) {
+                break;
+            }
+            $filenamePng = mb_strcut($filename . '.' . $i, 0, 245, 'utf-8') . '.fail.png';
```
Naktibalda: Append $i after strcut or it will be lost if $filename is too long.

Naktibalda:
```php
$filenamePng = mb_strcut($filename, 0, 244 - strlen($i), 'utf-8') . ".$i.fail.png";
```

src/Codeception/Module/WebDriver.php
```diff
@@ -374,9 +374,18 @@ public function _failed(TestInterface $test, $fail)
         $this->debugWebDriverLogs();
         $filename = preg_replace('~\W~', '.', Descriptor::getTestSignature($test));
         $outputDir = codecept_output_dir();
-        $this->_saveScreenshot($outputDir . mb_strcut($filename, 0, 245, 'utf-8') . '.fail.png');
-        $this->_savePageSource($outputDir . mb_strcut($filename, 0, 244, 'utf-8') . '.fail.html');
-        $this->debug("Screenshot and page source were saved into '$outputDir' dir");
+        $filenamePng = mb_strcut($filename, 0, 245, 'utf-8') . '.fail.png';
+        $filenameHtml = mb_strcut($filename, 0, 244, 'utf-8') . '.fail.html';
+        for ($i = 1; $i < 1000; $i++) {
+            if ((! file_exists($outputDir . $filenamePng)) && (! file_exists($outputDir . $filenameHtml))) {
```
DavertMik: actually this is a mess. No one will ever be able to guess which files belong to which build unless looking to their date. Current approach is probably not the best as well but at least there is only 1 file and not 1000 of them.

DavertMik:
This can't be accepted unless we change policy for naming reports.

Let's say we have 3 failed files from 1st run, 5 failed from 2nd and 1 from 3rd. Then by looking into them we will have no ideas what and when actually went wrong. 

If we add timestamp to filenames we will have dozens of screenshots created in a few runs. 

Current behavior is stupid, dead simple, and effective. I can't think of something better for now :( 
This improvement can be moved to extension (the similar stuff is done via RecorderExtension). 

Can you take the https://github.com/Codeception/Codeception/blob/2.2/ext/Recorder.php and make it worked for failing steps only and not for all steps. Then you can print to console failed step names + corresponding file names.

DavertMik:
Sorry, but I need to close this PR.
@kusnir I recommend you to create an extension, publish it on packagist and add it as [Addon](http://codeception.com/addons) on our site. This will make this functionality work. 
Thank you for submitting PR. Sorry we can't accept this one.

PR #3595, opened 2016-10-10
FIX can failed in console:

kusnir:
PHP Fatal error:  Call to a member function addFailure() on null in ...\Codeception\Scenario.php on line 72

Fatal error: Call to a member function addFailure() on null in ...\Codeception\Scenario.php on line 72

FATAL ERROR. TESTS NOT FINISHED.
Call to a member function addFailure() on null
in ...\Codeception\Scenario.php:72

Reproduce:
1. Run console: codecept console acceptance
2. Run in console: amOnPage('/');
3. Run in console: canSeeElement('#xxx');
... and the error should occurred.


DavertMik:
Thanks!

PR #3596, opened 2016-10-10
Remove backticks from documentation

gabriel403:
There's random backticks in a block in the documentation, so I've removed them!

### DIFF
--- a/docs/modules/WebDriver.md
+++ b/docs/modules/WebDriver.md
@@ -711,7 +711,7 @@ Fills a text field or textarea with the given string.
 
 ``` php
 <?php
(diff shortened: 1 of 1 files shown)


Naktibalda:
That documentation is auto-generated from https://github.com/Codeception/Codeception/blob/02ee5b75eed804a75c3bfc43b08a3fda25fc4f8a/src/Codeception/Lib/Interfaces/Web.php#L690

It looks like it was mangled by some kind of email address obfuscator.

gabriel403:
@Naktibalda heh
OK
Thanks I'll dump this PR then :)

PR #3598, opened 2016-10-10
Catch errors in _before for not to crash and continue test execution

DavertMik:
If exception is thrown in `test.before` tests will continue to be executed but test will be marked as error and skipped.


DavertMik:
Closed in favor of #3534

PR #3602, opened 2016-10-11
HTML report to include screenshots of failures from WebDriver

DavertMik:
* added `reports` section to MetaData in tests
* WebDriver to store html and png reports to metadata
* improved styling of HTML report

selection_154

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -61,7 +61,8 @@ class InnerBrowser extends Module implements Web, PageSourceSaver, ElementLocato
         }
         $filename = preg_replace('~\W~', '.', Descriptor::getTestSignature($test));
         $filename = mb_strcut($filename, 0, 244, 'utf-8') . '.fail.html';
-        $this->_savePageSource(codecept_output_dir() . $filename);
+        $this->_savePageSource($report = codecept_output_dir() . $filename);
+        $test->getMetadata()->addReport('html', $report);
     }
 
     public function _after(TestInterface $test)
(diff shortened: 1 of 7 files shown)


Naktibalda:
How are you going to support multiple screenshots per test? #3594

DavertMik:
> How are you going to support multiple screenshots per test? #3594

I won't. For multiple screenshots we have Recoder and API to build any custom extension to handle that.

DavertMik:
Added HTML reports to InnerBrowser

selection_156

rebas82:
1) when i run test by selenium server (chrome), screenshot is only from visible screen. When i run test by phantomjs, screenshot is frome complete page (i can see everything on page).

Is it posible have complete screenshot from selenium server as is from phantomjs?

2) some people dont like screenshots in html report, or have own function for doing fail screenshot (as me). Is it posible doing some settings to YML file, where tester set if they need failscreenshot or not need?

Naktibalda:
@rebas82 You should raise a separate ticket for these suggestions.

rebas82:
ok

Sereda-Fazen:
Hi everybody!
I've updated Codeception to 2.2.6. and when I run tests. I received in HTML Report the following
image

We together with my collegue added the feature in order to work in HTML Report.
1. In the file ResultReport/HTML.php we added the string

> $png = "<tr><td class='error'><div class='screenshot'>failure screenshot</div></td></tr>";

Currently it works but the path in the HTML Report is absolute

hadouke:
Is ther3 any updates on this? I also have trouble with showing screens in new report

PR #3606, opened 2016-10-12
better Example::offsetExists() behavior

adrian-enspired:
Differentiate between missing keys and those with NULL values.


DavertMik:
Thanks

PR #3610, opened 2016-10-14
Added Codeption\Module\Apc module

sergeyklay:


src/Codeception/Module/Apc.php
```diff
@@ -0,0 +1,235 @@
...
+     * ```
+     *
+     * @param $key
+     * @param $value
+     */
+    public function dontSeeInApc($key, $value = null)
+    {
+        if (null === $value) {
+            $this->assertFalse($this->exists($key), "The key '$key' exists in APC(u).");
+        } else {
```
DavertMik: you can have `return` one line above without `else`

src/Codeception/Module/Apc.php
```diff
@@ -0,0 +1,235 @@
...
+    /**
+     * Stores an item `$value` with `$key` on the APC(u).
+     *
+     * @param string $key
+     * @param mixed $value
+     * @param int $expiration
+     */
+    public function haveInApc($key, $value, $expiration = null)
+    {
+        $this->assertTrue($this->store($key, $value, $expiration));
```
DavertMik: there should not be assertions in `have*` methods. Just store it and return some key (if it is needed).
Also please provide an example in docblock

sergeyklay: I just repeating the Memcached logic
https://github.com/Codeception/Codeception/blob/2.2/src/Codeception/Module/Memcache.php#L183-L193

sergeyklay:
@DavertMik Done

DavertMik:
Thank you!

PR #3615, opened 2016-10-17
-- separates options from arguments in codecept run

### LINKED ISSUE #3614: -- for end of parameter (option) list does not work
I'd like to execute only my unit tests, skipping acceptance tests for the time being. This should be supported, according to [docs](http://codeception.com/docs/reference/Commands#Run):

`./vendor/bin/codecept run unit`

#### What do you get instead?
All three suites are executed:

```bash
Functional Tests (0) ---------------------------------------------------------
------------------------------------------------------------------------------

Unit Tests (1) ---------------------------------------------------------------
E RedirectTest: Redirect create (0.28s)
------------------------------------------------------------------------------

Acceptance Tests Tests (12) ---------------------------------------------------------
[…]
------------------------------------------------------------------------------
```

### Details

* Codeception version: 2.2.5
* PHP Version: 7.0
* Operating System: RHEL 7
* Installation type: Composer

Naktibalda:
Fixes #3614

This is a correct way to handle `--`
But it is potentially breaking, because commands like `codecept run --coverage-html -- unit --steps` will no longer work.
The correct way to write that one is `codecept run --coverage-html --steps -- unit`

### DIFF
--- a/src/Codeception/Command/Run.php
+++ b/src/Codeception/Command/Run.php
@@ -368,11 +368,12 @@ class Run extends Command
         $tokens = explode(' ', $request);
         foreach ($tokens as $token) {
             $token = preg_replace('~=.*~', '', $token); // strip = from options
-            if (strpos($token, '--') === 0 && $token !== '--') {
-                $options[] = substr($token, 2);
-                continue;
+            if ($token == '--') {
+                break; // there should be no options after ' -- ', only arguments
             }
-            if (strpos($token, '-') === 0) {
+            if (substr($token, 0, 2) === '--') {
+                $options[] = substr($token, 2);
+            } elseif ($token[0] === '-') {
                 $shortOption = substr($token, 1);
                 $options[] = $this->getDefinition()->getOptionForShortcut($shortOption)->getName();
             }


Naktibalda:
@DavertMik any comments?

DavertMik:
I think I will add some by the end of the week. I didn't have chance to look into it deeply

Naktibalda:
Actually there is no bc-break, any attempt to use `--`  resulted in InvalidArgumentException before this patch.

DavertMik:
Ok, so I have almost no arguments to not merge it. 

Except that `--` means passing variables directly to some underlying runner. 

When we do 

```
composer exec codecept run -- --group t
```

we are telling composer not to parse `--group` but bypass it to codecept runner. In case of Codeception we are not bypassing variables to no other app. We actually could pass them to PHPUnit, but we don't, we use only our runner and collect all variables here. 

So maybe it's not correct to use `--` like this. However, I don't see any other option, and I'm not a Linux expert so I can be mistaken in this case

Naktibalda:
We are using Symfony Console component, so it is best to adhere to its standards: http://symfony.com/doc/current/components/console/console_arguments.html

Also the Usage section of `codecept run` looks like this:
```
Usage:
  run [options] [--] [<suite>] [<test>]
```

DavertMik:
Thanks for pointing. Yes, you are absolutely right, here is the quote from http://docopt.org

> A double dash "--", when not part of an option, is often used as a convention to separate options and positional arguments, in order to handle cases when, e.g., file names could be mistaken for options. In order to support this convention, add "[--]" into your patterns before positional arguments.

> Usage: my_program [options] [--] <file>...
> Apart from this special meaning, "--" is just a normal command, so you can apply any previously-described operations, for example, make it required (by dropping brackets

Issue #3552, opened 2016-09-27
Laravel 5 - Add seeding capabilities


edno:
It would be nice to have seeding capabilities for the **Laravel5** module.

Currently, we can decide if we want to run the migration but there's no option for seeding.

Suggestion of implementation at module configuration level:
```yaml
modules:
    enabled:
        - Laravel5:
            seeders:
                  - UsersTableSeeder
                  - PostsTableSeeder
                  - CommentsTableSeeder
```
Ideally testing seeders would be to store under the `tests` folder instead of the `migrate` folder.

janhenkgerritsen:
Thanks for the suggestion!

edno:
Temporary workaround:

1- Update `autoload/classmap` within `composer.json` with your test seeders' path (e.g. `tests/_data`)
```json
"autoload": {
        "classmap": [
            "database",
            "tests/_data"
        ],
        "psr-4": {
            "App\\": "app/"
        }
    },
```
2- Use the `Laravel5` module's method `callArtisan` for running `db:seeder`:
```php
$I->callArtisan('db:seeder', ['class' => 'Acceptance\Seeder\AcceptanceDataSeeder']);
```

I recommend to use namespace for your test seeder, so there's no confusion and they cannot be loaded inadvertently.

Bouhnosaure:
Hello 
I've made a simple way for seeding the database you just set `run_database_seeds` to true but `run_database_migrations` need to be true too.

for the first comment with the ability to choose what seeds we want to run 
a simple way would be to set `APP_ENV` to testing in the `.env.testing` and detect in `DatabaseSeeder.php` if wich environement we are and run only the seeds we want

janhenkgerritsen:
Seeder capabilities were added with #3630.

mcblum:
If this is useful to anyone, I created this script which allows me to specify an order of operations for testing:
```bash
#!/usr/bin/env bash

# if a DB exists, delete and recreate it
if [ ! -f ./database/testing.sqlite ]; then
    touch ./database/testing.sqlite
else
    rm -f ./database/testing.sqlite
    touch ./database/testing.sqlite
fi

# if a log file exists, delete and recreate it
if [ -f ./storage/logs/laravel.log ]; then
    rm -f ./storage/logs/laravel.log
    touch ./storage/logs/laravel.log
fi

# migrate the test DB
php artisan migrate --env=testing --database testing

# seed the test DB
php artisan db:seed --database='testing'  --env=testing

[...]

PR #3625, opened 2016-10-20
Add seedings capabilities (#3552)

Bouhnosaure:
Hello 
I've made a simple way for seeding the database you just set run_database_seeds to true but run_database_migrations need to be true too.

for the first comment in #3552  with the ability to choose what seeds we want to run 
a simple way would be to set APP_ENV to testing in the .env.testing and detect in DatabaseSeeder.php if wich environement we are and run only the seeds we want.

### DIFF
--- a/src/Codeception/Module/Laravel5.php
+++ b/src/Codeception/Module/Laravel5.php
@@ -154,7 +154,7 @@ class Laravel5 extends Framework implements ActiveRecord, PartedModule
             // Must be called before database transactions are started
             $this->callArtisan('migrate');
 
-            if ($this->config['run_database_seeds']){
+            if ($this->config['run_database_seeds']) {
                 $this->callArtisan('db:seed');
             }
         }
--- a/src/Codeception/Module/Laravel5.php
+++ b/src/Codeception/Module/Laravel5.php
@@ -101,6 +101,7 @@ class Laravel5 extends Framework implements ActiveRecord, PartedModule
             [
                 'cleanup' => true,
                 'run_database_migrations' => false,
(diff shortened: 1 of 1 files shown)


janhenkgerritsen:
I have a few comments and questions.

Firstly, can you submit your PR to the 2.2 branch instead of master?

I also think this implementation is a bit too simple. I think it should at least be possible to specify a class name of the seeder you want to run. This way you can have different seeders for different environments. I think a good way to implement this is by adding another configuration option `database_seeder_class` that defaults to `DatabaseSeeder`, which is the name of the class used in a new Laravel installation.

I also think `run_database_seeds` should be changed into `run_database_seeder`.

Lastly, I think you should move the code outside the if statement that checks if migrations should run. This way you are not required to run the migrations, which provides more flexibility. You could for example use the DB module to setup your database and then run your database seeders via the Laravel5 module.

Let me know what you think.

Bouhnosaure:
Yes i can change the branche of my PR to 2.2

Hum you are right for the multiple seeder option,
I will do this soon, i think i have to make a loop throught the seeder classes in the option `database_seeder_class` because laravel doesn't permit to specify multiples classes in the command ([code](https://github.com/laravel/framework/blob/5.3/src/Illuminate/Database/Console/Seeds/SeedCommand.php#L72))

For the options names, it's okay too, i will change it for better names.

And sorry again for the PR into the master, i was a bit confused, i didn't see the changes in this [file](https://github.com/Codeception/Codeception/commit/2f7a28de429484b5efa03a6941aaf67f48dc8dc3) i had contribued to Laravel5 module but before this commit 😕

PR #3630, opened 2016-10-21
Seeder improvement in Laravel5 Module (#3552)

Bouhnosaure:
Hello again,
I've made a more simple way for seeding the database.
So now you can set these parameters in the config file :
```yml
class_name: FunctionalTester
modules:
    enabled:
        - Laravel5:
            run_database_migrations: true      # not required to run seeding
            run_database_seeder: true          # required to use this feature
            database_seeder_class:             # not required ( must be an array )
            - UserTableSeeder
            - AnotherTableSeeder
```
If you didn't specify any seeder, the default `DatabaseSeeder` will be used.
Seeders are runned in the order of the array.

@janhenkgerritsen what do you think ?

### DIFF
--- a/src/Codeception/Module/Laravel5.php
+++ b/src/Codeception/Module/Laravel5.php
@@ -110,7 +110,9 @@ class Laravel5 extends Framework implements ActiveRecord, PartedModule
             [
                 'cleanup' => true,
                 'run_database_migrations' => false,
+                'run_database_seeder' => false,
                 'database_migrations_path' => '',
+                'database_seeder_class' => ['DatabaseSeeder'],
                 'environment_file' => '.env',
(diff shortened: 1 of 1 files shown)


janhenkgerritsen:
Thanks for improving your PR. This is already a lot better, but I still have one request for a change.

The `database_seeder_class` option should just accept a single class name and not an array. This brings it more in line with how Laravel uses seeders, you specify a single seeder class and can call additional seeders from that class. Check out the relevant section in Laravel's documentation: https://laravel.com/docs/5.3/seeding#calling-additional-seeders

Bouhnosaure:
@janhenkgerritsen, i've corrected my PR, now it accept only one classe name instead of an array.
i hope it's better now !

Thanks !

PR #3635, opened 2016-10-22
Added commands to run the unit tests

damko:
it's a bit redundant but I think it might be useful for those jumping straight to the unit tests page.


DavertMik:
Yep, that's helpful!

PR #3637, opened 2016-10-24
Undefined variable  change to the correct  in MongoDb Module

hlogeon:
There were an issue with MongoDb Module. Variable `$enitiy` were undefined. I suppose that author meant `$entry` and just made a typo there. It doesn't affect anything except data import using `mongotype` dump type.


DavertMik:
Thanks for finding and fixing this tricky one bug.

PR #3641, opened 2016-10-25
Mongofill support

hlogeon:
If someone are using MongoDb with PHP7 or HHVM, he forced to use mongofill package(https://github.com/mongofill/mongofill). Mongofill has class MongoClient in the root namespace but it does't means that it's using legacy `mongo` driver instead of current `mongodb`. In order to make it work with Codeception MongoDB module I changed the way you check if user uses legacy driver by adding additional condition. So if there is `\MongoClient` class and there is no `mongofill` substring in its version -> use legacy mode. Normal mode otherwise.
Before fixing that cleaning of the database didn't work and I've got errors on duplicate keys every test which were using db.

### DIFF
--- a/src/Codeception/Lib/Driver/MongoDb.php
+++ b/src/Codeception/Lib/Driver/MongoDb.php
@@ -95,7 +95,7 @@ class MongoDb
      */
     public function __construct($dsn, $user, $password)
     {
-        $this->legacy = class_exists('\\MongoClient');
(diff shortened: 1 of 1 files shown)


hlogeon:
By the way I would like to ask what should I do to add Doctrine ODM(for MongoDB) as a core module of Codeception? I mean that I already made it and want to add it to this repository

DavertMik:
Thanks! Looks good, probably I will mention mongofill in docs for MongoDb

> Doctrine ODM(for MongoDB) as a core module of Codeception

That would be great! :+1: 
Please send Pull Request

PR #3648, opened 2016-10-27
Fix a minor typo with PHPDoc generation

Kussie:
This fixes an incredibly small typo in a couple of the PHPDoc blocks which have a plus sign instead of an asterisk.


DavertMik:
Thanks

PR #3654, opened 2016-10-29
Fixes #3551 - Monolog implementation

### LINKED ISSUE #3551: Can't manually log using the Logger class
I've required monolog/monolog with composer. I've also added Logger as an extension to `codeception.yml`. I cannot, however, find any way to manually log messages or variables using the Logger class.

Furthermore, when looking at the source code of `Logger.php` I see that the Monolog instance is actually stored inside the `protected` property `logger`.

Is it correct that there is no way to manually (like, during a test) log messages via this class, and that [information saying that you can do that](http://stackoverflow.com/a/21331033/1809762) is incorrect?

Farkie:
Hey,

I had a crack at https://github.com/Codeception/Codeception/issues/3551 - You can now use Monolog how you'd usually use it in any of your tests, plus they all log with the correct namespace.

### DIFF
--- a/ext/Logger.php
+++ b/ext/Logger.php
@@ -11,6 +11,7 @@ use Codeception\Extension;
 use Codeception\Test\Descriptor;
 use Monolog\Handler\RotatingFileHandler;
 use Codeception\Configuration as Config;
+use Psr\Log\LoggerInterface;
 
 /**
  * Log suites/tests/steps using Monolog library.
@@ -37,15 +38,15 @@ use Codeception\Configuration as Config;
 class Logger extends Extension
 {
     public static $events = [
-        Events::SUITE_BEFORE => 'beforeSuite',
-        Events::TEST_BEFORE => 'beforeTest',
-        Events::TEST_AFTER => 'afterTest',
-        Events::TEST_END => 'endTest',
-        Events::STEP_BEFORE => 'beforeStep',
-        Events::TEST_FAIL => 'testFail',
-        Events::TEST_ERROR => 'testError',
+        Events::SUITE_BEFORE    => 'beforeSuite',
+        Events::TEST_BEFORE     => 'beforeTest',
+        Events::TEST_AFTER      => 'afterTest',
+        Events::TEST_END        => 'endTest',
+        Events::STEP_BEFORE     => 'beforeStep',
+        Events::TEST_FAIL       => 'testFail',
+        Events::TEST_ERROR      => 'testError',
         Events::TEST_INCOMPLETE => 'testIncomplete',
-        Events::TEST_SKIPPED => 'testSkipped',
+        Events::TEST_SKIPPED    => 'testSkipped',
     ];
 
     protected $logHandler;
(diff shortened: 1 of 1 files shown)


ext/Logger.php
```diff
@@ -52,7 +54,7 @@ class Logger extends Extension
     /**
      * @var \Monolog\Logger
      */
-    protected $logger;
+    protected static $logger;
```
sergeyklay: Could you please explain why?

Farkie: Sure, I don't believe this to be a BC break as it was a protected method with no getter, so it was only ever an internal property. I've changed it to static so that you can actually use the logger in your tests, and not just codeception using it internally. It's also done statically so it's a singleton and doesn't need to create a new instance every time.

DavertMik: I think there is no other way to access logger from other place rather than making it static. Probably it is ok as extensions are loaded in the very beginning of `run` so they are available almost everywhere. We have static `$output` inside `Codeception\Util\Debug` and I don't see any other way on implementing `codecept_debug` there

ext/Logger.php
```diff
@@ -91,33 +119,40 @@ public function afterTest(TestEvent $e)
...
+        self::info("# Incomplete #");
     }
 
     public function beforeStep(StepEvent $e)
     {
-        $this->logger->info((string) $e->getStep());
+        self::info((string)$e->getStep());
+    }
+
+    public static function __callStatic($name, $arguments)
```
DavertMik: but I'm highly against this `__callStatic` it really breaks SOLID principle. This class is not a logger, it is an extension which wraps logger, so it shouldn't be used this way.

DavertMik:
Sorry for taking me that long to review the issue.

ext/Logger.php
Farkie: Hi @DavertMik, thanks for taking a look - do you mean reverse this change - https://github.com/Codeception/Codeception/pull/3654/commits/cbc56494a85b1fa742a4c225cc51db4402ac21d1

I was trying to get rid of wrapping the functions, but I can see why you might not like that (SOLID).

ext/Logger.php
```diff
@@ -63,26 +65,52 @@ public function _initialize()
...
         }
-        $this->path = $this->getLogDir();
+    }
+
+    public static function getLogger()
+    {
+        if (self::$logger) {
+            $logger = self::$logger;
+        } else {
+            $logger = self::getDefaultLogger();
```
DavertMik: Also I think logger should not be created for every purpose. It's ok if it would be created one time at extension start.
So `getLogger` should only return already created logger and not try to recreate it.

Here is the point. The logger should be accessible only if extension is loaded and initialize. If it's not enabled in config and `getLogger` still works that can be confusing.

ext/Logger.php
Farkie: Sure, I'll take a look when I get some time :)

DavertMik: However, this function should do *nothing* if the extension is not enabled.

ext/Logger.php
Farkie: Sure - I added a new Logger every time, so each test could be logged separately. Originally it was just the one Logger, but that seemed to get different outputs on the Logs, and I wanted to keep them the same. I'll take a look, see what I can come up with.

DavertMik: Logger should be created only on codeception events. Before suite or before init, whatever.

PR #3655, opened 2016-10-30
Fix for issue #3612

### LINKED ISSUE #3612: Yii2 module don't call _fixtures() method of Cest class
I'm tried to add `_fixture()` method to my Cest classes on any suites, but Yii2 module on 169 line: `method_exists($test, self::TEST_FIXTURES_METHOD)` always return false value and loading fixture is passing;
I'm try to debug this moment, on 169 line, $test contain "Codeception\Test\Case" class and have property $testClassInstance with my current Cest class. But when i tried check to any exist method by calling `method_exist($test, 'exist_method_name')` of this Cest class, that always return false value, only methods of "Codeception\Test\Case" can be finded (ex. calling method_exist($test, "test") return true).

* Codeception version: 2.2.5
* PHP Version: 5.5.37
* Operating System: os x
* Installation type: Composer
* Suite configuration:
codeception.yml
```yml
namespace: productEav\tests\backend
actor: Tester
paths:
    tests: .
    log: ./_output
    data: ./_data
    support: ./_support
settings:
    bootstrap: ../../../../../tests/codeception/backend/_bootstrap.php
    colors: true
    memory_limit: 1024M
    log: true
extensions:
    enabled:
config:
    test_entry_url: http://shop-back/index-test.php
```
unit.suite.yml
```yml
class_name: UnitTester
modules:
    enabled:
        - Asserts
        - productEav\tests\backend\Helper\Unit
        - Yii2:
            configFile: '../../../../tests/codeception/config/backend/unit.php'
```

primipilus:
Now in Codeception Yii2 Module Documentation at http://codeception.com/docs/modules/Yii2 on Fixture section now placed code example:
```php
<?php
// inside Cest file or Codeception\TestCase\Unit
public function _fixtures()
{
    return ['posts' => PostsFixture::className()]
}
```
For Cest file it does not work before this commit

Fixes #3612

### DIFF
--- a/src/Codeception/Module/Yii2.php
+++ b/src/Codeception/Module/Yii2.php
@@ -166,12 +166,26 @@ class Yii2 extends Framework implements ActiveRecord, PartedModule
         $this->app = $this->client->getApplication();
 
         // load fixtures before db transaction
-        if (method_exists($test, self::TEST_FIXTURES_METHOD)) {
-            $this->haveFixtures(call_user_func([$test, self::TEST_FIXTURES_METHOD]));
+        if ($test instanceof \Codeception\Test\Cest) {
+            $this->loadFixtures($test->getTestClass());
+        } else {
+            $this->loadFixtures($test);
         }
 
         if ($this->config['cleanup'] && $this->app->has('db')) {
             $this->transaction = $this->app->db->beginTransaction();
+        }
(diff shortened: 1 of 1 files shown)


samdark:
@DavertMik looks OK.

PR #3661, opened 2016-10-31
Fix corrupted console output for unicode characters

grassofhust:


Naktibalda:
@grassofhust Please make this change in 2.2 branch instead.

grassofhust:
@Naktibalda Done.

Naktibalda:
Thanks

PR #3664, opened 2016-11-01
Added services part to the Phalcon module

sergeyklay:


sergeyklay:
@DavertMik Unfortunately I found an annoying bug (d483c1f) which doesn't allow to do fully-fledged functional testing

PR #3665, opened 2016-11-01
Check if Yii2 cleanup option makes tests fail

Naktibalda:


Naktibalda:
I raised a PR to update the configuration of yii2-basic-app instead https://github.com/yiisoft/yii2-app-basic/pull/93

PR #3669, opened 2016-11-02
add gherkin docs

gimler:
I would suggest to put gherkin docs into the modules. I have add an example to demonstrate it. If it would excepted i can do more work. this has two benefits

1. codeception comes with a lot of default gherkin steps like behat does
2. you don't need boilerplate code in your *Tester.php classes like
```
    /**
     * Saving a screenshot
     *
     * @When I save a screenshot in :filename
     */
    public function iSaveAScreenshotIn($filename)
    {
        $this->makeScreenshot($filename);
    }
```

I would also suggest to orientate on behat|mink|behatch to make the migration as simple as possible.

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -544,6 +544,8 @@ class WebDriver extends CodeceptionModule implements
      * ?>
      * ```
      *
+     * @When I save a screenshot in :name
+     *
      * @param $name
      */
     public function makeScreenshot($name)


DavertMik:
> codeception comes with a lot of default gherkin steps like behat does

No, it's not :open_mouth: 

> you don't need boilerplate code in your *Tester.php classes like

You need them. The idea is to clear your business step from implementation. I save screenshot can't be trated as a business value or part of live documentation for the application. It is considered harmful to write *tests* with *gherkin* syntax. 

The idea is not to mix test steps with business steps. Expected application behavior with implementation of its expectation. So by default Gherkin steps are empty as they rely on business rules of your application.

Here is basic example:
```
Given I have 5 items in cart
When I add 1 more item
Then I should have 6 items in cart
```
Here we don't specify any UI element, not any technical details on how it should implemented or tested. However, this provides example of an application feature. Does it clicks or makes screenshot? It doesn't matter. This can be implemented for web, android app, as unit test, desktop app, etc, and tested there.

So I think this is not relevant.

> I would also suggest to orientate on behat|mink|behatch to make the migration as simple as possible.

The only good way to do this to have a tool that translates mink steps to WebDriver/PhpBrowser steps:

```
When I click "Submit" => $I->click('Submit')
```
However, I'm not sure they can be translated without issues.

PR #3675, opened 2016-11-02
Allow to get groups from scenario

frantzen:
Hi!
We use groups also to indicate that a test should, for instance, not being executed on firefox. So we do a:
`* @group non_firefox`

It would be great to have a getter for groups in the Scenario class.
Having that, we can do things like:

`        $groups = $scenario->getGroups();
`
`       if (in_array('non_firefox', $groups) and $scenario->current('browser') == 'firefox') {
            $scenario->skip();
        }`

Thanks for this great tool!
Lars

### DIFF
--- a/src/Codeception/Scenario.php
+++ b/src/Codeception/Scenario.php
@@ -51,6 +51,11 @@ class Scenario
         return $this->metadata->getFeature();
     }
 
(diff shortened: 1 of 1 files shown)


DavertMik:
Thanks! Looks helpful

PR #3678, opened 2016-11-02
Click title support

gimler:


Naktibalda:
Please make a pull request to 2.2 branch instead

DavertMik:
Thanks. Can you also implement this for `Codeception\Lib\InnerBrowser`?

tests/web/WebDriverTest.php
```diff
@@ -413,14 +413,21 @@ public function testWebDriverByLocators()
...
         $this->module->amOnPage('/form/checkbox');
         $this->module->checkOption(WebDriverBy::name('terms'));
+        
         $this->module->amOnPage('/');
         $this->module->seeElement(WebDriverBy::linkText('Test'));
         $this->module->click(WebDriverBy::linkText('Test'));
         $this->module->seeCurrentUrlEquals('/form/hidden');
+
+        $this->module->amOnPage('/form/example7');
+        $this->module->click('Chocolate Bar');
```
Naktibalda: A problem with this test is that it doesn't test your change.
It passes with PhpBrowser and unpatched WebDriver module, because `Chocolate Bar` matches
`<span>Buy Chocolate Bar</span>`
https://github.com/Codeception/Codeception/blob/2.2/tests/data/app/view/form/example7.php

PR #3689, opened 2016-11-05
[Sqlite] Use _ROWID_ to remove inserted rows after test

### LINKED ISSUE #3680: Sqlite database is not cleaned after last test
Run this test:
```php
class LastTestCleanedTest extends \Codeception\Test\Unit {
        /**
        * @var \UnitTester
        */
        protected $tester;

        public function testHaveInDatabase() {
                $this->tester->haveInDatabase('users', ['username' => 'John', 'email' => null]);
                $this->tester->seeInDatabase('users', ['email' => null]);
        }

        public function testDatabaseIsCleaned() {
                $this->tester->dontSeeInDatabase('users', ['email' => null]);
        }
}
```
After running ```codecept run``` the database does not show John as expected. However if I remove the second test like so:
```php
class LastTestCleanedTest extends \Codeception\Test\Unit {
        /**
        * @var \UnitTester
        */
        protected $tester;

        public function testHaveInDatabase() {
                $this->tester->haveInDatabase('users', ['username' => 'John', 'email' => null]);
                $this->tester->seeInDatabase('users', ['email' => null]);
        }
}
```
John does appear in the (sqlite) database. It should have been cleaned.

### Details

* Codeception version: 2.2.6
* PHP Version: 7.0.8
* Operating System: Lubuntu
* Installation type: Composer
* List of installed packages (`composer show`): none other than codeception/codeception and its dependencies
* Suite configuration:

```yml
# paste suite config here
actor: Tester
paths:
    tests: tests
    log: tests/_output
    data: tests/_data
    support: tests/_support
    envs: tests/_envs
settings:
    bootstrap: _bootstrap.php
    colors: true
    memory_limit: 1024M
extensions:
    enabled:
modules:
    config:
        Db:
            dsn: 'sqlite:tests/_data/sqlite.db'
            user: ''
            password: ''
            dump: tests/_data/dump.sql
```
tests/_data/dump.sql:
```sql
DROP TABLE IF EXISTS users;
CREATE TABLE USERS(username VARCHAR(20) NOT NULL PRIMARY KEY, email VARCHAR(20));
```

Naktibalda:
Fixes #3680 

`removeInserted` didn't work for sqlite, because I didn't know that lastInsertId returns ROWID if sqlite table has it (default behaviour).

### DIFF
--- a/src/Codeception/Lib/Driver/Sqlite.php
+++ b/src/Codeception/Lib/Driver/Sqlite.php
@@ -53,6 +53,10 @@ class Sqlite extends Db
     public function getPrimaryKey($tableName)
     {
         if (!isset($this->primaryKeys[$tableName])) {
+            if ($this->hasRowId($tableName)) {
+                return $this->primaryKeys[$tableName] = ['_ROWID_'];
+            }
+
             $primaryKey = [];
             $query = 'PRAGMA table_info(' . $this->getQuotedName($tableName) . ')';
             $stmt = $this->executeQuery($query, []);
@@ -68,5 +72,18 @@ class Sqlite extends Db
         }
 
         return $this->primaryKeys[$tableName];
+    }
+
+    /**
+     * @param $tableName
+     * @return bool
+     */
+    private function hasRowId($tableName)
+    {
+        $params = ['type' => 'table', 'name' => $tableName];
+        $select = $this->select('sql', 'sqlite_master', $params);
+        $result = $this->executeQuery($select, $params);
+        $sql = $result->fetchColumn(0);
+        return strpos($sql, ') WITHOUT ROWID') === false;
     }
 }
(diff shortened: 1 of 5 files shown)


DavertMik:
I'm approving but PHP 5.4 is not. Current Travis failure is related:

```
General error: 1 near "WITHOUT": syntax error  
```

Naktibalda:
I prefer dropping support for 5.4 :)

I haven't managed to find if `WITHOUT ROWID` is a new feature or if has always been there, but Travis made it obvious that version of sqlite bundled with PHP 5.4 does not support it.

To make tests pass I can:
a) Delete the tests (I don't like this option),
b) Create separate dump files for PHP 5.4 and newer versions and skip some tests on 5.4.

sergeyklay:
I vote for B

richardbrinkman:
The ```WITHOUT ROWID``` keyword exists since SQLite version 3.8.2 (2013-12-06).
http://sqlite.org/withoutrowid.html#compatibility

I would prefer option B as well.

DavertMik:
Great! Thank you
And dropping PHP 5.4 support is nice idea as well :)

PR #3686, opened 2016-11-05
Fix #3225

Mitrichius:
Fix #3225

**Html part** - only one error/fail was saved, added nested array to fix it.

**Console part** - only the first failed step used always. 
1. Renamed `$this->fails` to `$this->conditionalFails`, as it really contains only `ConditionalAssertion` steps.
2. Added `$this->failedStep`.
3. Changed fails' handling in `printScenarioFail()`.
4. In `writelnFinishedTest()` renamed `$conditionalFails` to `$conditionalFailsMessage` to separate it from `$this->conditionalFails`.
5. Also removed excess second argument from `printScenarioTrace()` call. 

**Fixed html**: [report.html.zip](https://github.com/Codeception/Codeception/files/573155/report.html.zip)

**Fixed console output**:
```
Codeception PHP Testing Framework v2.2.6
Powered by PHPUnit 4.8.27 by Sebastian Bergmann and contributors.

Api Tests (1) ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
CanCept: Perform actions and see result
Signature: CanCept
Test: tests/api/temp/CanCept.php
Scenario --
 I am on page "/"
 I can see element "//nothere"
 FAIL 

 I can see element "//nothereeither"
 FAIL 

 I can see element "//doesntexist"
 FAIL 

 PASSED 

--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------


Time: 1.05 seconds, Memory: 22.00MB

There were 3 failures:

---------
1) CanCept: Perform actions and see result
 Test  tests/api/temp/CanCept.php
 Step  Can see element "//nothere"
 Fail  Element located either by name, CSS or XPath element with '//nothere' was not found.

Scenario Steps:

 4. $I->canSeeElement("//doesntexist") at tests/api/temp/CanCept.php:9
 3. $I->canSeeElement("//nothereeither") at tests/api/temp/CanCept.php:8
 2. $I->canSeeElement("//nothere") at tests/api/temp/CanCept.php:7
 1. $I->amOnPage("/") at tests/api/temp/CanCept.php:5


---------
2) CanCept: Perform actions and see result
 Test  tests/api/temp/CanCept.php
 Step  Can see element "//nothereeither"
 Fail  Element located either by name, CSS or XPath element with '//nothereeither' was not found.

Scenario Steps:

 4. $I->canSeeElement("//doesntexist") at tests/api/temp/CanCept.php:9
[...]

PR #3697, opened 2016-11-07
Added test for #3686, minor refactoring

DavertMik:


src/Codeception/Module/Filesystem.php
```diff
@@ -265,7 +265,7 @@ public function seeFileFound($filename, $path = '')
             \PHPUnit_Framework_Assert::assertFileExists($file);
             return;
         }
-        $this->fail("$filename in $path");
+        $this->fail("File \"$filename\" not found at \"$path\"");
```
DavertMik: required for a better test message which we verify inside our test

sergeyklay:
```php
$this->fail("File '{$filename}' not found at '{$path}'");
```

?

DavertMik: Assertion message for `seeFileFound`, 

before it that was like: `Fail: file.txt in /home/davert/Codeception`.
now it is: '`Fail: File "file.txt" not found at "/home/davert/Codeception"`.

PR #3717, opened 2016-11-10
Add more configure options for webdriver path.

sven-carstens-udg:
The geckodriver has no option to set the base url. It will only generate failues if used with the default path "/wd/hub". It will work fine if we use the new optional path config option and set it to "/".


DavertMik:
Thank you!

PR #3718, opened 2016-11-10
Documentation Improvements - Chapter 06a

CJDennis:
Improved the grammar of Chapter 06a.


DavertMik:
You are awesome! Thank you very much. That was serious editing work and it is really appreciated.

Issue #3706, opened 2016-11-09
Yii2: @webroot dir


githubjeka:
#### What are you trying to achieve?
View page has code:
```
$this->registerJsFile('@web/js/d3Temperature.js',
    ['depends' => [\yii\web\JqueryAsset::className(), \common\assets\D3Asset::className()]]);
```

After run the command:
> codecept run functional

`@webroot` doesn't created, because https://github.com/Codeception/Codeception/blob/5654704dd7dde6d8b891666484c1f00d36b394f0/src/Codeception/Lib/Connector/Yii2.php#L255

#### What do you get instead?

``@webroot`` created

This fix https://github.com/Codeception/Codeception/pull/3387 isn't correct.

githubjeka:
I propose to revert #3387 by https://github.com/yiisoft/yii2-app-basic/pull/95

And delete mockAssetManager:
```php
private function mockAssetManager()
{
    $this->app->set('assetManager', Stub::make('yii\web\AssetManager'));
}
```

Because user-coder may want to check whether there is the necessary asset files in directories, through functional tests. Let he will decide to needs this feature for mock component  in his application youself.

samdark:
Checking files is a job of unit tests.

githubjeka:
I agree but for exampe:
```
$I->amOnPage('/');
$I->openFile(\Yii::$app->assetManager->getPublishedPath('@web/js/my-app'));
$I->seeInThisFile('data-url-create = "site/create"');
```

Are you suppose?

samdark:
Again, this is low level stuff and that's the thing for unit tests (integration tests to be precise). Functional tests are to verify more high-level user interaction scenarios.

PR #3719, opened 2016-11-10
[Yii2] Prevent publishing of assets and `@webroot` directory creation.

dmitry-kulikov:
Fix for #3706.

### DIFF
--- a/src/Codeception/Lib/Connector/Yii2.php
+++ b/src/Codeception/Lib/Connector/Yii2.php
@@ -275,6 +275,16 @@ class Yii2 extends Client
 
     private function mockAssetManager()
     {
-        $this->app->set('assetManager', Stub::make('yii\web\AssetManager', ['bundles' => false]));
+        $this->app->set(
+            'assetManager',
+            Stub::make(
+                'yii\web\AssetManager',
+                [
+                    'publish' => function () {
+                        return null;
+                    },
+                ]
+            )
+        );
     }
 }


src/Codeception/Lib/Connector/Yii2.php
```diff
@@ -275,6 +275,16 @@ protected function persistDb()
 
     private function mockAssetManager()
     {
-        $this->app->set('assetManager', Stub::make('yii\web\AssetManager', ['bundles' => false]));
```
DavertMik: I know that was a dirty fix but it worked for some cases. At least for me it didn't create `@webroot`. Are we 100% sure that `@webroot` won't be created ever again with this patch?

githubjeka: `@webroot` created via **protected** methods `yii\web\AssetManager::publishFile` and `yii\web\AssetManager::publishDirectory`, but Application uses **public** `yii\web\AssetManager::publish` that used `publishDirectory` and `publishFile` functions.

Why have you stopped to create this directory? Why didn't use `@webroot/assets/` that uses in the real application? You should set correct path to `@webroot/assets/` for that.

dmitry-kulikov: > Are we 100% sure that `@webroot` won't be created ever again with this patch?

With new patch `@webroot` won't be created until significant changes will happen in Yii.

Alternatively we can mock "low-level" protected methods
```php
Stub::make(
    'yii\web\AssetManager',
    [
        'publishFile' => function () {
            return null;
        },
        'publishDirectory' => function () {
            return null;
        },
    ]
);
```
It may be better at some conditions but I think generally it worse because chances that signature of 2 protected methods will be changed are higher than same for 1 public `yii\web\AssetManager::publish`.
Other problem may be overriding of AssetManager by developer in concrete project - if developer overrides AssetManager and code relies on new methods then it is hard to predict what will happen. Then we could use `Maybe` class here.

Also we could specify `basePath` and publish assets:
```php
Stub::make(
    'yii\web\AssetManager',
    [
        'basePath' => '/path/to/webroot/assets',
    ]
);
```
but what should be instead of `/path/to/webroot/assets` is unclear. Should developer specify it in config or it is constant path to `tests/_output/assets` (just example)?

githubjeka: By default is `'basePath' => '@webroot/assets'`. That why you should set correct `@webroot` alias [like in real application](http://www.yiiframework.com/doc-2.0/guide-concept-aliases.html#predefined-aliases). 
> @webroot, the Web root directory of the currently running Web application. **It is determined based on the directory containing the entry script**

githubjeka: [IMHO](https://github.com/Codeception/Codeception/issues/3706#issuecomment-260045843)

dmitry-kulikov:
Since #3706 fixed via specifying of `basePath` for AssetManager I don't see any sense in this PR. Probably make sense to remove `\Codeception\Lib\Connector\Yii2::mockAssetManager` but this not subject of this PR.

PR #3727, opened 2016-11-11
Fix grammar errors: it's should be its

jdq22:


jdq22:
I'm not sure why the builds are failing since I only modified documentation. This should still be merge-able. Please let me know if there's anything I can do.

samdark:
@jdq22 everything's fine. Tests are failing for another reason.

DavertMik:
Are you sure you want to patch 1.8 branch? Current Codeception branch is 2.2 and latest version is 2.2.6.
This change will be applied to 1.8 only... I'm not sure eveyone ever will notice that.
Also, those md files are generated from source code of PhpBrowser. So ideally it would be good if a patch goes to source code. At least it won't be overwritten on next documentation build

jdq22:
@DavertMik: Ah, thanks for letting me know. Didn't realize they were being auto-generated. Will be able to make the change tomorrow probably.

And yes, we're using 1.8 for PHP 5.3.

DavertMik:
On second thought, I don't see any problems in merging this. We are very unlikely to update 1.8 docs soon

PR #3730, opened 2016-11-13
Dataprovider

edno:
Core implementation of the `@dataprovider` annotation based on the extension **[codeception-dataprovider](https://github.com/edno/codeception-dataprovider/issues/2)**.
```php
     /**
      * @dataprovider __fileSource
      */
      public function filesExists(AcceptanceTester $I, Codeception\Example $example)
      {
          $I->amInPath($example['path']);
          $I->seeFileFound($example['file']);
      }

      /**
       * @return array
       */
      public static function __fileSource()
      {
          return[
              ['path' => ".", 'file' => "scenario.suite.yml"],
              ['path' => ".",  'file' => "dummy.suite.yml"],
              ['path' => ".",  'file' => "unit.suite.yml"]
          ];
      }
```

Unlikely the extenstion, the core implemention supports the mix of @dataprovider and @example annotation for a test.
```php
      /**
       * @dataprovider __fileSource
       * @example(path=".", file="skipped.suite.yml")
       */
       public function fileExists(AcceptanceTester $I, Codeception\Example $example)
       {
           $I->amInPath($example['path']);
           $I->seeFileFound($example['file']);
       }
```

This will deprecate the extension **codeception-dataprovider** once released.

There are 2 commits for this PR:
* code + tests
* documentation

### DIFF
--- a/docs/07-AdvancedUsage.md
+++ b/docs/07-AdvancedUsage.md
@@ -1,6 +1,6 @@
 # Advanced Usage
 
-In this chapter we will cover some techniques and options that you can use to improve your testing experience and stay with better organization of your project. 
+In this chapter we will cover some techniques and options that you can use to improve your testing experience and stay with better organization of your project.
 
 ## Cest Classes
 
@@ -28,7 +28,7 @@ class BasicCest
     }
 
     // tests
-    public function tryToTest(\AcceptanceTester $I) 
+    public function tryToTest(\AcceptanceTester $I)
     {    
     }
 }
@@ -46,7 +46,7 @@ As you see, we are passing Actor object into `tryToTest` method. It allows us to
 class BasicCest
 {
     // test
-    public function checkLogin(\AcceptanceTester $I) 
+    public function checkLogin(\AcceptanceTester $I)
     {
         $I->wantTo('log in to site');
         $I->amOnPage('/');
@@ -82,17 +82,17 @@ class SignUpCest
      * @var Helper\NavBarHelper
      */
     protected $navBar;
- 
+
     protected function _inject(\Helper\SignUp $signUp, \Helper\NavBar $navBar)
     {
         $this->signUp = $signUp;
         $this->navBar = $navBar;
     }
-    
+
     public function signUp(\AcceptanceTester $I)
     {
         $I->wantTo('sign up');
- 
+
         $this->navBar->click('Sign up');
         $this->signUp->register([
             'first_name'            => 'Joe',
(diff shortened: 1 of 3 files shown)


Naktibalda:
@edno please make changes suggested by Nitpick
And it would be better to make a pull request to 2.2 branch.
Master will not be released for many months.

DavertMik:
Thanks!
But yeah, please resend it to 2.2 branch.

docs/07-AdvancedUsage.md
```diff
@@ -217,6 +217,28 @@ These examples can be written using Doctrine-style annotation syntax as well:
   }
 ```
 
+You can also use the `@dataprovider` annotation for creating dynamic examples, using a public static method for providing example data:
```
DavertMik: why *static* method? To be compatible with PHPUnti it should be an instance method. And I don't see any technical limitation why not.

src/Codeception/Test/Loader/Cest.php
```diff
@@ -50,6 +52,27 @@ function ($v) {
...
+                            $file, "DataProvider $dataMethod for $testClass->$method is invalid or not callable."
+                            . PHP_EOL .
+                            "Make sure that the dataprovider method is a public static function."
+                        );
+                    }
+                    // allow to mix axample and dataprovider annotations
+                    if(count($examples)) {
+                        $examples = array_merge($examples, $testClass::$dataMethod());
+                    } else {
+                        $examples = $testClass::$dataMethod();
```
DavertMik: static call can be replaced with `call_user_func([$unit, $dataMethod])` to call dataprovider method.
Also dataprovider can be allowed to be marked as protected, so it won't need the `_` prefix in it. This way we can use `ReflectionHelper::invokePrivateMethod()` to call it. 

I'd suggest to have data providers as protected methods. It should be mentioned in documentations that they at least should be protected or starting with `_` prefix so they won't become tests by themselves

mikeuding: I agree on this. I was running into this "problem" when using the dataprovider extension. Other than that, i make full use of this extension at my current project so i would love to see this extension being imported into the core asap

DavertMik: Great. So could you do the improvements which I pointed to?

edno: Sounds like a reasonable request :relieved:

edno:
I close this PR since changes are available in PR #3737

src/Codeception/Test/Loader/Cest.php
edno: In the new PR, the method can be whatever the tester want, but recommendation are protected or public prefixed with `_`

docs/07-AdvancedUsage.md
edno: See comment below

src/Codeception/Test/Loader/Cest.php
mikeuding: Ah i am sorry guys, looks like i missed these last messages. If there is anything i can contribute to, i feel happy to help.

PR #3736, opened 2016-11-14
[WebDriver] Documented operation of waitForText method

davidjeddy:
ADDED notes to WebDriver->WaitForText() DocBlock about how the method operates.

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -1953,7 +1953,7 @@ class WebDriver extends CodeceptionModule implements
     /**
      * Waits up to $timeout seconds for the given string to appear on the page.
      *
-     * Can also be passed a selector to search in, bas a specifica as possible when using selectors.
+     * Can also be passed a selector to search in, be as specific as possible when using selectors.
      * WaitForText() will only watch the first instance of the matching selector / text provided.
      * If the given text doesn't appear, a timeout exception is thrown.
      *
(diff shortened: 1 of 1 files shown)


src/Codeception/Module/WebDriver.php
```diff
@@ -1952,7 +1952,9 @@ public function waitForElementNotVisible($element, $timeout = 10)
 
     /**
      * Waits up to $timeout seconds for the given string to appear on the page.
-     * Can also be passed a selector to search in.
+     *
+     * Can also be passed a selector to search in, bas a specifica as possible when using selectors.
```
Naktibalda: what's a meaning of `bas a specifica`?

davidjeddy: Me being a terrible typer, fixed.

src/Codeception/Module/WebDriver.php
```diff
@@ -1952,7 +1952,9 @@ public function waitForElementNotVisible($element, $timeout = 10)
 
     /**
      * Waits up to $timeout seconds for the given string to appear on the page.
-     * Can also be passed a selector to search in.
+     *
+     * Can also be passed a selector to search in, be as specific as possible when using selectors.
+     * WaitForText() will only watch the first instance of the matching selector / text provided.
```
Naktibalda: Method names aren't case sensitive, but PSR-2 requires them to be formatted this way.

PR #3739, opened 2016-11-15
Adds gitlab CI configuration

richardbrinkman:
Gitlab CI configuration was "work in progress". This Pull Request gives a sample .gitlab-ci.yml file that will run codecept with the built-in php webserver.

### DIFF
--- a/docs/12-ContinuousIntegration.md
+++ b/docs/12-ContinuousIntegration.md
@@ -121,7 +121,7 @@ Travis doesn't provide visualization for XML or HTML reports so you can't view r
 
 If a file ```.gitlab-ci.yml``` exists in the root of the git repository, gitlab will run a pipeline each time you push to the gitlab server. The file configures the docker image that will be called. Below is a sample which loads a php7 docker image, clones your files, installs composer dependencies, runs the built-in php webserver and finally runs codeception:
 
-```
+```yaml
 image: mychiara/php7
 
 # Select what we should cache
--- a/docs/12-ContinuousIntegration.md
+++ b/docs/12-ContinuousIntegration.md
@@ -122,7 +122,8 @@ Travis doesn't provide visualization for XML or HTML reports so you can't view r
 If a file ```.gitlab-ci.yml``` exists in the root of the git repository, gitlab will run a pipeline each time you push to the gitlab server. The file configures the docker image that will be called. Below is a sample which loads a php7 docker image, clones your files, installs composer dependencies, runs the built-in php webserver and finally runs codeception:
 
 ```yaml
-image: mychiara/php7
+# Select image from https://hub.docker.com/_/php/
+image: php:7.0
 
 # Select what we should cache
 cache:
@@ -130,10 +131,14 @@ cache:
   - vendor/
 
 before_script:
+# Install git and unzip (composer will need them)
+- apt-get update && apt-get install -qqy git unzip
+# Install composer
+- curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin --filename=composer
 # Install all project dependencies
 - composer install
 # Run webserver
-- php -S localhost:8085&>/dev/null&
+- php -S localhost:8085 --docroot public &>/dev/null&
 
 # Test
 test:
--- a/docs/12-ContinuousIntegration.md
+++ b/docs/12-ContinuousIntegration.md
@@ -119,7 +119,27 @@ Travis doesn't provide visualization for XML or HTML reports so you can't view r
 
 ## GitLab
 
-[Work In Progress]
+If a file ```.gitlab-ci.yml``` exists in the root of the git repository, gitlab will run a pipeline each time you push to the gitlab server. The file configures the docker image that will be called. Below is a sample which loads a php7 docker image, clones your files, installs composer dependencies, runs the built-in php webserver and finally runs codeception:
+
+```
+image: mychiara/php7
+
+# Select what we should cache
+cache:
+  paths:
+  - vendor/
+
+before_script:
+# Install all project dependencies
+- composer install
+# Run webserver
+- php -S localhost:8085&>/dev/null&
+
+# Test
+test:
+  script:
+  - vendor/bin/codecept run
+```
 
 ## Conclusion
 


docs/12-ContinuousIntegration.md
```diff
@@ -119,8 +119,28 @@ Travis doesn't provide visualization for XML or HTML reports so you can't view r
 
 ## GitLab
 
-[Work In Progress]
+If a file ```.gitlab-ci.yml``` exists in the root of the git repository, gitlab will run a pipeline each time you push to the gitlab server. The file configures the docker image that will be called. Below is a sample which loads a php7 docker image, clones your files, installs composer dependencies, runs the built-in php webserver and finally runs codeception:
+
+```
+image: mychiara/php7
```
DavertMik: mychiara? Maybe we can point to an offical php7 image?

docs/12-ContinuousIntegration.md
```diff
@@ -119,8 +119,28 @@ Travis doesn't provide visualization for XML or HTML reports so you can't view r
 
 ## GitLab
 
-[Work In Progress]
+If a file ```.gitlab-ci.yml``` exists in the root of the git repository, gitlab will run a pipeline each time you push to the gitlab server. The file configures the docker image that will be called. Below is a sample which loads a php7 docker image, clones your files, installs composer dependencies, runs the built-in php webserver and finally runs codeception:
+
+```
```
DavertMik: should be ```yaml

DavertMik:
Thanks, however I have few questions. Also If you have some screenshots of running Codeception on GitLab that would be great to have them there :+1:

docs/12-ContinuousIntegration.md
richardbrinkman: The official php images don't have composer installed. That's why I choose another one that has it.
Of course you can use an official one like php:7.0-apache and install composer in it, but that will be less efficient and makes the documentation slightly longer and more difficult.

docs/12-ContinuousIntegration.md
richardbrinkman: I have added yaml to the code block.

docs/12-ContinuousIntegration.md
richardbrinkman: Also git needs to be installed and therefore also: git-man less libcurl3-gnutls liberror-perl libpopt0 libx11-6 libx11-data libxau6 libxcb1 libxdmcp6 libxext6 libxmuu1 openssh-client patch rsync xauth

The minimal .gitlab-ci.yml using an official image will be:
```yaml
# Select image from https://hub.docker.com/_/php/
image: php:7.0-apache

# Select what we should cache
cache:
  paths:
  - vendor/

before_script:
# Install git and unzip (composer will need them)
- apt-get update && apt-get install -qqy git unzip
# Install composer
- curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin --filename=composer
# Install all project dependencies
- composer install
# Run webserver
- php -S localhost:8085 --docroot public &>/dev/null&

# Test
test:
  script:
  - vendor/bin/codecept run
```
This version will take around a minute longer to execute, then the one in the pull request.
Do you still prefer to use the official docker image?

schmunk42:
Here's another example: https://github.com/dmstr/docker-yii2-app/tree/master/tests

Will add some more info tomorrow....

docs/12-ContinuousIntegration.md
DavertMik: yes, I still prefer official because if this goes to official documentation this image should be reliable. I know nothing about its maintaining policy and I don't want to receive issues if something goes wrong with it. We just need to understand that people will do copy-pasting this configs today, tomorrow, and years after. And we need to be sure that this will work anyway.

That's why I prefer official. Also, didn't you check `codeception/codeception` image? It has composer built in as well

richardbrinkman: I have tried the codeception/codeception image. However it does not contain any shell. Even before the ```before_script``` section is run, I get this error:
```
  [Codeception\Exception\ConfigurationException]                
  Your configuration file `if [ -x /usr/local/bin/bash ]; then  
  	exec /usr/local/bin/bash                                     
  elif [ -x /usr/bin/bash ]; then                               
  	exec /usr/bin/bash                                           
  elif [ -x /bin/bash ]; then                                   
  	exec /bin/bash                                               
  elif [ -x /usr/local/bin/sh ]; then                           
  	exec /usr/local/bin/sh                                       
  elif [ -x /usr/bin/sh ]; then                                 
  	exec /usr/bin/sh                                             
  elif [ -x /bin/sh ]; then                                     
  	exec /bin/sh                                                 
  else                                                          
  	echo shell not found                                         
  	exit 1                                                       
  fi                                                            
  ` could not be found.                                         
```
So I guess I will use the php:7.0 image.

schmunk42:
> Just to note that here: This example is based on Docker runners - which is perfectly fine - which can use a single docker image and work within that, AFAIK. While my example is mainly using docker-compose on a custom runner.

But wouldn't it be better to use `codeception/codeception` in the example, the idea is to mount your PHP source code to that image, run composer install and then codecept. No need to install codeception at all in your project.

richardbrinkman:
I guess you have not read Davertmik's comments on the code. Click the  ```show outdated``` button above to expand his comments. He prefers to use an official docker image and not a custom made one like you propose.

I can only think of two official images: php:7.0 and codeception/codeception:2.2.

I have tried the codeception/codeception image. However it does not contain any shell. Even before the ```before_script``` section is run, I get this error:
```
  [Codeception\Exception\ConfigurationException]                
  Your configuration file `if [ -x /usr/local/bin/bash ]; then  
  	exec /usr/local/bin/bash                                     
  elif [ -x /usr/bin/bash ]; then                               
  	exec /usr/bin/bash                                           
  elif [ -x /bin/bash ]; then                                   
  	exec /bin/bash                                               
  elif [ -x /usr/local/bin/sh ]; then                           
  	exec /usr/local/bin/sh                                       
  elif [ -x /usr/bin/sh ]; then                                 
  	exec /usr/bin/sh                                             
  elif [ -x /bin/sh ]; then                                     
  	exec /bin/sh                                                 
  else                                                          
  	echo shell not found                                         
  	exit 1                                                       
  fi                                                            
  ` could not be found.                                         
```
So I guess I will use the php:7.0 image.

I prefer the current updated version of this pull request above creating a new docker image. Examples in documentation should be as simple as possible. Building a docker image is simply not always needed, just to run codecept on your code.

schmunk42:
> He prefers to use an official docker image and not a custom made one like you propose.

It's the image for this repo, which is based on the official image.

> I have tried the codeception/codeception image. However it does not contain any shell.

You can get a shell with

    docker run -it --entrypoint bash codeception/codeception

---

I noticed another thing, you run `codecept` from `vendor/bin`, while the approach I am describing does not involve installing Codeception at all in your project. Instead you can host-mount your project's source code into the `codeception/codeception` container and run the tests there from a Codeception installation in the container PATH.

This has some implications about the PHP version used and local development setup, for sure. I just wanted to point that out here, since I still think the "docker-way" of using Codeception should be directly from its image.

richardbrinkman:
I really don't understand what schmunk42 is trying to say.
It looks like your intention is to create a docker image to run your website in by running docker on your server.
My intention is to only use docker to check the php source of a website. People building a PHP website don't have to know anything about the docker technology and should not be bothered with making docker images and/or providing a Dockerfile.
schmunk42 pointed to his project https://github.com/dmstr/docker-yii2-app. I tried to understand what you want and imported it in gitlab. It simply fails because you try to run the ```docker``` command on an image that does not have ```docker``` installed. See https://gitlab.com/richardbrinkman/docker-yii2-app/builds/6414676 for the error.

The other thing about calling codecept by ```vendor/bin/codecept```, I can understand, but my personal opinion is that code checked at the Gitlab CI runner should be using the exactly same codecept version as the one everyone on the php-project is working with locally. It should not be the case that the result of a CI-test is different than the one a website-developer gets. Therefore, the best practise is having a composer.lock file in your project telling exactly which version of codeception to use.

schmunk42:
> schmunk42 pointed to his project https://github.com/dmstr/docker-yii2-app. I tried to understand what you want and imported it in gitlab. It simply fails because you try to run the docker command on an image that does not have docker installed. See https://gitlab.com/richardbrinkman/docker-yii2-app/builds/6414676 for the error.

Yes, that's what I mentioned [here](https://github.com/Codeception/Codeception/pull/3739#issuecomment-261210227) - you would need a runner like this: https://github.com/schmunk42/docker-gitlab-runner

I tried to review the log, but I am not able to login at the moment. See here for an example: https://git.hrzg.de/dmstr/docker-yii2-app or https://travis-ci.org/dmstr/docker-yii2-app

> My intention is to only use docker to check the php source of a website. People building a PHP website don't have to know anything about the docker technology and should not be bothered with making docker images and/or providing a Dockerfile.

That's also my point and why I was referring to `codeception/codeception`. I also agree to your last paragraph - I  just noted that we do that differently, and we save a full composer installation, including dependencies for every project we make, because we use Docker. Using `codeception/codeception` is simply faster - but nobody is forced to do that.

@richardbrinkman Could you point me to a public repo of yours where you are using your described workflow? Maybe I can set up a demo.

richardbrinkman:
I use https://gitlab.com/richardbrinkman/codeceptionTest to test the various .gitlab-ci.yml configurations. It has 3 branches: master will use the mychiara/php7 image I found somewhere. Since I was pointed to use a more official one, I branched off the officialdocker branch that uses the php:7.0 image. I also tried (in the codeceptionimage branch) to use the codeception/codeception image, but I could get the last one to pass. On the pipelines tab you can see that the tests on the master and the officialdocker branches passes, whereas on the codeceptionimage branch it fails.
As you can see, I'm not building my own runner. I'm just using the shared runners gitlab.com offers for free.

schmunk42:
I am getting a 404 for the URL to the test repo, even when logged in on GitLab.com. Could you double check if it is public?

richardbrinkman:
I'm sorry. The project was indeed private. I made it public. I also made the fork of docker-yii2-app public so that you can see the failed test as well.

schmunk42:
It's still empty :(

bildschirmfoto 2016-11-20 um 22 07 03

schmunk42:
As far as I see, this is one cause of the problems: https://gitlab.com/gitlab-org/gitlab-ci-multi-runner/issues/1421

Having `codecept` as entrypoint is mainly for shorter typing, like you'd use

    docker run codeception/codeception codecept run

without an entrypoint. 

@DavertMik We could change that or build a separate tag/image for this kind of usage.

I tried another approach with `docker-compose` resulting in [green build](https://gitlab.com/t-munk/docker-yii2-app/pipelines/5066031). But this is more an option for more complex setups, since it also runs very long at the moment.

I'll try a custom codeception image for debugging the entrypoint issue.

schmunk42:
How about this? https://gitlab.com/t-munk/docker-yii2-app/blob/feature/image-only-gitlab.com/.gitlab-ci.yml

Build: https://gitlab.com/t-munk/docker-yii2-app/builds/6445826

richardbrinkman:
Now both https://gitlab.com/richardbrinkman/codeceptionTest and https://gitlab.com/richardbrinkman/docker-yii2-app/builds/6414676 should be accessible to everyone.

@schmunk42 your links got a 404 error too. You have to set the visibility level to public **and** set the feature visibility of the repository and builds to 'everyone with access'.

herloct:
> My intention is to only use docker to check the php source of a website. People building a PHP website don't have to know anything about the docker technology and should not be bothered with making docker images and/or providing a Dockerfile.

@richardbrinkman i think it would be hard if we use docker and assume people doesn't need to learn docker at all.  

What if they need some extension like `pdo_mysql`?  
When using docker image, especially the one based on official php image, we couldn't use "usual" script like `apt-get install php-mysql`, but we need to use something like `docker-php-ext-install pdo_mysql` instead.  
What if the extension depends on another packages, or need to be configured first like `ext-gd`?

I think it would be better to start with some thing more basic, like using ubuntu or debian image.

schmunk42:
Sorry, I've updated the settings for the projects linked [here](https://github.com/Codeception/Codeception/pull/3739#issuecomment-261812397). Please try again.

> What if they need some extension like pdo_mysql?

@herloct Valid point, I also thought about that. Actually Codeception could come with all PHP extension it may need for its modules pre-installed. But there may still cases, where you need to build an own image.

schmunk42:
@richardbrinkman Please also have a look at https://gitlab.com/t-munk/codeceptionTest/builds/6480769

It would run without additional Codeception requirement, the failed unit tests should be just an autoloading issue.

Source: https://gitlab.com/t-munk/codeceptionTest/tree/480791b94bae0fa9f186aac95c8f87730c4bd626

schmunk42:
Btw: Experimental image without `ENTRYPOINT` https://hub.docker.com/r/schmunk42/codeception/tags/

glensc:
having struggled with [gitlab-ci-multi-runner#1421](https://gitlab.com/gitlab-org/gitlab-ci-multi-runner/issues/1421)

it's fine to override `ENTRYPOINT` if you emulate the interface GitLab CI requires:

```
ENTRYPOINT ["/composer.sh"]
```

```
# cat composer.sh 
#!/bin/sh

case "$1" in
        *sh)
                exec "$@"
                exit 1
        ;;
esac

exec composer --ansi "$@"
```

schmunk42:
So, what shall we do?

a) leave `ENTRYPOINT` as is
b) change it to `bash` - not only because of GitLab, but it may also be used in development

DavertMik:
I like having current entrypoint (not bash) but I'm not using Docker image so I'm not sure what is better for usability.

I think the shorter command is the simpler it is... So I choose a)

richardbrinkman:
With the entrypoint the codeception image cannot be used for CI and you are forced to create your own docker image and publish it somewhere or you have to use a general one like debian and install a lot each time you do a ```git push``` which makes it inefficient.
If you only use codeception locally you don't have to use docker at all. Typing ```codecept run``` is way shorter than ```docker run --rm -v /x:/y codeception/codeception run```
So I would prefer removing the entrypoint.

schmunk42:
> With the entrypoint the codeception image cannot be used for CI

This only applies to GitLab Runner or are other Docker CIs also affected?
I asked [here](https://gitlab.com/gitlab-org/gitlab-ci-multi-runner/issues/1421#note_19261677) for an ETA.

DavertMik:
> So I would prefer removing the entrypoint.

Ok, then could I vote for it too, but won't this break BC for current docker image users?

Maybe we can leave entrypoint but use the script you attached? Or maybe as you said have a small image that overrides official one? What if we just put that Dockerfile into this documentation section and ask users to use that custom image for Gitlab CI builds?

herloct:
I think current `ENTRYPOINT` is better.
We should treat Codeception images as an executables, like using normal `codecept`, not as a platform to run `composer install`, `php -S`, or any other thing.

richardbrinkman:
Just out of curiosity, may I ask what the purpose of running the codeception image as an executable is? It seems to me that it is an overkill to:

1. install docker
2. ```docker pull codeception/codeception``` (which effectively downloads a complete debian system + a second copy of php (and possibly a different version than your project uses and if your project uses a php module other than the ones in the codeception image you are out of luck (you have to create your own image then)))
3. be forced to use your database as a docker container as well and have a second version of say mysql
4. ```docker create --name mytest --link mysql:mysql codeception/codeception```
5. ```docker cp . mytest:/repo``` all your code inside the docker container (using volumes bind mounts is not really an option since docker runs effectively as root and you don't want your tests to run as root and the test logs etc written as root (which the user cannot remove))
6. ```docker start mytest run``` to run the test

The standard way of using codecept is way easier:

1. ```composer require --dev codeception/codeception``` (or ```wget http://codeception.com/codecept.phar```)
2. ```vendor/bin/codecept run```

schmunk42:
> docker pull codeception/codeception (which effectively downloads a complete debian system + a second copy of php (and possibly a different version than your project uses and if your project uses a php module other than the ones in the codeception image you are out of luck (you have to create your own image then)))

I mean that's just the Docker workflow, very often you already have these layers already pushed and you just need to download them only once (depending on your CI).
We're also discussing automated builds of other PHP versions here https://github.com/Codeception/Codeception/pull/3760 - I.e. [Alpine would be 53 MB](https://hub.docker.com/r/schmunk42/codeception/tags/)

> be forced to use your database as a docker container as well and have a second version of say mysql

That solely depends on your setup. You can use another DB from your CI or add other RDBMS via Docker containers.

> docker create --name mytest --link mysql:mysql codeception/codeception

I would suggest `docker-compose` here, but anyway in a "classic" setup you need to define this in your CI config, which is not really portable; Docker stacks are.

> docker cp . mytest:/repo all your code inside the docker container (using volumes bind mounts is not really an option since docker runs effectively as root and you don't want your tests to run as root and the test logs etc written as root (which the user cannot remove))

That can also be configured, if needed, AFAIK.

> docker start mytest run to run the test

Or `docker-compose run codecept run`, which can be configured to bring up all additional services you need, eg. Redis, Mongo, Selenium, ....

> The standard way of using codecept is way easier:
>     composer require --dev codeception/codeception (or wget http://codeception.com/codecept.phar)
> vendor/bin/codecept run

That is still an option to you, but you're adding a huge dependency to your project, which is even designed to run as a standalone binary. Pro side here: You can easily switch between Codeception versions.

herloct:
> The standard way of using codecept is way easier

Yes, but to do it we need to:

1. Install php and all required extensions.
2. Install mysql.
3. Install composer and install all dependencies.
4. Adjust our setting to point the mysql

schmunk42:
There is progress at https://gitlab.com/gitlab-org/gitlab-ci-multi-runner/issues/1421

Assuming that it gets implemented as discussed, an example `.gitlab-ci.yml` would look like this:

```
image: 
  name: codeception/codeception
  entrypoint: bash

build:
  stage: build
  script:
    - composer install
    - codecept run
```

I'd also add `--prefer-dist --optimize-autoloader` to composer install but the above is the most minimalistic version I could think of.

Bonus:

```
build:
  [...]
  artifacts:
    name: codeception
    paths:
      - tests/codeception/_output
```

Bonus II:

As a *temporary workaround* it should be possible to build an image, like so:

```
FROM codeception/codeception
ENTRYPOINT bash
```

And link that build to the `codeception/codeception` repository on DockerHub. So it would be always up-to-date. Only con: Tagging needs to be done manually.

oprudkyi:
Some suggestions to improve docs, based on https://gitlab.com/oprudkyi/laravel-bootstrap-adminlte-starter-kit/blob/master/.gitlab-ci.yml :

- real project will require special prebuilt image with all deps (because installing into docker is very slow ~10 minutes)
- cache doesn't work on shared/public runners (I have never seen it), so private/custom runner is needed if you want to save another 10 minutes (as well you need to 'redirect' caches of composer/npm/bower from /home/... into current dir by setting caches in composer.json,  .bowerrc,  .npmrc )
- 'composer install' won't work as root, it should be run as 'install --no-plugins --no-scripts' and then scripts should be run manually 
- phantomjs on docker is hanging very often (30%-50% of runs) https://github.com/ariya/phantomjs/issues/11526 (hanging by self can be avoided  by pageload_timeout https://github.com/Codeception/Codeception/pull/3874 )

schmunk42:
> real project will require special prebuilt image with all deps (because installing into docker is very slow ~10 minutes)

You need to run `install` with a `composer.lock` file in your repository. With https://github.com/hirak/prestissimo installation is usually done in less than 1 minute.

> 'composer install' won't work as root, it should be run as 'install --no-plugins --no-scripts' and then scripts should be run manually

There's a flag for that `COMPOSER_ALLOW_SUPERUSER=1`, see https://github.com/Codeception/Codeception/blob/2.2/Dockerfile#L29

oprudkyi:
> You need to run install with a composer.lock

in this particular case composer.lock isn't commited for reason 

> COMPOSER_ALLOW_SUPERUSER=1

for me it fires error (on php 7.1 only, so it's probably some issues with php 71 or Laravel)
```
/usr/local/bin/php composer.phar install
> php artisan clear-compiled

Warning: require(/builds/oprudkyi/laravel-bootstrap-adminlte-starter-kit/bootstrap/../vendor/autoload.php): failed to open stream: No such file or directory in /builds/oprudkyi/laravel-bootstrap-adminlte-starter-kit/bootstrap/autoload.php on line 17

Fatal error: require(): Failed opening required '/builds/oprudkyi/laravel-bootstrap-adminlte-starter-kit/bootstrap/../vendor/autoload.php' (include_path='.:/usr/local/lib/php') in /builds/oprudkyi/laravel-bootstrap-adminlte-starter-kit/bootstrap/autoload.php on line 17
```

schmunk42:
> in this particular case composer.lock isn't commited for reason

For which reason? Do you want to see if an `update` would work? Why would a prebuilt image help you in this case?

> for me it fires error (on php 7.1 only, so it's probably some issues with php 71 or Laravel)

You are missing `vendor/autoload.php` - maybe due to an (empty?) host-volume?

docs/12-ContinuousIntegration.md
```diff
@@ -119,8 +119,33 @@ Travis doesn't provide visualization for XML or HTML reports so you can't view r
...
+test:
+  script:
+  - vendor/bin/codecept run
+```
 
 ## Conclusion
 
-It is tringly recommended to use Continuous Integration system in development. Codeception is easy to install and run in any CI systems. However, each of them has their differences you should take into account. You can use different repoters to provide output in format expected by CI system.
\ No newline at end of file
+It is tringly recommended to use Continuous Integration system in development. Codeception is easy to install and run in any CI systems. However, each of them has their differences you should take into account. You can use different repoters to provide output in format expected by CI system.
```
KhairulA: It is strongly recommended to use Continuous Integration systems for development. Codeception is easy to install and run on any CI systems. However, each CI system has their differences. You can output different report formats expected by your CI system.

schmunk42:
After working with some other projects, I simply think there are many ways to do this. We might just document multiples ones.

@DavertMik I'd merge the current PR, it's just the current way to do it with a default GitLab runner, because of the outstanding entrypoint issue.

---

Addon: There can be many different ways use Codeception:
- install it in your project
- global installation from image
- use pre-built image

DavertMik:
> I'd merge the current PR

Wooohoo! Finally!

PR #3387, opened 2016-07-28
[Yii2] Fixed `@webroot` directory creation

DavertMik:
Proposed by @dmitry-kulikov

PR #3742, opened 2016-11-15
Yii2: remove mockAssetManager

githubjeka:
`mockAssetManager` prevents setting from test.config file.

https://github.com/yiisoft/yii2-app-advanced/pull/219
https://github.com/yiisoft/yii2-app-basic/pull/95

Now code from PR's doesn't work.

P.S. mock was introduced in [here](https://github.com/Codeception/Codeception/pull/3343/commits/6add1042ef297110c63adbb4281cada4e129b119)

### DIFF
--- a/src/Codeception/Lib/Connector/Yii2.php
+++ b/src/Codeception/Lib/Connector/Yii2.php
@@ -70,7 +70,7 @@ class Yii2 extends Client
         /** @var \yii\web\Application $app */
         $this->app = Yii::createObject($config);
         $this->persistDb();
-        $this->mockMailer($config);       
+        $this->mockMailer($config);
         \Yii::setLogger(new Logger());
     }
 
--- a/src/Codeception/Lib/Connector/Yii2.php
+++ b/src/Codeception/Lib/Connector/Yii2.php
@@ -70,8 +70,7 @@ class Yii2 extends Client
         /** @var \yii\web\Application $app */
         $this->app = Yii::createObject($config);
         $this->persistDb();
-        $this->mockMailer($config);
-        $this->mockAssetManager();
+        $this->mockMailer($config);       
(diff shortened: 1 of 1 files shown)


githubjeka:
Lets wait https://github.com/Codeception/Codeception/pull/3387#issuecomment-260633259

DavertMik:
Code works, but `@webroot` directory is created. That was the case. The most recent discussion about it was here https://github.com/Codeception/Codeception/pull/3719/files Looks like we still need to mock it in some way to disable creating this directory

githubjeka:
One way - or this PR or #3719. Choose your destiny.

samdark:
Tested it. With these changes assets are published to `/assets` which is definitely wrong.

First of all, let's find out why do we want asset publishing in functional tests. I don't see any reason for having them.

githubjeka:
> Tested it. With these changes assets are published to /assets which is definitely wrong.

Did you add the code https://github.com/yiisoft/yii2-app-basic/pull/95/files#diff-595e684beb311d7dead1be4469cfea82R17 ?

> First of all, let's find out why do we want asset publishing in functional tests. I don't see any reason for having them.

Maybe someone will need it for various checks. I understand that these checks should be part of unit tests. But someone need it in functional tests. If we mock this behavior, user will not has possible to disable it.

samdark:
No, I did not :) Anyway, the question remains.

samdark:
I think we want asset publishing disabled by default. Ability to enable it is a pro indeed.

samdark:
So this pull request is OK to be merged.

DavertMik:
Thanks!

githubjeka:
Thanks too.

githubjeka:
Ждите кучку issue от старожил: у меня создаётся @webroot/ ^_^

PR #3745, opened 2016-11-16
typo in service name

bscheshirwork:
use "codecept" above


Naktibalda:
@bscheshirwork please make this change in 2.2 branch instead.

PR #3746, opened 2016-11-16
typo in service name

bscheshirwork:
use "codecept" above

#3745

### DIFF
--- a/docs/12-ParallelExecution.md
+++ b/docs/12-ParallelExecution.md
@@ -41,7 +41,7 @@ We prepared a sample config with codeception, web server, database, and selenium
 ```yaml
 version: '2'
 services:
-  codeception:
+  codecept:
     image: codeception/codeception
     depends_on:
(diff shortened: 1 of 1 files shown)


DavertMik:
@schmunk42 is that ok?

schmunk42:
Yes, but it depends a bit on the final decision of https://github.com/Codeception/Codeception/pull/3739#issuecomment-261810192

In the current image there's a default entrypoint for `codecept` which makes commands a bit shorter to type. In the related issue, there's a discussion about changing that to `bash`.

Any thoughts about https://github.com/Codeception/Codeception/pull/3739#issuecomment-263076079? Would there be other commands than `codecept` to use from the image?

Faryshta:
this is the reason all the pr tests are failing?

Naktibalda:
If this change is still necessary please raise a new pull request to 2.3 branch.

PR #3750, opened 2016-11-17
Update LocalServer.php

ksiv:
Fix http 500 error caused by an empty header while using file_get_contents()

### DIFF
--- a/src/Codeception/Coverage/Subscriber/LocalServer.php
+++ b/src/Codeception/Coverage/Subscriber/LocalServer.php
@@ -33,7 +33,7 @@ class LocalServer extends SuiteSubscriber
     protected $c3Access = [
         'http' => [
             'method' => "GET",
-            'header' => ''
+            'header' => '"User-Agent:Codeception/2.2\r\n"'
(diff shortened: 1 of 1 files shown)


src/Codeception/Coverage/Subscriber/LocalServer.php
```diff
@@ -33,7 +33,7 @@ class LocalServer extends SuiteSubscriber
     protected $c3Access = [
         'http' => [
             'method' => "GET",
-            'header' => ''
+            'header' => '"User-Agent:Codeception/2.2\r\n"'
```
Naktibalda: Quotes don't look right.

ksiv:
concerning q 2. Removed version 
concerning q 1. it's not clear about constant. This is modification of a predefined array. there was 'method' predefined and I added 'header'. It's not important what to place into Header, the idea is to place something because it does not work with some modern servers without any header.  Take a look at web search "file_get_contents 500"

ksiv:
No positive resolution for more than a month.

PR #3753, opened 2016-11-17
Add strict CSS locator example for see()

weshooper:
It works, just isn't documented.


DavertMik:
Sure. However the documentation is generated from the source code. Please update docs for corresponding method at  https://github.com/Codeception/Codeception/blob/2.2/src/Codeception/Lib/Interfaces/Web.php

PR #3754, opened 2016-11-17
fixed typo

PhilippSchaffrath:
Just fixed a little typo.


Naktibalda:
This file is generated from the source code,
please update https://github.com/Codeception/Codeception/blob/2.2/src/Codeception/Module/REST.php#L705 instead.

PR #3755, opened 2016-11-17
Add strict CSS locator example for see()

weshooper:
It works, just isn't documented.

### DIFF
--- a/src/Codeception/Lib/Interfaces/Web.php
+++ b/src/Codeception/Lib/Interfaces/Web.php
@@ -56,10 +56,10 @@ interface Web
      *
      * ```php
      * <?php
-     * $I->dontSee('Login');                    // I can suppose user is already logged in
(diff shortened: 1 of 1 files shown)


src/Codeception/Lib/Interfaces/Web.php
```diff
@@ -59,6 +59,7 @@ public function see($text, $selector = null);
      * $I->dontSee('Login');                    // I can suppose user is already logged in
      * $I->dontSee('Sign Up','h1');             // I can suppose it's not a signup page
      * $I->dontSee('Sign Up','//body/h1');      // with XPath
+     * I->see('Sign Up', ['css' => 'body h1']); // with strict CSS locator
```
DavertMik: $I

weshooper: @DavertMik oops, fixed that - thanks!

PR #3760, opened 2016-11-19
[WIP] Add alpine base for Docker

herloct:
Add `php:7.0-alpine` instead of `php:7.0-cli` as a base, so it could have a lot smaller image.

Image size comparison:

| Base           | Docker Hub | HDD    |
|----------------|-----------:|-------:|
| php:7.0-cli    | 187 MB     | 456 MB |
| php:7.0-alpine |  42 MB     | 102 MB |


DavertMik:
@schmunk42 what do you think?

schmunk42:
Having different versions and images to choose from is definitely nice.

But I'd prefer a separate `Dockerfile-alpine` for this (or placing them in folders), tags should follow the same rules like here https://hub.docker.com/_/php/

herloct:
@schmunk42 I think so too.
But, can we use something like `Dockerfile-alpine` on Docker Hub?
I don't think placing them in folders is a good idea, since `Dockerfile` can't do something like `COPY ../composer.json composer.json` (can't copy folder outside of it's scope), so we need to copy the whole project to that new folder.

CMIIW

schmunk42:
@herloct From reading Docker docs, it first looked to me, that it would be possible to work around the build context issue, but I did not find a solution for this directly.
The only other solution with automated builds on DockerHub would be to place all `Dockerfile` into the project root, which s**ks. Or generating `Dockerfile`s in the build process, which is also not very nice.

---

Here's [another approach](https://github.com/schmunk42/Codeception/blob/6d6f22ceec943aec6b12cc670d1cb2f7770f63a9/build/docker-alpine/Dockerfile) not involving any `COPY` instruction, but installing Codeception with composer. What do you think?

herloct:
> Here's another approach not involving any COPY instruction, but installing Codeception with composer. What do you think?

@schmunk42 the only drawback i could think of was we're limited to released version of codeception.

Also we need to specify development version (or even commit hash) like `composer global require "codeception/codeception:2.2.x-dev" when we need to test the images against development version of codeception.

And if we still chose this approach, ignoring the development release, would fetching `codecept.phar` is better than using composer?

schmunk42:
> limited to released version of codeception.

We could also use `dev-master` or another branch alias, but I am not decided here.
Also, with the approach without `COPY` we could use a separate repo, just for building docker images.
Just downloading the `phar` would be also an option.

I think we should define different use-cases here. The current image is more kind of a development image. 
Using alpine with codecept.phar would be super slim and very lightweight - perfect for CI, like a single binary in a Docker image.

herloct:
> Also, with the approach without COPY we could use a separate repo, just for building docker images.

@schmunk42 I'm in for this.

> I think we should define different use-cases here. The current image is more kind of a development image.
Using alpine with codecept.phar would be super slim and very lightweight - perfect for CI, like a single binary in a Docker image.

@schmunk42 I'm agree with this.  
In fact, i'm always do this approach on my project's codeception images.

schmunk42:
That looks nice: https://github.com/herloct/docker-php-cs-fixer

@herloct Can create tags from the version pattern in the folder name automatically?

herloct:
> @herloct Can create tags from the version pattern in the folder name automatically?

Docker Image tags?  
Maybe we could use git tags and mapping it using regex and `{sourceref}` placeholder like instructed on Build Settings's page. But I'm not sure whether we could use it on folder name.  
I prefer using directory approach than git tags, so I still tagging it manually.

schmunk42:
Currently, I see no sane way how to implement this on DockerHub.

We could use GitLab.com for building the images (they offer free runners now), since it would allow a more complex setup. Or maybe the "1 free repo" of Docker Cloud.

Another pro on the GitLab side would be that we could also parse/create Dockerfiles for various versions during in the build stage.

herloct:
@schmunk42 Maybe we could use `Dockerfile-alpine` approach, set docker build on travis-ci, then push it to docker hub.

schmunk42:
Travis is already used by the standard testing setup, I talked to @DavertMik once about setting up dockerized tests in Travis, but the last decicision was not to rip apart the current working setup.

Would there be a way to run like two TravisCI test suites?

DavertMik:
We have `codeception/base` repo which mirrors the current one. Probably we can set up docker tests there. If you write a `.travis.yml` which will execute Codeception tests inside Docker we can make it to be executed from time to time... Right now the cloning happens once a month, just before the release.

glensc:
@schmunk42 https://github.com/Codeception/Codeception/pull/3760#issuecomment-262319005

i was about to suggest that. while gitlab.com also has docker registry, you can push to docker.io if you let it use your credentials.

i'll share one very complex build: https://gitlab.com/pld-linux/pld
- it builds base image in foreign context (pld linux image in debian!)
- build docker image needed to build base image
- use freshly built image to build image from scratch :)

but gitlab-ci build steps allow you to prepare content outside docker and then just `COPY` what you need. resulting very clean images without build time deps that usually is put to docker images.

glensc:
also, you can leave your project "home" in github, and setup gitlab home to mirror automatically from github, and still do ci jobs there.

DavertMik:
@glensc can gitlab CI be connected to GitHub like Travis does? 
also if get a gitlab mirror who will sponsor that?

glensc:
@DavertMik no, gitlab-ci (ci-runner) is tightly tied to gitlab. also even if you can run gitlab-ci-multi-runner with github, who host the runner? using gitlab.com the repo and runners are already there.

by sponsor you mean what? you can go with the free plan. your project fits the limits (<10G repo size)

but i meant such flow:
- you create gitlab.com/codeception group
- you create gitlab.com/codeception/codeception repo
- gitlab.com/codeception/codeceiption is git repo configured to mirror from github.com/codeception/codeception
- codeception repo has .gitlab-ci.yml committed to github repo, that gets mirrored to gitlab repo
- on repo sync (performed automatically by gitlab.com) the ci jobs get started
- indicate that gitlab.com/codeception/codeception is just a mirror, disable issues and merge requests and all other features by ci and docker registry.

there may be delay with the syncing, so i suggest fist develop the ci jobs directly in gitlab.com perhaps your own fork under your namespace, and when finished push to github (which will be pushed to gitlab), no further changes needed.

herloct:
@DavertMik @schmunk42 @glensc can't we just create new repo on github, something like `codeception/docker`, dedicated for docker images, and put the `.travis.yml` there?

glensc:
@herloct yes you can, but i was discussion option to use gitlab-ci instead of travis-ci or dockerhub-automated-builds.

surely the complex build i described can be done in travis as well. altho it's not that flexible as gitlab-ci, it does not have concept or "jobs", i.e several steps and several stages.

herloct:
@DavertMik @schmunk42 @glensc I'll try using `Dockerfile-alpine` combined with GitLab CI approach.

schmunk42:
I also think GitLab CI is the most flexible solution currently.

@DavertMik I think you need to register/reserve the namespace "Codeception" at gitlab.com - and also an additional user on DockerHub which has push access to codeception/codeception would be required.

Push tokens can be stored as project variables, but we still need to pay attention that they can not be exposed in PRs, I think Travis has a better solution here with secret variables.

Anyone has experience with this?

herloct:
> Push tokens can be stored as project variables, but we still need to pay attention that they can not be exposed in PRs, I think Travis has a better solution here with secret variables.

I think GitLab also has Secret Variables at https://gitlab.com/your_vendor/your_project/variables. But maybe it different from what Travis have.

glensc:
i can confirm that gitlab "variables" are also "secret" they are replaced with `xxxxx` in build output. it iterates over the variable values so it should match and replace them appearing everywhere.

.gitlab-ci.yml
```diff
@@ -0,0 +1,63 @@
...
+  - docker:dind
+
+variables:
+  # @TODO: should be codeception/codeception
+  ALPINE_IMAGE: herloct/codeception-test:$CI_BUILD_REF_NAME-alpine
+  # @TODO: should be registry.gitlab.com/codeception/codeception
+  ALPINE_IMAGE_TEST: registry.gitlab.com/herloct/codeception:$CI_BUILD_REF_NAME-alpine-test
+
+before_script:
+  - docker login -u gitlab-ci-token -p $CI_BUILD_TOKEN registry.gitlab.com
```
glensc: use `$CI_REGISTRY` variable for `registry.gitlab.com`

.gitlab-ci.yml
```diff
@@ -0,0 +1,63 @@
+---
+image: docker:latest
+services:
+  - docker:dind
+
+variables:
+  # @TODO: should be codeception/codeception
+  ALPINE_IMAGE: herloct/codeception-test:$CI_BUILD_REF_NAME-alpine
+  # @TODO: should be registry.gitlab.com/codeception/codeception
+  ALPINE_IMAGE_TEST: registry.gitlab.com/herloct/codeception:$CI_BUILD_REF_NAME-alpine-test
```
glensc: use `$CI_REGISTRY_IMAGE` variable, and remove TODO

herloct:
I have created GitLab repository at https://gitlab.com/herloct/Codeception/ which is mirroring my codeception's fork on GitHub.

And also the images are being pushed to my docker repository at https://hub.docker.com/r/herloct/codeception-test

Dockerfile-alpine
```diff
@@ -0,0 +1,57 @@
+FROM php:7.0-alpine
+
+MAINTAINER Tobias Munk tobias@diemeisterei.de
```
glensc: `MAINTAINER` is deprecated use `LABEL` instead

https://docs.docker.com/engine/userguide/labels-custom-metadata/
https://docs.docker.com/engine/reference/builder/#label

Dockerfile-alpine
```diff
@@ -0,0 +1,57 @@
...
+        --install-dir=/usr/local/bin
+RUN composer global require --optimize-autoloader \
+        "hirak/prestissimo"
+
+# Prepare application
+WORKDIR /repo
+
+# Install vendor
+COPY ./composer.json /repo/composer.json
+RUN composer install --prefer-dist --optimize-autoloader
```
glensc: `--no-dev` perhaps, do you need dev dependencies in the image?

Dockerfile-alpine
```diff
@@ -0,0 +1,57 @@
...
+
+# Add source-code
+COPY . /repo
+
+ENV PATH /repo:${PATH}
+ENTRYPOINT ["codecept"]
+
+# Prepare host-volume working directory
+RUN mkdir /project
+WORKDIR /project
```
glensc: you should add `/project` as `VOLUME` and then `mkdir` is superfluous as `VOLUME` directive will mkdir itself

Dockerfile-alpine
herloct: Is it okay to use `LABEL` now instead of `MAINTAINER`?

Because reading it [here](https://github.com/docker/docker/blob/master/docs/deprecated.md#maintainer-in-dockerfile), the deprecation start at v1.13.0, while v1.13.0 is currently in pre-release state (RC2).

glensc: `LABEL` has existed for some time already. the `MAINTAINER` has had instrumentation ever, therefore it's deprecated (not removed!).

.gitlab-ci.yml
```diff
@@ -6,11 +6,10 @@ services:
 variables:
   # @TODO: should be codeception/codeception
   ALPINE_IMAGE: herloct/codeception-test:$CI_BUILD_REF_NAME-alpine
```
glensc: `DOCKERHUB_ALPINE_IMAGE` to be clear what var is

herloct: Should we use Secret Variables for `DOCKERHUB_ALPINE_IMAGE`?
So we could merge it later without changing the `.gitlab-ci.yml`.

glensc:
On 24.11.2016 17:33, Herloct wrote:
>
> *@herloct* commented on this pull request.
>
> ------------------------------------------------------------------------
>
> In .gitlab-ci.yml <https://github.com/Codeception/Codeception/pull/3760>:
>
> > @@ -6,11 +6,10 @@ services:
>   variables:
>     # @TODO: should be codeception/codeception
>     ALPINE_IMAGE: herloct/codeception-test:$CI_BUILD_REF_NAME-alpine
>
> Should we use Secret Variables for |DOCKERHUB_ALPINE_IMAGE|?
> So we could merge it later without changing the |.gitlab-ci.yml|.
>

[i can't find this post in webpage anymore, so replying via email]

if you intend to have this customizable, yes, define the variable in 
project settings and adjust ci config accordingly to fail or do nothing 
if variable is not present:

   test -n "$DOCKERHUB_ALPINE_IMAGE" && exit 0 # or exit 1 depending if 
you want to succeed or fail

but i think you want only the "project/path" being configurable, so the 
":tag" part comes from ci:

DOCKERHUB_IMAGE_PROJECT=herloct/codeception-test


-- 
glen

glensc:
btw, there's new way of specifying docker login credentials in gitlab-ci-multi-runner 1.8.0:

[v1.8.0](https://gitlab.com/gitlab-org/gitlab-ci-multi-runner/blob/master/CHANGELOG.md)
 Add support for using private docker registries [!386](https://gitlab.com/gitlab-org/gitlab-ci-multi-runner/merge_requests/386)

personally i do not understand it, why it's better to include whole auth.json than just use variables for each specific login/password pair.

i think it's usable in gitlab.com as they always deploy the latest version

herloct:
> if you intend to have this customizable, yes, define the variable in project settings and adjust ci config accordingly to fail or do nothing if variable is not present

Looks like we could define variable on both YAML variables and secret variables, the secret variables would override the YAML variables.

.gitlab-ci.yml
```diff
@@ -0,0 +1,95 @@
...
+
+docker_build:alpine:
+  <<: *dind_before_script
+  stage: build
+  script:
+    - >
+      docker build --rm
+      --tag $ALPINE_IMAGE_TEST
+      --file Dockerfile-alpine .
+    - docker push $ALPINE_IMAGE_TEST
```
schmunk42: Do you need to push the image in build, because you load it from the registry in other stages?

herloct: Because I can't find any other way to share the built image on other stages (test and deploy stage).

herloct:
Now i have a weird problem on GitLab.  
When deleting some unused branch on GitHub, the unused branch keep showing up on GitLab.  
Even when I manually deleting it on GitLab, it keep coming back when synced with GitHub.

herloct:
I wonder if we could change this:

```dockerfile
COPY . /repo
```

Into something like this:

```dockerfile
RUN git clone -b 2.2 --single-branch git@github.com:Codeception/Codeception.git /repo
```

So that we could put `Dockerfile`s on dedicated directory, something like this:

```
docs/
docker/
  alpine/
    Dockerfile
  Dockerfile
ext/
...
```

And we doesn't really need GitLab anymore.

.gitlab-ci.yml
glensc: you should:
- create docker image tag in gitlab.com registry named by branch+commit hash
- pull that in test stage
- tag & push the final image in publish stage

see this excellent example:
https://gitlab.com/markpundsack/docker-example

glensc:
https://github.com/Codeception/Codeception/pull/3760#issuecomment-262869088

search and report issue here: https://gitlab.com/gitlab-org/gitlab-ce/issues

glensc:
>    composer install --prefer-dist --optimize-autoloader --working-dir /repo && \

do you need to cleanup composer cache as well after this?

https://github.com/Codeception/Codeception/pull/3760#issuecomment-262894271
haven't paid slightest attention what this issue is about, can't comment :D

.gitlab-ci.yml
herloct: @glensc yup, already done that, except the commit hash part.

herloct:
> do you need to cleanup composer cache as well after this? 

Yes, of course.

> haven't paid slightest attention what this issue is about, can't comment :D

:'(

.gitlab-ci.yml
schmunk42: I'd also run the image building completely on GitLab for now.
Including a cleanup at the end.

If everything works fine, we can also push the images to DockerHub.

DavertMik:
Sorry, I was a bit lost in this discussion. To my personal view there is no big issue with loading a bigger image, so probably that's not as important. And yes, the automatic build can only work for one repo.

schmunk42:
@DavertMik I agree. We are i.e. using Alpine for our PHP applications, but faced a lot of problems with it.
For debugging and development I'd also recommend to use Debian as a base.

PR #3765, opened 2016-11-22
Fixed action inclusion when includeInheritedActions is set to false

omnilight:
When you set in your $includeInheritedActions = false, then no actions are included at all.

This PR fixes that


DavertMik:
Thanks!

PR #3770, opened 2016-11-26
Bugfix for Symfony Module

olecarl:
Bugfix for 
"Symfony bootstrap file not found... Please specify path to bootstrap file using `var_path`"


DavertMik:
Sorry, it's not a bugfix but a breaking change. This default configuration options were written for Symfony 2.x, I think you are using Symfony 3.x. Defaults should not be changed as we will break current user configs.

I think that in this case documentation of this module should be changed to make a notice on using Symfony 3.x

PR #3772, opened 2016-11-26
Removed generated files from git

Naktibalda:


Naktibalda:
For some odd reason one actor is not regenerated in the sandbox.

DavertMik:
Ok, let me check this

DavertMik:
I have a fix for that. 
Probably Action classes are not rebuilt because sandbox is never execute for all suites but per specific tests only... Or maybe this happens because we are copying folders for each run. Anyway

DavertMik:
I created another PR with the fix.

I think we should use organization repos for PRs so it would be easier to improve our patches.

PR #3773, opened 2016-11-26
Test click link by title

Naktibalda:
Correct test for #3678 
and fix for InnerBrowser.

WebDriverTest will pass after #3678 is merged.


Naktibalda:
WebDriverTest should pass now, because I merged  #3678 and rebased this branch.

PR #3775, opened 2016-11-28
[Laravel5] Added vendor_dir option

AdrianSkierniewski:
Hi,

I've added this option to cover the case when I want to test Laravel package separately. To do that, I'm using `orchestra/testbench` to bootstrap simple Laravel app. In my example, I want to use my package vendor dir instead `orchestra/testbench/fixture/vendor`.

By the way, why package name is set to `codeception/base` instead `codeception/codeception` in `composer.json`? I had some problems with installing my forked version to test if everything works as expected :)


janhenkgerritsen:
Thanks for the PR!

As for you question about the package name, can you answer that one @DavertMik?

Issue #3716, opened 2016-11-10
"#" comments in .env params file not parsed correctly


ivokund:
`\Codeception\Configuration::prepareParams` uses `parse_ini_file()` to parse both `.ini` and `.env` files. 
It makes sense to parse `.ini` files with `parse_ini_file`, but should it not use something like https://github.com/vlucas/phpdotenv to parse `.env` files? 

My .env contains comments starting with `#` that are not parsed correctly by `parse_ini_file`. 

To reproduce, 

1. Add this to `codeception.yml`

```yaml
params:
    - ../.env
```

2. Add a comment containing `"` or `(` to your .env file, for example:
```bash
# MySQL access configuration (required)
```

3. Run `codecept run` and get
```
PHP Warning:  syntax error, unexpected '(' in <dir>/.env on line 1
 in <dir>/vendor/codeception/codeception/src/Codeception/Configuration.php on line 689
```

Codeception 2.2.5

DavertMik:
What can I say about it.... Oooops! Yes, the formats are different but parse_ini worked fine for .env too. 
Is there a specification to .env format? Is there a way to parse it better without installing additional packages?

sergeyklay:
* https://github.com/vlucas/phpdotenv
* https://github.com/josegonzalez/php-dotenv

DavertMik:
@sergeyklay 

>  Is there a way to parse it better **without installing additional packages**

DavertMik:
I think it is actually a bug/feature/change in PHP itself. See [the changelog for PHP 7.0](http://php.net/manual/en/function.parse-ini-file.php)

DavertMik:
For our puposes we can use just http://php.net/manual/en/function.parse-ini-string.php and strip comments.
@ivokund do you wish to send PR with such fix?

ivokund:
I guess it would be a bit like pretending to support .env files, because some features (like nested variables) would still not be supported. I see three other alternatives here (besides using an external library or incorporating code from that library into Codeception).

**First**, perhaps if would make more sense to allow a PHP callback as a value for `params`, in which custom logic (such as loading env variables using `vlucas/phpdotenv`) could be added. 

Something like

```yaml
params:
    - \app\tests\EnvLoader::loadEnvVariables
```

In my case, loadEnvVariables would look something like.
```php
$dotEnv = new Dotenv\Dotenv(APP_ROOT);
$dotEnv->load();
return $_ENV;
```

[...]

PR #3786, opened 2016-11-30
Refactored params loader. 

DavertMik:
Using vlucas/phpdotenv for dot.env params loader.
Refactored params loader out of Configuration class.
Fxes #3716

### DIFF
--- a/src/Codeception/Lib/ParamsLoader.php
+++ b/src/Codeception/Lib/ParamsLoader.php
@@ -1,0 +1,93 @@
+<?php
+namespace Codeception\Lib;
+
+use Codeception\Exception\ConfigurationException;
+use Symfony\Component\Yaml\Yaml;
+
+class ParamsLoader
+{
+    protected $paramStorage;
+    protected $paramsFile;
+
+    public function load($paramStorage)
+    {
+        $this->paramsFile = null;
+        $this->paramStorage = $paramStorage;
+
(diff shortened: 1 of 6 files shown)


DavertMik:
Anyway, I'm not happy with current implementation. It looks hardcoded for me :( 
Maybe it could be changed for 3.0

src/Codeception/Lib/ParamsLoader.php
```diff
@@ -0,0 +1,85 @@
...
+    {
+        if (!class_exists('Dotenv\Dotenv')) {
+            throw new ConfigurationException(
+                "`vlucas/phpdotenv` library is required to parse .env files.\n" .
+                "Please install it via composer: composer require vlucas/phpdotenv"
+            );
+        }
+        $dotEnv = new \Dotenv\Dotenv(codecept_root_dir(), $this->paramStorage);
+        $dotEnv->load();
+        return $_ENV;
```
EricTendian: Hey @DavertMik - sorry to revive this, but was wondering why `$_ENV` is being returned here as opposed to `$_SERVER` in `loadEnvironmentVars`. The reason I ask is because for the ini setting [`variables_order`](http://us.php.net/manual/en/ini.core.php#ini.variables-order) the recommended development value and production value is `GPCS`, so `$_ENV` will be empty. Happy to make a PR if needed.

PR #3790, opened 2016-11-30
Update GroupManager.php

ksiv:
1. Fixed comment to suite site documentation.
As of http://codeception.com/docs/07-AdvancedUsage#Group-Files
g_* pattern would not load any 'group_N' files

2. Fixed code so we now can not only load data groups, but also run these by their names.
Without this fix one with g1 and g9 files loaded by pattern would not be able to call -g g9 because internally g9 will be turned to g2 by iterator. 
Surprisingly enough we used old version without any issues for awhile. 

Thank you very much for the comment in the code. A rare treat.

### DIFF
--- a/src/Codeception/Lib/GroupManager.php
+++ b/src/Codeception/Lib/GroupManager.php
@@ -28,10 +28,15 @@ class GroupManager
      * proceeds group names with asterisk:
      *
      * ```
-     * "tests/_log/g_*" => [
-     *      "tests/_log/group_1",
-     *      "tests/_log/group_2",
-     *      "tests/_log/group_3",
+     * "tests/groups/g*" => [
+     *      "tests/groups/g1",
+     *      "tests/groups/g2",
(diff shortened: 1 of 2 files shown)


DavertMik:
> Thank you very much for the comment in the code. A rare treat.

Yep, I don't do that often, but when I do I try to add it with real examples )

ksiv:
tests/unit/Codeception/Lib/GroupManagerTest.php:testGroupsByDifferentPattern
this test is made out of assumption that adding iterator to group name is a correct behaviour. 
If it is - it might be added to documentation for everyone to know it. (and we will have to use internal patch) 
If adding iterator to a group prefix is incorrect behaviour - this test is incorrect also and might be deleted. 

p.s. we used groups for awhile for parallel execution and these worked fine just because we generated those as g1,g2,g3 it would not work  even if decided to start series from g0.

ksiv:
fixed documentation instead

PR #3803, opened 2016-12-03
Codeception\Step\Comment::__toString() must return a string

### LINKED ISSUE #3296: Gherkin Method Codeception\Step\Comment::__toString() must return a string value
I want to use the new Gherkin feature, but with my php5.6 and php7 an error occured.

`[PHPUnit_Framework_Exception] Method Codeception\Step\Comment::__toString() must return a string value`

I analyzed it and the function runStep in src/Codeception/Test/Gherkin.php called

`$this->getScenario()->comment(null);`

Then throws the __toString in src/Codeception/Step/Comment.php the exception. 

I fixed it with following code

`return (string)$this->getAction();`

DavertMik:
Method Codeception\Step\Comment::__toString() must return a string
Fixes #3296

### DIFF
--- a/src/Codeception/Step/Comment.php
+++ b/src/Codeception/Step/Comment.php
@@ -8,17 +8,17 @@ class Comment extends CodeceptionStep
 {
     public function __toString()
     {
-        return $this->getAction();
+        return (string)$this->getAction();
     }
 
     public function toString($maxLength)
     {
-        return $this->getAction();
+        return (string)$this->getAction();
     }
 
     public function getHumanizedAction()
     {
-        return $this->getAction();
+        return (string)$this->getAction();
     }
 
     public function getHtml($highlightColor = '#732E81')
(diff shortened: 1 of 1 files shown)


src/Codeception/Step/Comment.php
```diff
@@ -8,17 +8,17 @@ class Comment extends CodeceptionStep
...
     public function __toString()
     {
-        return $this->getAction();
+        return (string)$this->getAction();
     }
 
     public function toString($maxLength)
     {
-        return $this->getAction();
+        return (string)$this->getAction();
```
edno: Why having a mandatory parameter $maxLength that is not use? I would be better to have toString() and getHumanizedAction() calling $this->__toString() that is doing the type casting.

DavertMik: thanks for noticing

PR #3809, opened 2016-12-05
Trying to move HHVM builds to Wercker CI

DavertMik:
Continuing @Naktibalda's work on making Codeception to be HHVM tested.
I think HHVM setup should be moved aside of Travis because
* it bloats our .travis.yml and 
* makes build to be run 2 times slower (20 minutes vs 10). 
* makes Travis builds unstable, we will need to check every time if that was not HHVM that caused Travis fail

So as I was telling earlier I thought on taking another CI and use it for our purposes. Unfortunately there is no HHVM-specific CIs but we can use Docker containers. Wercker looks almost good for that purpose. This PR enabled basic CLI and core unit tests, but shouldn't be hard to run HHVM web server and run PhpBrowser tests as well. For running server I'm using [HHVM-Proxygen](https://github.com/hhvm/hhvm-docker/tree/master/hhvm-latest-proxygen)

The only limitation I see now is that Wercker uses outdated `--links` Docker protocol to connect services which is not nice and easy to configure. I don't know is there a nice way to run Mysql/Redis tests inside it, but we can hope for good. 

Let this PR stay here as an alternative for Travis setup, if we see it covers our needs we can replace Travis HHVM setup with wercker.

Builds should be seen at https://app.wercker.com/davert/Codeception


DavertMik:
CLI + Unit tests successfully passed https://app.wercker.com/davert/Codeception/runs/build/5844e59e00937d01006694c0?step=5844e5b79b175d00015c45aa

DavertMik:
Added PostgreSQL, MySQL, Redis to tests. 
Duration: 16 minutes, 45 seconds.

I those tests are slow because of file operations happening in cli tests. Probably it's somewhere in copy or delete dir. We use PHP methods for copying/deleting dirs and probably if we would switch to linux commands we would improve the speed dramatically.

The only thing what is missing in current wercker setup is framework tests, but I don't see them to be really important, especially as not all the frameworks declare hhvm support.

PR #3816, opened 2016-12-05
Added seeNumRecords to Laravel5 module

dmoreno:
This is equivalent to seeNumRecords in DB module

### DIFF
--- a/src/Codeception/Module/Laravel5.php
+++ b/src/Codeception/Module/Laravel5.php
@@ -1061,7 +1061,7 @@ class Laravel5 extends Framework implements ActiveRecord, PartedModule
      */
     public function grabNumRecords($table, $attributes = [])
     {
-        return (class_exists($table))? $this->countModels($table, $attributes) : $this->countRecords($table, $attributes);
+        return class_exists($table)? $this->countModels($table, $attributes) : $this->countRecords($table, $attributes);
     }
 
     /**
--- a/src/Codeception/Module/Laravel5.php
+++ b/src/Codeception/Module/Laravel5.php
@@ -1033,14 +1033,35 @@ class Laravel5 extends Framework implements ActiveRecord, PartedModule
         if (class_exists($table)) {
             $currentNum = $this->countModels($table, $attributes);
             if ($currentNum != $expectedNum) {
-                $this->fail("The number of found $table ($currentNum) does not match expected number $expectedNum in with " . json_encode($attributes));
+                $this->fail("The number of found $table ($currentNum) does not match expected number $expectedNum with " . json_encode($attributes));
             }
         } else {
             $currentNum = $this->countRecords($table, $attributes);
             if ($currentNum != $expectedNum) {
-                $this->fail("The number of found records ($currentNum) do not match expected number $expectedNum in table $table with " . json_encode($attributes));
+                $this->fail("The number of found records ($currentNum) does not match expected number $expectedNum in table $table with " . json_encode($attributes));
             }
         }
+    }
+
+    /**
+     * Retrieves number of records from database
+     * You can pass the name of a database table or the class name of an Eloquent model as the first argument.
+     *
+     * ``` php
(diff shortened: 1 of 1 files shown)


src/Codeception/Module/Laravel5.php
```diff
@@ -1050,6 +1075,62 @@ protected function findRecord($table, $attributes = [])
...
+    }
+
+    /**
+     * @param string $table
+     *
+     * @return EloquentModel
+     */
+    protected function getQueryBuilderFromTable($table)
+    {
+        return $query = $this->app['db']->table($table);
```
janhenkgerritsen: The `$query` variable is unnecessary here

src/Codeception/Module/Laravel5.php
```diff
@@ -1013,20 +1013,45 @@ public function grabRecord($table, $attributes = [])
...
+     * @param string $table
+     * @param array $attributes
+     * @part orm
+     */
+    public function seeNumRecords($expectedNum, $table, $attributes = [])
+    {
+        if (class_exists($table)) {
+            $currentNum = $this->countModels($table, $attributes);
+            if ($currentNum != $expectedNum) {
+                $this->fail("The number of found $table ($currentNum) does not match expected number $expectedNum in with " . json_encode($attributes));
```
janhenkgerritsen: Missing text between "in" and "with" in error message.

src/Codeception/Module/Laravel5.php
```diff
@@ -1013,20 +1013,45 @@ public function grabRecord($table, $attributes = [])
...
+            }
+        } else {
+            $currentNum = $this->countRecords($table, $attributes);
+            if ($currentNum != $expectedNum) {
+                $this->fail("The number of found records ($currentNum) do not match expected number $expectedNum in table $table with " . json_encode($attributes));
```
janhenkgerritsen: Change "do not match" to "does not match" in error message.

src/Codeception/Module/Laravel5.php
```diff
@@ -1050,6 +1075,62 @@ protected function findRecord($table, $attributes = [])
     }
 
     /**
+     * @param string $modelClass
+     * @param array $attributes
+     * @return array
```
janhenkgerritsen: Return type should be integer.

src/Codeception/Module/Laravel5.php
```diff
@@ -1050,6 +1075,62 @@ protected function findRecord($table, $attributes = [])
...
+            $query->where($key, $value);
+        }
+
+        return $query->count();
+    }
+
+    /**
+     * @param string $table
+     * @param array $attributes
+     * @return array
```
janhenkgerritsen: Return type should be integer.

dmoreno:
@janhenkgerritsen, thanks for your code review. `grabNumRecords` method is a good idea. I have applied all your proposed changes.

src/Codeception/Module/Laravel5.php
```diff
@@ -1033,17 +1033,38 @@ public function seeNumRecords($expectedNum, $table, $attributes = [])
...
+     * ```
+     *
+     * @param string $table
+     * @param array $attributes
+     * @return integer
+     * @part orm
+     */
+    public function grabNumRecords($table, $attributes = [])
+    {
+        return (class_exists($table))? $this->countModels($table, $attributes) : $this->countRecords($table, $attributes);
```
janhenkgerritsen: Can you remove the extra parentheses around the `class_exists` call?

dmoreno:
@janhenkgerritsen, done!

janhenkgerritsen:
Thanks for adding the `grabNumRecords` method and applying the changes. I see you already pushed the change for the parentheses :)

The previous build failed, but this was not related to your changes.

janhenkgerritsen:
I added the changes from this PR to the changelog in #3822. Thanks for your work!

PR #3825, opened 2016-12-08
Fix bug #3824 by removing code that added extra option.

### LINKED ISSUE #3824: Innerbrowser submitForm uses wrong value for select.
### Problem:
`getFormFromCrawler() ` in the InnerBrowser adds an empty option to select elements that have no empty options.

This results in `submitForm()` submitting this empty value in stead of the default.

Testform:
```html
<form method="post">
    <select name="test">
        <option value="_none">- None -</option>
    </select>
</form>
```

There is no option without a value and no default selected option so the `getFormFromCrawler()` method adds one:
```html
    <select name="test">
        <option value=""></option>
        <option value="_none">- None -</option>
    </select>
```

The `submitForm()` method now submits `test=""` in stead of the correct `test="_none"`

### Solution:
1. Find out why extra `<option>` adding code was added in commit c048980e318850c154d789d5ce5d69952a6cef60
2. Create another solution for whatever problem the extra `<option>` fixed.
3. Remove the extra `<option>` adding code.

### Details:
* Codeception version: 2.2.6

JorisVanEijden:
I don't know what the extra `<option>` was for so I expect this PR to maybe fail a test somewhere.

Other than that it adds a test for the bug and removes the code causing the bug.

Fixes #3824

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -829,12 +829,6 @@ class InnerBrowser extends Module implements Web, PageSourceSaver, ElementLocato
         foreach ($shouldDisable as $field) {
             $field->parentNode->removeChild($field);
         }
-        $selectNonMulti = $cloned->filterXPath('//select[not(@multiple) and not(option[@value=""])]');
-        $opt = new \DOMElement('option');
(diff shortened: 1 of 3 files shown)


DavertMik:
Thank you!

PR #3836, opened 2016-12-15
use PDO attribute to control resultset array index casing

ashnazg:


ashnazg:
Looks like driver classes might be inconsistent with casing expectations.  Will replace this PR with one that only targets Oci, which is where I had my issue.

PR #3838, opened 2016-12-15
use PDO attribute to enforce lowercase resultset array indices

ashnazg:


ashnazg:
If setting this on the PDO connection itself seems too bold, then perhaps a more targeted change for the issue I encountered is `s/column_name/COLUMN_NAME/` on line 91 in `getPrimaryKey()`.

ashnazg:
Since other interactions with Oci insists on uppercase table and column names, any use of lowercase would be inconsistent.  See fix on PR #3840 instead of this one.

PR #3839, opened 2016-12-15
use user_* views rather than all_* views

ashnazg:
Querying the `all_*` views can result in false matches.  In our context, we're only interested in what the `user_*` views can return, since it is "us" that created the objects in the first place (`dump.sql`).


Naktibalda:
I have a problem with "In our context".
Is your context generic enough to apply to all users of Codeception that are using Oracle?

Tables could be created by another user if `populate: false` is set in configuration of Db module.

ashnazg:
That's a fair point @Naktibalda... I have not used the Db module in a case where I was not populating the Db via the module.  In that context, I suppose there is not a valid way to avoid false positives on the primary key lookup.

PR #3844, opened 2016-12-16
📚 Fixed doc formatting issue

shashikreddy:


DavertMik:
Sorry, this formatting issue is deeper than it seems. It is in our documentation parser located in [Robofile](https://github.com/Codeception/Codeception/blob/2.2/RoboFile.php). Fixing it one time won't help as documentation is regenerated before each release :(

PR #3845, opened 2016-12-17
Autostart webservices when run the tests

Anton-Burtyka:
It's very inconvenient to run manually php built-in server before tests execution. These changes allow to start any web server before tests execution and kill it afterward.


Naktibalda:
Thanks, but you implemented a feature not for Codeception but for Codeception test suite.
It seems like an overkill.

Codeception already has a [PhpBuiltinServer](https://github.com/tiger-seo/PhpBuiltinServer) extention `to start and stop PHP built-in web server for your tests`, could you try to use it instead?

DavertMik:
Thanks for the PR!

The thing you are trying to implement with environment modules is already can be done using extensions. Probably that was not as useful before the latest release, but from 2.2.7 you can enable extensions per suite and per environment as well. That means that the same config can be implemented using currently available extension and suite config.

inside `unit.suite.yml`:

```yml
extensions:
    enabled:
        - Codeception\Extension\PhpBuiltinServer
    config:
        Codeception\Extension\PhpBuiltinServer:
            hostname: localhost
            port: 8000
            autostart: true
            documentRoot: tests/_data   
```

I'm not sure it is currently possible to start 2 extensions at a time, probably not.

I think it is a problem from our side that per suite extensions were only mentioned in release notes but were not described in documentation. That should be improved

PR #3852, opened 2016-12-19
add checkbox name attrbitute support

gimler:
Add WebDriver::checkOption name attribute support.


DavertMik:
Thanks!
We need to wait while CI build stabilizes and then we can merge your PR.

Issue #3862, opened 2016-12-21
@depends doesnt work with @dataprovider


Fuin:
Hello

I have an @depends as well as an @dataprovider annotation on one of my tests. 

/**
     * @param \Step\Acceptance\Login $I
     *
     * @depends ProfileManagement\RegistrationCest:registerWithRequiredFields
     * @dataprovider userProvider
     */
    public function editPassword(\Step\Acceptance\Login $I, Example $example)
    {
        $I->amOnPage(HomePage::$URL);
        $I->login();
        $I->waitForElementVisible(ProfilePage::$profileChangePasswordFirstField);
        $I->fillField(ProfilePage::$profileChangePasswordFirstField, $example['password']);
        $I->fillField(ProfilePage::$profileChangePasswordSecondField, $example['password']);
        $I->click(ProfilePage::$profileSaveNewPasswordButton);
        $I->waitForElementVisible(ProfilePage::$profileMessageContainer);
        $I->canSee('Passwort würde erfolgreich geändert', ProfilePage::$profileMessageContainer);
    }

It works well for that one test however all following tests which have the @depends annotation are skipped with the message

> Dependent test for ProfileManagement\UpdateProfileCest:editPassword not found

If I add the @dataprovider annotation to all tests with @depends its working

 Codeception version: 2.2.7
 PHP Version: 5.7.3
 Operating System: Windows 7
 Installation type: Composer

edno:
Confirmed, however, in my case only the last test is skipped even if it does not have a `depends` annotation.

edno:
@Fuin can you try in your project the fix proposed.
The fix is located in *vendor/Codeception/[src/Codeception/SuiteManager.php](https://github.com/Codeception/Codeception/pull/3863/files#diff-a55be34a09d77e19e04792511523d214)*

Fuin:
Thanks, I've added the dataprovider to the rest of the project and all tests are working fine. 
Is this going to be added to the 2.2.8 Release and when is it?

edno:
If fine with @DavertMik then yes it will be available in 2.2.8.

Naktibalda:
The fix was merged.

PR #3863, opened 2016-12-21
Fix "@depends doesnt work with @dataprovider" #3862

edno:
Fix for the issue "**@depends doesnt work with @dataprovider**" #3862


edno:
@DavertMik before merging this PR (if you think it can be merged), I suggest to wait for @Fuin's feedback on #3862

DavertMik:
Thanks! Let's wait for it

edno:
Positive feedback from @Fuin, so the PR is ready to go 🚀

DavertMik:
Ok, well done. Thank you!

PR #3868, opened 2016-12-22
fix printScenarioFail with multiple scenarios

gimler:
When you have run a feature file and mulitple scenarios fails.
Then in the exception message the step is always the last one.

I am not sure what `conditionalFails` means. See change.

### DIFF
--- a/tests/cli/RunCest.php
+++ b/tests/cli/RunCest.php
@@ -355,11 +355,18 @@ EOF
         $I->seeInShellOutput('PASSED');
     }
 
-    public function runIncompleteGherkinTest(CliGuy $I)
+    public function reportsCorrectFailedStep(CliGuy $I)
     {
         $I->executeCommand('run scenario File.feature -v');
         $I->seeInShellOutput('OK, but incomplete');
         $I->seeInShellOutput('Step definition for `I have only idea of what\'s going on here` not found in contexts');
+    }
+
+    public function runFailingGherkinTest(CliGuy $I)
+    {
+        $I->executeCommand('run scenario Fail.feature -v --no-exit');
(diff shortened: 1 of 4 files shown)


DavertMik:
I'm not sure for the fix as it goes to Console which is the very general class and probably may cause side effects but I will try to reproduce the bug and fix it.

gimler:
I think i can provide a testcase for it

DavertMik:
> I think i can provide a testcase for it

Looking forward to see it. Thanks!

Naktibalda:
@gimler you promised to implement a test for your fix.
Hint: it should go to cli suite, please make something similar to https://github.com/Codeception/Codeception/blob/2.2/tests/cli/RunSkippedCept.php

gimler:
@DavertMik @Naktibalda ready for review

tests/cli/RunCest.php
```diff
@@ -354,6 +354,13 @@ public function runIncompleteGherkinTest(CliGuy $I)
         $I->seeInShellOutput('Step definition for `I have only idea of what\'s going on here` not found in contexts');
     }
 
+    public function runFailingGherkinTest(CliGuy $I)
```
Naktibalda: This test should be named `reportsCorrectFailedStep` to explain what it is testing.

robsch:
Detected the same issue in another situation. I Hope the fix gets applied soon.

tests/cli/RunCest.php
gimler: done

Naktibalda:
This change has broken the output of failed step: http://phptest.club/t/seeelement-wierd-fail-message/1470/3

PR #3872, opened 2016-12-26
fix Codeception-3866 grabValueFrom doesn't work

wumouse:
fix Codeception-3866 grabValueFrom doesn't work


wumouse:
this fix is not compatible with fill textarea

wumouse:
it's not compatible for \TestsForWeb::testFillFieldSquareBracketNames

PR #3873, opened 2016-12-26
fix #3866 grabValueFrom doesn't work with fillField

### LINKED ISSUE #3866: grabValueFrom doesn't work with fillField
>test code is:
```php
$I->fillField('username', 'admin');
codecept_debug($I->grabValueFrom('[name=username]'));
$I->click('Login');
```
> run with ``--debug`` option, will get output like:
```bash
   I grab value from "[name=username]"
  
   I click "Login"
```
> no value ouputed here. notice that the empty line is my expected output location of value.

> I found the issue in last line of method `CodeCeption\Lib\InnerBrowser::fillField`:
```php
$input->getNode(0)->nodeValue = htmlspecialchars($value);
```
> the point is input field must be have no nodeValue, and `grabValueFrom` use the right way to get value. I think the right code should be like:
```php
$input->getNode(0)->setAttribute('value', htmlspecialchars($value));
``` 
> I tried to change to this, it works

> I think a test for `CodeCeption\Lib\InnerBrowser` will be good.

wumouse:
fix issue 3866

### DIFF
--- a/tests/unit/Codeception/Module/TestsForWeb.php
+++ b/tests/unit/Codeception/Module/TestsForWeb.php
@@ -693,6 +693,23 @@ abstract class TestsForWeb extends \Codeception\TestCase\Test
         $this->assertEquals('oldfag', $result);
     }
 
+    /**
+     * @see https://github.com/Codeception/Codeception/issues/3866
+     */
+    public function testGrabValueFromWithFillField()
+    {
+        $this->module->amOnPage('/form/bug3866');
+        $this->module->fillField('empty', 'new value');
+        $result = $this->module->grabValueFrom('#empty');
+        $this->assertEquals('new value', $result);
+        $this->module->fillField('empty_textarea', 'new value');
+        $result = $this->module->grabValueFrom('#empty_textarea');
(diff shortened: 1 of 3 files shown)


wumouse:
#3866

tests/unit/Codeception/Module/TestsForWeb.php
```diff
@@ -691,6 +691,17 @@ public function testGrabValueFrom()
         $this->module->amOnPage('/form/select');
         $result = $this->module->grabValueFrom('#age');
         $this->assertEquals('oldfag', $result);
+
+        $this->module->amOnPage('/form/bug3866');
```
DavertMik: big thanks for this PR, and especially for test, but could you move this to separate test so it would be easier to find it? Thanks

wumouse: 👌

wumouse:
@DavertMik Please review the new commit, thank you

wumouse:
@Naktibalda Oh, yes! it fixed `seeInField` too.

PR #3874, opened 2016-12-26
Wedriver pageload timeout

oprudkyi:
adds support of pageload_timeout config to WebDriver

the amount of time to wait for a page load to complete before throwing an error
calls facebook/webdriver/lib/WebDriverTimeouts.php::pageLoadTimeout($seconds)

This patch allows to reduce issues from phantomjs random freezing, especially within CI environment 
https://github.com/ariya/phantomjs/issues/11526

patch based on suggestion from  https://github.com/ariya/phantomjs/issues/11526#issuecomment-115135376

### DIFF
--- a/docs/modules/WebDriver.md
+++ b/docs/modules/WebDriver.md
@@ -114,6 +114,7 @@ you should use a tunnel application provided by a service.
 * `capabilities` - Sets Selenium2 [desired capabilities](https://github.com/SeleniumHQ/selenium/wiki/DesiredCapabilities). Should be a key-value array.
 * `connection_timeout` - timeout for opening a connection to remote selenium server (30 seconds by default).
 * `request_timeout` - timeout for a request to return something from remote selenium server (30 seconds by default).
+* `pageload_timeout` - amount of time to wait for a page load to complete before throwing an error (default 0 seconds).
 * `http_proxy` - sets http proxy server url for testing a remote server.
 * `http_proxy_port` - sets http proxy server port
 * `debug_log_entries` - how many selenium entries to print with `debugWebDriverLogs` or on fail (15 by default).
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -1194,7 +1194,7 @@ class WebDriver extends CodeceptionModule implements
             );
             $this->sessions[] = $this->_backupSession();
             $this->webDriver->manage()->timeouts()->implicitlyWait($this->config['wait']);
-            if(!is_null($this->config['pageload_timeout'])) {
+            if (!is_null($this->config['pageload_timeout'])) {
                 $this->webDriver->manage()->timeouts()->pageLoadTimeout($this->config['pageload_timeout']);
             }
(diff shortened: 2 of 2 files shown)


docs/modules/WebDriver.md
```diff
@@ -114,6 +114,7 @@ you should use a tunnel application provided by a service.
 * `capabilities` - Sets Selenium2 [desired capabilities](https://github.com/SeleniumHQ/selenium/wiki/DesiredCapabilities). Should be a key-value array.
 * `connection_timeout` - timeout for opening a connection to remote selenium server (30 seconds by default).
 * `request_timeout` - timeout for a request to return something from remote selenium server (30 seconds by default).
+* `pageload_timeout` - amount of time to wait for a page load to complete before throwing an error (default 0 seconds).
```
DavertMik: strange but in default is `null` in this case. A bit confusing.

oprudkyi: it's 0 (wait forever) for phantomjs (it seems like so) , though spec says 
https://w3c.github.io/webdriver/webdriver-spec.html#dfn-session-page-load-timeout
> Unless stated otherwise it is 300,000 milliseconds.

so it implemented in way if value doesn't set by user then used default from underling layer

DavertMik: ok, I got it. Thanks

DavertMik:
Thank you, it should be useful

loesalleman:
Since 2.2.8 my tests stopped running with the following error:
Undefined index: pageload_timeout

So I suppose there is no default for this setting? I added this config setting as pageload_timeout: 0, but the tests now fail, with a timeout exception: 
```
[Facebook\WebDriver\Exception\TimeOutException]
  timeout: cannot determine loading status
  from timeout: Timed out receiving message from renderer: -0.001
    (Session info: chrome=55.0.2883.87)
```

Only when I set pageload_timeout to some higher value (I used 5) my test can run again. 

So for me the problem is now solved but I do not think setting this config option should be forced this way.

Naktibalda:
@loesalleman Please run codeception with `-vvv` flag and provide a stack trace of undefined index error.

oprudkyi:
@loesalleman 

> Undefined index: pageload_timeout

according to source - https://github.com/Codeception/Codeception/blob/2.2.8/src/Codeception/Module/WebDriver.php if you have not set pageload_timeout then it's ignored (your tests may call WebDriver directly though, so you probably should grep over your 'tests' directory to be sure if some helper doesn't help here)

>I added this config setting as pageload_timeout: 0
>Timed out receiving message from renderer: -0.001
>(Session info: chrome=55.0.2883.87)

hm, it seems like chrome sees 0 as no timeout at all 

>Only when I set pageload_timeout to some higher value (I used 5) my test can run again.

you probably may use 30 to avoid possible interference of slow running remote test servers

loesalleman:
@oprudkyi , Thanks! I must have gotten the error because I do an extend on the Webdriver class and do some stuff to be able to set other custom config settings. I should have thought of this in the first place and make a better solution :).

PR #3882, opened 2016-12-29
[Filesystem] dontSeeFileFound searches in path

### LINKED ISSUE #3877: dontSeeFileFound method doesn't work
I try to test if file exists in filesystem. I include Filesystem module for that. When I use **seeFileFound** method everything looks fine, but if I use **dontSeeFileFound** it always return success.

#### What do you get instead?

I get success as an output when I check for same file when I use **seeFileFound** and **dontSeeFileFound** methods, but first one is good second one is not.

### Details

* Codeception version: 
* PHP Version: 7.0
* Operating System: Ubuntu 16.04 LTS
* Installation type: Phar || Composer
* Framework: Laravel

```yml
# paste suite config here
```
class_name: FunctionalTester
modules:
    enabled:
        # add framework module here
        - \Helper\Functional
        - Asserts
        - Db
        - Cli
        - Filesystem

        - Laravel5:

Naktibalda:
Fixes  #3877

### DIFF
--- a/src/Codeception/Module/Filesystem.php
+++ b/src/Codeception/Module/Filesystem.php
@@ -251,7 +251,6 @@ class Filesystem extends CodeceptionModule
         }
 
         $path = $this->absolutizePath($path);
-        $this->debug($path);
         if (!file_exists($path)) {
             $this->fail("Directory does not exist: $path");
         }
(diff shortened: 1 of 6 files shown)


DavertMik:
Thank you!

PR #3890, opened 2017-01-03
 enable incomplete/skipped/risky/warning for logger 

mario-naether:
- enable "logIncompleteSkipped" for JUnit logger
- enable config "beStrictAboutTestsThatDoNotTestAnything" and "beStrictAboutOutputDuringTests"


edno:
I suggest to make `beStrictAboutTestsThatDoNotTestAnything` and `beStrictAboutOutputDuringTests ` configurable instead of hardcoding the value to **true**.
Also why not extending this feature to all parameters for [risky tests](https://phpunit.de/manual/current/en/risky-tests.html)?

Naktibalda:
I think that it should be configurable with current behavior as default.
It isn't nice to change rules in the patch release.

mario-naether:
I made the phpunit settings configurable in codeception settings and updated main yml file + documentation.

edno:
You should default the parameters to **false** in [Codecept.php](https://github.com/Codeception/Codeception/blob/86770e89d266557c20dd0b1de5390e706f4770c1/src/Codeception/Codecept.php#L39), otherwise that will generate errors will running tests if the parameters are not set in `codeception.yml`.

mario-naether:
ok, I added the default values as **false** in options array.

mario-naether:
ok, I fixed the last stuff and added an method-check for old phpunit version

codeception.yml
```diff
@@ -12,6 +12,10 @@ settings:
     memory_limit: 1024M
     log: true
     lint: true
+    reportUselessTests: false
```
DavertMik: by convention and by other settings we use underscore_keys in yaml config files

docs/reference/Configuration.md
```diff
@@ -63,6 +63,23 @@ settings:
...
+    # See https://phpunit.de/manual/current/en/risky-tests.html#risky-tests.useless-tests
+    reportUselessTests: false
+    
+    # PHPUnit can be strict about output during tests.
+    # See https://phpunit.de/manual/current/en/risky-tests.html#risky-tests.output-during-test-execution
+    disallowTestOutput: false
+    
+    # PHPUnit can be strict about tests that manipulate global state.
+    # See https://phpunit.de/manual/current/en/risky-tests.html#risky-tests.global-state-manipulation
+    beStrictAboutChangesToGlobalState: false
```
DavertMik: tooooooooooooooooooooo long

what does it mean to be strict?
`no_global_changes`, `avoid_global_changes`,... 
naming can be improved

src/Codeception/Codecept.php
```diff
@@ -37,29 +37,33 @@ class Codecept
...
+        'excludeGroups'                     => null,
+        'filter'                            => null,
+        'env'                               => null,
+        'fail-fast'                         => false,
+        'ansi'                              => true,
+        'verbosity'                         => 1,
+        'interactive'                       => true,
+        'no-rebuild'                        => false,
+        'quiet'                             => false,
+        'reportUselessTests'                => false,
```
DavertMik: this is not a configuration options but a runner options passed with parameters like `--fail-fast`, `--no-rebuild`. This is a config option so probably it should be set inside `Codeception\Configuration` class.

Yes, I know that might sound confusing :upside_down_face:

docs/reference/Configuration.md
mario-naether: this is a copy of the officially phpunit documentation

codeception.yml
mario-naether: ok, i changed the names settings to underscore.

src/Codeception/Codecept.php
mario-naether: yes, i understand this and moved to Configuration.php

DavertMik:
Yes, looks good. Thank you!

Naktibalda:
@mario-naether @edno Would you miss `disallow_test_output`, `log_incomplete_skipped` and `report_useless_tests` settings if they were missing in Codeception 5?

Naktibalda:
I reimplemented `report_useless_tests` and made it better than before: https://github.com/Codeception/Codeception/pull/6339

edno:
> @mario-naether @edno Would you miss `disallow_test_output`, `log_incomplete_skipped` and `report_useless_tests` settings if they were missing in Codeception 5? [26b54ae](https://github.com/Codeception/Codeception/commit/26b54aec8c7cd652304e8d02147c9f7137dc2e0f)

I can work around if need be.

PR #3893, opened 2017-01-03
avoid Notice

ashnazg:
`PHP Notice:  Undefined offset: 4`

### DIFF
--- a/src/Codeception/Subscriber/Console.php
+++ b/src/Codeception/Subscriber/Console.php
@@ -536,7 +536,9 @@ class Console implements EventSubscriberInterface
             }
         } elseif ($this->isWin() && (php_sapi_name() === "cli")) {
             exec('mode con', $output);
-            preg_match('/^ +.* +(\d+)$/', $output[4], $matches);
+            if (isset($output[4])) {
+                preg_match('/^ +.* +(\d+)$/', $output[4], $matches);
+            }
             if (!empty($matches[1])) {
                 $this->width = (int) $matches[1];
             }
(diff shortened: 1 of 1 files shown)


src/Codeception/Subscriber/Console.php
```diff
@@ -536,7 +536,9 @@ protected function detectWidth()
             }
         } elseif ($this->isWin() && (php_sapi_name() === "cli")) {
             exec('mode con', $output);
-            preg_match('/^ +.* +(\d+)$/', $output[4], $matches);
+            if (isset($output[4])) {
```
samdark: Looks risky. @DavertMik needs your attention.

ashnazg: The `PHP Notice` occurs in the `preg_match` line when `$output[4]` is _not_ set.  The patch just makes sure that it _is_ set before trying to use it.  It does solve the `undefined offset` notice that I was seeing on my Windows box.

src/Codeception/Subscriber/Console.php
```diff
@@ -536,7 +536,9 @@ protected function detectWidth()
...
+                preg_match('/^ +.* +(\d+)$/', $output[4], $matches);
+            }
             if (!empty($matches[1])) {
```
edno: The condition `!empty($matches[1])` should nested in `isset($output[4])` since `$matches` can only exist if `preg_match` statement has been executed.

ashnazg: I thought about that too, but did not see a notice at runtime.  Turns out that `empty()` avoids it if the variable doesn't exist:  `empty() does not generate a warning if the variable does not exist.` (http://php.net/manual/en/function.empty.php)

ashnazg: Actually, that makes sense logically, without regard to the notice or not for `$matches`.  I'll update it.

ashnazg:
bump... any feedback on this one?  I have something on hold at work that's waiting for this one.

Issue #3607, opened 2016-10-13
Graceful termination not works in PHP 7.0


DavertMik:
It looks tests can't be interrupted by Ctrl-C when running in PHP 7
Accrording to https://bugs.php.net/bug.php?id=71448 there is no way to make pcntl signal handling for PHP 7.0 (correct me if I'm wrong) and that's not a good thing when tests can't be stopped.

I propose to disable Graceful termination for PHP 7.0.x

PR #3907, opened 2017-01-11
PHP 7.1 - Used new pcntl_async_signals function to handle graceful termination

AdrianSkierniewski:
Hi,

This PR is related to #3607 and it should enable graceful termination for PHP >= 7.1
PHP 7.1 introduced the new way to handle signals asynchronously [pcntl_async_signals](http://php.net/manual/en/migration71.new-features.php#migration71.new-features.asynchronous-signal-handling)


DavertMik:
Thanks!

PR #3912, opened 2017-01-13
[PhpBrowser] Ensure sessions have independent cookies, fixes #3911

### LINKED ISSUE #3911: Friends don't have their own cookies in PhpBrowser

Two users logged into a web site and one performs an action which affects the other. In very pseudo code, something like this:

```
$fred=$I->haveFriend();
$I->login("grace");
$fred->does(function(AcceptanceTester $I){
    $I->login("fred");
    $I->doSomething();
});
$I->seeTheConsequences();
```

#### What do you get instead?

The problem is that friends don't seem to have independent cookies. When the code gets to `$I->seeTheConsequences()` the main tester ($I) has hijacked Fred's session because Fred's call to `$I->login("fred")` has replaced the cookie from `$I->login("grace")`.

### Details

* Codeception version: 2.2.7
* PHP Version: 5.6
* Operating System: Mac OS X
* Installation type: Phar
* List of installed packages: none via composer; don't believe it is relevant anyway
* Suite configuration:

```yml
class_name: AcceptanceTester
modules:
    enabled:
        - \Module\PhpBrowser:
            url: http://localhost:8989/WebInterface.php
```

insightfuls:


DavertMik:
Thanks for your work. Really impressive. I need some time to review it.

src/Codeception/Coverage/Subscriber/LocalServer.php
```diff
@@ -148,18 +149,52 @@ function ($h) {
...
+
+        // putting in configuration ensures the cookie is used for all sessions of a MultiSession test
+
+        $cookies = $this->module->_getConfig('cookies');
+        if (!$cookies|| !is_array($cookies)) {
+            $cookies = [];
+        }
+
+        $cookie = null;
+        foreach ($cookies as &$cookie) {
```
DavertMik: what does this loop actually do? According to what I see it only breaks and continues

src/Codeception/Coverage/Subscriber/LocalServer.php
```diff
@@ -148,18 +149,52 @@ function ($h) {
...
+        $this->module->_setConfig(['cookies' => $cookies]);
     }
 
     protected function fetchErrors()
     {
-        $error = $this->module->grabCookie(self::COVERAGE_COOKIE_ERROR);
+        try {
+            $error = $this->module->grabCookie(self::COVERAGE_COOKIE_ERROR);
+        } catch (ModuleException $e) {
+            // when a new session is started we can't get cookies because there is no
```
DavertMik: good catch!

src/Codeception/Coverage/Subscriber/LocalServer.php
```diff
@@ -148,18 +149,52 @@ function ($h) {
...
+            }
+        }
+        if (!$cookie || $cookie['Name'] !== self::COVERAGE_COOKIE) {
+            $cookie = &$cookies[];
+            $cookie['Name'] = self::COVERAGE_COOKIE;
+        }
+
+        $cookie['Value'] = $value;
+
+        $this->module->_setConfig(['cookies' => $cookies]);
```
DavertMik: I agree with this way but could the code above be simplified or documented? 
This things especially with `&` are pretty hard to understand.

src/Codeception/Coverage/Subscriber/LocalServer.php
insightfuls: It results in $cookie being set to a reference to the cookie we're interested in changing. You're probably right and I should rewrite it to something simpler.

insightfuls:
I think it's better now.

DavertMik:
Thank you! This was really brave to dig into deepest darkness of Codeception and fix the bug there :+1:

PR #3913, opened 2017-01-14
[Travis] Run tests on PHP 7.1

Naktibalda:


Naktibalda:
Phalcon is not compatible with PHP 7.1 yet
https://github.com/phalcon/cphalcon/issues/12055

.travis.yml
```diff
@@ -8,15 +8,15 @@ php:
 
 matrix:
   include:
+    - php: hhvm
```
DavertMik: +1 more reason to move HHVM to separate CI. 
I will try to work on Wercker to support all current test cases and then we could try to move it out. 
I think moving that away would improve the overall stability of Travis builds

DavertMik:
Please look into my PR: https://github.com/Codeception/Codeception/pull/3809
I didn't reinvent the wheel, I moved most of your commands for HHVM to wercker.yml and made them running there. Also I granted you admin rights on wercker ci.

.travis.yml
DavertMik: HHVM took 35 minutes to run the tests 0.o
I assume it's not a travis issue, they are slow on wercker too. Probably the slowdown happens for filesystem operations, cli tests run painfully slow (tried them locally and on wercker too)

PR #3914, opened 2017-01-15
[2.2] Support php files in ParamsLoader.php

arrilot:
My use case: In a project we use simple `.env.php` file instead of `vlucas/phpdotenv` and want to load params (specifically APP_URL) from it instead of duplicating configs.

It also fixes a bug when trying to load params from `.env.yml` and `.env.ini` results in an error "vlucas/phpdotenv library is required to parse .env files." which makes little sense to me.


DavertMik:
Thanks for this feature and bigfix. I will update documentation for adding PHP files as param storage

Issue #3957, opened 2017-01-25
Merge parameters instead of overwriting in Rest module.


albertleao:
First ever pull request so bare with me.

The $params class variable in the REST module is overwritten during the `execute()` function. By merging $parameters with $this->params instead of overwriting it completely, the user will have the ability to preset a parameter for all requests. For example beforehand they will be able to set CSRF tokens.

PR #3958, opened 2017-01-25
merged params in rest module

albertleao:
#3957

### DIFF
--- a/src/Codeception/Module/REST.php
+++ b/src/Codeception/Module/REST.php
@@ -468,6 +468,12 @@ EOF;
             $url = $this->config['url'] . $url;
         }
 
+        // allows for parameters to be preset
+        if (is_array($parameters)) {
+            $this->params = array_merge($parameters, $this->params);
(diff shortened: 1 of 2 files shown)


Naktibalda:
@albertleao Please merge changes from 2.2 branch (or rebase) to get the tests pass.

albertleao:
Great point. I will attempt another pull with a `setDefaultParameters`

Naktibalda:
Do you really need a function for that?
The same result could be achieved this way:
```php
$defaultParams = ['token' => '1234567'];
$I->sendPOST('/', ['foo' => 1] + $defaultParams);
$I->sendPOST('/', ['bar' => 2] + $defaultParams);
```

albertleao:
Yes I realize the same functionality can be achieved the way you described. But instead of having to to have  
` $defaultParams = ['token' => '1234567']; `
and the + for every test, I could set it up in my actor so that it can be defaulted for any type of test suite.

PR #3960, opened 2017-01-26
MultiDb module

iamdevice:
module for work with multiple databases.
based on original Db module

### DIFF
--- a/src/Codeception/Module/MultiDb.php
+++ b/src/Codeception/Module/MultiDb.php
@@ -1,0 +1,540 @@
+<?php
+
+namespace Codeception\Module;
+
+use Codeception\Configuration;
+use Codeception\Exception\ModuleConfigException;
+use Codeception\Exception\ModuleException;
+use Codeception\Lib\Driver\Db as Driver;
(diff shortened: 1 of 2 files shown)


src/Codeception/Module/MultiDb.php
```diff
@@ -0,0 +1,540 @@
...
+     *
+     * ```php
+     * <?php
+     * $I->amConnectedToDatabase('primary');
+     * ?>
+     * ```
+     *
+     * @param $database string
+     */
+    public function amConnectedToDatabase($database) {
```
iamdevice: Oh, my mistake

DavertMik:
Thanks for the PR. I think we discussed something similar with a team a few years ago but no one came with the solution. However, I don't think this should be the part of main package. MultiDb serves specific case so can I ask you to publish it as standalone package on packagist and list it on Addons page.

DavertMik:
As I said I'd rather not accept as it is not general purpose module, and I'd rather not to take respondibility to support this module inside our core. I'd like to encourage you to release this on packagist and add a link to Addons. Thanks

PR #3968, opened 2017-01-31
#3967: Yii2 Connector. Allow to run without db configured

unVooDoo:
#3967 Allows to run Yii2 tests without configuring 'db' component

PR #3973, opened 2017-01-31
Fixed #3961 (removed bad changes after #3834)

### LINKED ISSUE #3961: Yii2 model is lost when doing functional test and the action returns redirect response
Bug is caused by pull request #3834 in v.2.2.8. I'm testing an action which redirects to itself in certain condition. However before the redirect is done the transaction is roll backed and the model that I use for testing is lost.

iRipVanWinkle:
I think issue #3834 need reopen and will research

### DIFF
--- a/src/Codeception/Lib/Connector/Yii2.php
+++ b/src/Codeception/Lib/Connector/Yii2.php
@@ -58,9 +58,6 @@ class Yii2 extends Client
 
     public function resetApplication()
     {
-        if ($this->app->db !== null && $this->app->db->transaction !== null) {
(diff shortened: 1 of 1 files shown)


Naktibalda:
Have you seen #3968 ?

iRipVanWinkle:
@Naktibalda yes i have seen it. The #3968 will not fix #3961 and I think the need in the #3968  appeared after merged #3834

DavertMik:
Yes, thanks. I merged that one by mistake, sure transaction should be rolled back between tests only

PR #3975, opened 2017-01-31
Merge 2.2 to master

Naktibalda:


DavertMik:
:+1: 
Maybe this should be automated for release updates? What do you think?

Naktibalda:
@DavertMik Can you automate conflict resolution?
Currently there is a conflict in CHANGELOG which needs to be resolved every time, because you put an entry for 3.0.0 in master branch.

PR #3977, opened 2017-02-01
Fix test coloring

gimler:
When a Feature title has some special chars in it like `/` or `-` the test coloring is broken.
```
Feature: Tests-special chars

  Scenario: test 1
    Given I am on "/"
```


DavertMik:
Really good catch. Thank you!

PR #3978, opened 2017-02-01
added ability to use custom params from console in tests

yuriy-chabaniuk:
It is small test plan for this.

1. Run test with option -p or --params
```
php codecept run -p "site-id: 123456, group: admin" -p "duty: free"
```

In the test it should be available like following: 

```
.....
$I->amOnPage('/');
.....
$I->fillField('.site-id-input', \Codeception\Configuration::getParam('site-id'));
.....
```

Also I updated file tests/_support/Helper/Acceptance.php like so

```
namespace Helper;

class Acceptance extends \Codeception\Module
{

    public function getParam($key)
    {
        return Configuration::getParam($key);
    }
}
```

And I was able to use this: 

```
.....
$I->amOnPage('/');
.....
$I->fillField('.site-id-input', $I->getParam('site-id'));
.....
```

It probably is better way how to use params in the tests, and if you have ideas how to do it better let me know.

### DIFF
--- a/src/Codeception/Command/Run.php
+++ b/src/Codeception/Command/Run.php
@@ -114,6 +114,8 @@ class Run extends Command
             new InputArgument('suite', InputArgument::OPTIONAL, 'suite to be tested'),
             new InputArgument('test', InputArgument::OPTIONAL, 'test to be run'),
             new InputOption('override', 'o', InputOption::VALUE_IS_ARRAY  | InputOption::VALUE_REQUIRED, 'Override config values'),
+            new InputOption('params', 'p', InputOption::VALUE_IS_ARRAY  | InputOption::VALUE_REQUIRED, 'Add custom 
+            params in tests'),
             new InputOption('report', '', InputOption::VALUE_NONE, 'Show output in compact style'),
             new InputOption('html', '', InputOption::VALUE_OPTIONAL, 'Generate html with results', 'report.html'),
             new InputOption('xml', '', InputOption::VALUE_OPTIONAL, 'Generate JUnit XML Log', 'report.xml'),
@@ -214,6 +216,11 @@ class Run extends Command
         // update config from options
         if (count($this->options['override'])) {
             $config = $this->overrideConfig($this->options['override']);
+        }
+
+        if (count($this->options['params'])) {
+            $this->options['params'] = $this->parseParams($this->options['params']);
+            Configuration::attachParams($this->options['params']);
         }
 
         if (!$this->options['colors']) {
--- a/src/Codeception/Command/Shared/Config.php
+++ b/src/Codeception/Command/Shared/Config.php
@@ -24,13 +24,35 @@ trait Config
         return Configuration::suites();
     }
 
+    protected function parseParams($params)
+    {
+        $updatedParams = [];
+
+        foreach ($params as $match) {
+            $match = explode(',', str_replace(array(', '), ',', $match));
+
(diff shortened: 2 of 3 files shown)


DavertMik:
The implementation is pretty ok but we already have [Params concept](http://codeception.com/docs/06-ModulesAndHelpers#Dynamic-Configuration-With-Parameters) so it would be confusing to have them in different places.

I think what you trying to achieve can be implemented using [environment configuration](http://codeception.com/docs/07-AdvancedUsage#Environments) and a [custom helper](http://codeception.com/docs/06-ModulesAndHelpers#Helpers): 

```php
namespace Helper;

class Acceptance extends \Codeception\Module
{
   protected $config = ['site_id' => null];

   function enterSiteId($field)
   {
       $this->getModule('WebDriver')->fillField($field, $this->config['site_id']);
    }
}
```

this helper cna be configured in `acceptance.suite.yml`:

```yml
modules:
    - WebDriver:
        url: localhost
        browser: chrome
    - \Helper\Acceptance:
        site_id: 1 
```

and reconfigured with environments.

Probably this way is a bit complcated than your params, but to me it looks cleaner and more reliable, as it moves the params to configuration.

yuriy-chabaniuk:
Thank you @DavertMik for fast review. 

Params concept that you currently have looks pretty solid and flexible. Also environment concept sounds great for this case. But there is some difference why neither is suitable for me. 

I will explain why. 
We have platform to quick and delivery sites for real estate. 
Lets imagine that we have script like this to test important functionality: 
```
.....
$I->amOnPage('/');
.....
$I->fillField('.site-id-input','123456');
// Do something
```

Using this script we should test functionality on different sites (without updating test for specific site)

Then lets imagine that we have 50 new sites per day. And we need to test all of them. 

Using current 'Params concept' or environment configuration with helpers, as you described above,  require as to change (update) config file each time when we need to test new site. Am I right or I missed something. 

Using dynamic params directly from the console should:
1. give as ability to run tests without updating any config files.
2. run sites for different sites at once. (in multiple streams)
3.

DavertMik:
We have `-o` `--override` option which can dynamically update the config. http://codeception.com/docs/reference/Commands#Run

Just put your site_id to Helper's config as shown above and change them on run

vedro-compota:
this was cool PR, It's really problem to get user parameter from terminal in codeception now....

Naktibalda:
@vedro-compota Have you seen this?

https://codeception.com/docs/06-ModulesAndHelpers#Dynamic-Configuration-With-Parameters

vedro-compota:
@Naktibalda yes, but there I've only red about params from .env/config (as I understand), not from command line (terminal).
As result I started use overriding config params instead of using custom console params (e.g. http://fkn.ktu10.com/?q=node/11401)

Naktibalda:
It is very easy to set environment variables in command line:
```sh
URL=https://example.org ./vendor/bin/codecept run`
```

vedro-compota:
@Naktibalda thank you!

BobRay:
> URL=https://example.org codecept run acceptance

> 'URL' is not recognized as an internal or external command,
operable program or batch file.

This happens in the Cmder console emulator on Windows 10.

Is there any other way to   pass in an environment variable?

Naktibalda:
Try `set URL https://example.org/` on separate line.
Though you should get a better terminal, e.g. Git Bash, or install Linux on WSL.

BobRay:
> Environment variable URL "https://example.org/" not defined

PR #3983, opened 2017-02-02
Pass error message from wait*() methods to until()

disc:
Sometimes, when test failed because some element on page is missing I receive TimeOutException exception with empty message. Now I will pass error message to wait*() methods like $I->waitForText('foo', 10, 'Text foo was not found');

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -1951,15 +1951,16 @@ class WebDriver extends CodeceptionModule implements
      * @param $element
      * @param \Closure $callback
      * @param int $timeout seconds
+     * @param string $errorMessage
      * @throws \Codeception\Exception\ElementNotFound
      */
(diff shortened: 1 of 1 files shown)


thespecial:
nicely done!

DavertMik:
Sorry but this breaks the compatibility with other methods. There are no thing like custom error messages in Codeception modules. Probably it's not a good thing, however, this is way it is now. 

Inside our actions and assertions we try to generate error message based on action's name and its arguments. Inside `waitForElement` this can be printed as

```php
sprintf("waited for %d secs but element %s still not found",
  $secs,
  \Codeception\Util\Locator::humanReadableString($selector)
);
```

DavertMik:
@disc Thank you! Can you extend this to all wait* methods?

disc:
@DavertMik It's not necessary, because other methods throws NoSuchElementException inside until() method

DavertMik:
I see, ok, that makes sense.

DavertMik:
Thank you!

Issue #1884, opened 2015-04-29
Prevent output binary data


jumale:
- **_Known Linux issue:**_ [terminal is broken](http://unix.stackexchange.com/questions/79684/fix-terminal-after-displaying-a-binary-file) after binary data output.
- **_Codeception issue:**_ controllers may return binary response which causes the problem in "verbose" or "debug" mode.
- **_How to reproduce:**_
  - create a controller which returns binary data, for example ZIP archive
  - create a functional test using REST module and send there a GET request to  the controller
  - run the test with -vvv or --debug
- **_Solution:**_ if it's possible, it would be nice to check that response contains binary data, and do not output it [here](https://github.com/Codeception/Codeception/blob/2.0/src/Codeception/Module/REST.php#L396) (maybe show something like string `[Binary Data]` instead).

P.S.: I got this issue in REST module, but probably it can happen in some another modules

DavertMik:
How to do you propose to detect binary data?

jumale:
@DavertMik ok, I've discovered internet and looks like there is no easy solution to define the binary data.
But I found a way how it will be enough for me. All is needed from codeception - allow to define a callback to intercept debug messages. Like that:
```php
<?php
namespace Codeception\Util;

class Debug
{
    /**
     * @var callable
     */
    public static $interceptDebug = null;

    public static function debug($message)
    {
        if (!self::$output) {
            return;
        }
        // modified part
        if (is_callable(static::$interceptDebug)) {
            $message = call_user_func(static::$interceptDebug, $message);
        }
        self::$output->debug($message);
    }

    // ...
}
```

then I can create a helper with the next method:
```php
<?php
public function expectBinaryResponse()
{
    \Codeception\Util\Debug::$interceptDebug = function ($message) {
        if (preg_match('#\[Response\]#', $message)) {
            $message = '[Response] [Binary Data]';
        }
    
        return $message;
    };
}
```

and then in my test:
```php
$I = new Tester($scenario);
$I->expectBinaryResponse();
// ...
```

I know it looks a bit ugly, but I have no other idea how to hack it :)

DavertMik:
Thanks, I will try to think on this too

codedokode:
You could also replace unprintable characters 00-1f with codes like `\x00` so the data are still readable.

spikyjt:
The linked PR did not fix this issue, it only fixed an issue where the request data included binary data.

This issue is about the response having binary data.

[...]

PR #3985, opened 2017-02-02
hide binary response data in debug #1884

spikyjt:
As per https://github.com/Codeception/Codeception/issues/1884#issuecomment-276941773

### DIFF
--- a/src/Codeception/Module/REST.php
+++ b/src/Codeception/Module/REST.php
@@ -498,7 +498,13 @@ EOF;
             $this->debugSection("Request", "$method $url " . $requestData);
             $this->response = (string)$this->connectionModule->_request($method, $url, [], $files, [], $parameters);
         }
-        $this->debugSection("Response", $this->response);
+        $response = $this->response;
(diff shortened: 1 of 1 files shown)


src/Codeception/Module/REST.php
```diff
@@ -498,7 +498,13 @@ protected function execute($method, $url, $parameters = [], $files = [])
             $this->debugSection("Request", "$method $url " . $requestData);
             $this->response = (string)$this->connectionModule->_request($method, $url, [], $files, [], $parameters);
         }
-        $this->debugSection("Response", $this->response);
+        $response = $this->response;
```
DavertMik: could you name as `printedResponse` so it would be obvious that this is can't be used as real response

spikyjt: Yeah that makes sense.

I'll also rework it to move the formatting out to a couple of methods, so that the code is not repeated.

DavertMik:
Awesome, thank you!

PR #3988, opened 2017-02-03
Add methods to bind an (non-existent) exchange to a (non-existent) queue in AMPQ module

erikverheij:
The added methods allow users to bind an (non-existent) exchange to a (non-existent) queue in AMPQ module. This is useful when you want to test some pub/sub implementation with a producer that pushes messages to an exchange (and not directly to a queue).


Naktibalda:

DavertMik:
Thank you!

Issue #3990, opened 2017-02-03
PHP Fatal error:  Cannot redeclare class App (in _ide_helper.php)


timbroder:
#### What are you trying to achieve?

Run my first unit test

#### What do you get instead?

> Provide console output if related. Use `-vvv` mode for more details.

```bash
composer exec codecept run unit usa_invoice_report.feature -vvv
Reading ./composer.json
Loading config file /Users/timbroder/.composer/config.json
Loading config file /Users/timbroder/.composer/auth.json
Loading config file ./composer.json
Checked CA file /private/etc/ssl/cert.pem: valid
Executing command (/Users/timbroder/workspace/kidfund/web): git branch --no-color --no-abbrev -v
Reading /Users/timbroder/.composer/composer.json
Loading config file /Users/timbroder/.composer/config.json
Loading config file /Users/timbroder/.composer/auth.json
Loading config file /Users/timbroder/.composer/composer.json
Loading config file /Users/timbroder/.composer/auth.json
Reading /Users/timbroder/.composer/auth.json
Reading /Users/timbroder/workspace/kidfund/web/vendor/composer/installed.json
Reading /Users/timbroder/.composer/vendor/composer/installed.json
Running 1.3.2 (2017-01-27 18:23:41) with PHP 5.6.29 on Darwin / 16.3.0
> __exec_command: codecept 'run' 'unit' 'usa_invoice_report.feature'
Executing command (CWD): codecept 'run' 'unit' 'usa_invoice_report.feature'
Codeception PHP Testing Framework v2.2.8
Powered by PHPUnit 4.8.27 by Sebastian Bergmann and contributors.

Unit Tests (1) ---------------------------------------------------------------
- First Feature: First ScenarioPHP Fatal error:  Cannot redeclare class App in /Users/timbroder/workspace/kidfund/web/_ide_helper.php on line 1096
PHP Fatal error:  Cannot mix bracketed namespace declarations with unbracketed namespace declarations in /Users/timbroder/workspace/kidfund/web/vendor/symfony/debug/Exception/FatalErrorException.php on line 12
```
> Provide test source code if related

```php
public function firstScenario()
    {
        print("HI");
    }
```

At the top of my _ide_helper.php file, there is an exit that doesn't seem to process with codeception. This file should be ignored

```php
/**
 * A helper file for Laravel 5, to provide autocomplete information to your IDE
 * Generated for Laravel 5.2.45 on 2017-02-03.
 *
[...]

PR #3991, opened 2017-02-03
Update Laravel5 database_migrations_path default

timbroder:
To be inline with Laravel

Addresses #3990

### DIFF
--- a/src/Codeception/Module/Laravel5.php
+++ b/src/Codeception/Module/Laravel5.php
@@ -45,7 +45,7 @@ use Illuminate\Support\Collection;
  * * cleanup: `boolean`, default `true` - all database queries will be run in a transaction,
  *   which will be rolled back at the end of each test.
  * * run_database_migrations: `boolean`, default `false` - run database migrations before each test.
- * * database_migrations_path: `string`, default `` - path to the database migrations, relative to the root of the application.
+ * * database_migrations_path: `string`, default `database/migrations` - path to the database migrations, relative to the root of the application.
  * * run_database_seeder: `boolean`, default `false` - run database seeder before each test.
  * * database_seeder_class: `string`, default `` - database seeder class name.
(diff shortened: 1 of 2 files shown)


DavertMik:
Just to make sure, is that compatible with ALL Laravel 5s? 5.1, 5.2, 5.3, 5.4?

timbroder:
Confirmed:

https://github.com/laravel/laravel/tree/master/database/migrations
https://github.com/laravel/laravel/tree/5.0/database/migrations
https://github.com/laravel/laravel/tree/5.1/database/migrations
https://github.com/laravel/laravel/tree/5.2/database/migrations
https://github.com/laravel/laravel/tree/5.3/database/migrations

(there is no 5.4 branch yet)

src/Codeception/Module/Laravel5.php
```diff
@@ -114,7 +114,7 @@ public function __construct(ModuleContainer $container, $config = null)
             [
                 'cleanup' => true,
                 'run_database_migrations' => false,
-                'database_migrations_path' => '',
+                'database_migrations_path' => 'database/migrations',
```
edno: I would suggest to use Laravel approach (see example in [NotificationTableCommand](https://github.com/laravel/framework/blob/5.4/src/Illuminate/Notifications/Console/NotificationTableCommand.php#L77)) rather than a static value.
Example:
```php
$this->config['database_migrations_path'] = $this->app->databasePath().'/migrations';
```

DavertMik:
@edno thanks for review.
I asked @janhenkgerritsen to look into it as well. Unfortunately he will be able to do that only tomorrow, while today I plan to release next version. If anyone stumbles upon the same issue, they can change the `database_migrations_path` in config.

timbroder:
@DavertMik ok. I can update per @edno but will wait to see if @janhenkgerritsen has any changes and do it all at once

PR #3993, opened 2017-02-04
binary output checks for REST module

### LINKED ISSUE #3992: Add methods to test binary output for REST module.
Following on from #1884 I think it would be useful to be able to test that a binary response is as expected.

I am already using this in an extension to the module. The result is compared using a hash of the response data.

We could use base64 encoding, but this would be cumbersome with large files.

spikyjt:
closes  #3992

### DIFF
--- a/src/Codeception/Module/REST.php
+++ b/src/Codeception/Module/REST.php
@@ -1220,6 +1220,30 @@ EOF;
 
     /**
      * Checks if the hash of a binary response is exactly the same as provided.
+     * Parameter can be passed as any hash string supported by hash(), with an
+     * optional second parameter to specify the hash type, which defaults to md5.
+     *
+     * Examples:
+     *
+     * ```php
+     * <?php
+     * $I->seeBinaryResponseEquals("8c90748342f19b195b9c6b4eff742ded");
+     * ?>
+     * ```
+     *
+     * ```php
+     * <?php
+     * $fileData = file_get_contents("test_file.jpg");
+     * $I->seeBinaryResponseEquals(md5($fileData));
+     * ?>
+     * ```
+     *
+     * ```php
+     * <?php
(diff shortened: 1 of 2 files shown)


src/Codeception/Module/REST.php
```diff
@@ -1219,6 +1219,34 @@ public function dontSeeXmlResponseIncludes($xml)
     }
 
     /**
+     * Checks if the hash of a binary response is exactly the same as provided.
```
DavertMik: a code example would be very helpful here. You can take one from your tests and add here.

spikyjt: done (3 of them!)

src/Codeception/Module/REST.php
```diff
@@ -1219,6 +1219,58 @@ public function dontSeeXmlResponseIncludes($xml)
...
+     * <?php
+     * $fileData = file_get_contents("test_file.jpg");
+     * $I->seeBinaryResponseEquals(md5($fileData));
+     * ?>
+     * ```
+     *
+     * ```php
+     * <?php
+     * $fileData = '/9j/2wBDAAMCAgICAgMCAgIDAwMDBAYEBAQEBAgGBgUGCQgKCgkICQkKDA8MCgsOCwkJDRENDg8QEBEQCgwSExIQEw8QEBD/yQALCAABAAEBAREA/8wABgAQEAX/2gAIAQEAAD8A0s8g/9k='; // very small jpeg
+     * $I->seeBinaryResponseEquals(hash("sha256", base64_decode($fileData)));
```
DavertMik: yep, but use `sha256` but the method still uses `md5`!

src/Codeception/Module/REST.php
```diff
@@ -1219,6 +1219,58 @@ public function dontSeeXmlResponseIncludes($xml)
...
+     */
+    public function seeBinaryResponseEquals($hash, $algo = 'md5')
+    {
+        $responseHash = hash($algo, $this->connectionModule->_getResponseContent());
+        $this->assertEquals($hash, $responseHash);
+    }
+
+    /**
+     * Checks if the hash of a binary response is not the same as provided.
+     *
```
DavertMik: and one example here, please :)

DavertMik:
Please fix the example and I'm merging.

DavertMik:
Ok, I will update that by myself. Thanks

Issue #3987, opened 2017-02-03
Add option cleanup for DataFactory module


alexpts:
I suggest add option 'cleanup' for DataFactory module. Default value is TRUE.
For set 'cleanup' to FALSE module don`t need remove models on '_after' hook.

In my case other module wrap test on transaction on '_before' hook and rollback db on '_after' hook.

It will possible ьore flexibility to customize the database cleanup and config via other molules.

Now already have problem with orm module:
```php
if ($this->ormModule->_getConfig('cleanup')) {
    return; // don't delete records if ORM is set with cleanup
}
```

After add `clenup` option:
```php
if (!$this->config['cleanup']) {
    return;
}
```

DavertMik:
Sorry, I didn't understand the problem.

The DataFactory depends on ORM module (Laravel, Yii2, Doctrine), with `cleanup` option.
When ORM config have `cleanup: true`, DataFactory won't perform a cleanup by itself. In other cases it will. 

So cleanup option of DataFactory depends on cleanup option in a module it uses

alexpts:
I need disable cleanup for ORM module (Yii2) and DataFactory.

DavertMik:
so how records be cleared between tests?

Actually, don't ask me, just do whatever you want. 1) [extend DataFactotry](http://codeception.com/docs/06-ModulesAndHelpers#Extending-a-Module) in Helper, 2) change `_after` method, include your helper instead of DataFactory.

alexpts:
Another module clears the database between tests.

1) I've already done so, but it seemed to me this way is not very correct
I'm worried that in future versions of changed behavior and inheritance I lose backward compatibility.

Also I noticed that you can move make object $story to new method

```php
protected function getStore() 
{
    return $this->ormModule instanceof DataMappe
        ? new RepositoryStore($this->ormModule->_getEntityManager())
        : new ModelStore()
}
```

DavertMik:
> Also I noticed that you can move make object $story to new method

Yep. Things like a good idea. 

Ok, if you can please make PR with this change and adding a cleanup option which is `false` by default. Thanks

alexpts:
Ok, Thanks!
By default cleanup option will set to true for save current behavior. 

I will prepare PR and close this issue after add prepare.

alexpts:
I prepared PR - https://github.com/Codeception/Codeception/pull/3996

PR #3996, opened 2017-02-05
Issue #3987. Add option cleanup for skip cleanup

alexpts:
Add option cleanup for DataFactory. If option cleanup in DataFactory config set to FALSE that don`t need delete records by self.
Also move create default story for new method for more flexible extend it. 

Default option cleanup is TRUE for save old behavior.

### DIFF
--- a/src/Codeception/Module/DataFactory.php
+++ b/src/Codeception/Module/DataFactory.php
@@ -176,7 +176,7 @@ EOF;
      */
     protected function getStore()
     {
-        return $this->ormModule instanceof DataMappe
+        return $this->ormModule instanceof DataMapper
             ? new RepositoryStore($this->ormModule->_getEntityManager()) // for Doctrine
             : new ModelStore();
     }
--- a/src/Codeception/Module/DataFactory.php
+++ b/src/Codeception/Module/DataFactory.php
@@ -157,10 +157,7 @@ EOF;
 
     public function _beforeSuite($settings = [])
     {
-        $store = null;
-        if ($this->ormModule instanceof DataMapper) { // for Doctrine
-            $store = new RepositoryStore($this->ormModule->_getEntityManager());
-        }
+        $store = $this->getStore();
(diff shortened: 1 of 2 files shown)


src/Codeception/Module/DataFactory.php
```diff
@@ -173,6 +170,16 @@ public function _beforeSuite($settings = [])
             }
         }
     }
+    
+    /**
+     * @return StoreInterface
+     */
+    protected function getStore()
+    {
+        return $this->ormModule instanceof DataMappe
```
edno: Incorrect class name for `DataMapper` (`r` is missing, and maybe some test cases since the code passed the current tests).

DavertMik: Yeah, this module really lacks integration tests :(
To be honest it is pretty hard to set up them.

alexpts: Thanks for CR.
I will fix typo error, but tests...

I can\`t research and add functional tests now. 
Do we will can to add this PR without test? Now I don\`t find tests for DataFactory module - https://github.com/Codeception/Codeception/tree/2.2/tests/unit/Codeception/Module

src/Codeception/Module/DataFactory.php
```diff
@@ -173,6 +170,16 @@ public function _beforeSuite($settings = [])
...
     }
+    
+    /**
+     * @return StoreInterface
+     */
+    protected function getStore()
+    {
+        return $this->ormModule instanceof DataMapper
+            ? new RepositoryStore($this->ormModule->_getEntityManager()) // for Doctrine
+            : new ModelStore();
```
Naktibalda: What is this ModelStore class? There is no use statement for it.

alexpts: You right, here must be null instead default ModelStore

alexpts: @Naktibalda, I fixed it

alexpts:
Thanks :)

Issue #4002, opened 2017-02-07
Couldn't parse a test when start with functionals


punchi:
#### What are you trying to achieve?
Test functional and unit tests

#### What do you get instead?

[Codeception\Exception\TestParseException]
Couldn't parse test '/home/runner/proitac/tests/unit/UDTECest.php'
After finishing functionals

```bash
Semaphore (43 min left): ~/alfredo $ php codecept.phar run functional,unit
Codeception PHP Testing Framework v2.2.7
Powered by PHPUnit 5.7.2 by Sebastian Bergmann and contributors.

Functional Tests (129) ------------------------------------------------------------------------------------------------------------------------------------
✔ FAdminBodegaCest: Listar bodega (0.29s)
✔ FAdminBodegaCest: Crear un(a) bodega y ok (0.48s)
...
✔ FVentasCest: Venta una venta nuevo (0.40s)
✔ FVentasCest: Eliminar una venta existente (0.32s)
-----------------------------------------------------------------------------------------------------------------------------------------------------------


  [Codeception\Exception\TestParseException]
  Couldn't parse test '/home/runner/proitac/tests/unit/UDTECest.php'


run [-o|--override OVERRIDE] [--report] [--html [HTML]] [--xml [XML]] [--tap [TAP]] [--json [JSON]] [--colors] [--no-colors] [--silent] [--steps] [-d|--debug] [--coverage [COVERAGE]] [--coverage-html [COVERAGE-HTML]] [--coverage-xml [COVERAGE-XML]] [--coverage-text [COVERAGE-TEXT]] [--no-exit] [-g|--group GROUP] [-s|--skip SKIP] [-x|--skip-group SKIP-GROUP] [--env ENV] [-f|--fail-fast] [--no-rebuild] [--] [<suite>] [<test>]

Semaphore (41 min left): ~/alfredo $ php codecept.phar run unit UDTECest
Codeception PHP Testing Framework v2.2.7
Powered by PHPUnit 5.7.2 by Sebastian Bergmann and contributors.

Unit Tests (3) --------------------------------------------------------------------------------------------------------------------------------------------
✔ UDTECest: Consulta dte (4.96s)
✔ UDTECest: Genera dte (5.27s)
✔ UDTECest: Genera dte2 (5.02s)
-----------------------------------------------------------------------------------------------------------------------------------------------------------


Time: 15.57 seconds, Memory: 30.50MB

OK (3 tests, 4 assertions)
[...]

PR #4007, opened 2017-02-09
Add parse error to TestParseException in PHP7

Naktibalda:
When \ParseError was thrown during test execution , the actual message was lost.

This change helps to debug issues like #4002


DavertMik:
Thanks!

PR #4009, opened 2017-02-10
Added option --filter for executing filtered tests only

disc:
Allow executing only filtered tests like PHPUnit --filter option:

`codecept run unit --filter testGetUser`
`codecept run --filter testGetUser`

### DIFF
--- a/src/Codeception/Command/Run.php
+++ b/src/Codeception/Command/Run.php
@@ -77,6 +77,7 @@ use Symfony\Component\Console\Output\OutputInterface;
  *  --ansi                Force ANSI output.
  *  --no-ansi             Disable ANSI output.
  *  --no-interaction (-n) Do not ask any interactive question.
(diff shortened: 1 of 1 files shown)


DavertMik:
Ok, I may understand your point but you can specify a test by its name or group...
even though this PR is far from being complete, by adding this option in command you are not passing it to PHPUnit runner

disc:
--filter option override filter value inside Codeception and it passed to PHPUnit runner inside
src/Codeception/PHPUnit/Runner.php:prepareSuite

disc:
@DavertMik What do you think about override --filter solution?

Naktibalda:
`codecept run unit --filter testGetUser` is equivalent to `codecept run unit :testGetUser`

disc:
@Naktibalda Thank you! I didn't know about this case!

PR #4010, opened 2017-02-10
Added option --repeat for executing tests repeatedly

disc:
Added --repeat (-r as short version) option for executing tests repeatedly

`codecept run unit tests/unit/UserTest.php:getUserByName --repeat 3`
`codecept run unit tests/unit/UserTest.php:getUserByName -r 3`

### DIFF
--- a/src/Codeception/Command/Run.php
+++ b/src/Codeception/Command/Run.php
@@ -77,6 +77,7 @@ use Symfony\Component\Console\Output\OutputInterface;
  *  --ansi                Force ANSI output.
  *  --no-ansi             Disable ANSI output.
  *  --no-interaction (-n) Do not ask any interactive question.
+ *  --repeat (-r)         Runs the test(s) repeatedly.
  * ```
  *
  */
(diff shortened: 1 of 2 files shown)


DavertMik:
...why?

disc:
@DavertMik Like in PHPUnit --repeat. Sometimes I need to start a test couple of times when I looking for random failure bug in test

edno:
Why not just using an helper?

disc:
Can you provide an example, please? 11:28 AM, February 11, 2017, Greg <notifications@github.com>:Why not just using an helper?

—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or mute the thread.


  
  




Отправлено из мобильной Яндекс.Почты: http://m.ya.ru/ymail

DavertMik:
Yep, but in this case you are restarting the whole suite no matter of the results of execution!

disc:
@DavertMik I see what a problem. I will implement another one solution for repeat.

Issue #4006, opened 2017-02-09
Console completion


gdscei:
It would be great to add autocompletion in bash for commands, and arguments to those commands (suites, tests).

There's a drop-in package available for Symfony Console apps: https://github.com/stecman/symfony-console-completion
It would be great though to configure this so suites & tests (maybe even groups?) had autocompletion too.

gdscei:
@DavertMik I'll be glad to contribute to get this done. How do you suggest I put a library like symfony console completion in composer? As a suggestion or a required composer package? 
If I put it as a suggestion, I'd probably have to create a seperate package/extension for codeception, to not get class not found errors (or nasty `if(class_exist())`, but that doesn't seem like the right way to go...)

Naktibalda:
It is a small package, adding it as required dependency should be fine.

PR #4014, opened 2017-02-10
Added console completion for commands (#4006)

gdscei:
This is a PR for #4006. I've branched this off 2.2, let me know if you want it somewhere else.

Currently working auto-completion:
- All commands
- All options (not the option values)
- Suite, test and path arguments


src/Codeception/Command/Completion.php
```diff
@@ -0,0 +1,53 @@
...
+                ConsoleCompletion::TYPE_ARGUMENT
+            ),
+            new ShellPathCompletion(
+                ConsoleCompletion::ALL_COMMANDS,
+                'test',
+                ConsoleCompletion::TYPE_ARGUMENT
+            ),
+        ]);
+    }
+}
```
gdscei: Fixed in b158e48bd3e847663c40ea3469ee7fc0fc297439

docs/07-AdvancedUsage.md
```diff
@@ -664,6 +664,12 @@ groups:
 This will load all found `p*` files in `tests/_data` as groups. Group names will be as follows p1,p2,...,pN.
 
 
+## Shell autocompletion
+For bash and zsh shells, you can use auto completion by executing the following in your shell (or add it to your bash/zshrc):
```
Naktibalda: This line should say `to your .bashrc/.zshrc`

docs/07-AdvancedUsage.md
```diff
@@ -664,6 +664,12 @@ groups:
...
+````bash
+source <([codecept location] _completion --generate-hook)
```
Naktibalda: I have tested your change and I got  a problem with `[codecept location]`.

I added `source <( ~/projects/codeception/codecept _completion --generate-hook)` to .bashrc.
Autocompletion only works if I specify an absolute path `/home/user/projects/codeception/codecept`,
it doesn't work for `~/projects/codeception/codecept` and `./codecept`.

I think that it would work better if I had codecept in the PATH and ran `source <(codecept _completion --generate-hook)` instead. But it would be good to make autocompletion work for local composer installation.

gdscei:
@Naktibalda Actually, you can already do that:
`[full codecept path] _completion --generate-hook --program vendor/bin/codecept`

Does that work for you as you expect? If so, I can add it to the docs.

Naktibalda:
Actually it is best to use `--program codecept` -  it makes autocompletion work for all commands ending with codecept.
Please update documentation.

Also I think that this cool new feature shouldn't be hidden at the bottom of Advanced Usage page.
@DavertMik Would it be all right to add a short description or a link to [Getting Started](http://codeception.com/docs/02-GettingStarted) page?

gdscei:
@Naktibalda Note that it still technically completes from the original codecept you used for the 'full codecept path' part - this works fine as long as the working directory is in a project directory, because then codeception config will work fine. But different versions of codeception across projects could cause issues (commands that don't exist/changed in versions).

I could try and get the completion function that's outputted with the `--generate-hook` to work off of the composer you're trying to run (so if you did: `vendor/bin/codecept run`, it would actually complete from that codecept install).

gdscei:
@Naktibalda I've added that to this PR. You can now do the following:

`source <([full codecept path] _completion --generate-hook --program codecept --use-vendor-bin)`

This will work for any projects that have a regular codeception set-up (with codeception included as composer package in the project).

Let me know if that works as you'd expect, and I'll wait for an answer from @DavertMik on the docs.

DavertMik:
Thanks, I really like the implementation but I didn't have a chance to try it yet.

gdscei:
@Naktibalda The Travis build failed for something not related to this PR, namely: https://github.com/Naktibalda/codeception-zend-expressive-tests/issues/4

Naktibalda:
I know about that and raised an issue https://github.com/zendframework/zend-expressive/issues/446

Naktibalda:
Travis build has passed, but SemaphoreCI is blocked for some other reason.
Anyway, you need to extend the documentation, a new commit will start a new build.

gdscei:
@Naktibalda I've pushed the doc change, let me know if you need anything else updated.

src/Codeception/Command/Completion.php
```diff
@@ -0,0 +1,79 @@
...
+                ConsoleCompletion::TYPE_ARGUMENT
+            ),
+        ]);
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        if ($input->getOption('generate-hook') && $input->getOption('use-vendor-bin')) {
+            global $argv;
+            $argv[0] = 'vendor/bin/' . basename($argv[0]);
```
Naktibalda: I think that `use-vendor-bin` option is unnecessary and confusing (and actually broken, because `$argv[0]` is not used if `program` option is given),  using `--program vendor/bin/codecept` gives the same result.
Please document effects of using different values of `program` option.

Also I think that it would be good to suggest to use  ```eval `codecept _completion --generate-hook --program codecept` ``` for quick evaluation of code completion without modifying rc files.

gdscei: @Naktibalda  There is a difference: although it will autocomplete with just `--program`, it will actually use the codeception version you used to generate the hook, to autocomplete your codeception. I'll give an example scenario:

Say I use autocomplete like so:
`/var/www/myproject/vendor/bin/codecept _completion --generate-hook --program codecept`

In the complete function from the output of `--generate-hook`, it will use the following to autocomplete:
`/var/www/myproject/vendor/bin/codecept _completion`
(see https://github.com/stecman/symfony-console-completion/blob/master/src/HookFactory.php#L148)

With my change, it will complete using the following:
`vendor/bin/codecept _completion`

If you didn't use the `--use-vendor-bin` option, you could run into an issue where it will autocomplete a different project's codeception commands using the original project's codeception.

I could make this behaviour default if that's what you think would be best? But I thought it would be best to leave it open, in case people have different set-ups for their codeception (global codeception for example)

I'll update the docs with more info, including the `eval` option you mentioned.

Naktibalda: OK, it probably works, but it is still confusing (and it must be clearly documented).
I am not sure if this feature solves or creates more issues.
All versions in the wild do not have `_completion` command, so adding new version to one project will break `tab` behaviour for all projects that are using older versions:
```
$ ./vendor/bin/codecept r
                                                                  
  [Symfony\Component\Console\Exception\CommandNotFoundException]  
  Command "_completion" is not defined. 
```
Using a single source for completion works for all projects, new commands and options aren't added frequently and they are never removed.

gdscei: I understand the issue it can cause too yes. It really depends on the project set-up of the user. I hope I've documented it clearly enough (seems like it from the approval - thanks!)

DavertMik:
Thanks for the great pull request. Thanks to @Naktibalda for reviewing it.
I had some concerns regarding to the functionality: should it be kept in core or to be a part of a standalone package as it is done fro symfony, composer, etc. After a short discussion we decided to keep it in the core, but probably this functionality will be decoupled in 3.0 (whenever it comes out).

Thank you!

PR #4015, opened 2017-02-12
Fixtures behavior compability with `yii2-codeception`

leandrogehlen:
Same behavior on load fixtures: 
https://github.com/yiisoft/yii2-codeception/blob/master/TestCase.php#L85


Naktibalda:
Please make pull request to 2.2 branch instead.

PR #4016, opened 2017-02-13
Fixtures behavior compability with `yii2-codeception`

leandrogehlen:
The same behavior on load fixtures:
https://github.com/yiisoft/yii2-codeception/blob/master/TestCase.php#L85


boboldehampsink:
yii2-codeception is deprecated

leandrogehlen:
@boboldehampsink Yes, the features was moved to yii2 module in codeception, and this implementation was not, then the behavior is different, and this change it's necessary.

boboldehampsink:
Ah yeah, my bad 👍

samdark:
@DavertMik looks OK to merge.

PR #4018, opened 2017-02-15
- g:step gives ambigous warning. add distinct command to docs

B4rb4ross4:
If you execute g:step you get an ambigous error:

> [Symfony\Component\Console\Exception\CommandNotFoundException]       
>   Command "g:step" is ambiguous (generate:stepobject, gherkin:steps). 

I corrected the docs from g:step to g:stepobject


B4rb4ross4:
I have to apologize. I stumbled into this error once before and didn't fix it. So I kind of felt bad now ....

Issue #2923, opened 2016-03-25
loadSessionSnapshot does not work with coverage enabled


shlainn:
Hiya,

I have some working acceptance tests and am now trying to get code coverage for them. As recommended by the Guides, I use an @before directive to handle login before each test, and use saveSessionSnapshot to avoid having to go through login for each test. 

This works nicely until I run codeception with --coverage. What happens then is that the first test in the file goes through nicely, but all subsequent tests, which rely on loadSessionSnapshot to skip login, do not work any more. 

Is this behavior known?

shlainn:
I did some more digging, maybe someone with more experience with codeception can make sense of this. I am running 
`Codeception PHP Testing Framework v2.1.5
Powered by PHPUnit 4.8.21 by Sebastian Bergmann and contributors.`
I ran one of my Cests, once with the **--coverage** flag, and once without and compared the **-vvv** outputs. Without **--coverage** each subsequent test in the Cest continues where the previous test had left off - same page, same cookies, same everything. 
With **--coverage** it resets to the base url for each test. This difference breaks those of my tests which assume that they are still on the same page - because without **--coverage** they still would be.

boboldehampsink:
This is still not working. I am on Codeception 2.2.9. Commenting lines 161-162 in `Codeception\Coverage\Subscriber\LocalServer` makes it work again:

```php
$this->module->amOnPage(basename($this->settings['c3_url']));
$this->module->setCookie(self::COVERAGE_COOKIE, $value);
```

boboldehampsink:
The fix mentioned here fixes bc79819 but does not fix the problem.

boboldehampsink:
I have found this is because calling `$this->module->amOnPage('/')` before every suite resets the session on the server. We should somehow restore the session snapshot before calling this (with a hook?), or have the ability to call startCoverageCollection() manually.

lowerends:
I can confirm this is still happening in version 2.2.10. This version contains #4020.

boboldehampsink:
#4020 is different fix

iBotPeaches:
Thanks for the tips in this. I recently upgraded to 2.3.3 and implemented code coverage and saw my tests failing due to `loadSessionSnapshot`. I think it was my fault at bad test design, but enforcing all my tests to do `$I->amOnPage('/whatever')` before execution fixed this. 

Prior to this, I was depending on the order of tests of where they ended for the next one to begin. (Yes, I know that sounds horrible, that has been resolved).

[...]

PR #4020, opened 2017-02-17
Fix coverage cookie check, mentioned in #2923

boboldehampsink:
There's a double check on the same cookie.


boboldehampsink:
No idea either, only know that it was wrong before ;-) Coincidental notice.

DavertMik:
Thanks. I will merge it once I get a chance to try it by myself

DavertMik:
Thanks!

PR #4025, opened 2017-02-21
Looking for group files can be really slow

pitpit:
This is caused by use of `Finder::path()` at https://github.com/Codeception/Codeception/blob/2.2/src/Codeception/Lib/GroupManager.php#L43-L47 when `groups` parameter is set in codeception.yml

Finder is referencing every files in project root dir, then filters them according to string passed to `path()`. If a lot of files stand in project root dir, it can takes more than 2minutes to find a few group files.

Here's a patch to get rid of `Finder::path()`. It reduces elapsed time from several minutes to few seconds.

### DIFF
--- a/src/Codeception/Lib/GroupManager.php
+++ b/src/Codeception/Lib/GroupManager.php
@@ -42,14 +42,13 @@ class GroupManager
             }
             $files = Finder::create()->files()
                 ->name(basename($pattern))
-                ->path(dirname($pattern))
                 ->sortByName()
-                ->in(Configuration::projectDir());
+                ->in(Configuration::projectDir().dirname($pattern));
 
             $i = 1;
             foreach ($files as $file) {
                 /** @var SplFileInfo $file * */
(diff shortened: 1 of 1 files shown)


pitpit:
testing fails on https://github.com/Codeception/Codeception/blob/2.2/tests/unit/Codeception/Module/PhpBrowserTest.php#L283 which is not related to this PR

----
Guzzle seems to not support request URL without starting slash anymore.
Could be fixed adding `$page = '/'.ltrim($page, '/');` at https://github.com/Codeception/Codeception/blob/2.2/src/Codeception/Lib/InnerBrowser.php#L330

Naktibalda:
#4028

Naktibalda:
Please merge 2.2 branch to make the build pass.

pitpit:
done

laupiFrpar:
+1

pitpit:
If you could have a look on this, please

DavertMik:
Thank you for this performance improvement. Looks nice and elegant

PR #4026, opened 2017-02-21
fixes Guzzle exception `The path of a URI with an authority must start with a slash "/" or be empty`

pitpit:


pitpit:
canceling it, seems to have to much impact

PR #4027, opened 2017-02-21
Explain the $files paramater of sendPOST better

Putr:


Naktibalda:
Thanks.

I was sure that files structure was documented somewhere, but apparently I was wrong.
Anyway, it isn't necessary to specify all properties of $_FILES array,
key-value pairs `['file' => codecept_data_dir('example_img.jpg')]` work too.

Putr:
Yeah, only found it on a number of stackoverflow threads.

But note: I tried the form you suggested. Did not seem to work. Quote a few people online also complained that it's not working so I just assumed that the API changed.

PS: I also noticed that this structure removes the original file on execution. Didn't find any mention of it online tho, but I'm very sure it's happening.

Naktibalda:
> PS: I also noticed that this structure removes the original file on execution. Didn't find any mention of it online tho, but I'm very sure it's happening.

Is it specific to certain structure? I think that it shouldn't make a difference,
but this could happen when the framework module is used and the framework or application code moves the file from original location.

Putr:
The `['file' => codecept_data_dir('example_img.jpg')]` structure does not seem to work at all.

The structure I added to the documentation does, but I've found it removes the file. I'm doing functional API tests on a symfony2.8 framework.
The fact that sendPOST let's the framework work with the source file seems weird to me :/ as in this case the POST functionality can't be reliably tested (since the file keeps being removed).

Naktibalda:
So you have to copy the source file to temporary location before uploading it, then the framework can do whatever it wants with the file.

Putr:
Yeah, thanks for the tip. But this should probably be documented (the fact that it exposes the file directly into the tested code). Do I add the warning in this pull request?

Naktibalda:
`['file' => codecept_data_dir('example_img.jpg')]` works with [PhpBrowser ](https://github.com/Codeception/Codeception/blob/2.2.9/tests/unit/Codeception/Module/PhpBrowserRestTest.php#L220-L231)and all [Zend Frameworks](https://github.com/Naktibalda/codeception-rest-testcase/blob/master/PostMethodFileUploadWithKeyValueArrayCept.php) that I have tested.

DavertMik:
Looks like file uploading with sendPOST may not work as expected in Symfony 2.x
@Putr could you provide and environment to we could reproduce the issue and fix it.

Naktibalda:
Replaced by #4151

Issue #4031, opened 2017-02-23
Too Many Connections error (Laravel5)


rmblstrp:
Similar Issues: #2958, #1591, #2060

Running a large suite of tests with models that have different connection settings throws the "Too Many Connections" error when running tests locally or in a CI pipeline.

The irony of the issue is that the source code makes specific mention of the "Too Many Connections" issue and tries to close open connections.  However, it only closes the the default connection and ignores other open connections.
```php
// disconnect from DB to prevent "Too many connections" issue
if (isset($this->app['db'])) {
    $this->app['db']->disconnect();
}
```
### Details

* Codeception version: 2.2.8
* PHP Version: 5.6.22
* Operating System: Amazon Linux AMI release 2014.03
* Installation type: Composer
* List of installed packages
```
aws/aws-sdk-php                       3.19.32 AWS SDK for PHP - Use Amazon Web Services in your PHP project
barryvdh/laravel-debugbar             v2.2.2  PHP Debugbar integration for Laravel
barryvdh/laravel-ide-helper           v2.1.4  Laravel IDE Helper, generates correct PHPDocs for all Facade classes, to improve auto-completion.
behat/gherkin                         v4.4.1  Gherkin DSL parser for PHP 5.3
chobie/jira-api-restclient            v1.0.0  JIRA REST API.
classpreloader/classpreloader         3.0.0   Helps class loading performance by generating a single PHP file containing all of the autoloaded files for a specifi...
codeception/c3                        2.0.9   CodeCoverage collector for Codeception
codeception/codeception               2.2.8   BDD-style testing framework
codeception/mockery-module            0.2.2   Mockery Module for Codeception
dnoegel/php-xdg-base-dir              0.1     implementation of xdg base directory specification for php
doctrine/annotations                  v1.2.7  Docblock Annotations Parser
doctrine/cache                        v1.6.0  Caching library offering an object-oriented API for many cache backends
doctrine/collections                  v1.3.0  Collections Abstraction library
doctrine/common                       v2.6.1  Common Library for Doctrine projects
doctrine/dbal                         v2.5.4  Database Abstraction Layer
doctrine/inflector                    v1.1.0  Common String Manipulations with regard to casing and singular/plural rules.
doctrine/instantiator                 1.0.5   A small, lightweight utility to instantiate objects in PHP without invoking their constructors
doctrine/lexer                        v1.0.1  Base library for a lexer that can be used in Top-Down, Recursive Descent Parsers.
[...]

PR #4032, opened 2017-02-23
4031: Close all Laravel DB connections after test execution

rmblstrp:
This fixes the "Too Many Connections" issue when running a large test suite which has multiple connection names that point to the same database (i.e. local dev, CI pipeline).

Issue: #4031

### DIFF
--- a/src/Codeception/Module/Laravel5.php
+++ b/src/Codeception/Module/Laravel5.php
@@ -187,17 +187,23 @@ class Laravel5 extends Framework implements ActiveRecord, PartedModule
      */
     public function _after(\Codeception\TestInterface $test)
     {
-        if (! $this->applicationUsesDatabase()) {
-            return;
-        }
+        if ($this->applicationUsesDatabase()) {
+            $db = $this->app['db'];
 
-        if (isset($this->app['db']) && $this->config['cleanup']) {
-            $this->app['db']->rollback();
-        }
+            if ($db instanceof \Illuminate\Database\DatabaseManager) {
+                if ($this->config['cleanup']) {
+                    $db->rollback();
+                }
 
-        // disconnect from DB to prevent "Too many connections" issue
-        if (isset($this->app['db'])) {
-            $this->app['db']->disconnect();
+                /**
+                 * Close all DB connections in order to prevent "Too many connections" issue
+                 *
+                 * @var \Illuminate\Database\Connection $connection
+                 */
+                foreach ($db->getConnections() as $connection) {
+                    $connection->disconnect();
+                }
+            }
         }
     }
 


janhenkgerritsen:
Thanks for the PR!

@DavertMik can you retry the Wercker build? Nothing happens if I do that. If that build succeeds this PR can be merged.

rmblstrp:
@janhenkgerritsen @DavertMik I branched off the 2.2.9 label.  Should I rebase off the head of the 2.2 branch to fix the issue?

Naktibalda:
Yes, that issue is fixed in 2.2

michaelrios:
This would be a great improvement!

rmblstrp:
Rebased and ready to go

aschwin:
I'm a bit confused. How can one set up Codeception with multiple connections and DB's?

rmblstrp:
@aschwin For each test written for Laravel a new fresh instance of the Laravel application is created.  So if you have database connections opened on a connection name other than the default those database connections were being left open.  This fix is to ensure that those open connections are closed after each test in order to avoid a "too many connections" error.

mindplay-dk:
> For each test written for Laravel a new fresh instance of the Laravel application is created. So if you have database connections opened on a connection name other than the default those database connections were being left open. This fix is to ensure that those open connections are closed after each test in order to avoid a "too many connections" error.

@rmblstrp my question is why do application instances accumulate in the first place?

I'm going to guess it's because `*Tester` instances accumulate?

But this seems like (sorry but) really bad design - why wouldn't you dispose of each tester (and any per-test dependencies) when you finish an individual test?

Why would you allow every instance of every system under test to build up in memory?

Every test-suite, as it grows, will need more and more system resources - not just for database connections, but for every instance of every service in any application container.

Moreover, this design requires every individual framework integration to apply an individual fix for the PDO connection limit issue - I could find at least 3 different pull requests for different frameworks; addressing this as each framework user re-discovers the same issue again and again seems rather symptomatic, and in my opinion points to a general design issue.

Note that I don't use one of these major frameworks in the first place - this issue not only affects those frameworks, but essentially any project that uses a DI container.

Now that we've identified the problem, of course I can fix it for each of our projects - but it's really a pretty sub-par work-around in the first place, since it only addresses the issue for PDO specifically, doesn't address any other potential memory/resource issues, puts unnecessary extra load on the CI server, and so on.

I'd really like to see a more complete, long-term solution to this general problem.

mindplay-dk:
For the record, we finally found a simple, general solution to this problem.

We simply added a module with a call to [`gc_collect_cycles`](https://www.php.net/manual/en/function.gc-collect-cycles.php) (more [here](https://www.php.net/manual/en/features.gc.collecting-cycles.php)) from the `_after()` callback.

The fix you've applied here (and for other frameworks elsewhere) specifically deal with leaking too many PDO instances - but this general fix is actually much better in terms of memory usage, as it allows the *entire* DI container's contents to be garbage-collected after running a test.

We benchmarked this solution on a very large test-suite, and it is in fact slightly *faster*, not slower.

We first isolated the problem by mixing in a `trait` that counts instances, like this:

```php
trait Tracking
{
    private $__id;

    private static $__next_id = 1;

    private static $__tracked = [];

    protected function track()
    {
        $this->__id = self::$__next_id++;

        self::$__tracked[$this->__id] = true;

        var_dump("CREATED " . get_class($this));
    }

    public function __destruct()
    {
        unset(self::$__tracked[$this->__id]);

        $count = count(self::$__tracked);

        var_dump("DESTROYED " . get_class($this) . " (COUNT: {$count})");
    }
}
```

Adding `use Tracking` and a call to `$this->track()` from the constructor of your DI container, you can see the real problem illustrated clearly, as hundreds of instances accumulate in-memory and get garbage-collected much too rarely - in some cases every 30-60 seconds of running integration tests that open and destroy hundreds of PDO instances.

Add the module with the `_after()` callback as described, and you'd see everything getting properly collected after each test.

Garbage-collection in PHP doesn't work the way I thought it did - I would have expected cycles to get collected automatically e.g. as soon as a DI container falls out of scope. That's not the case. Garbage collection apparently is immediate only for objects that don't contain cycles - anything that contains cycles gets garbage-collected very rarely, depending on memory usage. Our test-suites don't consume a lot of memory, so in our case, that's as rare as 30-60 seconds. (Clearly this aspect of PHP was designed/optimized to control memory usage and doesn't take into account things like PDO connections which have other resources limitations besides memory usage.)

Note that this *doesn't* address the issue of Codeception caching all the test-instances in-memory: if a test or module caches a DI container in a property (and doesn't explicitly unset it from an `_after` callback) you will still have problems.

(This doesn't affect us, because we bootstrap our containers explicitly in our integration tests (e.g. `$I->createContainer()`) but it looks like caching container instances in modules is the more common approach for various framework integrations, so you do need to clean up your containers from `_after`.)

Anyhow, I'm not submitting a PR or anything, since our problem is solved - but I hope this information might be helpful to you, to improve and simplify framework-integrations.

For one, from what I can tell from this Laravel integration, it doesn't take into account *other* DB connections besides the "default" framework connection - if you have other connection/PDO instances registered, e.g. in migration scripts or importers etc., you'll have the same problems again.

Either way, this solution is not specific to the PDO problem and improves memory usage overall (and you can probably get rid of a lot of framework-specific complexity/details) so it might be worth considering.

PR #4043, opened 2017-03-01
Adds transaction capability to Db.php and boolean transactions config

vaneatona:
Allow the Db module to utilize database transactions by config.

The default is false and everything will work as usual.

### DIFF
--- a/src/Codeception/Module/Db.php
+++ b/src/Codeception/Module/Db.php
@@ -142,7 +142,8 @@ class Db extends CodeceptionModule implements DbInterface
         'populate' => true,
         'cleanup' => true,
         'reconnect' => false,
-        'dump' => null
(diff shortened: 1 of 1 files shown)


Naktibalda:
What is the use case?

DavertMik:
I understand what you are trying to do, but unfortunately this won't work.

Db module creates its own connection to database, separate to the application. Whenever you wrap anything in transaction it won't be visible to your application. If you wanted to use transactions for cleanup, please use one of the ORM (Doctrine, Laravel, Yii2) modules inside functional/integration testing. 

The reason only ORM can handle that is because MySQL does not support nested transactions, so any commit inside a test will commit the active transaction and break the test. So even Db module could get connection from application it is a big chance it won't be working.

PR #4046, opened 2017-03-01
fix deprecation warning in cli test suite

gimler:
The following command `./codecept run cli tests/cli/BuildCept.php` gives deprecation warning.

```
DEPRECATION: Codeception: $scenario->group() has been deprecated and removed. Use annotations to pass scenario params /home/gfranke/Codeception/src/Codeception/Scenario.php:150
```


DavertMik:
Thank you!

PR #4051, opened 2017-03-07
Split CHANGELOG to per-branch files

### LINKED ISSUE #4050: Per-branch changelog files
Changelog conflicts are very annoying when maintaining multiple branches, e.g. 2.2 and master.
Splitting CHANGELOG.md to per-branch files would eliminate a major source of conflicts.

PhpUnit style is a good one: CHANGELOG-2.0.md, CHANGELOG-2.1.md, CHANGELOG-2.2.md

Naktibalda:
Closes #4050


DavertMik:
Thanks, that's really good idea.

This will also require an update to Robofile so the Releases page on the website was built from multiple sources. I think it is better to update a template so it could display the changes on several.

Naktibalda:
I updated RoboFile.php

DavertMik:
Thank you!

PR #4063, opened 2017-03-16
Included suite tests fixes

DavertMik:
This fixes a case where you have mixed local suites (under tests folder) and included suite from other applications (via `include` path).

A reproducible example with a test provided. 
Fix also applied to `build` command


tests/data/included_mix/src/foo/AcmePack/tests/_support/_generated/UnitTesterActions.php
```diff
@@ -0,0 +1,472 @@
+<?php  //[STAMP] b822c13a1ed14d868ece1edfdbc8bd14
+namespace AcmePack\_generated;
```
Naktibalda: Please delete this file and add it to .gitignore

Naktibalda:
Thanks, we should add a rule to ignore generated files to bootstrap template.

DavertMik:
Yes, we have a line like 

```
file_put_contents('.gitignore', file_get_contents('.gitignore') . "\ntests/_output/*");
```

but this should be improved to include generated files. Thanks for the suggestion

PR #4064, opened 2017-03-16
Fix bug with docblock causing incorrect parameter error in IDE

M1ke:
When using the following code in a test

```
Locator::elementAt('.my-selector', 2)
```

PhpStorm reports a parameter type mismatch error due to reading the words immediately after the parameter names in the docblock as a type. This can be fixed by adding scalar types to the docblock which makes no change to the code operation.

Also fixed another warning about an unneeded fully qualified class name in the same file.


DavertMik:
Thanks

Issue #3949, opened 2017-01-24
_bootstrap.php does not introduce variable for tests


PunchyRascal:
#### What are you trying to achieve?
Have some variables ready in tests, like the bootstrap file suggests:
// Here you can initialize variables that will be available to your tests
#### What do you get instead?
undefined variables

### Details

* Codeception version: 2.1.8
* PHP Version: 7.0.13
* Operating System: Win10
* Installation type: Composer

I am talking about `codeception/acceptance/_bootstrap.php`

DavertMik:
yes, sorry this was working for Cept files in early days of Codeception but since we got other formats, they are not support. This is technical limitation of PHP. I think in next versions _bootstrap will be disabled or deprecated.

It is recommended to have shared variables in Actor class, or inside PageObjects.

PR #4065, opened 2017-03-18
Command Cleanup

DavertMik:
$input->getOption('config') no longer provides a config file, it is loaded inside Application, so it's always `null` inside a command class. The custom config is already loaded by that time. This is the reason why nothing was broken ).

Also I removed the `module: config:` section from the bootstrap template. They are never used and may be confusing

### DIFF
--- a/src/Codeception/Command/Bootstrap.php
+++ b/src/Codeception/Command/Bootstrap.php
@@ -140,16 +140,6 @@ class Bootstrap extends Command
             ],
             'extensions' => [
                 'enabled' => ['Codeception\Extension\RunFailed']
-            ],
-            'modules'  => [
-                'config' => [
-                    'Db' => [
-                        'dsn'      => '',
-                        'user'     => '',
-                        'password' => '',
-                        'dump'     => 'tests/_data/dump.sql'
-                    ]
-                ]
             ]
         ];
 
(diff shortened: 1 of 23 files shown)


src/Codeception/Command/Bootstrap.php
```diff
@@ -140,16 +140,6 @@ public function createGlobalConfig()
             ],
             'extensions' => [
                 'enabled' => ['Codeception\Extension\RunFailed']
-            ],
```
DavertMik: not too related for this PR, but better now than never to remove this default config part
it has no use for most setups

src/Codeception/Command/GenerateSuite.php
```diff
@@ -61,14 +61,17 @@ public function execute(InputInterface $input, OutputInterface $output)
...
         }
 
-        $this->buildPath($dir . $suite . DIRECTORY_SEPARATOR, $config['settings']['bootstrap']);
+        if ($config['settings']['bootstrap']) {
+            $this->buildPath($dir . $suite . DIRECTORY_SEPARATOR, $config['settings']['bootstrap']);
+
+            // generate bootstrap
+            $this->save(
+                $dir . $suite . DIRECTORY_SEPARATOR . $config['settings']['bootstrap'],
+                "<?php\n// Here you can initialize variables that will be available to your tests\n",
```
Naktibalda: This comment is no longer true, please remove it #3949

DavertMik:
Yes, but thanks for looking. I will update the code and merge

Issue #4062, opened 2017-03-16
Cannot change Lavarel environment variables after first call


OleksandrVladymyrov:
Hi!

Could you please help me to solve my problem. I write API and use Codeception to test it.
In some cases, I need to send 2 or more calls to API but before that, I should change Lavarel environment variables(LEV). When I try to set it at first all is OK and inside my code(at the controller) I have the value of  LEV the same as I set up. But after the first call(doesn't important method(get, post, ...)) the value of it takes on a default value. Below the part of code from the test:

        $I->getApplication()->make('config')->set(['test_value1' => 11]);

        $I->sendGET('/test');
        $I->seeResponseCodeIs(200);
        $I->seeResponseContainsJson([
            'error' => false,
            'data' => [
                0 => 11
            ]
        ]);

        $I->getApplication()->make('config')->set(['test_value1' => 15]);

        $I->sendGET('/test');
        $I->seeResponseCodeIs(200);
        $I->seeResponseContainsJson([
            'error' => false,
            'data' => [
                0 => 15
            ]
        ]);

As a result of the second query, I get
'data' => [
    0 => null
]

null -is the default value for this LEV

But I expected - 15

What do I make wrong?
Thanks in advance

janhenkgerritsen:
Before a test runs the Laravel application object is initialized, and this also is the application object that is used for the first request in a test. 

However, to support multiple requests in a single test the application object is also re-initialized on each next request. You can see the relevant source code here:

https://github.com/Codeception/Codeception/blob/2.2/src/Codeception/Lib/Connector/Laravel5.php#L109

This means that the application object is reset during the second `$I->sendGET('/test');` call in your test, and that's why your config value change is not sticking.

To work around this problem you should create a separate test for each request.

OleksandrVladymyrov:
First of all thanks for your quick answer.
Do you see a way to change LEV in the same test with a couple requests?

janhenkgerritsen:
Not right now. But you are not the first one with a similar issue, so I'll probably implement something that can help in these types of situations. Currently I am thinking of a hook that will be called after the application object is initialized, either via an event or by registering closures. But I'll have to investigate that a bit to see if this will not create any new problems. So I cannot say when something like this will be implemented.

OleksandrVladymyrov:
Thanks a lot.
I look forward to a speedy decision

janhenkgerritsen:
I created a solution for this issue, please review and provide feedback :)

[...]

PR #4068, opened 2017-03-20
Application handler functionality

janhenkgerritsen:
A generic solution for handling issues such as reported in #4062.

The names of the new methods are `haveApplicationHandler` and `clearApplicationHandlers`, I am open to suggestions for better names.

An example test using this functionality would be:

```
<?php
$I = new FunctionalTester($scenario);

$I->haveApplicationHandler(function($app) {
    $app->make('config')->set(['test_value' => 10]);
});
$I->sendGET('/test-value');
$I->see('Test value is 10');

$I->haveApplicationHandler(function($app) {
    $app->make('config')->set(['test_value' => 15]);
});
$I->sendGET('/test-value');
$I->see('Test value is 15');

$I->clearApplicationHandlers();
$I->sendGET('/test-value');
$I->see('Test value is 5'); // 5 is the default value
```

### DIFF
--- a/CHANGELOG-2.2.md
+++ b/CHANGELOG-2.2.md
@@ -1,3 +1,7 @@
+#### 2.2.10
+
+* [Laravel5] Added `haveApplicationHandler` and `clearApplicationHandlers` methods. See #4068. By @janhenkgerritsen
+
 #### 2.2.9
 
 * [Laravel5] **Laravel 5.4 support** by @janhenkgerritsen
--- a/src/Codeception/Lib/Connector/Laravel5.php
+++ b/src/Codeception/Lib/Connector/Laravel5.php
@@ -68,6 +68,11 @@ class Laravel5 extends Client
     private $instances = [];
 
     /**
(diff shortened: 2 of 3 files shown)


DavertMik:
What else can be done inside `haveApplicationHandler`? 
What other purposes than changing config?

janhenkgerritsen:
You can change whatever you want via the Laravel application object, this is just a generic implementation to give users this ability.

One other example would be overriding bindings in the container, however, the Laravel module already has a few dedicated methods for this like `haveBinding`, `haveInstance`, etc.

I could also have created a dedicated method such as `haveConfigValue`, but I think this generic approach is better. I don't think the use case of changing config values between multiple requests in the same test warrants a new method, and this new generic functionality also allows users to do other things we haven't thought of right now.

One other option is to implement this functionality  as a hidden API method and not as public methods, so users have to access it via a helper. But personally I like the current implementation better.

DavertMik:
Thanks

PR #4071, opened 2017-03-22
Fixed Issue/4059

### LINKED ISSUE #4059: SQL dump silently fails if query is not terminated by semicolon
Importing a dump that includes a single `INSERT` statement, not terminated by semicolon - it's expected to run just fine on PDO; in the end, it's a single command.

#### What do you get instead?
Nothing. Pure wind.

#### Investigation
Investigating the issue, it seems related to a silent test done by [`Driver\Db`](https://github.com/Codeception/Codeception/blob/2.2/src/Codeception/Lib/Driver/Db.php#L125), which basically doesn't run queries not terminated by the given delimiter.

To me, this feels like a not needed protection that might have bad, silent side effects - like it happened to me. So I would expect one of those:
1. some reason on why this is needed since the underlying driver (PDO) is completely capable of running SQL queries without a delimiter in the very end. I noticed this is cited as needed in the Db Module doc, but without reasoning.
2. that guard to be taken out, so queries can run freely without a delimiter :)

### Details

* Codeception version: ^2.2 (dev-master, c8df4d72f3c33a4ae58d3274fe2072402078b472)
* PHP Version: 7.0.x
* Operating System: Ubuntu
* Installation type: Composer

Naktibalda:
Run the last statement in dump file even if it doesn't end with delimiter

Fixes #4059

### DIFF
--- a/src/Codeception/Lib/Driver/Db.php
+++ b/src/Codeception/Lib/Driver/Db.php
@@ -125,8 +125,13 @@ class Db
             if (substr($query, -1 * $delimiterLength, $delimiterLength) == $delimiter) {
                 $this->sqlToRun = substr($query, 0, -1 * $delimiterLength);
                 $this->sqlQuery($this->sqlToRun);
-                $query = "";
+                $query = '';
             }
+        }
+
+        if ($query !== '') {
+            $this->sqlToRun = $query;
+            $this->sqlQuery($this->sqlToRun);
         }
     }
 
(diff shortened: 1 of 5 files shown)


DavertMik:
I restarted the build on Semaphore few times but it is still failing :(

DavertMik:
I think this is related https://github.com/Codeception/Codeception/issues/4073

DavertMik:
The WebDriver failure turned out to be unrelated so I'm merging

PR #4055, opened 2017-03-09
Fixed issue #3653

Naktibalda:
Fixes #3653

PR #4074, opened 2017-03-23
Fixed calling flush on null

Jurigag:
Sometimes when some exception happens during tests and before `_before` method is run in memcache module then in `_after` method flush is done on null causing other error.

### DIFF
--- a/src/Codeception/Module/Memcache.php
+++ b/src/Codeception/Module/Memcache.php
@@ -78,6 +78,10 @@ class Memcache extends CodeceptionModule
      */
     public function _after(TestInterface $test)
     {
+        if (empty($this->memcache)) {
+            return;
+        }
+
         $this->memcache->flush();
         switch (get_class($this->memcache)) {
             case 'Memcache':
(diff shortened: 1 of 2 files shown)


Jurigag:
Test seems unrelated.

src/Codeception/Module/Memcache.php
```diff
@@ -78,14 +78,16 @@ public function _before(TestInterface $test)
...
     {
-        $this->memcache->flush();
-        switch (get_class($this->memcache)) {
-            case 'Memcache':
-                $this->memcache->close();
-                break;
-            case 'Memcached':
-                $this->memcache->quit();
-                break;
+        if (!empty($this->memcache)) {
```
DavertMik: better to change the code to 

```php
if (empty($this->memcache)) {
    return;
}
```

DavertMik:
Yes, unrelated. Just one minor changein the code and I'm merging

src/Codeception/Module/Memcache.php
Jurigag: Okay, will change it later.

Naktibalda:
> when some exception happens during tests and before _before method is run in memcache module then in _after method flush is done on null causing other error.

@Jurigag #4055 fixed this issue for all modules.

Naktibalda:
Actually it didn't, please ignore me.

Jurigag:
Well to be honest this is dirty hack anyway - for example there is this code - https://github.com/phalcon/cphalcon/blob/master/tests/unit/Cache/Backend/ApcCest.php#L36 when this exception is thrown exactly the problem which i described happens, i don't know codeception well enough to figure out why it's happening, this is just workaround for now if someone could fix it properly.

DavertMik:
Thanks

PR #4075, opened 2017-03-23
Added new method seeOptionValueIsSelected

vjandrea:
This method checks only in the value attribute of the selected option. Probably naming it 'seeSelectedOptionHasValue' could be more meaningful but i used the former for legacy reasons.

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -1408,6 +1408,15 @@ class InnerBrowser extends Module implements Web, PageSourceSaver, ElementLocato
         $this->assertEquals($optionText, $value);
     }
 
+    public function seeOptionValueIsSelected($selector, $optionText)
+    {
+        $selected = $this->matchSelectedOption($selector);
+        $this->assertDomContains($selected, 'selected option');
+
+        $value = $selected->getNode(0)->getAttribute('value');
+        $this->assertEquals($optionText, $value);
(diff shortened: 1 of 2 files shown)


DavertMik:
Thanks, however to be accepted this method should be implemented for WebDriver module as well. The test you has added are executed for both PhpBrowser and WebDriver, and fails in its context:

```
- WebDriverTest: See selected option valuePHP Fatal error:  Call to undefined method Codeception\Module\WebDriver::seeOptionValueIsSelected() in /home/runner/Codeception/tests/unit/Codeception/Module/TestsForWeb.php on line 299

Fatal error: Call to undefined method Codeception\Module\WebDriver::seeOptionValueIsSelected() in /home/runner/Codeception/tests/unit/Codeception/Module/TestsForWeb.php on line 299
```
https://semaphoreci.com/codeception/codeception/branches/pull-request-4075/builds/1

vjandrea:
Ok, I'll work on this as soon as I manage to see Selenium work again with Firefox, right now they're rioting 😞 If i can't test locally then it's hard to continue

dwenaus:
This looks great to me. thanks for taking the time to implement it. :)

DavertMik:
Ok, I will merge it and fix the build

DavertMik:
Sorry, this PR was sent to wrong branch so I can't get it merged

rogoit:
How is the status here?

Issue #3829, opened 2016-12-10
Symfony bootstrap.php.cache file should not be mandatory


iquito:
When upgrading to the recently released Symfony 3.2 there were some problems with the bootstrap file (see https://github.com/symfony/symfony/issues/20753 - I had this exact problem) which fueled a discussion within Symfony to remove the bootstrap.php.cache file completely (see https://github.com/symfony/symfony/issues/20668) and this will probably happen with the next Symfony version or even sooner, because the bootstrap file is not needed anymore with PHP 7.

I have already removed the usage of bootstrap.php.cache in my projects - this was always an optimization within Symfony and is not strictly necessary when using composer, because all files can be autoloaded regularly. Unfortunately, Codeception now throws the following error:

    [Codeception\Exception\ModuleRequireException]                                                                                  
    [Symfony] module requirements not met --                                                                                        
                                                                                                                                  
    Symfony bootstrap file not found in /path-to-my-symfony-bundle/../../../var/bootstrap.php.cache  
                                                                                                                                  
    Please specify path to bootstrap file using `var_path` config option                                                            
                                                                                                                                  
    If you are trying to load bootstrap from a Bundle provide path like:                                                            
                                                                                                                                  
    modules:                                                                                                                        
        enabled:                                                                                                                    
        - Symfony:                                                                                                                  
            var_path: '../../app'                                                                                                   
            app_path: '../../app'  

Codeception should be able to run without a specific need for bootstrap.php.cache.

* Codeception version: 2.2.7
* PHP Version: 7.1.0
* Operating System: Debian Linux
* Installation type: Composer

raistlin:
On PHP 7.x you can avoid loading bootstrap.php.cache, some simple hacks around https://github.com/Codeception/Codeception/blob/2.2/src/Codeception/Module/Symfony.php#L137 should be enough. But we should not remove the possibility to load it if the file exists, it will run tests very slowly.

My proposal will be, if ```var\bootstrap.php.cache``` exists load it. If it fails, try to load ```app\autoload.php```.
Something like https://github.com/symfony/symfony-demo/blob/master/web/app.php#L13-L16

Merge requests are welcome.

PR #4081, opened 2017-03-27
Make symfony bootstrap.php.cache optional for php version > 7

patrickjahns:
As per https://github.com/Codeception/Codeception/issues/3829 - this PR allows for not requiring bootstrap.php.cache any longer for php versions > 7 - but will still use it if is available

### DIFF
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -134,20 +134,8 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
 
     public function _initialize()
     {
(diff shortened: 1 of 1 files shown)


Naktibalda:
Does it actually depend on a version of PHP or on a version of Symfony?

patrickjahns:
According to #3829  and  most of the issues on the symfony tracker, class cache won't be needed in php7. And is also being deprecated 
see: https://github.com/symfony/symfony/issues/20668

As mentioned in #3829 I left it there to be optionally included - but we can also discuss to completely deprecate it in the future?

DavertMik:
Ok, makes sense. Thanks for the Pull Request

PR #4086, opened 2017-03-28
[ZendExpressive] Support Zend Expressive 2.0

Naktibalda:
Migration to Zend Expressive 2.0 creates 2 new config files that must be included explicitly

### DIFF
--- a/src/Codeception/Module/ZendExpressive.php
+++ b/src/Codeception/Module/ZendExpressive.php
@@ -53,10 +53,22 @@ class ZendExpressive extends Framework
     public function _initialize()
     {
         $cwd = getcwd();
-        chdir(Configuration::projectDir());
-        $this->container = require Configuration::projectDir() . $this->config['container'];
+        $projectDir = Configuration::projectDir();
+        chdir($projectDir);
+        $this->container = require $projectDir . $this->config['container'];
+        $app = $this->container->get('Zend\Expressive\Application');
+
+        $pipelineFile = $projectDir . 'config/pipeline.php';
+        if (file_exists($pipelineFile)) {
+            require $pipelineFile;
+        }
+        $routesFile = $projectDir . 'config/routes.php';
+        if (file_exists($routesFile)) {
+            require $routesFile;
+        }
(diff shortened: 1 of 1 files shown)


src/Codeception/Module/ZendExpressive.php
```diff
@@ -53,10 +53,22 @@ class ZendExpressive extends Framework
...
     {
         $cwd = getcwd();
-        chdir(Configuration::projectDir());
-        $this->container = require Configuration::projectDir() . $this->config['container'];
+        $projectDir = Configuration::projectDir();
+        chdir($projectDir);
+        $this->container = require $projectDir . $this->config['container'];
+        $app = $this->container->get('Zend\Expressive\Application');
+
+        $pipelineFile = $projectDir . 'config/pipeline.php';
```
DavertMik: How about adding this to the config of a module with the default value?

src/Codeception/Module/ZendExpressive.php
```diff
@@ -53,10 +53,22 @@ class ZendExpressive extends Framework
...
+        if (file_exists($pipelineFile)) {
+            require $pipelineFile;
+        }
+        $routesFile = $projectDir . 'config/routes.php';
```
DavertMik: probably this can be added to config as well

src/Codeception/Module/ZendExpressive.php
Naktibalda: Migration tool creates these files, so the names are standard.
https://github.com/Naktibalda/codeception-zend-expressive-tests/pull/5/commits/5c3d43cd64576d6e03f929f192627d6d57864cc6#diff-db58f9aa96436a4e18ea969a09999aa9

If I added a parameter, I would have to throw exception if the file doesn't exist, right?
Requiring any files would break compatibility with 1.x (users would have add a new setting to fix their tests).
Not requiring any files by default, would add a manual step for users of 2.x.

DavertMik: Ok, got it

PR #4089, opened 2017-03-29
Port some functions to Lumen from Laravel5

kt81:
(Proposal)
Lumen's IoC container looks very similar to Laravel's one.
so I think several functions can be shared with each modules.


janhenkgerritsen:
I'll have to look into this when I have some more time, but I can't really say when that will be.

janhenkgerritsen:
Thanks for this PR, there is only one change I'd like you to make before merging it. Can you change the file for the trait for the modules to the following:

`src/Codeception/Lib/Shared/LaravelCommon.php`

This way it follows the same convention as the trait for the connector classes, which is located in `src/Codeception/Lib/Connector/Shared/LaravelCommon.php`.

kt81:
@janhenkgerritsen 
Thank you for reviewing. I moved the file so that it may follow the convention.

PR #4092, opened 2017-03-30
support dataProvider - fix case

ivan1986:
https://phpunit.de/manual/current/en/writing-tests-for-phpunit.html
default in phpunit is "dataProvider" support both for now ("dataprovider" need mark deprecated?)


DavertMik:
Thanks. Looks good, esp with BC layer. I think it can be merged

Issue #4091, opened 2017-03-30
attachFille() causes exception [InvalidArgumentException] file not found or not readable


ahmednawazbutt:
#### [InvalidArgumentException] file not found or not readable:

I run the below command in terminal
````codecept run --steps tests/acceptance/models/FileCest:uploadFolderInvalidValues````
and get the below output:
```bash
---------
1) FileCest: Submit a model with invalid values to see errors in submission
 Test  tests/acceptance/models/FileCest.php:uploadFolderInvalidValues
                                                               
  [InvalidArgumentException] file not found or not readable:   
                                                               

Scenario Steps:

 13. $I->attachFile("#directoryuploadform-uploadedfile","/home/anb-pc/Projects/TestProject/common/modules/ofilesgallery/tests/_data/files/flower.jpg") at tests/acceptance/models/FileCest.php:154
 12. $I->see("Upload Folder","h4.modal-title") at tests/acceptance/models/FileCest.php:152
 11. $I->wait("1") at tests/acceptance/models/FileCest.php:150
 10. $I->click("#btn_upload_directory") at tests/acceptance/models/FileCest.php:149
 9. $I->dontSee("No record Found!","div.alert-info") at tests/acceptance/models/FileCest.php:57
 8. $I->see("Files Manager") at tests/acceptance/models/FileCest.php:56

#1  Codeception\Module\WebDriver->attachFile
#2  /home/anb-pc/Projects/TestProject/common/modules/ofilesgallery/tests/_support/_generated/AcceptanceTesterActions.php:1491
#3  /home/anb-pc/Projects/TestProject/common/modules/ofilesgallery/tests/acceptance/models/FileCest.php:154
#4  common\modules\ofilesgallery\tests\models\FileCest->uploadFolderInvalidValues

```
> Here is the Source code for the method I am trying to test.

```php
    public function uploadFolderInvalidValues(AcceptanceTester $I)
    {
        $I->wantTo('Submit a Model with invalid values to see errors in submission');
        
        $I->amOnPage(Url::toRoute('/ofilesgallery/file/index'));    //  Root page
        $I->see( Yii::t('main', 'Files Manager') );
        $I->dontSee( Yii::t('main', 'No record Found!'), 'div.alert-info' );
        
        $I->click('#btn_upload_directory');
        $I->wait('1');  //  Wait untill the model-popup Apears
        
        $I->see(Yii::t('main', 'Upload Folder'), 'h4.modal-title');
        
[...]

PR #4096, opened 2017-04-01
[Webdriver] fixed attachFile messages when the file does not exist

Naktibalda:
attachFile method of WebDriver throws confusing message `[InvalidArgumentException] file not found or not readable:  ` when the file does not exist #4091 .
It was a fault of `realpath` function which returns FALSE when the file does not exist.
I fixed it to throw `[InvalidArgumentException] File does not exist: /path/to/tests/_data/filename.ext`.

And a different message `[InvalidArgumentException] File is not readable: /path/to/tests/_data/filename.ext`
when the file exists but is not readable.

Also I updated InnerBrowser to throw the same messages,
it used to throw `PHPUnit_Framework_AssertionFailedError` `file filename.ext not found in Codeception data path. Only files stored in data path accepted`


DavertMik:
Thank you

PR #4105, opened 2017-04-04
[Asserts] Add delta parameter to assertEquals() assertNotEquals() methods

spideyfusion:
Currently it's not possible to assert floating-point numbers reliably using the `assertEquals` and `assertNotEquals` methods inside the **Asserts** module.

This commit adds an additional default parameter to the above mentioned methods while following PHPUnit's method prototypes. There shouldn't be any backwards incompatible changes introduced here.

### DIFF
--- a/src/Codeception/Module/Asserts.php
+++ b/src/Codeception/Module/Asserts.php
@@ -10,27 +10,57 @@ class Asserts extends CodeceptionModule
 {
 
     /**
-     * Checks that two variables are equal.
+     * Checks that two variables are equal. If you're comparing floating-point values,
+     * you can specify the optional "delta" parameter which dictates how great of a precision
+     * error are you willing to tolerate in order to consider the two values equal.
+     *
+     * Regular example:
+     * ```php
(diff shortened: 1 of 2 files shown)


Naktibalda:
I can't tell you if this change will be accepted, but it would be better to make pull request to 2.2 branch.

spideyfusion:
@Naktibalda Thanks for the feedback. I changed the target branch of this PR.

src/Codeception/Module/Asserts.php
```diff
@@ -15,10 +15,11 @@ class Asserts extends CodeceptionModule
      * @param        $expected
      * @param        $actual
      * @param string $message
+     * @param float  $delta
```
DavertMik: Please add some documentation about what delta is and how to use it. Add it to this docblock.

spideyfusion: I updated the documentation and provided some usage examples.

DavertMik:
Thanks, merging

Issue #4106, opened 2017-04-05
Yii 2.0 inner browser and nested transaction


bscheshirwork:
#### What are you trying to achieve?
Run yii2 functional test
#### What do you get instead?
```
root@379af4bbfaed:/project/backend# cd backend/; codecept run functional CreateUserCest
bash: cd: backend/: No such file or directory
Codeception PHP Testing Framework v2.2.10
Powered by PHPUnit 5.7.17 by Sebastian Bergmann and contributors.

Backend\tests.functional Tests (1) -------------------------------------------------------------------------------------------------------------------------------------------------------------------------
E CreateUserCest: Ensure that user creation works (0.13s)
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
1) CreateUserCest: Ensure that user creation works
 Test  tests/functional/CreateUserCest.php:createUser
                                                                                                                                                                       
  [yii\db\Exception] SQLSTATE[42000]: Syntax error or access violation: 1305 SAVEPOINT LEVEL1 does not exist
The SQL being executed was: ROLLBACK TO SAVEPOINT LEVEL1  
                                                                                                                                                                       

Scenario Steps:

 9. $I->click("form[action*="create"] button[type=submit]") at tests/_support/Page/CreateUser.php:47
 8. $I->fillField("#user-password","") at tests/_support/Page/CreateUser.php:46
 7. $I->fillField("#user-email","") at tests/_support/Page/CreateUser.php:45
 6. $I->fillField("#user-username","") at tests/_support/Page/CreateUser.php:44
 5. $I->amOnPage("/user/admin/create") at tests/_support/Page/CreateUser.php:42
 4. // I am going to try to create user with empty fields

#1  /project/vendor/yiisoft/yii2/db/Command.php:856
#2  /project/vendor/yiisoft/yii2/db/Schema.php:428
#3  /project/vendor/yiisoft/yii2/db/Transaction.php:194
#4  /project/vendor/dektrium/yii2-user/models/User.php:279
#5  /project/vendor/dektrium/yii2-user/controllers/AdminController.php:233
#6  dektrium\user\controllers\AdminController->actionCreate
#7  /project/vendor/yiisoft/yii2/base/InlineAction.php:57
#8  /project/vendor/yiisoft/yii2/base/Controller.php:156
#9  /project/vendor/yiisoft/yii2/base/Module.php:523
#10 /project/vendor/yiisoft/yii2/web/Application.php:102


[...]

PR #4108, opened 2017-04-06
more info about _fixtures

bscheshirwork:
#4106


DavertMik:
Please update dockblock in Yii2 module itself, as this .md file is generated from dockblock

PR #4115, opened 2017-04-09
Fixed PHP Notice for Db::loadDump

sergeyklay:
```
PHP Notice:  Undefined property: Codeception\Lib\Driver\MySql::$sqlToRun in
.../vendor/codeception/codeception/src/Codeception/Module/Db.php on line 299
```


Naktibalda:
Oops, I missed that one.

Naktibalda:
PhpStorm didn't show this usage of sqlToRun property, so I removed it from the drivers.

Naktibalda:
It was caused by 7e239f747f518b0cc400548a35926706742230db
I provided a better patch in #4120

PR #4117, opened 2017-04-10
edits to first paragraph to make it easier to read & understand

bou2you:
I'm a big fan of BDD and found it a shame that the opening paragraph to the BDD Codeception doc page had grammatical errors.

I find this easier to read as a native English speaker.

What do you think? Does it still convey what you were trying to get across?


DavertMik:
Yes, looks good. Thank you

PR #4120, opened 2017-04-10
[Db] loadDump reports sql statement which caused error

Naktibalda:
This functionality was accidently broken by removal of sqlToRun property in 2.2.10
Replaces #4115


Naktibalda:
Wercker build has failed because HHVM uses a different message format for PDO exceptions.
I will remove non-matching part from expected message.

DavertMik:
Good. Thanks

PR #4125, opened 2017-04-11
Steps with PyString and with inline string argument considered the same

### LINKED ISSUE #4121: PyString not considered as step arg by gherkin:snippets and when validating steps before running them
### What are you trying to achieve?

I have two the same steps but first has inline argument and second PyString argument:

**InlineExample.feature**

```gherkin
  Scenario: Inline argument
    ...
    Then I see output "test"
    ...
```
**PyStringExample.feature**
```gherkin
  Scenario: PyString argument
    ...
    Then I see output
      """
      First line
      Second line
      """
    ...
```

I want them all to be considered as the same steps:

1. `gherkin:snippets` command generates same snippet for this step for both feature:

```php
    /**
     * @Then I see output :arg1
     */
     public function iSeeOutput($arg1)
     {
        throw new \Codeception\Exception\Incomplete("Step `I see output :arg1` is not defined");
     }

```

2. Suppose I defined all steps for `InlineExample.feature` including `I see output` step.
When I run `PyStringExample.feature` then Codeception sees that `I see output` already defined and shows only really incompleted steps.

### What do you get instead?

Now Codeception considers same steps with inline and PyString argument as different steps:

1. For `InlineExample.feature` step generated with argument (`public function iSeeOutput($arg1)`) and for `PyStringExample.feature` we get snippet without arg (`public function iSeeOutput()`)
2. If I defined all steps for `InlineExample.feature` and then run `PyStringExample.feature` then Codeception shows `I see output` step as incompleted. 

### Details

* Codeception version: 2.2.10

kuntashov:
Fixes #4121

### DIFF
--- a/src/Codeception/Test/Gherkin.php
+++ b/src/Codeception/Test/Gherkin.php
@@ -7,6 +7,7 @@ use Behat\Gherkin\Node\ScenarioInterface;
 use Behat\Gherkin\Node\StepNode;
 use Behat\Gherkin\Node\TableNode;
 use Codeception\Lib\Di;
+use Codeception\Lib\Generator\GherkinSnippets;
 use Codeception\Scenario;
 use Codeception\Step\Comment;
 use Codeception\Step\Meta;
@@ -91,7 +92,7 @@ class Gherkin extends Test implements ScenarioDriven, Reported
     protected function validateStep(StepNode $stepNode)
     {
         $stepText = $stepNode->getText();
-        if (\Codeception\Lib\Generator\GherkinSnippets::stepHasPyStringArgument($stepNode)) {
+        if (GherkinSnippets::stepHasPyStringArgument($stepNode)) {
             $stepText .= ' ""';
         }
         foreach ($this->steps as $pattern => $context) {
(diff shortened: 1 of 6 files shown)


src/Codeception/Test/Gherkin.php
```diff
@@ -91,6 +91,9 @@ public function test()
     protected function validateStep(StepNode $stepNode)
     {
         $stepText = $stepNode->getText();
+        if (\Codeception\Lib\Generator\GherkinSnippets::stepHasPyStringArgument($stepNode)) {
```
Naktibalda: It would be better to add `use \Codeception\Lib\Generator\GherkinSnippets;` line at the top.

kuntashov: @Naktibalda Thanks, done.

DavertMik:
Looks good

kuntashov:
Found bug, please do not merge.
Working on it.

DavertMik:
@kuntashov Ok, ping when you think it is ready.

kuntashov:
@DavertMik  Fixed in https://github.com/Codeception/Codeception/pull/4125/commits/f6f7dec422cf9b0795c2b27d3efe6633e0a70cac

DavertMik:
Thank you

PR #4127, opened 2017-04-12
about _fixtures

bscheshirwork:
#4106 #4108

### DIFF
--- a/src/Codeception/Module/Yii2.php
+++ b/src/Codeception/Module/Yii2.php
@@ -288,6 +288,35 @@ class Yii2 extends Framework implements ActiveRecord, PartedModule
      * ]);
      * ```
      *
+     * Note: if you need to load fixtures before the test (probably before the cleanup transaction is started; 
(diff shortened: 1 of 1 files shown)


src/Codeception/Module/Yii2.php
```diff
@@ -288,6 +288,35 @@ public function amLoggedInAs($user)
...
+     * public function _fixtures(){
+     *     return [
+     *         'user' => [
+     *             'class' => UserFixture::className(),
+     *             'dataFile' => codecept_data_dir() . 'user.php'
+     *         ]
+     *     ];
+     * }
+     * ```
+     * instead of
```
DavertMik: "...instead of defining them in `_before`"

That's all. No need for code example below.

bscheshirwork: little bit expand
"...instead of defining `haveFixtures` in Cest `_before`"
may be?

DavertMik: yes, but without code sample. It may be confusing

DavertMik:
Ok, thanks for the update. Merging it.

PR #4135, opened 2017-04-17
Fixed output of failed step

Naktibalda:
This part of output was missing for regular steps.
```
 Step  See file found "games.zip"
 Fail
```

http://phptest.club/t/seeelement-wierd-fail-message/1470


DavertMik:
Thank you!

Issue #4134, opened 2017-04-15
Add markup validation functionality to acceptance tests


Kolyunya:
#### What are you trying to achieve?
Automatically validate page markup.

#### What do you get instead?
No page markup validation API available in the testing framework core.

#### Details
Having a valid page markup is absolutely crucial for HTML page rendering. I would like to propose to add a markup validation functionality to acceptance tests API.

The World Wide Web Consortium is so kind to provide us with such a wonderful code-quality audit tool like the [markup validator](https://validator.w3.org/). It's even kinder to provide us with an [API to the validator](https://validator.w3.org/docs/api.html). I can see no reason not to make use of it in acceptance tests. This would increase the HTML code quality drastically.

The process I propose to automate is when an acceptance tester
* opens a page;
* copies page source to the validator;
* checks if the validator output has any errors or warnIngs;
* reports if markup contains any errors or warnings.

#### Solution
I made a [draft implementation](https://github.com/Codeception/Codeception/pull/4136) of the markup validator module and I would be more than happy to work on the pull request if you would like to adopt my proposal.

Module usage is as easy as this:
```php
$I->amOnPage('/');
$I->validateMarkup();
```
That's it. Just one method call which will inform you if your markup is invalid.

The module is configurable. You can ignore warnings if needed or ignore particular errors by regular expressions. We can come up with some other configuration options.

Thank you!

Kolyunya:
I refactored the module and added test coverage.

Kolyunya:
[The first release of a module](https://github.com/Kolyunya/codeception-markup-validator) is complete. Hope it will be useful to someone!

/cc @edno

PR #4136, opened 2017-04-17
WIP: Implement markup validator module (#4134)

Kolyunya:
#### Implementation details
A new `PageSourceViewer` interface is added to be able to retrieve page source from abstract configurable module. If you don't want to introduce a new interface we can rollback to the first commit where page source retrieval process is hardcoded inside the `MarkupValidatorModule`.

#### Module configuration
- `pageSourceViewer` - a string representing a name of a page source viewer. E.g. `PhpBrowser` or `WebDriver` or any other module implementing `PageSourceViewer`.
- `ignoredErrors` - a list of ignored errors (matching by a regular expression).
- `ignoreWarnings` - a boolean indicating whether to ignore warnings or not.

#### Roadmap
- [x] Add `PageSourceViewer` to be able to get page source from.
- [x] Implement `PageSourceViewer` in `PhpBrowser`.
- [x] Implement `PageSourceViewer` in `WebDriver`.
- [x] Implement `MarkupValidatorModule`.
- [x] Add tests coverage of the `MarkupValidatorModule`.
- [ ] Add tests coverage of the `PhpBrowser::_getPageSource()`.
- [ ] Add tests coverage of the `WebDriver::_getPageSource()`.
- [ ] Approve `PageSourceViewer` name.
- [ ] Approve `MarkupValidatorModule` name.
- [ ] Approve module configuration options.
- [ ] Review and approve code.
- [ ] Review and approve tests.
- [ ] Update documentation.
- [ ] Update changelog.

### DIFF
--- a/src/Codeception/Module/MarkupValidator.php
+++ b/src/Codeception/Module/MarkupValidator.php
@@ -3,9 +3,8 @@ namespace Codeception\Module;
 
 use Exception;
 use stdClass;
+use Codeception\Lib\Interfaces\PageSourceViewer;
 use Codeception\Module;
-use Codeception\Module\PhpBrowser;
-use Codeception\Module\WebDriver;
 use GuzzleHttp\Client;
 
 /**
@@ -13,6 +12,7 @@ use GuzzleHttp\Client;
  * Requires either the `PhpBrowser` or the `WebDriver` module.
  *
  * Configuration options:
+ *  - pageSourceViewer
  *  - ignoreWarnings
  *  - ignoredErrors
  */
@@ -45,21 +45,27 @@ class MarkupValidator extends Module
      */
     protected function getCurrentPageMarkup()
     {
-        try {
-            $markup = $this->getMarkupFromPhpBrowser();
-            return $markup;
-        } catch (Exception $exception) {
-            // Wasn't able to get markup from the PhpBrowser.
+        $pageViewerConfigKey = 'pageSourceViewer';
+        if (!isset($this->config[$pageViewerConfigKey])) {
+            throw new Exception(sprintf(
+                'Missing mandatory configuration parameter "%s".',
+                $pageViewerConfigKey
+            ));
         }
 
-        try {
-            $markup = $this->getMarkupFromWebDriver();
-            return $markup;
-        } catch (Exception $exception) {
-            // Wasn't able to get markup from the WebDriver.
+        $pageViewerName = $this->config[$pageViewerConfigKey];
+        if (!$this->hasModule($pageViewerName)) {
+            throw new Exception(sprintf(
+                '"%s" page viewer was not found.',
+                $pageViewerName
+            ));
         }
 
-        throw new Exception('Unable to obtain current page markup.');
+        /* @var $pageViewer PageSourceViewer */
+        $pageViewer = $this->getModule($pageViewerName);
+        $pageMarkup = $pageViewer->_getPageSource();
+
+        return $pageMarkup;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


tests/unit/Codeception/Module/MarkupValidatorTest.php
```diff
@@ -0,0 +1,402 @@
...
+    public function testGetCurrentPageMarkupNoPhpBrowserNoWebDriver()
+    {
+        $this->module = $this
+            ->getMockBuilder('\Codeception\Module\MarkupValidator')
+            ->setConstructorArgs(array(make_container()))
+            ->enableProxyingToOriginalMethods()
+            ->getMock()
+        ;
+
+        $this->expectException('Exception');
```
Naktibalda: This test fails on PHP 5.4 with `PHP Fatal error:  Call to undefined method MarkupValidatorTest::expectException() in /home/travis/build/Codeception/Codeception/tests/unit/Codeception/Module/MarkupValidatorTest.php on line 40`

Codeception 2.2 supports PhpUnit 4.8, please use `setExpectedException` instead.

src/Codeception/Module/MarkupValidator.php
```diff
@@ -0,0 +1,184 @@
...
+        $pageViewerName = $this->config[$pageViewerConfigKey];
+        if (!$this->hasModule($pageViewerName)) {
+            throw new Exception(sprintf(
+                '"%s" page viewer was not found.',
+                $pageViewerName
+            ));
+        }
+
+        /* @var $pageViewer PageSourceViewer */
+        $pageViewer = $this->getModule($pageViewerName);
```
Kolyunya: The module depends on any other module which can provide current page source code - not only `WebDriver`. I think that it would be nice to have a generic `PageSourceViewer` anyway since `PageSourceSaver` is already present. This module is a great use-case example for that.

Otherwise this module will have to know each of the dependent modules implementation details to obtain page sources from them. That is not nice. It's better to work on the interface level.

Naktibalda: Yes, so the configuration can be set to `depends: PhpBrowser`, `depends: Symfony`, `depends: Phalcon`, etc.

Kolyunya:
@Naktibalda thanks for the review! I'm going to update the PR accordingly in a while.

> It would be easier to accept a change which adds one method to WebDriver and another to InnerBrowser.
Why would we duplicate the very same functionality in two separate modules?

Naktibalda:
> Why would we duplicate the very same functionality in two separate modules?

Don't duplicate it, put all the code somewhere in `Codeception/Lib` and call it from 2 places.

src/Codeception/Module/MarkupValidator.php
Kolyunya: Markup validator will have to know implementation details of each dependent module to get page source from them which is not good.

I think it's better to extract common functionality to an interface. This will allow client code independent from implementation details of dependent modules. This will lower module coupling which is always good.

DavertMik:
Thanks for pull request!
Thanks @Naktibalda for a review.

I really like the implementation but I'm not sure that Codeception needs this as base functionality. I don't remember requests for markup validation feature so probably this is not what we need at first place. We can't receive all new modules, as by accepting them we will need to support them and maintain in future. 

So I'd propose the following: 

1. If you need new interface for modules, I can accept this change. 
2. Module with markup validation functionality please publish as a standalone package and add to Addons page as @Naktibalda suggested.

Kolyunya:
@DavertMik I see, you want to keep the core as small as possible but I just thought that this functionality is something absolutely fundamental and generic in the scope of web-testing. Anyway thank you for a suggestion. I'm going to make a third-party module then.

Regarding the interface. Do you think we should stick to a separate interface or add a method to say `Codeception\Lib\Interfaces\Web`?

DavertMik:
> Regarding the interface. Do you think we should stick to a separate interface or add a method to say Codeception\Lib\Interfaces\Web?

I think `Interfaces\Web` is ok. All web modules can return source of a page.

Kolyunya:
@Naktibalda @DavertMik thank you guys for the review and advises. I finally made [the module](https://github.com/Kolyunya/codeception-markup-validator). Any feedback is very appreciated! Thank you!

PR #4141, opened 2017-04-18
Add a method to `Codeception\Lib\Interfaces\Web`

Kolyunya:
Extracted from #4136.

Added `Codeception\Lib\Interfaces\Web::grabPageSource()` method. Implemented the method in `Codeception\Lib\InnerBrowser` and `Codeception\Module\WebDriver`. Added test coverage for scenarios when page is and is not loaded.

I tried to be consistent with the rest of the code base. Is `grabPageSource()` OK?

There's one problem: `Selenium` adds `xmlns="http://www.w3.org/1999/xhtml"` while `PhantomJS` does not do that. I'm not quite sure how to handle that gracefully. I'm not quite happy with the 
```php
$sourceActualRaw = $this->module->grabPageSource();
$sourceActual = str_replace('xmlns="http://www.w3.org/1999/xhtml"', '', $sourceActualRaw);
```
and `isPhantom()` is `protected`...

### DIFF
--- a/tests/web/WebDriverTest.php
+++ b/tests/web/WebDriverTest.php
@@ -998,4 +998,40 @@ class WebDriverTest extends TestsForBrowsers
         $this->module->switchToIFrame();
         $this->module->see('Iframe test');
     }
+
+    public function testGrabPageSourceWhenNotOnPage()
+    {
+        $this->setExpectedException(
+            '\Codeception\Exception\ModuleException',
+            'Current url is blank, no page was opened'
+        );
+        $this->module->grabPageSource();
+    }
+
+    public function testGrabPageSourceWhenOnPage()
+    {
+        $this->module->amOnPage('/minimal');
+        $sourceExpected =
+<<<HTML
+<!DOCTYPE html>
+<html>
+    <head>
+        <title>
+            Minimal page
+        </title>
+    </head>
+    <body>
+        <h1>
+            Minimal page
+        </h1>
+    </body>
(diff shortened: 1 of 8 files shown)


Naktibalda:
Please make a new pull request to 2.2 branch, master will not be released for months.

Kolyunya:
@Naktibalda rebased to `2.2`.

tests/web/WebDriverTest.php
```diff
@@ -998,4 +998,40 @@ public function testSwitchToIframe()
...
+            Minimal page
+        </h1>
+    </body>
+</html>
+
+HTML
+        ;
+        $sourceActualRaw = $this->module->grabPageSource();
+        // `Selenium` adds the `xmlns` attribute while `PhantomJS` does not do that.
+        $sourceActual = str_replace('xmlns="http://www.w3.org/1999/xhtml"', '', $sourceActualRaw);
```
Naktibalda: Does this replace cause a validation issue if the page is actually XHTML?

Kolyunya: I'm not quite sure I got the point of the question. The test page is an `HTML5` page. `Selenium` adds this attribute for some reason I'm not aware of. And this replace is just a hack to make the test pass for `Selenium`.

Kolyunya: I think this method should return unmodified source of a page but I'm not sure if it's possible to implement.

Issue #3995, opened 2017-02-04
Support for PHPUnit 6.0


neufeind:
PHPUnit 6.0 was recently released. Make using 6.0 possible in the 2.2-branch (composer.json has <6.0 currently).

samdark:
It has all the classes namespaced so it won't be easy...

neufeind:
Just found another issue mentioning bigger changes in PHPUnit 6.0 which might make it hard (impossible?) to support 5.x and 6.x both. Though it would be great for other frameworks currently using Codeception (and maybe not really wanting to wait until 3.0).
https://github.com/Codeception/Codeception/issues/3972#issuecomment-277348788

DavertMik:
Yep, technically it is impossible to support it in nearest future. According to PHPUnit changelog there is no good reason to do that, as it currently brings no new features. But yes, once we will be close to 3.0 we will switch to PHPUnit 6.x

lastzero:
We are using PHPUnit 6, but due to Codeception we can not upgrade our frontend project and are stuck with 5.7 as it seems :(

I upgraded my own [test-tools](https://github.com/lastzero/test-tools) within a day, no issues. Just change the _ to \ basically.

So official statement is: If you're using PHPUnit 6.0, wait for Codeception 3.0, right? If we want to use both before, we have to fork?

lastzero:
Note: PHPUnit 5.7 offers a class name alias for forward compatibility. So the issue is not as big as it seems at first. Good work Sebastian :)

orlangur:
> bigger changes in PHPUnit 6.0 which might make it hard (impossible?) to support 5.x and 6.x both

It seems not possible easily to me with preserving current constraints:
> "phpunit/phpunit": ">4.8.20 <6.0"

However, it's easy to support both `5.4+` and `6.0` due to [forward-compatible changes](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-5.4.md#added).

For my projects I'm going to automate migration in scope of `php-cs-fixer` project first in order to reduce manual work (most massive changes are `PHPUnit_Framework_TestCase` -> `phpunit\framework\TestCase` and `getMock` -> `createMock` replacements). But it could be done pretty easily manually as well if you are in a hurry.

DavertMik:
[...]

PR #4142, opened 2017-04-18
Add support for PHPUnit 6 using class_alias in shim.php

MontealegreLuis:
I added support for PHPUnit 6 using `class_alias` as discussed in #3995 

* Tests are passing in Travis https://travis-ci.org/MontealegreLuis/Codeception/builds/223307775
* I had to include `PHPUnit_Util_Log_JSON`, `PHPUnit_Util_Log_TAP` and `PHP_Util_String` in `phpunit5-loggers.php` given they were removed in PHPUnit 6.
* There's a new method in the PHPUnit listener `addWarning`, so there are now 2 class definitions in `Codeception\PHPUnit\Listener`, one for PHPUnit 5 and another for 6, which are loaded conditionally.
* Method `setExpectedException` was removed in PHPUnit 6 from `PHPUnit\Framework\TestCase` so I added it back to the classes that inherit from `TestCase`. It calls the parent method if it exists, otherwise it will call the new methods `expectException`, `expectExceptionMessage`, and `expectExceptionCode`.
* `UnitCept` is checking for an exception class name so there's one more if statement there.

Thanks in advance!


composer.json
```diff
@@ -18,8 +18,9 @@
         "php": ">=5.4.0 <8.0",
         "ext-json": "*",
         "ext-mbstring": "*",
-        "phpunit/phpunit": ">4.8.20 <6.0",
-        "phpunit/php-code-coverage": ">=2.2.4 <5.0",
+        "phpunit/phpunit": ">4.8.20 <6.1",
```
Naktibalda: <7.0 should be fine, right?

composer.json
```diff
@@ -18,8 +18,9 @@
...
+        "phpunit/php-code-coverage": ">=2.2.4 < 4.1 || ^5.0",
+        "phpunit/phpunit-mock-objects": ">2.3 <3.5 || ^4.0",
```
Naktibalda: Have you got a reason to exclude php-code-coverage 4.1 and phpunit-mock-objects 3.5?

composer.json
```diff
@@ -31,7 +32,7 @@
         "symfony/css-selector": ">=2.7 <4.0",
         "symfony/dom-crawler": ">=2.7.5 <4.0",
         "behat/gherkin": "~4.4.0",
-        "sebastian/comparator": "~1.1",
+        "sebastian/comparator": ">1.1 <1.2 || ^2.0",
```
Naktibalda: Same here, why are you excluding 1.2?

shim.php
```diff
@@ -10,6 +10,44 @@ function toXPath($cssExpr, $prefix = 'descendant-or-self::') {
     }
 }
 
+// Add aliases for PHPUnit 6
+
+namespace {
+    if (version_compare(PHP_VERSION, '7.0.0') >= 0 && class_exists('PHPUnit\Framework\TestCase')) {
```
Naktibalda: I think that `if (!class_exists('PHPUnit_Framework_TestCase') && class_exists('PHPUnit\Framework\TestCase'))` condition would be more accurate.

src/Codeception/PHPUnit/Listener.php
```diff
@@ -10,119 +11,246 @@
...
 use Symfony\Component\EventDispatcher\EventDispatcher;
 
-class Listener implements \PHPUnit_Framework_TestListener
-{
-    /**
-     * @var \Symfony\Component\EventDispatcher\EventDispatcher
-     */
-    protected $dispatcher;
+// PHPUnit 6 definition for the TestListener class includes a new method `addWarning`
+if (interface_exists('PHPUnit\Framework\TestListener')) {
```
Naktibalda: It is an overkill to provide 2 versions of the class, simply add `addWarning` method, it doesn't matter if it is missing in the earlier versions of the interface.

src/Codeception/Test/Unit.php
```diff
@@ -75,6 +75,21 @@ protected function _after()
     }
 
     /**
+     * If the method exists (PHPUnit 5) forward the call to the parent class, otherwise
+     * call `expectException` instead (PHPUnit 6)
+     */
+    public function setExpectedException($exception, $message = '', $code = null)
+    {
+        if (class_exists('PHPUnit\Framework\TestCase')) {
```
Naktibalda: `method_exists($this, 'expectException')` is more accurate too, isn't it?

src/Codeception/Test/Unit.php
```diff
@@ -75,6 +75,21 @@ protected function _after()
...
+            $this->expectException($exception);
+            $message !== '' && $this->expectExceptionMessage($message);
+            $code !== null && $this->expectExceptionCode($code);
```
Naktibalda: Please don't do this, use if statement instead:
```php
if ($message !== '') {
    $this->expectExceptionMessage($message);
}
```

tests/unit/Codeception/Lib/Driver/MysqlTest.php
```diff
@@ -51,6 +51,21 @@ public function tearDown()
         }
     }
 
+    /**
+     * If the method exists (PHPUnit 5) forward the call to the parent class, otherwise
+     * call `expectException` instead (PHPUnit 6)
+     */
+    public function setExpectedException($exception, $message = '', $code = null)
```
Naktibalda: Ugh, you added this method to 8 test files.
The most elegant solution would be to migrate tests (all?) from PhpUnit_Framework_Testcase to `Codeception\Test\Unit`, but more pragmatic solution would be to add a trait.

composer.json
MontealegreLuis: I wasn't sure about this one, I read this comment on the issue https://github.com/Codeception/Codeception/issues/3995#issuecomment-284078436 and I tried to use a small range for the versions. If that's not a problem for you all, I can use a wider set of versions then.

composer.json
MontealegreLuis: Same reason for this one. What would be your preferred set of values for these versions?

shim.php
MontealegreLuis: I found some weird behaviour when running the tests on PHP 5.6, which installs PHPUnit 5.7.*, I think. When I tried with the condition `class_exists('PHPUnit\Framework\TestCase')` it was returning true, which is incorrect, because this class is only present in PHPUnit 6, which can only be installed in PHP 7. I think it's because the class can be autoloaded. It's probably safe to check for PHP's version only.

tests/unit/Codeception/Lib/Driver/MysqlTest.php
MontealegreLuis: There are some test classes not inheriting from `Codeception\Test\Unit`. That's why I didn't think of moving that method up. I think the trait is better. Unless you think that all those test should extend `Codeception\Test\Unit`.

src/Codeception/Test/Unit.php
MontealegreLuis: Ok, Ill change it.

src/Codeception/Test/Unit.php
MontealegreLuis: Yes, you're right I'll make the change.

src/Codeception/PHPUnit/Listener.php
MontealegreLuis: Yeah, now that I see it a second time, the solution was way simpler. I'll change it,

composer.json
MontealegreLuis: I was just being cautious, as in the other ones. Please let me know what range would work better.

composer.json
Naktibalda:
```
"phpunit/php-code-coverage": ">=2.2.4 < 6.0",
"phpunit/phpunit-mock-objects": ">2.3 <5.0",
```

composer.json
Naktibalda: `~1.1` actually allowed 1.2.* but you forbidden it.
```
">1.1 <3.0"
```

shim.php
Naktibalda: It was a `forward compatibility layer`, wasn't it? that's why you need to check `PHPUnit_Framework_TestCase`.

tests/unit/Codeception/Lib/Driver/MysqlTest.php
Naktibalda: `Codeception\Test\Unit` extends `\PHPUnit_Framework_TestCase`, so changing parent class should be straightforward. Please to update all tests if it doesn't cause any issues.

composer.json
```diff
@@ -18,9 +18,9 @@
         "php": ">=5.4.0 <8.0",
         "ext-json": "*",
         "ext-mbstring": "*",
-        "phpunit/phpunit": ">4.8.20 <6.1",
-        "phpunit/php-code-coverage": ">=2.2.4 < 4.1 || ^5.0",
-        "phpunit/phpunit-mock-objects": ">2.3 <3.5 || ^4.0",
+        "phpunit/phpunit": ">4.8.20 <7.0",
+        "phpunit/php-code-coverage": ">=2.2.4 < 6.0",
```
Naktibalda: Please delete that space between `<` and `6`

MontealegreLuis: Done.

src/Codeception/Test/Unit.php
```diff
@@ -75,6 +75,25 @@ protected function _after()
     }
 
     /**
+     * If the method exists (PHPUnit 5) forward the call to the parent class, otherwise
+     * call `expectException` instead (PHPUnit 6)
+     */
+    public function setExpectedException($exception, $message = '', $code = null)
+    {
+        if (method_exists($this, 'expectException')) {
+            $this->expectException($exception);
```
DavertMik: We can show a deprecation method warning using https://github.com/Codeception/Codeception/blob/2.2/src/Codeception/Lib/Notification.php#L13

Does it make sense?

MontealegreLuis: Thanks to all of you. I love this project! Yes, it makes sense. Is `PHPUnit\Framework\TestCase::setExpectedException deprecated in favor of expectException, expectExceptionMessage, and expectExceptionCode` a good message to use? This warning should only be shown when using PHPUnit 6 (not in the else of this condition)? Or it should be shown always (before the condition)?

Naktibalda: @DavertMik it would show a lot of warnings for Codeception test suite, and we can't switch until we drop support for PHPUnit 4.

orlangur:
Very bad implementation approach :smile: 

By the way, any idea why test suite is not compatible with PHPUnit 5.4? https://travis-ci.org/orlangur/Codeception/jobs/208363815

Naktibalda:
@orlangur Please try 5.7

orlangur:
@Naktibalda of course I can try... I'm trying to say that there are big holes in supported PHPUnit versions even before PHPUnit6 support with new hardcode.

So, it's better to drop PHPUnit 4, PHP <5.6 support and do some grooming instead of increasing fragility.

Naktibalda:
Thanks for telling us.

We will do that in 2.3 or 3.0, whatever comes first.

src/Codeception/Test/Unit.php
```diff
@@ -75,6 +75,25 @@ protected function _after()
     }
 
     /**
+     * If the method exists (PHPUnit 5) forward the call to the parent class, otherwise
+     * call `expectException` instead (PHPUnit 6)
+     */
+    public function setExpectedException($exception, $message = '', $code = null)
+    {
+        if (method_exists($this, 'expectException')) {
```
Naktibalda: This code does not match comments.
The matching code would be
```php
if (is_callable('parent::setExpectedException')) {
  parent::setExpectedException($exception, $message, $code);
} else {
 //...
}
```

src/Codeception/Test/Unit.php
Naktibalda: @MontealegreLuis I think that the warning should be shown if `expectException` exists.

Naktibalda:
@MontealegreLuis As you may have noticed, your work is only a temporary measure.
A large part of it will have to be removed once we drop support for PHP <5.6 and PHPUnit <5.7
All use cases of shimmed classes will be replaced with new names and all deprecated method calls will be replaced with new method calls.
Would you like to do that as well?

MontealegreLuis:
Sure thing, I'd love to do that!!

src/Codeception/Test/Unit.php
MontealegreLuis: Ok @Naktibalda will do that.

src/Codeception/Test/Unit.php
MontealegreLuis: I'll change it.

enumag:
This looks like a good compromise until Codeception 3 is ready. Looking forward to see this in a stable release.

src/Codeception/Test/Unit.php
MontealegreLuis: This one and the deprecation warning are done.

DavertMik:
This PR will be part of Codeception 2.3
There still can be issues with PHP 6,0 support so If they happen we encourage to restrict to phpunit ^5.7.0 manually in composer.json

GawainLynch:
I see 2.3.0 was just released, but without the changes to [`composer.json`](https://github.com/Codeception/Codeception/blob/2.3.0/composer.json#L21) to re-enable the higher constraint… will this be a 2.4 target now?

Naktibalda:
@DavertMik This was a driver for 2.3.0 and you missed it. Can it go out in 2.3.1?

DavertMik:
No, it was announced for 2.3 so it should be there. Looks like there was a
merge which reverted the constraints. I will update it today

23 трав. 2017 р. 07:25 "Gawain Lynch" <notifications@github.com> пише:

PR #4143, opened 2017-04-19
FIX log path override

kusnir:
When you run:
`php vendor\codeception\codeception\codecept run -vvv --override="paths: log: tests/_output/Acceptance_TestCest_test" --xml=report.xml tests\acceptance\TestCest:test`
it generates report and error screenshots to tests/_output and not to tests/_output/Acceptance_TestCest_test

... and override by configuration also does not work - this commit fix only override by parameter. Try add to tests/acceptance.suite.yml:
```yml
env:
    development:
        paths:
            log: tests/_output/runner1
```
and run:
`php vendor\codeception\codeception\codecept run -vvv --env=development --xml=report.xml tests\acceptance\TestCest:test`
it generates report and error screenshots to tests/_output and not to tests/_output/Acceptance_TestCest_test

### DIFF
--- a/src/Codeception/Configuration.php
+++ b/src/Codeception/Configuration.php
@@ -570,7 +570,14 @@ class Configuration
      */
     public static function append(array $config = [])
     {
-        return self::$config = self::mergeConfigs(self::$config, $config);
+        self::$config = self::mergeConfigs(self::$config, $config);
+
+        self::$logDir = self::$config['paths']['log'];
+        self::$dataDir = self::$config['paths']['data'];
+        self::$supportDir = self::$config['paths']['support'];
+        self::$testsDir = self::$config['paths']['tests'];
+
+        return self::$config;
     }
 
     public static function mergeConfigs($a1, $a2)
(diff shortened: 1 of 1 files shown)


src/Codeception/Configuration.php
```diff
@@ -570,7 +570,22 @@ public static function isEmpty()
      */
     public static function append(array $config = [])
     {
-        return self::$config = self::mergeConfigs(self::$config, $config);
+        self::$config = self::mergeConfigs(self::$config, $config);
+        
+        if (isset(self::$config['paths']['log'])) {
```
DavertMik: I think this checks are not needed as this config should be available by default

kusnir: Should, but without them some checks were not successful ... probably in some test cases are not defined all paths (see first commit checks)

kusnir: I was thinking about overriding by environment and it is not good idea because the results (HTML/XML reports) will be divided to separate folders by suites and I do not think it is desirable state. I think, user wants to set log path for run and not for suite. And for run can by path set by codeception.yml or --override parameter.

DavertMik:
Ok, thanks!

PR #4146, opened 2017-04-19
Display the location of unexpected exceptions

mytskine:
When an unexpected exception occured, there was no indication of the
file where it was raised. This commit displays the file and line number
in the console report, just under the exception name.

See #4145.

I could not run Codecept's test, because it complained of a missing mongodb extension, though `php -i` confirms that ext/mongo is available. I did not want to spend too much time on this.

### DIFF
--- a/src/Codeception/Subscriber/Console.php
+++ b/src/Codeception/Subscriber/Console.php
@@ -394,6 +394,7 @@ class Console implements EventSubscriberInterface
 
         if (!$isFailure) {
             $message->prepend("[$class] ")->block('error');
+            $message->append("\n" . OutputFormatter::escape($e->getFile()) . ":" . $e->getLine());
         }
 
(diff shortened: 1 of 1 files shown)


DavertMik:
Could you show me how this error message currently looks in console?

mytskine:
As noted in the related issue, this is very probably Yii2-specific, so I'm not sure this PR is still valid in the general case. Anyway, it doesn't change much, so I could be useful in any case. It adds one line under the red "exception" error block. Here is a picture (where the path is hidden), with this new suplementary line. Please note the wrong stack trace at the bottom.

codeception_yii2_exception-trace

Naktibalda:
I think that this change is not necessary.

Naktibalda:
I proposed a different way to improve `yii\base\ErrorException` in original issue.

Issue #2054, opened 2015-06-16
Bus error (core dumped) error


saiger:
Sometimes I got "Bus error (core dumped)" message, my steps:
- Split tests and run it by Robo using your manual(http://codeception.com/docs/12-ParallelExecution),
  sometimes one of thread is terminating wth next output info:

"[ParallelExec] Output for  vendor/bin/codecept run --group p8 --env p8 acceptance 
[12:07:45][Step 3/4] Codeception PHP Testing Framework v2.0.12
[12:07:45][Step 3/4] Powered by PHPUnit 4.5.1 by Sebastian Bergmann and contributors.
[12:07:45][Step 3/4] [Groups] p8 
[12:07:45][Step 3/4] 
[12:07:45][Step 3/4] Bus error (core dumped)"

Codeception version: 2.0.12

aleksandrzen:
+1

jiexaspb:
+1

aleksandrzen:
@DavertMik, hello, do you have any news about this bug?

saiger:
Fixed

I started run from one folder, need to do next:
cd folder1 && codecepction run p1
cd folder2 && codeception run p2

EvgeniySpinovSell:
I'm having the same issue with robo + codeception 2.1.1

But I didn't get the fix. I need to enter random directories before starting codeception? What if I have 20 parallel threads?

ivanfetischev:
You need run every thread from its directory

EvgeniySpinovSell:
That is like workaround, but not sure it is a proper fix. In my case tests doesn't have spare directories. So I'm like should traverse through folder tree and just enter any random directory of the project?

And is it a codeception or robo issue?

ivanfetischev:
> just enter any random directory of the project
No

> And is it a codeception or robo issue?
It is not trouble of codeception or robo.

When you running codeception in many threads, every thread trying to create file AcceptanceTester.php, if you running 20 parallel threads in one directory, 20 processes trying to create one file. If you create directory for every thread and run every thread from its directory, it helps to solve problem

EvgeniySpinovSell:
Thanks for detailed explanation, Ivan. Will have to workaround it with temp directories during tests. Probably this case should be reflected in docs for parallel testing.

EvgeniySpinovSell:
Suggested solution doesn't seems to be working. Let me describe the setup a bit:
- I create a folder p1, p2, p3 .... for each test group
[...]

PR #4148, opened 2017-04-19
[WIP] Generate Actions class in a thread-safe way

pitpit:
It intends to fix following issue : https://github.com/Codeception/Codeception/issues/2054

It writes content on a temporary locked file before erasing the main class file (less non thread-safe disk instructions).

I need some concerned people to test it and confirm `Bus Errors` do not occur anymore.
Sorry it's it very empirical...

On your project, to quickly switch to this branch :

```
composer config repositories.codeception/codeception git git@github.com:pitpit/Codeception.git
composer require codeception/codeception:dev-bugfix/bus-error
```

### DIFF
--- a/src/Codeception/Command/Shared/FileSystem.php
+++ b/src/Codeception/Command/Shared/FileSystem.php
@@ -66,8 +66,8 @@ trait FileSystem
             return false;
         }
 
-        //writing in a temporary file in order to reduce time between the file acquisition
-        //and the wrting, it should recude risk to occur `Bus Error` when running tests in parallel.
+        //Write in a locked temporary file in order to reduce non thread-safe disk instructions,
+        //  and reduce risk of `Bus Errors` when running tests in parallel.
         //@see https://github.com/Codeception/Codeception/issues/2054#event-1046859271
         $tmpFilename = $filename . '.tmp';
         $success = @file_put_contents($tmpFilename, $contents, LOCK_EX);
--- a/src/Codeception/Command/Shared/FileSystem.php
+++ b/src/Codeception/Command/Shared/FileSystem.php
@@ -82,9 +82,6 @@ trait FileSystem
             for ($i = 0; ($i < 3 && !$success); $i++) {
                 usleep(500000);
                 $success = @rename($tmpFilename, $filename);
-                if ($success) {
-                    break;
-                }
             }
             if (false === $success) {
                 throw new \RuntimeException(sprintf('Unable to move class from %s to %s(tried 3 times)', $tmpFilename, $filename));
--- a/src/Codeception/Command/Shared/FileSystem.php
+++ b/src/Codeception/Command/Shared/FileSystem.php
@@ -66,18 +66,29 @@ trait FileSystem
             return false;
         }
 
-        //Write in a locked temporary file in order to reduce non thread-safe disk instructions,
-        //  and reduce risk of `Bus Errors` when running tests in parallel.
+        //Write in a temporary file in order to reduce non thread-safe disk instructions,
+        //and reduce risk of `Bus Errors` when running tests in parallel.
         //@see https://github.com/Codeception/Codeception/issues/2054#event-1046859271
-        $tmpFilename = $filename . '.tmp';
-        $success = @file_put_contents($tmpFilename, $contents, LOCK_EX);
+        $tmpFilename = $filename . '.' . uniqid() . '.tmp';
+        $success = @file_put_contents($tmpFilename, $contents);
         if (false === $success) {
+            //no space left on device, IO errors...
             throw new \RuntimeException(sprintf('Unable to generate temporary class %s', $tmpFilename));
         }
 
         $success = @rename($tmpFilename, $filename);
         if (false === $success) {
-            throw new \RuntimeException(sprintf('Unable to generate class %s', $filename));
+            //wait 0,5s and try 3 times before throwing an excception
+            for ($i = 0; ($i < 3 && !$success); $i++) {
+                usleep(500000);
+                $success = @rename($tmpFilename, $filename);
+                if ($success) {
+                    break;
+                }
+            }
+            if (false === $success) {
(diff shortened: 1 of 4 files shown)


DavertMik:
I didn't encounter it so I don't know how to reproduce this error.

src/Codeception/Command/Shared/FileSystem.php
```diff
@@ -60,12 +60,26 @@ protected function removeSuffix($classname, $suffix)
...
     {
         if (file_exists($filename) && !$force) {
             return false;
         }
-        file_put_contents($filename, $contents, $flags);
+
+        //Write in a locked temporary file in order to reduce non thread-safe disk instructions,
+        //  and reduce risk of `Bus Errors` when running tests in parallel.
+        //@see https://github.com/Codeception/Codeception/issues/2054#event-1046859271
+        $tmpFilename = $filename . '.tmp';
```
ironsmile: I would've used [tempnam](https://secure.php.net/manual/en/function.tempnam.php) instead of concatenating the constant `.tmp`.

ironsmile:
I am starting to use this branch in my CI server. Would report my results with it if anything interesting pops up.

src/Codeception/Command/Shared/FileSystem.php
pitpit: you're right about using a ramdom name, to avoid collision but `tempnam` create the file on the disk, which is not-needed (file_put_contents actually create the file).

I'm upgrading this code to use uniqid() in $tmpFilename

ironsmile: Maybe you're confusing [tempnam](https://secure.php.net/manual/en/function.tempnam.php) with [tmpfile](https://secure.php.net/manual/en/function.tmpfile.php). The latter is the one which creates a file and returns a resource. The former just returns a unique filename as a string. No file is created at this point. Which is exactly what you are trying to do with `uniqid`.

pitpit:
It looks like I don't have `Bus error` anymore but sometimes Codeception\Command\Shared\FileSystem::save throws an exception when renaming file to target destination probably because of an IO collision.  I rewrite the code a bit, to retry 3 times before definitely throwing the exception.

@ironsmile Could you upgrade the branch on your CI server to this new version ?

ironsmile:
My CI server would always get the latest commit in the branch. So yes, I would try the new version.

ironsmile:
Well, the exception did happen:

```
+ ../vendor/bin/robo parallel:all-static output/codecept-default.junit.xml
 [ParallelExec] ../vendor/bin/codecept run --skip-group api --group parallel_1 --xml paracept/_results/result_1.xml acceptance
 [ParallelExec] ../vendor/bin/codecept run --skip-group api --group parallel_2 --xml paracept/_results/result_2.xml acceptance
 [ParallelExec] ../vendor/bin/codecept run --skip-group api --group parallel_3 --xml paracept/_results/result_3.xml acceptance
 [ParallelExec] ../vendor/bin/codecept run --skip-group api --group parallel_4 --xml paracept/_results/result_4.xml acceptance
 [ParallelExec] ../vendor/bin/codecept run --skip-group api --group parallel_5 --xml paracept/_results/result_5.xml acceptance
 [ParallelExec] Output for ../vendor/bin/codecept run --skip-group api --group parallel_3 --xml paracept/_results/result_3.xml acceptance:

Codeception PHP Testing Framework v2.2.10
Powered by PHPUnit 4.8.31 by Sebastian Bergmann and contributors.
[Groups] parallel_3 

 [ParallelExec]  
                                                                               
  [RuntimeException]                                                           
  Unable to move class from /raid1/jenkins/workspace/Tests/te  
  sts/./support/_generated/AcceptanceTesterActions.php.58f9b08d626ea.tmp to /  
  raid1/jenkins/workspace/Tests/tests/./support/_generated/Ac  
  ceptanceTesterActions.php(tried 3 times)      
```
I find this approach not good enough. We're trading the SIGBUS for the equally likely RuntimeError. The end result is the same - randomly failing tests.

pitpit:
Hi @ironsmile, in our side we don't have `Bus Error` anymore. Neither RuntimeError.
Are you sure you composer.lock do not require previous commit on my branch ?

Anyway, I'm looking for another approach for a safer fix

ironsmile:
Yes, I can confirm the run in question was using the latest commit - bd00be3. Here is an excerpt from the `composer install`:

```
- Installing codeception/codeception (dev-bugfix/bus-error bd00be3): Cloning bd00be3c16
```

Our CI always does a clean install before tests and because of that there is no `composer.lock` and `vendor/` directory at all.

pitpit:
sorry I did not have time to rework this PR.
For my part, we are using this fix since 1 month and `Bus error` never happened again. We neither faced RuntimeException when creating the Action class while Codeception tests are ran around 20 times a day.

@ironsmile Did you observe more RuntimeException than previous Bus Error ? Or is it worst ?

@DavertMik I think we could merge it waiting for a better way to solve this issue.

ironsmile:
@pitpit, I cannot judge the frequency of RuntimeExceptions against the Bus errors. We've reverted back to the stable version after the first few RuntimeExceptions.

DavertMik:
@ironsmile maybe you can suggest how this fix can be improved? 
according to your words it's better not to merge it right now...

ironsmile:
My suggestion would be to altogether abandon generating source code in files and subsequently running this code. This approach is bound to have problems when there are more than one Codeception processes  running at a time. There is no way I can think of to guarantee that one process would not start to read the file while an other one is in the middle of writing in it.

That being said, it might be too hard to re-design Codeception in such a way. Then an alternative approach would be to make sure the generated file always have a different name for different processes. This might be achieved by including the process id (PID) in the name of the file. As far as I know there is no OS which would allow for two processes with the same PID to coexist and thus it pretty much is a unique number in any point in time.

pitpit:
closed in favor of --no-rebuild parameter https://github.com/consolidation/Robo/pull/603

PR #4149, opened 2017-04-19
[WebDriver] Restart browser on _reconfigure

### LINKED ISSUE #1192: _reconfigure not works as expected in WebDriver
http://codeception.com/docs/06-ModulesAndHelpers#Runtime-Configuration suggests

``` terminal
Runtime Configuration
If you want to reconfigure module in run time, you can use the _reconfigure method of the module. You may call it from helper class and pass in all the fields you want to change.
```

``` php
<?php
$this->getModule('WebDriver')->_reconfigure(array('browser' => 'chrome'));
?>
```

but this is not changing anything when called from a helper class (too late to configure browser), when the helper class is loaded the browser is already launched.

See #1188 for more details.

Naktibalda:
Fixes #1192

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -292,6 +292,8 @@ class WebDriver extends CodeceptionModule implements
     protected $sslProxy;
     protected $sslProxyPort;
 
+    private $reconfigured = false;
+
     /**
      * @var RemoteWebDriver
      */
@@ -341,6 +343,7 @@ class WebDriver extends CodeceptionModule implements
     public function _reconfigure($config)
     {
         parent::_reconfigure($config);
+        $this->reconfigured = true;
         $this->webDriver->quit();
         $this->_initialize();
         $this->_initializeSession();
@@ -377,7 +380,8 @@ class WebDriver extends CodeceptionModule implements
 
     public function _after(TestInterface $test)
     {
-        if ($this->config['restart']) {
+        if ($this->config['restart'] || $this->reconfigured) {
+            $this->reconfigured = false;
             $this->cleanWebDriver();
             return;
         }
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -343,6 +343,10 @@ class WebDriver extends CodeceptionModule implements
     public function _reconfigure($config)
     {
         parent::_reconfigure($config);
+        if (count($config) === 1 && array_key_exists('url', $config)) {
+            //don't restart browser if only url was changed
+            return;
+        }
         $this->reconfigured = true;
         $this->webDriver->quit();
         $this->_initialize();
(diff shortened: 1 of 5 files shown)


DavertMik:
1. If only `url` is changed there is no need to restart the browser. 
2. You can implement that in `onReconfigure` method

Should we restart browser every time or check for options other than `url` to do so? In both cases we need to mention this in documentation

Naktibalda:
1. Isn't it better to use `amOnUrl` to change base URL for a single test?
2. I've seen `onReconfigure` and didn't like it because it is called before `validateConfig`.
3. Probably I have to restart browser browser in `_after` too to restore configuration after `_reconfigure`.

Naktibalda:
Ok, I've looked at the code. A special exception for `url` is necessary.

DavertMik:
I think you don't need `reconfigured` boolean property for this case. The module itself keeps initial configuration, so it can be reverted. Did you see http://codeception.com/docs/reference/Module#_resetConfig method? 

> I've seen onReconfigure and didn't like it because it is called before validateConfig.

True. Actually it checks for required fields only and throws an exception if some fields are missing. 
I think this should not happen as we already have all required fields preconfigured and you can't remove them, just replace.

Naktibalda:
The alternative is to compare 2 arrays and comparing 2 nested arrays is more complicated than using Boolean. Should I do that?

src/Codeception/Module/WebDriver.php
```diff
@@ -338,6 +340,19 @@ public function _before(TestInterface $test)
...
 
+    public function _reconfigure($config)
+    {
+        parent::_reconfigure($config);
+        if (count($config) === 1 && array_key_exists('url', $config)) {
+            //don't restart browser if only url was changed
+            return;
+        }
+        $this->reconfigured = true;
+        $this->webDriver->quit();
```
DavertMik: I found an edge case here.

If we have a chrome browser running and `restart: false`.
When we are changing browser to firefox and restarting browser in all next tests we will have firefox running.
While `_reconfigure` should rollback its changes in the end of a test.

tests/support/WebHelper.php
```diff
@@ -5,4 +5,8 @@
 
 class WebHelper extends \Codeception\Module
 {
+    public function changeBrowser($browser)
+    {
+        $this->getModule('WebDriver')->_reconfigure(['browser' => $browser]);
```
DavertMik: I think we should better add another method to restart a browser. 

```php
$this->getModule('WebDriver')->_restart(); // just restart
$this->getModule('WebDriver')->_restart(['browser' => $browser]); // reconfigure + restart
```

Naktibalda: It's ok with me, but I would like to pass it to you as you have a better idea about implementation.

DavertMik: Ok

PR #4151, opened 2017-04-20
Adds doc about different manners to POST files

igorsantos07:
This has been [noted in SO](http://stackoverflow.com/questions/28437633/send-file-to-restful-service-in-codeception) years ago, but there's still no docs on this, so...

### DIFF
--- a/src/Codeception/Module/REST.php
+++ b/src/Codeception/Module/REST.php
@@ -303,11 +303,14 @@ EOF;
     /**
      * Sends a POST request to given uri.
      *
-     * Parameters and files (as array of filenames) can be provided.
+     * Parameters and files can be provided separately.
      *
      * @param $url
      * @param array|\JsonSerializable $params
-     * @param array $files
+     * @param array $files A list of filenames or "mocks" of $_FILES (each entry being an array with the following
+     *                     keys: name, type, error, size, tmp_name (pointing to the real file path)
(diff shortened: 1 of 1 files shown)


src/Codeception/Module/REST.php
```diff
@@ -303,11 +303,33 @@ public function amBearerAuthenticated($accessToken)
...
-     * Parameters and files (as array of filenames) can be provided.
+     * Parameters and files can be provided separately.
      *
      * @param $url
      * @param array|\JsonSerializable $params
-     * @param array $files
+     * @param array $files A list of filenames or "mocks" of $_FILES (each entry being an array with the following
+     *                     keys: name, type, error, size, tmp_name (pointing to the real file path). Each key works
+     *                     as the "name" attribute of a file input field.
+     * @example
```
DavertMik: Please add this example before `@annotations`
Please use the markdown code formatting as in other docblocks with examples. Current `@example` tag is not supported by our publishing tool and will be broken on a website

igorsantos07: hmmm ok. could you point me to any place with an example of those examples (aha)? I looked quickly around, saw none so I wrote the best I could think of.

igorsantos07: hope it works now :)
(can we mark any review comment as "solved"? this sounds too advanced for GitHub uh?)

Issue #3626, opened 2016-10-20
$scenario->current('name') is everytime empty


rebas82:
in my cest test i use
$scenario->current('env')
this working perfectly, but when i use
$scenario->current('name')  
this is everytime empty :-(

v.2.2.5

Naktibalda:
Why do you need it?
One way to fix this issue is to remove `$scenario->current('name');` from the docs, have you got a good reason for a fix?

rebas82:
I doing to HTML report fail info with link and screenshot (by comment()  and hook Fail). I will like use saved file from _output (. PNG and. Html),  this file contains: class and test name and string fail.  I know how get class, and don't have test name 😎.  When I don't have test name, I must use makesreenshot() with my own name from_output/debug

Naktibalda:
Do you mean `public function _failed(\Codeception\TestInterface $test, $fail) {}`

Then you can get the test name from `$test->getMetadata()->getName()`

Naktibalda:
Fixed

PR #4154, opened 2017-04-20
Fixed $scenario->current('name')

### LINKED ISSUE #4144: $scenario->current('name') returns Null
In my StepObject I want to make a screenshot every time before I click on the login-Button for debugging reasons
I have several tests which are doing the login action so in order to get a screenshot for every single test and to know from which test that screenshot is I want to include the name of the test to the name
Currently it looks like this

```
$I = $this; 
$scenario = $I->getScenario();
 $name = $scenario->current('name');
 $I->makeScreenshot('login_Data_' . $name);

```

it looks like current('name') always returns NULL so the name is not added and the screenshot is getting overwritten every time a new one is taken.

Naktibalda:
Fixes #4144 and #3626

### DIFF
--- a/src/Codeception/Test/Metadata.php
+++ b/src/Codeception/Test/Metadata.php
@@ -94,7 +94,7 @@ class Metadata
                 return $this->current[$key];
             }
             if ($key === 'name') {
-               return $this->getName();
+                return $this->getName();
             }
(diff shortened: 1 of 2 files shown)


DavertMik:
Thank you

PR #4156, opened 2017-04-22
[AMQP] Documented parameters of declareQueue, declareExchange and bin…

Naktibalda:
Documented parameters of declareQueue, declareExchange and bindQueueToExchange

Also created a test for these methods.


DavertMik:
Thank you

PR #4168, opened 2017-04-25
make WebDriver:findClickable() public, so that it can be used in Helpers

tiger-seo:


DavertMik:
Please add documentation for `findClickable` similar to https://github.com/Codeception/Codeception/blob/2.2/src/Codeception/Lib/Interfaces/ElementLocator.php#L7 `findClickable` has signature different than `findElements` , it requires context element first so usage example should be added.

In the same manner please make `findCheckable` as public.

findClickable and findCheckable should not be added to ElementLocator interface, as they require WebDriver-only argument.

DavertMik:
Doooooocumentation!

src/Codeception/Module/WebDriver.php
```diff
@@ -821,11 +821,21 @@ public function click($link, $context = null)
...
+     * Locates clickable element.
+     *
+     * Use it in Helpers or GroupObject or Extension classes:
+     *
+     * ```php
+     * <?php
+     * $module = $this->getModule('WebDriver');
+     * $page = $module->webDriver;
+     * $el = $module->_findClickable($page, 'Click Me');
+     *
```
Naktibalda: You missed ``` here.

DavertMik: @tiger-seo please do better documentation next time

DavertMik:
Ok, I will fix docs locally

PR #4171, opened 2017-04-26
Add ability to load the dump using a tool/command

brutuscat:
The db module now accepts a `dumptool` configuration item
where the user can enter a command to execute his/her prefered
tool or wrapper script.
The dsn is parsed into variables and are made available, with
the variables from the module configuration, to the command
for interpolation, allowing the executing of commands like:

`dumptool: 'mysql -u $user -p $password -h $host -D $dbname < $dump'`

The Db module has been structured a bit more in order to make sense of the suite & test events rather than having spaghetti like code all around the place. 

The sqlite module has been cleaned up to respect the reconnect
settings and to aviod dependency on the order of the execution
of its parent method. Mainly because there is no need to "reload"
the connection when touching the sqlite db file...

### DIFF
--- a/src/Codeception/Lib/Driver/Sqlite.php
+++ b/src/Codeception/Lib/Driver/Sqlite.php
@@ -8,7 +8,6 @@ class Sqlite extends Db
 {
     protected $hasSnapshot = false;
     protected $filename = '';
-    protected $con = null;
 
     public function __construct($dsn, $user, $password)
     {
@@ -22,19 +21,24 @@ class Sqlite extends Db
         parent::__construct($this->dsn, $user, $password);
     }
 
+    public function __destruct()
+    {
+        file_put_contents($this->filename, '');
+    }
+
     public function cleanup()
     {
-        $this->dbh = null;
-        file_put_contents($this->filename, '');
-        $this->dbh = self::connect($this->dsn, $this->user, $this->password);
+        $this->dbh->exec('PRAGMA foreign_keys = OFF;');
+        $this->dbh->exec('PRAGMA writable_schema = 1;');
+        $this->dbh->exec('DELETE FROM sqlite_master WHERE type IN (\'table\', \'index\', \'trigger\');');
+        $this->dbh->exec('PRAGMA writable_schema = 0;');
+        $this->dbh->exec('PRAGMA foreign_keys = ON;');
     }
 
     public function load($sql)
     {
         if ($this->hasSnapshot) {
-            $this->dbh = null;
             file_put_contents($this->filename, file_get_contents($this->filename . '_snapshot'));
-            $this->dbh = new \PDO($this->dsn, $this->user, $this->password);
         } else {
             if (file_exists($this->filename . '_snapshot')) {
                 unlink($this->filename . '_snapshot');
--- a/src/Codeception/Module/Db.php
+++ b/src/Codeception/Module/Db.php
@@ -71,9 +71,33 @@ use Codeception\TestInterface;
  *              populate: true
  *              cleanup: false
  *              reconnect: true
+ *              dumptool: 'mysql -u $user -p $password -h $host -D $dbname < $dump'
  *
  * ## SQL data dump
  *
+ *  You now have two ways of loading the dump into your database. You could use the legacy way,
+ *  where the codeception program reads the dump file and passes it to the specific driver so it populates the
+ *  database (this method is usually slow in particular if your dump is "big") or you could now
+ *  configure a dump tool command (e.g. using the mysql client binary) that, after parsing the PDO
+ *  dsn string, gives you access to some placeholder variables for a very simple command completion
+ *  template.
+ *
+ *  Usage of the `dumptool` is simple, just set the command that you want to use to load the dump
+ *  into the database with the parsed variables from the config and/or dsn string. All the configuration
+ *  options will also be available and take priority over the dsn.
+ *  Most dsn have a `keyword=value` format, you should expect to have a variable named as the
+ *  keyword with the full value inside it. If you need further parsing you could write your own
+ *  script to wrap the real dump tool.
+ *
+ *
+ *  PDO dsn elements for the supported drivers:
+ *  * MySQL: [PDO_MYSQL DSN](https://secure.php.net/manual/en/ref.pdo-mysql.connection.php)
+ *  * SQLite: [PDO_SQLITE DSN](https://secure.php.net/manual/en/ref.pdo-sqlite.connection.php)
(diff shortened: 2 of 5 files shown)


src/Codeception/Module/Db.php
```diff
@@ -288,6 +330,93 @@ protected function cleanup()
 
     protected function loadDump()
     {
+        if ($this->config['dumptool']) {
+            $this->loadDumpUsingTool();
```
DavertMik: could you move that to separate class inside `Lib/Driver/DumpTool`?

src/Codeception/Module/Db.php
```diff
@@ -288,6 +330,93 @@ protected function cleanup()
...
+
+        $vars = array_merge($dsnVars, $config);
+        foreach ($vars as $key => $value) {
+            $vars['$'.$key] = $value;
+            unset($vars[$key]);
+        }
+        return str_replace(array_keys($vars), array_values($vars), $dumptool);
+    }
+
+    protected function execute($command)
```
DavertMik: `execute` sounds too general
can you rename it to something better

src/Codeception/Module/Db.php
```diff
@@ -232,28 +243,58 @@ private function disconnect()
...
 
+    public function _beforeSuite($settings = [])
+    {
+        $this->connect();
+
+        // FIXME: This needs proper structure to avoid cooking more spaghetti code
+        // since it only concerns to the "legacy" dump loading procedure.
+        // The db drivers should be revised to see if they are coupled on
+        // having the $sql variable loaded at this point in time of the execution or not...
+        if (!$this->config['dumptool']
```
DavertMik: I don't like spaghetti code as well, so please move dump loaders to corresponding classes

src/Codeception/Lib/Driver/Sqlite.php
```diff
@@ -22,19 +21,24 @@ public function __construct($dsn, $user, $password)
...
-        $this->dbh = null;
         file_put_contents($this->filename, '');
-        $this->dbh = self::connect($this->dsn, $this->user, $this->password);
+    }
+
+    public function cleanup()
+    {
+        $this->dbh->exec('PRAGMA foreign_keys = OFF;');
+        $this->dbh->exec('PRAGMA writable_schema = 1;');
+        $this->dbh->exec('DELETE FROM sqlite_master WHERE type IN (\'table\', \'index\', \'trigger\');');
```
DavertMik: Is that really needed? 

Probably Sqlite tests are failing because of this change: https://travis-ci.org/Codeception/Codeception/jobs/225994707#L2651

brutuscat: I fixed this already 😣 I'll try to see if I forgot to push the right merge or what...

src/Codeception/Module/Db.php
brutuscat: Well, I called it execute because the php function that is wrapping is `exec` any ideas?

src/Codeception/Module/Db.php
```diff
@@ -71,9 +71,33 @@
...
+ *  configure a dump tool command (e.g. using the mysql client binary) that, after parsing the PDO
+ *  dsn string, gives you access to some placeholder variables for a very simple command completion
+ *  template.
+ *
+ *  Usage of the `dumptool` is simple, just set the command that you want to use to load the dump
+ *  into the database with the parsed variables from the config and/or dsn string. All the configuration
+ *  options will also be available and take priority over the dsn.
+ *  Most dsn have a `keyword=value` format, you should expect to have a variable named as the
+ *  keyword with the full value inside it. If you need further parsing you could write your own
+ *  script to wrap the real dump tool.
```
brutuscat: Got it, will do!

src/Codeception/SuiteManager.php
```diff
@@ -75,7 +75,10 @@ public function __construct(EventDispatcher $dispatcher, $name, array $settings)
```
brutuscat: Yes, I kinda have an issue with that hahaha... Will do!

brutuscat:
@DavertMik do you think I should change the name of "dumper" to populator or something like that? Because it actually does the opposite of dumping a db, it is to populate the db with the dump.

src/Codeception/Lib/Driver/Sqlite.php
brutuscat: @DavertMik also, the "cleanup" was emptying the file `file_put_contents($this->filename, '');` and then force a reconnection to avoid a "schema changed" error. That was coupled with the Db module implementation, so after I change it to make "more sense" this coupling broke. All drivers "SQL the cleanup" all the way...

DavertMik:
@brutuscat yes, Populator sounds better to me. So use it as a namespace

DavertMik:
@brutuscat btw, I plan to include this PR in next release. I hope you will have a time soon to finish it

Naktibalda:
@brutuscat and please rename `dump tool` to `load tool` or `processor` everywhere where some variation of the term is used.

brutuscat:
@DavertMik when the release is due?

Naktibalda:
@brutuscat on last Sunday, so ASAP now, but most likely it will take another week.

DavertMik:
@brutuscat any updates? ;)

brutuscat:
@DavertMik sorry man... I'm still struggling migrating our @Vreasy test suite to [Codeception@v2.2](https://github.com/Codeception/Codeception/tree/2.2) since its pretty big and we started using this project 4 years ago :) 

I was also thinking about trying to have a simpler version of this PR, one that does not touches a lot of the legacy code. In particular, the issues with sqlite is that since we are "playing" with the actual db file, it breaks the reconnect functionality... Ideas? Suggestions?

Once I catch up with our company test suite I'll be able to deliver a proper new feature to Codeception.

DavertMik:
> sqlite is that since we are "playing" with the actual db file, it breaks the reconnect functionality... Ideas? Suggestions?

Don't support SQLite at all 🤗

I don't see why someone would use 3rd party tool to populate sqlite database if cleaning it up and reloading data is just removing and copying the file. So don't worry about that one, please make it work for Postgres and MySQL. That's important.

And please-please, finish this PR by the end of a week, as I'd really really love to see it in 2.3, on which I'm working on.

DavertMik:
@brodkin  I'm scheduling this PR for Codeception 2.3 which I plan to release next week.

brutuscat:
@DavertMik I'll give it a try today ;) I'll try to keep simple and leave the refactoring for later.

brutuscat:
Moved to #4220

PR #4172, opened 2017-04-26
Allow to set only host

SG5:
If you have enabled pretty url and try to set only host in url
```
modules:
    enabled:
      - Yii1:
          appPath: ./../../public/index-test.php
          url: http://localhost/
```
you will get warning (and exception in error handler) `strpos(): Empty needle`


DavertMik:
Thanks

Issue #4174, opened 2017-04-27
MongoDB module legacy condition result is wrong


retnek:
If on the server installed both mongo extension (legacy mongo and mongodb) the legacy check condition returns true.

in Driver\MongoDb.php 98 line:
```php
$this->legacy = class_exists('\\MongoClient') && strpos(\MongoClient::VERSION, 'mongofill') === false;
```

I think should be better solution for test extension loaded state, something like this:
```php
$this->legacy = extension_loaded('mongodb') === false;
```

Naktibalda:
Please make a pull request.

PR #4178, opened 2017-04-28
improve mongo legacy check

retnek:
Hello,

I think this is a better way for decide wich mongo have to use. If you use both legacy _mongo_ and the current _mongodb_, the legacy value will be false.

Issue #4174


Naktibalda:
I think that your change breaks MongoDb module on HHVM, please take a look at #3641 and https://github.com/mongofill/mongofill

retnek:
Good point. I fixed that.

DavertMik:
Thanks

PR #4182, opened 2017-04-30
[WebDriver] seeLink compares relative links correctly

### LINKED ISSUE #4107: seeLinks() Relative URLs Check
Hey guys,

I'd like to re-open a question stated in  [this request](https://github.com/Codeception/Codeception/issues/3463)

I've updated codeception from version 2.2.1 to 2.2.10 and got the same issue.

If it's not a bug in the code, then at least docs are misleading on this question, especially providing an example where relative URL is in the href, but still seeLink() method asserts it's found:

    $I->seeLink('Logout','/logout'); // matches <a href="/logout">Logout</a>

**In reality codeception doesn't work this way, or am I missing a point?**

An example from my setup:

 html:
```
    <li class="active"><a href="/index-test.php/master/apps">Apps</a></li>
```
this test fails:
```
    $I->seeLink('Apps', '/index-test.php/master/apps')
```

The way Codeception interprets this internally:
```
    // Found element href attribute
    ~/.composer/vendor/codeception/codeception/src/Codeception/Module/WebDriver.php:937:
string(44) "http://mydomain.dev/index-test.php/master/apps"

    // Passed $url to check
    ~/.composer/vendor/codeception/codeception/src/Codeception/Module/WebDriver.php:938:
    string(27) "/index-test.php/master/apps"
```

Thanks.

Naktibalda:
Fixes #4107

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -955,10 +955,12 @@ class WebDriver extends CodeceptionModule implements
             $this->fail("No links containing text '$text' were found in page " . $this->_getCurrentUri());
         }
         if ($url) {
+            $expectedUrl = $this->getAbsoluteUrlFor($url);
             $nodes = array_filter(
                 $nodes,
-                function (WebDriverElement $e) use ($url) {
-                    return trim($e->getAttribute('href')) == trim($url);
+                function (WebDriverElement $e) use ($expectedUrl) {
+                    $elementHref = $this->getAbsoluteUrlFor($e->getAttribute('href'));
+                    return $elementHref === $expectedUrl;
                 }
             );
             if (empty($nodes)) {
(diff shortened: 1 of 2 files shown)


src/Codeception/Module/WebDriver.php
```diff
@@ -976,17 +978,39 @@ public function dontSeeLink($text, $url = null)
...
 
+
+    /**
+     * Returns an absolute URL for the passed URI with the current URL
+     * as the base path.
+     *
+     * @param string $uri the absolute or relative URI
+     * @return string the absolute URL
+     */
+    private function getAbsoluteUrlFor($uri)
```
DavertMik: Let's move it to `Codeception\Util\Uri`

Naktibalda: The last line calls a method of `Codeception\Util\Uri`.
The first 2 lines are WebDriver specific. (It is funny that Selenium always returns an absolute url for href, but it can return a relative url for current uri).
The check for empty url should be moved to `Uri::mergeUrls`, but I have copied it from InnerBrowser and was too lazy to refactor/didn't want to break anything else.

There are similar methods in Guzzle and Guzzle6 connector too.

Naktibalda: I will restructure this code because I don't like that `_getCurrentUri` makes HTTP​ request to Selenium on every call to `getAbsoluteUrlFor`.

I will introduce a private method `filterNodesByHref`  instead to reduce code duplication in `seeLink` and `dontSeeLink` methods.

DavertMik:
Good. Thank you

PR #4189, opened 2017-05-04
Update 07-AdvancedUsage.md

MarlonSchultz:
@dataProvider Annotation is not parsed in CamelCase for CEST´s
needs to be lowercase @dataprovider

camelCase leads to InjectionException in \Codeception\Lib\Di.php
[InjectionException] Failed to inject dependencies in instance of 'YourCest'. Failed to create instance of 'Codeception\Example'. Parameter 'data' must have default value.

### DIFF
--- a/docs/07-AdvancedUsage.md
+++ b/docs/07-AdvancedUsage.md
@@ -230,12 +230,12 @@ These examples can be written using Doctrine-style annotation syntax as well:
   }
 ```
 
-You can also use the `@dataProvider` annotation for creating dynamic examples, using a protected method for providing example data:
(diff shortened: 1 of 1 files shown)


Naktibalda:
Isn't it a bug?

MarlonSchultz:
If the annotation is supposed to work like in phpunit - then yes, it is a bug.

I couldnt find the lexer settings for annotation parsing at a quick glance.  Would need to check inner workings of the Di.php reflection about that to make a PR that would actually fix it. Assuming this is an actual bug.

DavertMik:
Ok, for now let's merge it, but yes, this bug should be fixed, so Di won't use `dataProvider` as injection rule

PR #4193, opened 2017-05-05
Handle joins such as INNER JOIN s.order order which fail with expecte…

jfxninja:
…d literal due to reserved key words

Consider the following 

       $I->seeInRepository(
            'OrderItem',
            array(
                'order' => array(
                    'recipient_email'=> self::MEMBER_EMAIL
                )
            )
        );

Generates the following query:
SELECT s FROM  OrderItem s INNER JOIN s.order order WHERE order.recipient_email = :order__recipient_email

Currently fails with expected literal. 

The suggested change would produce
SELECT s FROM  OrderItem s INNER JOIN s.order _order WHERE _order.recipient_email = :_order__recipient_email

### DIFF
--- a/src/Codeception/Module/Doctrine2.php
+++ b/src/Codeception/Module/Doctrine2.php
@@ -414,7 +414,7 @@ EOF;
                         $qb->innerJoin("$alias.$key", $key);
                         foreach ($val as $column => $v) {
                             if (is_array($v)) {
-                                $this->buildAssociationQuery($qb, $map['targetEntity'], $column, $v);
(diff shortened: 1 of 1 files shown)


jfxninja:
Didnt have the desired effect. Looking at an alternative.

Naktibalda:
Please raise an issue even if you are working on the patch.

jfxninja:
Sorry didn't read the guidelines properly the first time... 

Hopefully the new issue and new pull request are inline with the right way to do it! Thanks!

PR #4195, opened 2017-05-05
Safe prefix aliases for buildAssociationQuery()

### LINKED ISSUE #4194: Entity aliases should be prefixed to avoid reserved word collisions
```php   
$I->seeInRepository(
        'OrderItem',
        array(
            'order' => array(
                'recipient_email'=> self::MEMBER_EMAIL
            )
        )
    );
```
Generates the following query:
```sql
SELECT s FROM OrderItem s INNER JOIN s.order order WHERE order.recipient_email = :order__recipient_email
```
This fails with expected literal since due to the use of order as an alias.

Suggest prefixing the alias to generate the following
```sql
SELECT s FROM OrderItem s INNER JOIN s.order _order WHERE _order.recipient_email = :_order__recipient_email
```

jfxninja:
Consider the following

```php
$I->seeInRepository(
'OrderItem',
array(
'order' => array(
'recipient_email'=> self::MEMBER_EMAIL
)
)
);
```

Generates the following query:

```
SELECT s FROM OrderItem s INNER JOIN s.order order WHERE order.recipient_email = :order__recipient_email
```

Currently fails with expected literal due to the use of the word order as an alias.

The suggested change would produce

```
SELECT s FROM OrderItem s INNER JOIN s.order _order WHERE _order.recipient_email = :_order__recipient_email
```

Fixes #4194

### DIFF
--- a/src/Codeception/Module/Doctrine2.php
+++ b/src/Codeception/Module/Doctrine2.php
@@ -411,14 +411,14 @@ EOF;
             if (isset($data->associationMappings)) {
                 if ($map = array_key_exists($key, $data->associationMappings)) {
                     if (is_array($val)) {
-                        $qb->innerJoin("$alias.$key", $key);
+                        $qb->innerJoin("$alias.$key", "_$key");
                         foreach ($val as $column => $v) {
                             if (is_array($v)) {
                                 $this->buildAssociationQuery($qb, $map['targetEntity'], $column, $v);
                                 continue;
                             }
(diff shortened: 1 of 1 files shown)


DavertMik:
Strange that Doctrine didn't apply quotes to params... 
Thanks for the fix

jfxninja:
Happy to help! Any idea when this will be available in a tag so I can include with cmoposer? Thanks

DavertMik:
Yes, I'm working on a release! )

PR #4201, opened 2017-05-08
Allows use `InitDbFixture` feature

leandrogehlen:


DavertMik:
Thanks!

markoheijnen:
How do you now remove InitDbFixture? Or the better question, how does this work with a non-relational database like MongoDB.

leandrogehlen:
The behavior is the same or [yii2-codepection/DbTestCase.php#L32](https://github.com/yiisoft/yii2-codeception/blob/master/DbTestCase.php#L23).

This feature just was moved to this repository

markoheijnen:
That could be, still new to Yii. I know my code worked before but that could have been luck. So in my case I'm using MongoDB for everything.

What I am seeing is that InitDbFixture can't be removed and so it's running also with MongoDB which doesn't have a `checkIntegrity()` method. So it's trying to call `yii\mongodb\Command::checkIntegrity()` but fails.

leandrogehlen:
@samdark , I could create a option `initdb` to define if the `InitDbFixture` should be used.
What do you think?

PR #4208, opened 2017-05-10
fix #4207 unload fixtures when `cleanup == true`

### LINKED ISSUE #4207: [yii2] unloading fixtures only if `cleanup` configuration equals true.

after setting up `cleanup: false` configuration for the yii2 module still unloads fixtures and that might cascade on other operations

```php
$I->haveFixtures(['store' => StoreFixture::class]);
$model = new Sale(['store_id' => 1]);
$this->assertTrue($model->save());
```

I expect that sale to be available after the test ends to be handled on other tests.

#### What do you get instead?

The fixtures always unload after every test so it means any foreign key will either cascade the delete to the store record or throw a db exception.

* Codeception version: 2.2.6 exact commit https://github.com/Codeception/Codeception/commit/080dba649f5dc0f4defa326b0e237a9aa084ad99#diff-938f756d763fc21aace78bb43ea26870R154

`$fixture->unloadFixtures();` must only be called if `$this->config['cleanup']` equals true.

Faryshta:


Faryshta:
@samdark please review

src/Codeception/Module/Yii2.php
```diff
@@ -196,7 +199,9 @@ public function _before(TestInterface $test)
      */
     private function loadFixtures($test)
     {
-        if (method_exists($test, self::TEST_FIXTURES_METHOD)) {
+        if (empty($this->loadedFixtures)
```
samdark: That would remove ability to load fixtures twice. Is that intended?

Faryshta: it empties all the fixtures after unloading them at line 222 when `cleanup` is true. so this line is meaningless when `cleanup` equals true.

on the other case this only forbids to load the fixtures defined in the `_fixtures()` method twice. thats intended for when you have two tests on the same cest class

```php
public function _fixtures()
{
     // ...
}

public function a()
{
    $I->haveFixtures(...); // would work normally.     
     // save some data
}

public function b()
{
    // reloading the fixtures defined in _fixtures() means the data saved by
    // test `a` will be lost

    $I->haveFixtures(...); // would work normally.
}
```

src/Codeception/Module/Yii2.php
```diff
@@ -210,14 +215,17 @@ public function _after(TestInterface $test)
...
 
-        foreach ($this->loadedFixtures as $fixture) {
-            $fixture->unloadFixtures();
-        }
-        $this->loadedFixtures = [];
+        if ($this->config['cleanup']) {
+            foreach ($this->loadedFixtures as $fixture) {
+                $fixture->unloadFixtures();
+            }
+            $this->loadedFixtures = [];
```
Faryshta: @samdark this empties the fixtures after every test with cleanup activated.

src/Codeception/Module/Yii2.php
samdark: OK.

src/Codeception/Module/Yii2.php
samdark: Yep. Thanks for clarifying it.

src/Codeception/Module/Yii2.php
Faryshta: @samdark now i got a doubt. what if two cest classes define a `_fixtures()` method?

will the first one run but the second wouldn't? are all cest classes handled by the same instance of yii2 module?

Issue #4211, opened 2017-05-11
haveFakeRepository in the Doctrine module doesn't work


laszlo-karpati:
#### What are you trying to achieve?
Use the doctrine module's haveFakeRepository action.

#### What do you get instead?
MyTestCest.php
``` .php
$I->haveFakeRepository('Entity\User', array('findByUsername' => function($username) {  return null; }));
```

console output:
```bash
...
Scenario --
 I have fake repository "AppBundle:UniqueItem",{"getUniqueContentBySlug":"Clo...}
  [Warning] Repository can't be mocked, the EventManager class doesn't have "repositories" property
```

As I see, they removed the "repositories" property from the EntityManager, and made it accessible through the repositoryFactory:
https://github.com/doctrine/doctrine2/commit/7eb744126ba5081ea51441a893dd950046777383

### Details

* Codeception version: 2.2.10
* PHP Version: 5.6
* Operating System: Ubuntu 14.04
* Installation type: Composer
* Suite configuration:

unit.suit.yml:
```yml
# paste suite config here
class_name: UnitTester
modules:
    enabled:
        - \AppBundle\Helper\Unit
        - Symfony2:
            app_path: '../../app'
            var_path: '../../app'
        - Doctrine2:
            depends: Symfony2
            cleanup: true
        - Asserts:
```

PR #4212, opened 2017-05-11
fix haveFakeRepository in the doctrine module

laszlo-karpati:
Fix https://github.com/Codeception/Codeception/issues/4211

### DIFF
--- a/src/Codeception/Module/Doctrine2.php
+++ b/src/Codeception/Module/Doctrine2.php
@@ -268,16 +268,37 @@ EOF;
                 $methods
             )
         );
+
         $em->clear();
         $reflectedEm = new \ReflectionClass($em);
-        if ($reflectedEm->hasProperty('repositories')) {
-            $property = $reflectedEm->getProperty('repositories');
-            $property->setAccessible(true);
-            $property->setValue($em, array_merge($property->getValue($em), [$classname => $mock]));
+
+        if ($reflectedEm->hasProperty('repositoryFactory')) {
+            $repositoryFactoryProperty = $reflectedEm->getProperty('repositoryFactory');
+            $repositoryFactoryProperty->setAccessible(true);
+            $repositoryFactory = $repositoryFactoryProperty->getValue($em);
+
+            $reflectedRepositoryFactory = new \ReflectionClass($repositoryFactory);
+
+            if ($reflectedRepositoryFactory->hasProperty('repositoryList')) {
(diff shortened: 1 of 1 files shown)


DavertMik:
Probably this has been working in earlier versions of Doctrine
Could you check when they did change their internal API?

laszlo-karpati:
@DavertMik I wrote about it in the connected issue as well:

they changed this part of their internal API with this commit:
https://github.com/doctrine/doctrine2/commit/7eb744126ba5081ea51441a893dd950046777383

DavertMik:
Could you leave the current implementation so for not to break things for users of old Doctrine?

laszlo-karpati:
@DavertMik Sure,
I've just pushed the backward compatibility.

DavertMik:
Thanks

polki-bil:
Hello, we are still having problem. Symfony generates proxy class for EntityManager, so Doctrine2 module can't find repositories nor repositoriyFactory. 
class:
    `DoctrineORMEntityManager_00000000136fffc9000 ....`
and has only properties (shortened): 
```
    object(ReflectionProperty)[2677]
    public 'name' => string 'valueHolder593fb92bd0e38185253777' (length=33)
    object(ReflectionProperty)[3536]
    public 'name' => string 'initializer593fb92bd0e4f480946614' (length=33)
    object(ReflectionProperty)[3539]
    public 'name' => string 'publicProperties593fb92bd0dee937024105' (length=38)

```
We are using Symofny 3.3 and Doctrine 4.5
We tried various Symfony configs to prevent from generating this, but with no success. 
With this code added it works:
   ` $ReflectedEm->getParentClass()->hasProperty('repositoriyFactory');`

Thanks for any help

laszlo-karpati:
Hi @polki-bil!

_I think you've mistyped your Doctrine's version, and you are using v2.5._

As I see Docrine was adding the entityManager's spl_object_hash to the repositoryList key's name in the DefaultRepositoryFactory since version 2.5.  

They added it with this commit, released in version 2.5:
https://github.com/doctrine/doctrine2/commit/a9847533e5d3c2efc15284fb93fb57a443978cf5

I'm using Doctrine v2.4 right now, but if I'll have time for it in the next days, I will try to figure out something for it.

Can you try it with the [v2.4's DefaultRepositoryFactory class](https://github.com/doctrine/doctrine2/blob/2.4/lib/Doctrine/ORM/Repository/DefaultRepositoryFactory.php)?

PR #4216, opened 2017-05-13
Installation Templates

DavertMik:
This feature allows to customize Codeception setup and create custom installation recipes.

Setup can be customized using the Template classes which are extended from `Codeception\InitTemplate`.
Then it can interactively ask questions and generate configuration, helpers, directories.

The single suite template for acceptance tests is included.

selection_203

The installation code is pretty easy to work with: https://github.com/Codeception/Codeception/blob/04ddcef2c1065e84dfc7f0e4b8a5b285dd5ad0c1/src/Codeception/Template/Acceptance.php#L66


DavertMik:
> This command has a very similar functionality to bootstrap, names are synonymous too. I think that it would be better to implement it as a --template option to bootstrap command and package current bootstrap functionality as a default template.

I was thinking on it as well, however, there are reasons to make it as the standalone command.

1. `bootstrap` is non-interactive and creates 3 suites by default. However, those 3 suites are not regularly used, as people do not always plan to have 3 layers of testing. They do not understand the purpose of some suites.
2. using `bootstrap --template` is not an option; we expect that this command would be used by newcomers for a customized setup. It will be hard to explain them to use this option every time they need to set up a project. If they forget this option - they get a completely different setup.
3. we could refactor/replace bootstrap with init, however, there are so many tutorials on the internet on bootstrap command so I'd like to keep it the way it is. 

So the idea is: 
* `bootstrap` stays as opinionated non-interactive setup with a default template
* `init` is for interactive customized setups with no default template

DavertMik:
Bootstrap is now a template as well.
selection_204

PR #4220, opened 2017-05-15
Db dump populator

brutuscat:
Continuation from #4171

Add ability to load the dump using a populator command. The db module now accepts a populator configuration item where the user can enter a command to execute his/her prefered tool or wrapper script to load the dump.

The dsn is parsed into variables and are made available, with the variables from the module configuration, to the command for interpolation, allowing the executing of commands like:

`populator: 'mysql -u $user -p $password -h $host -D $dbname < $dump'`

The Db module test has been structured in order to test individual drivers features.

TODO: 
- [ ] Add all driver's tests
- [ ] Rebase once accepted


brutuscat:
@DavertMik I wanted to leave this with the less amount of changes possible. So I added some abstract test to assert for the db driver features. 

Could you take it from here?

DavertMik:
Yes, thank you! I will look into it in the evening

brutuscat:
@DavertMik added some more abstract test for the cleanup and the load dump. But Sqlite still is giving errors... I guess it is because some permission issue on travis, but I can't reproduce it locally, could you check it out? 

Also some tests I broke because of the phpunit version and the php it seems... But overall after the final test refactor to use instance variables instead of static, everything seems to be working.

I just need your expertise with travis to solve the populator not working for some reason (getting exit code 3 when executing the command) and tiny help wrapping some code to support older versions of php and phpunit.

DavertMik:
Yeah looks like a permission issue:

```
#define SQLITE_PERM         3   /* Access permission denied */
```
I will pull this branch, do some refactoring and create another PR

src/Codeception/Module/Db.php
```diff
@@ -71,9 +72,32 @@
  *              populate: true
  *              cleanup: false
  *              reconnect: true
+ *              populator: 'mysql -u root -h $host -D $dbname < $dump'
  *
  * ## SQL data dump
  *
+ *  You now have two ways of loading the dump into your database. You could use the legacy method,
+ *  where codeception reads the dump file and passes it to the specific driver so it populates the
```
Naktibalda: It is Codeception, not codeception.

src/Codeception/Module/Db.php
```diff
@@ -71,9 +72,32 @@
  *              populate: true
  *              cleanup: false
  *              reconnect: true
+ *              populator: 'mysql -u root -h $host -D $dbname < $dump'
```
Naktibalda: Example with pgsql would be usefull.

tests/unit/Codeception/Module/Db/DbCleaningTest.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+abstract class DbCleaningTest extends \PHPUnit_Framework_TestCase
```
Naktibalda: Wouldn't it be better to convert all these function specific files to traits and combine then to a single Test class per database type?

DavertMik: Very good suggestion. 

Also we need MySQL tests there as well, and Postgress too...

DavertMik:
> Support older versions of php and phpunit.

Mark test as skipped the very old ones (5.4, 5.5).
As for PHPUnit it is recommended to extend tests from `Codeception\Test\Unit` as it provides a compatibility layer

DavertMik:
I started to work on this in my branch 
https://github.com/Codeception/Codeception/tree/brutuscat-db-dump-populator

If you want to add something more send PR to it.
I'm trying to fix tests and refactor the module and populator itself.
Current tests look pretty ugly as there were too much reflection magic

tests/unit/Codeception/Module/Db/DbCleaningTest.php
brutuscat: I was just following the "best practice" for PHPUnit, is the trait as a collection of test something that you guys do? I'm not a PHPUnit guy neither but I found that the abstract class is how PHP folks like to do it. In Ruby I would do "it_behaves_like" using rspec shared examples.... https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples

brutuscat:
👍 about the "Current tests look pretty ugly as there were too much reflection magic
" I was trying to follow the "privatization" of methods that I saw around the Db module...

Anything else I could help?

DavertMik:
No. I think I finished everything I wanted to change and updated docs.
Thanks for your work, Populator parser looks very good and I works with sqlite/mysql/postgres

PR #4225, opened 2017-05-17
Minor improvements

ThomasLandauer:


DavertMik:
> Please explain: What's the advantage of having a *public* method for the data provider? Why would somebody want to have this?

The test is prepared for execution by external class. So it needs to run the public method to get all data out of it and make corresponding tests

docs/05-UnitTests.md
```diff
@@ -233,7 +233,7 @@ function testUserNameCanBeChanged()
 
 In both examples you should not be worried about the data persistence between tests.
 The Doctrine2 and Laravel5 modules will clean up the created data at the end of a test.
-This is done by wrapping a test in a transaction and rolling it back afterwards.
+This is done by wrapping each test in a transaction and rolling it back afterwards.
```
DavertMik: yes, this is correct. Each test

docs/07-AdvancedUsage.md
```diff
@@ -267,20 +269,12 @@ Alternatively, the `@dataprovider` can also be a public method starting with `_`
     */
     public function staticPages(AcceptanceTester $I, \Codeception\Example $example)
     {
-        $I->amOnPage($example['url']);
-        $I->see($example['title'], 'h1');
-        $I->seeInTitle($example['title']);
+        // ...
```
DavertMik: why did you remove examples?

ThomasLandauer:
@DavertMik: About the public data provider:  
You mean, by having it public, you can use the data provider for a different test class too?  
If yes, please tell me the syntax for accessing another class in the `@dataprovider` annotation, and I'll include it.

docs/07-AdvancedUsage.md
ThomasLandauer: I deleted the parts which are the same as in the above code block. Here, only the syntax of the public data provider is important, the rest is obvious :-)

DavertMik:
> You mean, by having it public, you can use the data provider for a different test class too?

No, data provider is used by a [Loader](https://github.com/Codeception/Codeception/blob/2.2/src/Codeception/Test/Loader/Cest.php) class. It takes a test class, and creates tests out of it. If `@depends` annotation found - separate tests are created from it. So the loader needs to access the dataprovider method to get the data.

However, there is no technical difference right now, as to access this method we use `ReflectionHelper::invokePrivateMethod` (line 65) so even private methods can be used.

readme.md
```diff
@@ -83,25 +83,23 @@ Run CLI utility:
...
 
-this will create a default directory structure and default test suites
+This will create a default directory structure and default test suites.
+
+## Documentation
+
+[Documentation](http://codeception.com/docs/01-Introduction)
 
-See Documentation for more information.
+Documentation is included within the project. Look for it in the ['docs' directory](https://github.com/Codeception/Codeception/tree/master/docs).
```
DavertMik: Yes, but master branch stands for the "bleeding edge". Current stable branch is 2.2, once 2.3 is released it will be 2.3, and so on. So there is no permalink for that :(

ThomasLandauer:
I'm not sure we're talking about the same thing ;-)

My question:  
On https://github.com/Codeception/Codeception/blob/2.2/docs/07-AdvancedUsage.md#examples you say
> Alternatively, the @dataprovider can also be a public method starting with _ prefix so it will not be considered as a test:

What's the advantage of doing it like in this codeblock, compared to the codeblock above (i.e. protected function)?  
Is there a useful usecase for having the dataprovider public (i.e. callable from outside)?  
If yes, it should be explained there. If no, the entire codeblock could be deleted :-)

readme.md
ThomasLandauer: Yeah, OK. My main point: The "Documentation" link should lead to the actual documentation website - not to the docs directory.

Can you undo this part? Or do I have to create a separate commit? I'm no GitHub expert.....

DavertMik:
> Is there a useful usecase for having the dataprovider public (i.e. callable from outside)?
If yes, it should be explained there. If no, the entire codeblock could be deleted :-)

Short answer: There is no difference

ThomasLandauer:
OK, so please merge this pull request, and I'll start a new one for getting rid of this code block :-)

DavertMik:
OK!

PR #4229, opened 2017-05-19
Redis module default configuration must be more foolproof.

hchonan:
According to Document ( http://codeception.com/docs/modules/Redis ) and comments of module source, it says Codeception flushes the database "At the beginning of every suite" ( `cleanupBefore: 'suite'` ) . However default setting of `cleanupBefore` is set to "At the beginning of every test" ( `cleanupBefore: 'test'` ), so description of documentation is not actually true.

And flushing database without any attention seems dangerous behavior, so I turned default configuration to safe-side.

* If this change accepted, Document ( http://codeception.com/docs/modules/Redis ) have be updated.

### DIFF
--- a/docs/01-Introduction.md
+++ b/docs/01-Introduction.md
@@ -10,10 +10,10 @@ but with tests you can cover the most important parts of your app and at least b
 
 There are plenty of ways to test your application.
 The most popular paradigm is [Unit Testing](http://en.wikipedia.org/wiki/Unit_testing).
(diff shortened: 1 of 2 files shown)


DavertMik:
Yes, this will be consistent with Db in 2.3.

DavertMik:
Thanks. 

Please note, this fix will be available starting from Codeception 2.3.0

hchonan:
Thanks for quick reaction!

PR #4230, opened 2017-05-19
db dump populator

DavertMik:
Initially developed by @brutuscat 
Replaces #4220 

* updated tests (added MySQL and Postgres db tests) 
* refactored populator class
* updated documentation
* [BREAKING] Default `cleanup` and `populate` set to false in Db module


src/Codeception/Lib/DbPopulator.php
```diff
@@ -0,0 +1,108 @@
...
+    /**
+     * @var array
+     */
+    protected $config;
+
+    /**
+     * Constructs a DbPopulator object for the given command and Db module.
+     *
+     * @param string  $command  The parameterized command to evaluate and execute later.
+     * @param Codeception\Module\Db|null $dbModule The Db module used to build the populator command or null.
```
Naktibalda: Docblock doesn't match signature

src/Codeception/Module/Db.php
```diff
@@ -28,35 +29,23 @@
...
- * Supported but not tested.
+ * Also available:
  *
  * * MSSQL
  * * Oracle
  *
  * Connection is done by database Drivers, which are stored in the `Codeception\Lib\Driver` namespace.
- * [Check out the drivers](https://github.com/Codeception/Codeception/tree/2.1/src/Codeception/Lib/Driver)
+ * [Check out the drivers](https://github.com/Codeception/Codeception/tree/2.2/src/Codeception/Lib/Driver)
  * if you run into problems loading dumps and cleaning databases.
```
Naktibalda: It's time to update this link to 2.3

src/Codeception/Module/Db.php
```diff
@@ -71,9 +60,74 @@
...
+ *
+ * ```
+ * modules:
+ *    enabled:
+ *       - Db:
+ *          dsn: 'pgsql:host=localhost;dbname=testdb'
+ *          user: 'root'
+ *          password: ''
+ *          cleanup: true # run populator befor each test
+ *          populator: 'pg_restore -u $user -h $host -D $dbname < tests/_data/db_backup.dump'
```
Naktibalda: This command only works if CWD is project dir.
Would it be possible to add a variable for codecept_data_dir()?
e.g. ` < $dataDir/dump.sql`

src/Codeception/Module/Db.php
```diff
@@ -139,10 +193,11 @@ class Db extends CodeceptionModule implements DbInterface
      * @var array
      */
     protected $config = [
-        'populate' => true,
-        'cleanup' => true,
+        'populate' => false,
+        'cleanup' => false,
```
Naktibalda: Why are you changing default values?

tests/unit/Codeception/Module/Db/TestsForDb.php
```diff
@@ -0,0 +1,177 @@
...
+        // just for "table not found" across all the drivers
+        $this->expectException(\PDOException::class);
+
+        $this->module->dontSeeInDatabase('users', ['name' => 'davert']);
+    }
+
+    public function testHaveAndSeeInDatabase()
+    {
+        $user_id = $this->module->haveInDatabase('users', ['name' => 'john', 'email' => 'john@jon.com']);
+        $group_id = $this->module->haveInDatabase('groups', ['name' => 'john', 'enabled' => false]);
```
Naktibalda: Underscores aren't allowed in variable names.

src/Codeception/Module/Db.php
DavertMik: Yes. They are dangerous as they are now. It took me few years to conclude that. 
I should not forget to add this in changelog.
This may break some BC but changing config values shoudn't be a big deal...

src/Codeception/Module/Db.php
DavertMik: You can use specify dump in config option and use  `$dump` variable.
However pg_restore uses binary format for dumps, so if you remove the `populator` config option, Codeception won't be able to load that dump

src/Codeception/Lib/DbPopulator.php
```diff
@@ -0,0 +1,109 @@
...
+     * $config = ['name' => 'Mauro'];
+     *
+     * // With the above parameters it will return `'Hello Mauro'`.
+     * ```
+     *
+     * @param string $command The command to be evaluated using the given config
+     * @param array $config The configuration values used to replace any found $keys with values from this array.
+     * @return string The resulting command string after evaluating any configuration's key
+     */
+    protected function buildCommand($command)
```
tiger-seo: since this method only returns, please rename it to `getBuildCommand()`

DavertMik: Thanks

tiger-seo: it doesn't change the state, i.e. pure method

DavertMik: it takes command, changes it, returns a command. Yes, it's pure function but I don't see how it is supposed to affect the naming. `getCommand` is not supposed to transform something

docs/09-Data.md
```diff
@@ -35,9 +32,40 @@ Use <a href="http://codeception.com/docs/06-ModulesAndHelpers#Dynamic-Configurat
...
+          user: 'root'
+          password: ''
+          cleanup: true # run populator before each test
+          populate: true # run populator before all test
+          populator: 'mysql -u $user $dbname < tests/_data/dump.sql'
+```
+
+See the []Db module reference](http://codeception.com/docs/modules/Db#SQL-data-dump) for more examples.
+
+To ensure database dump is loaded before all tests add `populate: true`. To clean current database and reload dump between tests use `cleanup: true`.
```
brutuscat: 👍 to add this description of the behavior expected. This has never been documented and now the intention is clear.

src/Codeception/Lib/DbPopulator.php
brutuscat: I agree wit @DavertMik prefixing it with `get` gives the implicit impression that it is just a property of the object. But this is not the case, the "value" is what the user configured and this method uses the context of data (e.g. the config of the Db module) to obtain an output after computing what needs (and what is available) be evaluated on the populator config's value.

src/Codeception/Lib/DbPopulator.php
```diff
@@ -0,0 +1,109 @@
...
+        if (!$this->config['dump']) {
+            codecept_debug("[Db] No dump file found. Skip loading the dump using the populator command.");
+            return false;
+        }
+
+        $command = $this->getBuiltCommand();
+        codecept_debug("[Db] Executing Populator: `$command`");
+
+        exec($command, $output, $exitCode);
+        codecept_debug("[Db] Populator Finished. Exit code: `$exitCode`");
```
brutuscat: @DavertMik tiny detail, we could move this debug line after it is sure that the exception won't be thrown )e.g. after the if and before the return)

src/Codeception/Module/Db.php
```diff
@@ -69,11 +58,78 @@
...
+ *          dsn: 'pgsql:host=localhost;dbname=testdb'
+ *          user: 'root'
+ *          password: ''
+ *          dump: 'tests/_data/db_backup.dump'
+ *          populate: true # run populator before all tests
+ *          cleanup: true # run populator before each test
+ *          populator: 'pg_restore -u $user -h $host -D $dbname < $dump'
+ * ```
+ *
+ *  Variable names taken from config and DSN which have a `keyword=value` format, so you should expect to have a variable named as the
```
brutuscat: @DavertMik grammar: Variable names **are been** taken

docs/09-Data.md
DavertMik: The behavior has changed by the way. By default cleanup and populate was disabled. 
This makes sense, because the first Codeception version didn't have methods like haveInDatabase so there was no way to create data in database, so it was creating it from dump.

src/Codeception/Lib/DbPopulator.php
tiger-seo: @DavertMik it doesn't change anything, it only reads provided input variable with current class state and returns some value by its internal logic;  I offered to add `get`, i.e. `getBuildCommand`, not `getCommand`, as you wrote;  `getBuildCommand` takes input template, modifies its copy and returns it, that's all,  but if you insist, ok, go the way you want

@brutuscat maybe you are used to treat anything with `get`-suffix only as simple getter for class properties, but I, and many many others, for instance Symfony developes, don't think that it is the only use case, generally it's common to treat method names as follows:
- `get`-methods that are designed to return some data and do not change this or any others state
- `set`-methods that are designed to change this or some other object state (properties)
- action-method, e.g.  `buildCommand`, `doRequest`, `handle`, etc. - procedures that execute some logic or call other series of actions

brutuscat: @tiger-seo yes, get and set are either for property values or just to retrieve something where no computation happens. At least that's the implicit "logic" that I would think a method has.

PR #4232, opened 2017-05-21
Deleted example for *public* dataprovider function

ThomasLandauer:
See https://github.com/Codeception/Codeception/pull/4225#issuecomment-302549188

BTW: Is http://codeception.com/docs/07-AdvancedUsage not updated in real-time from `07-AdvancedUsage.md`? I still see the old version there...

### DIFF
--- a/docs/07-AdvancedUsage.md
+++ b/docs/07-AdvancedUsage.md
@@ -249,7 +249,7 @@ You can also use the `@dataprovider` annotation for creating dynamic examples, u
     /**
      * @return array
      */
-    protected function pageProvider()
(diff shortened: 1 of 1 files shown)


Naktibalda:
but why?

ThomasLandauer:
Why delete it? => Because it's useless! Giving a detailed example for a useless scenario only confuses readers (since they think it must be important)

DavertMik:
Yes, better to delete it as it adds nothing more than the basic example.

ThomasLandauer:
Well, OK, then please merge this :-)

PR #4239, opened 2017-05-22
[REST] Don't fail dontSee*JsonPath assertions when response is scalar…

### LINKED ISSUE #4237: JsonPath fails in 2.2.12
JsonPath related method failing in codecept 2.2.12, working in 2.2.11.

#### What do you get instead?
See output below.

```bash
 Step  Don't see response json matches json path "$.error"
 Fail  JSON response is not an array: true
```
Response: true

```php
$I->dontSeeResponseJsonMatchesJsonPath('$.error');
```
### Details

* Codeception version: 2.2.12, flow/jsonpath 0.3.4
* PHP Version: 7.1
* Operating System: linux
* Installation type: Composer
* Suite configuration:

```yml
# paste suite config here
```

Naktibalda:
… value

Closes #4237

### DIFF
--- a/src/Codeception/PHPUnit/Constraint/JsonContains.php
+++ b/src/Codeception/PHPUnit/Constraint/JsonContains.php
@@ -31,6 +31,9 @@ class JsonContains extends \PHPUnit_Framework_Constraint
     protected function matches($other)
     {
         $jsonResponseArray = new JsonArray($other);
+        if (!is_array($jsonResponseArray->toArray())) {
+            throw new \PHPUnit_Framework_AssertionFailedError('JSON response is not an array: ' . $other);
+        }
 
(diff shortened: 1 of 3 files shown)


src/Codeception/PHPUnit/Constraint/JsonContains.php
```diff
@@ -31,6 +31,9 @@ public function __construct(array $expected)
     protected function matches($other)
     {
         $jsonResponseArray = new JsonArray($other);
+        if (!is_array($jsonResponseArray->toArray())) {
+            throw new \PHPUnit_Framework_AssertionFailedError('JSON response is not an array: ' . $other);
```
DavertMik:
```php
$this->fail('JSON response is not an array: ' . $other)
```

DavertMik:
Ok, thanks!

PR #4241, opened 2017-05-23
Added link to Db Module

ThomasLandauer:


ext/Recorder.php
```diff
@@ -16,8 +16,6 @@
  * Saves a screenshot of each step in acceptance tests and shows them as a slideshow on one HTML page.
  * Activated only for suites with WebDriver module enabled.
  *
- *  ![recorder](http://codeception.com/images/recorder.gif)
```
DavertMik: why do you remove the picture? Its nice ;)

ThomasLandauer: No, it's really annoying! Even worse than most ads ;-) And it doesn't add any value. Everybody knows what a screenshot looks like...

DavertMik: Then please make it as a link.
Recorder is much easier to understand when you see it in action.

ThomasLandauer: OK, good compromise :-) Did that.

DavertMik:
Thanks

PR #4248, opened 2017-05-24
[WebDriver] Failing test for sticky header

ThomasLandauer:


Naktibalda:
This test is 3 years old now,
It would be great if you could come up with a fix.
Or at least make new pull request to https://github.com/Codeception/module-webdriver

PR #4249, opened 2017-05-24
Minor changes

ThomasLandauer:


DavertMik:
> I got them in `_data`, not in `_output`. But I have Symfony. So if this is Symfony-specific, please explain.

No, you are right, the command has changed and I absolutely forgot I wrote about it in Getting Started. So everything is fine )

PR #4252, opened 2017-05-24
Db cleanup fix

DavertMik:
While refactoring Db module I accidentally hit the old API and this made accidentally call database cleanups before each test without considering a `cleanup` option. This happened because `_cleanup` method was called before each test as part of the initialization process of a module (no one used it actually).

### DIFF
--- a/src/Codeception/Module.php
+++ b/src/Codeception/Module.php
@@ -202,10 +202,6 @@ abstract class Module
     {
     }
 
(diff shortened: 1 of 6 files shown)


DavertMik:
This fix does not include tests because it was triggered by `Subscriber\Module->before`, so it could not be located by unit test. The bug was found in Phalcon internal tests

Issue #3976, opened 2017-02-01
Codeception 3.0 Roadmap


DavertMik:
Here starts a discussion of what we'd like to see in Codeception 3.0. Any proposals are welcome, they can probably be added to a list. Currently I see the main goals for next major release are *simplification* and *decoupling*. By simplification I mean:

- [ ] ability to use Codeception for acceptance tests only (single suite)

installation should provide an option to use one suite only and for one suite-setup there can probably be an option to merge `acceptance.suite.yml` config with main codeception config

- [ ] moving most of important modules to the separate repositories #1233

in this case installing codeception for laravel and webdriver will probably require this config:

```json
{
   "require-dev": {
        "codeception/base": "^3.0.0",
        "codeception/laravel-module": "^3.0.0",
        "codeception/webdriver-module": "^3.0.0"
   }
}
```

codeption/codeception will still be a main repo for developement. Also `codeception/codeception` will still provide all-in-one-package. Modules will be [split](https://help.github.com/articles/splitting-a-subfolder-out-into-a-new-repository/) to own repositories. By doing so will allow us keep separate changelogs and release things faster, on per module bases.

- [ ] programmatic interface: allow developers to create their own runners for PHP script. 

That also involves refactoring of some messy places.

- [ ] drop support for PHP < 7 #3972 
- [ ] YAML test format (for non-PHP developers)
- [ ] **Deprecation of Cept** format.

By deprecation I mean keeping abilty to run it but no longer mention it in documentation, remove generators for it 

...(more will come)...

NavyCoat:
Yeah, I will lovely see version made only for acceptance testing! :D

DavertMik:
@NavyCoat you don't need to wait for 3.0, it is already planned for Codeception 2.3 :)

Faryshta:
>  ability to use Codeception for acceptance tests only (single suite)

I love codeception cest classes for unit testing and functional testing. How about dividing the suites in repos?

- codeception/acceptance-suite
- codeception/functional-suite
- codeception/api-suite
- codeception/unit-suite

the idea of decoupling the modules is awesome too, it would ease the instal process a lot.

>  drop support for PHP < 7 

awesome. Specially if you can jump to php7.1 directly since php 7.0 will love active support at the end of this year.

zend and yii2 are taking this route too specially for the support for nullable type hints.

ThomasLandauer:
I was just about to suggest "Deprecation of Cept" when I found this issue :-)  
Reasons:
[...]

PR #4259, opened 2017-05-25
Changed code example from Cept to Cest

ThomasLandauer:


samdark:
Why?

ThomasLandauer:
Cause Cept is awaiting deprecation :-) https://github.com/Codeception/Codeception/issues/3976

DavertMik:
Thanks @ThomasLandauer 
That's a really good update

docs/04-FunctionalTests.md
```diff
@@ -250,10 +267,43 @@ error_level: "E_ALL & ~E_STRICT & ~E_DEPRECATED"
 
 `error_level` can also be set globally in `codeception.yml` file.
 
+## Accessing external URL's with Guzzle
+
+[Guzzle](http://docs.guzzlephp.org/en/latest/) is a library for accessing URL's through PHP's [cURL library](http://php.net/manual/book.curl.php). Guzzle is automatically installed with Codeception. You can easily access it from any test file:
+
+```php
+// ExternalCest.php
```
DavertMik: should start with `<?php`

docs/04-FunctionalTests.md
```diff
@@ -250,10 +267,43 @@ error_level: "E_ALL & ~E_STRICT & ~E_DEPRECATED"
 
 `error_level` can also be set globally in `codeception.yml` file.
 
+## Accessing external URL's with Guzzle
```
DavertMik: Sorry, this is not a valid part for Functional tests.
1. Functional tests are executed inside a framework with no external calls. 
2. You are trying to reimplement PhpBrowser (which is described in Acceptance Testing chapter)
3. Yes, you can use Guzzle but in most cases you need to make some custom HTTP call to 3rd party you should place this in a Helper and use like: `$I->sendCustomRequest()`. 
4. Mixing scenario-driven `$I->` style with plain PHP code is only applicable for unit testing; However, you are not testing Guzzle library here. 

I hope this explains philosophy a bit

docs/04-FunctionalTests.md
ThomasLandauer: I left it out on purpose. I copied the style from Symfony's code blocks, see e.g. http://symfony.com/doc/current/controller.html

Reasons:

* That a PHP framework is written in PHP should be obvious :-)

* More important is the question "In which file is this supposed to go?" That's why the file name should be included in every code block.

But if you *really* want the `<?php`, I'd write it like that:
```php
<?php
// ExternalCest.php
```

docs/04-FunctionalTests.md
ThomasLandauer: In advance: The code I posted worked for me! Is that Symfony-specific?

So I'm not sure what you mean:

1. It's wrong because it doesn't work. My answer: See above :-)
1. It's wrong because you shouldn't do it this way. My answer: Then how should it be done?

My main point: I see a need for accessing external url's in functional tests; people are certainly asking this question. So there should be an answer. If this pull request is the wrong answer, then let's change it to the right! :-)

DavertMik: The code is right and the suggestion works. However, it should not be the part of this specific chapter. This is a general advice not actually related to functional tests. It should be placed into [Cookbook](http://phptest.club/c/codeception/cookbook)

docs/04-FunctionalTests.md
DavertMik: Yes, please do. Our syntax on website is very sensitive to `<?php` header and won't highlight PHP code without it :upside_down_face:

docs/04-FunctionalTests.md
ThomasLandauer: Well, to me it's a functional testing topic, cause: Unit testing has nothing to do with URLs at all. And in Acceptance testing, there's no need for Guzzle, since you have a browser anyway. Right? => So Functional is the only one which is left ;-)

Didn't know that a "Cookbook" exists ;-) But I must say: One of the main problems of the docs is that they are scattered around 200 places - each with different layout, different wording, different commit policy. So if you have a question, you cannot go to one place and look for an answer, but you have to figure out first which place is the right to start looking.  
So I'd suggest to integrate those 7 cookbook articles into the main docs: The 2 WebDriver articles to the WebDriver Module, the 1 Symfony article to the Symfony Module, and so on...

docs/04-FunctionalTests.md
ThomasLandauer: Done :-)

docs/04-FunctionalTests.md
DavertMik: No, it's not possible. Cookbooks contain current solution for very specific problem, while documentation is a general, constantly updated. You explain a very specific problem, and as I said this is not about functional tests at least in the term of Codeception. Also, functional tests are supposed to test the system from inside, not outside, so this example is not applicable here .

Btw, instead of using Guzzle you can use PhpBrowser for the same case (which was already explained in the previous chapter). 

Please remove that part from this PR and I will merge it. Thanks

ThomasLandauer:
Guzzle/PhpBrowser: Yeah, meanwhile I noticed that ;-)  
Question: Is it possible to use PhpBrowser **and** WebDriver? Do/can you have two `acceptance.suite.yml`s for that?

DavertMik:
Thanks

PR #4262, opened 2017-05-26
Fix autoloading with PHPUnit 6

enumag:
My builds on Travis are failing because composer autoloading doesn't work correctly for PHPUnit 6.

```
PHP Fatal error:  Interface 'PHPUnit_Framework_Test' not found in ...\vendor\codeception\codeception\src\Codeception\TestInterface.php on line 7
```

So far the only fix I found was to add `"files": ["vendor/codeception/codeception/shim.php"]` to my `autoload-dev` in composer.json. This commit should fix the issue.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -64,7 +64,10 @@
         "psr-4":{
             "Codeception\\": "src\\Codeception",
             "Codeception\\Extension\\": "ext"
-        }
+        },
+        "files": [
+            "shim.php"
+        ]
     },
     "extra": {
         "branch-alias": {


Naktibalda:
How do you run your tests?

enumag:
I checked my builds again. Tests are actually passing. This error happens when analyzing the code using [phpstan](https://github.com/phpstan/phpstan).

Naktibalda:
shim.php is included by [autoload.php](https://github.com/Codeception/Codeception/blob/2.3/autoload.php#L45).

enumag:
Which is not autoloaded either.

Naktibalda:
That's correct, it is included by [codecept file](https://github.com/Codeception/Codeception/blob/2.3/codecept#L7).

enumag:
I know. PHPStan doesn't use that obviously which is why it is failing.

DavertMik:
Anyway, for such cases or for case when a custom runner is implemented, this patch makes sense

Naktibalda:
Shouldn't we take it out of autoload.php now?
It will probably cause a class already exists error in some edge case.

DavertMik:
@enumag sorry, this patch caused a deeper problem. Now, `shim` is loaded everywhere, even outside the Codeception. So it may affect the execution of other applications not related to Codeception.
I need to revert this PR

enumag:
Understood. I'll just add it to my autoload-dev then. Thanks.

PR #4263, opened 2017-05-26
Update WebDriver:makeScreenshot to accept noname

eXorus:
Hello,

Currently, makeScreenshot of WebDriver require a name to save a file.

This little PR accept noname.

$I->makeScreenshot('edit_page'); //saved to: tests/_output/debug/edit_page.png
$I->makeScreenshot(); //saved to: tests/_output/debug/2017-05-26_14-24-11.png

What do you think ?

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -640,13 +640,16 @@ class WebDriver extends CodeceptionModule implements
      * $I->amOnPage('/user/edit');
      * $I->makeScreenshot('edit_page');
      * // saved to: tests/_output/debug/edit_page.png
+     * $I->makeScreenshot();
+     * // saved to: tests/_output/debug/2017-05-26_14-24-11.png
      * ?>
      * ```
      *
      * @param $name
      */
-    public function makeScreenshot($name)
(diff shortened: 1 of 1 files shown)


DavertMik:
Looks good to me :+1:

src/Codeception/Module/WebDriver.php
```diff
@@ -640,13 +640,16 @@ public function _savePageSource($filename)
...
+     * // saved to: tests/_output/debug/2017-05-26_14-24-11.png
      * ?>
      * ```
      *
      * @param $name
      */
-    public function makeScreenshot($name)
+    public function makeScreenshot($name = null)
     {
+        $name = ($name) ? $name : date("Y-m-d_H-i-s");
```
DavertMik: In case a test will be executed in parallel this may produce a conflict between 2 files.
how about adding `uniqid` to the name?

also please refactor this code using `if` operator.

eXorus:
@DavertMik There is no conflict with parallel execution because the screenshot will be replaced. But I added uniqid to be sure that all screenshot will be saved.

I have kept the date to be able to see the order.

Issue #4257, opened 2017-05-25
Wrong documentation of InnerBrowser::haveHttpHeader()


ThomasLandauer:
See https://github.com/Codeception/Codeception/blob/2.3/src/Codeception/Lib/InnerBrowser.php#L283

The function is called `haveHttpHeader()`, but the example in the code block reads:
```php
$I->setHeader('X-Requested-With', 'Codeception');
```

I don't change anything, since I don't know what the intended name is. But I must say that `haveHttpHeader()` is a **really bad** name for an action, cause "have" sounds like an assertion! Any name starting with "set" is certainly better.

Naktibalda:
In Codeception assertions start with `see`,
`have` methods are used to setup preconditions.

Another example is [Db::haveInDatabase](https://github.com/Codeception/Codeception/blob/2.3.1/src/Codeception/Module/Db.php#L397)

DavertMik:
We have another strange naming which is `expect`, you expect this to be assertion but it is a comment :laughing: :rofl: #self-irony

But for `have` it's totally fine as it describes what you have in the beginning. For preconditions we use 2 prefixes: `have` and `am`, so ths scenario would follow Arrange-Act-Assert syntax:

```php
// arrange
$I->haveUser('davert');
$I->haveProduct('iPhone');
$I->amLoggedAsAdmin();
$I->amOnPage('/users/davert');

// act
$I->click('Add Product');
$I->selectOption('iPhone');

// assert
$I->see('iPhone', '.products');
```

Something like that

ThomasLandauer:
Please fix the **error** that I mentioned in the first sentence!

For the question if "have" and "expect" are good names, I'll start a new issue :-)

ThomasLandauer:
Was fixed with https://github.com/Codeception/Codeception/pull/4347

PR #4265, opened 2017-05-27
Explain the naming philosophy

ThomasLandauer:
I don't know if this is the right place. But there should be a short introduction like this. Please make this section (i.e. anchor) the link target for codeception.com "User Centric Tests => Learn more"

Details:

* grab: Is there an easier exmple you can think of? One that doesn't require the Asserts Module?

* Missing in the list is "have" and maybe "expect". Maybe I can convince you to change those :-) See https://github.com/Codeception/Codeception/issues/4257

I'd like to rewrite the next section (Actors) too. Questions:

* I guess the term "Actor" is a leftover from previous versions of Codeception? More precise: Today, it's not really important (anymore) to know the term "Actor", is it?

* What do you mean by "Actor classes are not written but generated from suite configuration." => That they are *generated* automatically?

### DIFF
--- a/docs/02-GettingStarted.md
+++ b/docs/02-GettingStarted.md
@@ -5,6 +5,30 @@ and bootstrapped your first test suites. Codeception has generated three of them
 They are well described in the [previous chapter](http://codeception.com/docs/01-Introduction). Inside your __/tests__ folder you will have three `.yml` config files and three directories
 with names corresponding to these suites: `unit`, `functional`, `acceptance`. Suites are independent groups of tests with a common purpose.
 
+## The Codeception Syntax
+
+Codeception follows simple naming rules to make it easy to remember (as well as easy to understand) its method names.
+
+* **Actions** start with a plain english verb, like "click" or "fill". Examples:
+    ```php
+    $I->click('Login');
+    $I->fillFiled('#input-username', 'John Dough');
+    $i->pressKey('#input-remarks', 'foo');
+    ```
+* **Assertions** always start with "see" or "dontSee". Examples:
+    ```php
+    $I->see('Welcome');
+    $I->seeInTitle('My Company');
+    $i->seeElement('nav');
+    $i->dontSeeElement('#error-message');
+    $i->dontSeeInPageSource('<section class="foo">');
+    ```
+* Methods that just *grab* something off the page, but don't process it, start with "grab". The return value of those are meant to be saved as variables and used later. Example:
+    ```php
+    $method = $I->grabAttributeFrom('#login-form', 'method');
+    $I->assertEquals('post', $method);
+    ```
+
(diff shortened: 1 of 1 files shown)


docs/02-GettingStarted.md
```diff
@@ -5,6 +5,30 @@ and bootstrapped your first test suites. Codeception has generated three of them
 They are well described in the [previous chapter](http://codeception.com/docs/01-Introduction). Inside your __/tests__ folder you will have three `.yml` config files and three directories
 with names corresponding to these suites: `unit`, `functional`, `acceptance`. Suites are independent groups of tests with a common purpose.
 
+## The Codeception Syntax
+
+Codeception follows simple naming rules to make it easy to remember (as well as easy to understand) its method names.
+
+* **Actions** start with a plain english verb, like "click" or "fill". Examples:
+    ```php
```
DavertMik: please add `<?php` our website highlighter is very strict about this :)

docs/02-GettingStarted.md
```diff
@@ -5,6 +5,30 @@ and bootstrapped your first test suites. Codeception has generated three of them
...
+    ```
+* **Assertions** always start with "see" or "dontSee". Examples:
+    ```php
+    $I->see('Welcome');
+    $I->seeInTitle('My Company');
+    $i->seeElement('nav');
+    $i->dontSeeElement('#error-message');
+    $i->dontSeeInPageSource('<section class="foo">');
+    ```
+* Methods that just *grab* something off the page, but don't process it, start with "grab". The return value of those are meant to be saved as variables and used later. Example:
```
DavertMik: We call them **grabbers** :)

docs/02-GettingStarted.md
ThomasLandauer: Done. But I don't like it, since this way it looks like you can use the shown code right below the opening `<?php` tag....

docs/02-GettingStarted.md
ThomasLandauer: Done.

ThomasLandauer:
> No, this term is really important. Actor is the most important class in functional/acceptance tests. It proxies actions to modules and can contain some custom methods.

Yeah, OK, but this is the **Getting Started** page ;-) People want to write their first test, not fully understand the inner workings of the framework.

> It's better to use special term, instead of methods, because not all methods of a Module become actions. For instance, _restart is a method of Webdriver, but is not an action. It can be used only from Helpers

Yeah, that's what I'm saying :-)  
The heading "Actions" (e.g. at http://codeception.com/docs/modules/Symfony#Actions ) is auto-generated somehow? Please change that to "Methods".

DavertMik:
Ok, let's merge it!

Issue #4258, opened 2017-05-25
Add assertion to check HTTP headers


ThomasLandauer:
I can't find any assertion to check if an HTTP header has an expected value in a functional test.

In a PHPUnit test that extends Symfony's `WebTestCase`, you can do:
```php
$this->assertTrue(
    $client->getResponse()->headers->contains(
        'Content-Type',
        'text/javascript; charset=UTF-8'
    )
);
```

I can't even find a method to get the full HTTP response. Since the PhpBrowser Module uses Guzzle, it should be possible to implement something like `getResponse()`?

Naktibalda:
You could use [REST module](http://codeception.com/docs/modules/REST#seeHttpHeader) for that.
Don't forget to configure it with:
```
REST:
  depends: Symfony
```

ThomasLandauer:
Thanks, I'll take a look!

But I have a general question in advance: Can you combine any module with any other? Don't some of them have the same methods, so that they'd override each other? => This should be explained somewhere :-)

Naktibalda:
No, only some modules, like REST and Doctrine can depend on others to get their work done.
There is a ConflictsWithModule interface too, so you can't have WebDriver and some framework module enabled at the same time, and you can't have two framework modules enabled.
Also REST conflicts with SOAP.

It's documented somewhere.

ThomasLandauer:
> It's documented somewhere.

Well, that's not the best sign, if not even you do know where that documentation is ;-)

Seriously, @DavertMik I'd suggest to add a new section to every module's documentation, namely "Compatible with", followed by a bulleted list of compatible modules, with links to them. Can/will you invent a system that somehow generates this list automatically? Or should I start manually?

I would locate it between the "Status" and the "Config" section.

BTW: Is "Status" really necessary anymore? As far as I can see, most are stable anyway. All others could include a red warning bar at the top. Then "Status" could be renamed to "Maintainer" and moved to a sidebar.

Naktibalda:
What do you mean by "compatible"? That they do not conflict?

ThomasLandauer:
Yes.

Naktibalda:
That's the default relation between modules.
There are only 2 groups of modules that conflict:
a) framework modules + WebDriver.
b) REST with SOAP.

All other combinations are allowed.

ThomasLandauer:
[...]

PR #4266, opened 2017-05-27
Starting to implement module compatibility documentation

ThomasLandauer:
See https://github.com/Codeception/Codeception/issues/4258#issuecomment-304460781

### DIFF
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -22,6 +22,10 @@ use Symfony\Component\VarDumper\Cloner\Data;
  *
  * * Maintainer: **raistlin**
  * * Stability: **stable**
+ * 
+ * ## Compatibility
+ *
+ * This module is compatible with all other Codeception modules, except WebDriver and the modules for other PHP frameworks.
  *
  * ## Config
  *
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -16,7 +16,7 @@ use Symfony\Component\VarDumper\Cloner\Data;
  * * Complete: Requests are emulated through Symfony's [DomCrawler](http://symfony.com/doc/current/components/dom_crawler.html) and [HttpKernel](https://symfony.com/doc/2.7/components/http_kernel.html)
  * * Certain parts only: [PhpBrowser Module](http://codeception.com/docs/modules/PhpBrowser) or [WebDriver Module](http://codeception.com/docs/modules/WebDriver) is used for requests
  *
- * ## Complete
+ * ## Functional Tests
  *
  * ### Config
  *
@@ -37,9 +37,9 @@ use Symfony\Component\VarDumper\Cloner\Data;
  *
  * ### Compatibility
  *
- * When loaded this way, this module is compatible with all other Codeception modules, except WebDriver and the modules for other PHP frameworks.
+ * When used in a functional test, this module is compatible with all other Codeception modules, except WebDriver and the modules for other PHP frameworks.
  * 
- * ## Certain Parts Only
+ * ## Acceptance Tests
  * 
  * ### Config
  *
@@ -57,7 +57,7 @@ use Symfony\Component\VarDumper\Cloner\Data;
  * ```
  *
  * Available Parts:
- * * `services`: Access Symfony's services
+ * * `services`: Access Symfony's services. Symfony parameters can be accessed with `$I->grabService('kernel')->getContainer()->getParameter('my_parameter');`
  * * `orm`: Access Symfony's ORM
  *
  * ## Demo Project
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -12,84 +12,67 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\VarDumper\Cloner\Data;
 
 /**
- * This module uses Symfony Crawler and HttpKernel to emulate requests and test response.
+ * This module makes it possible to access Symfony features. It can be loaded in two ways:
+ * * Complete: Requests are emulated through Symfony's [DomCrawler](http://symfony.com/doc/current/components/dom_crawler.html) and [HttpKernel](https://symfony.com/doc/2.7/components/http_kernel.html)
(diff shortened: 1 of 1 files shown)


src/Codeception/Module/Symfony.php
```diff
@@ -22,6 +22,10 @@
  *
  * * Maintainer: **raistlin**
  * * Stability: **stable**
+ * 
+ * ## Compatibility
+ *
+ * This module is compatible with all other Codeception modules, except WebDriver and the modules for other PHP frameworks.
```
DavertMik: Mentioning all possible option would make a mess :(

ThomasLandauer: OK, please tell me! How can you load a module partially??

Naktibalda:
```
Symfony:
  part: orm
```

ThomasLandauer: 1. OK, so how many parts do exist? I see `services` here http://codeception.com/docs/modules/Symfony#Parts and you're telling me `orm`. Anything else?

1. And is it true for all parts that their purpose is compatibility with WebDriver and PhpBrowser? Or do some of them have a different use case?

1. And can you load more than one part? If yes, how's the syntax?

Naktibalda: 1. I didn't check if it exists.
  [Laravel5](http://codeception.com/docs/modules/Yii2#Parts)and Yii2 modules have ORM parts,
  [Yii2](http://codeception.com/docs/modules/Yii2#Parts) has 4 parts actually,
2. No, it wasn't the original purpose. I think that parts functionality was created to reduce the size of generated files. That's why REST module has Xml and Json parts.

ThomasLandauer: See https://github.com/Codeception/Codeception/pull/4266/commits/903222a4e8dc3acccaa53dfdf2a0ba1606e018b5

More questions :-)

4. `var_path`: This is only required for Symfony 3 directory structure? Which path is to be used? Current documentation ("where bootstrap.php.cache is") is ambiguous, since I have multiple `bootstrap.php.cache`s. Best guess (i.e.) suggested value should probably be `var`, not `app`, right? I'm wondering, cause I have `app` in my config (Symfony 3 directory structure), and it still works.
4. What does `environment: 'local_test'` mean?
4. I omitted this link: http://lakion.com/blog/how-did-we-speed-up-sylius-behat-suite-with-blackfire  What's the relation to this module? Does the module use Blackfire?
4. Is it true for all parts that they are to be loaded in `acceptance.suite.yml`?
4. When loading a Symfony part, the other Symfony parameters (app_path, var_path, etc.) are not required? Or is `funcitonal.suite.yml` always required and they are taken from there?
4. `part: orm` only makes sense in combination with Doctrine? Or it there anything like a Symfony ORM by itself?
4. When loading a Symfony Part, it's not just *possible* but *mandatory* to load one of [PhpBrowser, WebDriver]?

Please merge what I have so far (if it doesn't contain a severe error) - I need to see a preview of what I already have..... :-)

Naktibalda: `4.` @DavertMik should know, because he implemented it.
`7.` no, features are not tied to specific suite names.
`8.` these parameters aren't required, because they have default values.
`9.` part names are module specific, please refer to Parts section of documentation.
`10.` no.

src/Codeception/Module/Symfony.php
```diff
@@ -12,80 +12,67 @@
...
+ *             var_path: 'app' # Required for Symfony 3 directory structure only
+ *             # Optional parameters:
+ *             environment: 'local_test'
+ *             debug: true
+ *             em_service: 'doctrine.orm.entity_manager' # Use a different than the standard EntityManager to pair with Doctrine Module
+ *             cache_router: 'false' # Enable router caching between tests in order to increase performance
+ *             rebootable_client: 'true' # Reboot kernel before each request
+ * ```
  *
- * * app_path: 'app' - specify custom path to your app dir, where bootstrap cache and kernel interface is located.
```
DavertMik: No, the list of configuration options should be kept!

src/Codeception/Module/Symfony.php
```diff
@@ -12,80 +12,67 @@
...
+ *         - Symfony:
+ *             part: 'services' # 'orm' or 'services'
+ *         - Doctrine2:
+ *             depends: Symfony
+ *         - WebDriver:
+ *             url: http://www.example.com/
+ *             browser: chrome
  * ```
  *
- * ### Symfony 3.x Directory Structure
```
DavertMik: I think Symfony 3 notice should be kept as well

src/Codeception/Module/Symfony.php
```diff
@@ -12,80 +12,67 @@
 use Symfony\Component\VarDumper\Cloner\Data;
 
 /**
- * This module uses Symfony Crawler and HttpKernel to emulate requests and test response.
+ * This module makes it possible to access Symfony features. It can be loaded in two ways:
+ * * Complete: Requests are emulated through Symfony's [DomCrawler](http://symfony.com/doc/current/components/dom_crawler.html) and [HttpKernel](https://symfony.com/doc/2.7/components/http_kernel.html)
+ * * Certain parts only: [PhpBrowser Module](http://codeception.com/docs/modules/PhpBrowser) or [WebDriver Module](http://codeception.com/docs/modules/WebDriver) is used for requests
```
DavertMik: Ok, I got the idea but I don't like the proposed structure. You explain configuration from the technical point but not from point of a use case. 

Instead of "Complete" it should be called: "Functional Testing", with the configuration for functional tests.
Then explain how to load Symfony for acceptance tests with WebDriver or PhpBrowser.

src/Codeception/Module/Symfony.php
ThomasLandauer: I just moved them into the codeblock; everything is still there :-)

src/Codeception/Module/Symfony.php
ThomasLandauer: I found only one difference to Symfony 2, and explained it in the codeblock too.

src/Codeception/Module/Symfony.php
ThomasLandauer: Yeah!! I fully understood this only afterwards... :-)

DavertMik:
This is outdated due to Symfony4 support

PR #4267, opened 2017-05-28
Overview table for test suites

ThomasLandauer:
This is meant to replace the current "Pros" and "Cons" sections, for better overview.

Points which I left out:
* "can be shown to your clients and managers": Doesn't make sense to me, since you can anything to anyobdy - as long as they are willing to take a look ;-)
* "most stable in support: less affected by changes in source code or technologies": Not true for me, since acceptance test is most affected by changes in *browser* source code or *JavaScript* technologies
* "fewer checks can lead to false-positive results": Don't understand
* "not stable in execution: rendering and JavaScript issues can lead to unpredictable results": You mean that Selenium is sometimes running unstable? I wouldn't mention that here.
* "can provide more detailed reports": Don't understand. Which reports?
* "you can still show this code to managers and clients": Contradiction to above :-)
* "stable enough: only major code changes, or moving to other framework, can break them": Well, any error can break them - that's the very purpose of having tests ;-)
* "by emulating the browser you might get more false-positive results": Don't understand. Which for example?
* "can cover rarely used features": True for any test, if you include those rarely used features ;-)
* "can test the stability of the application core" and "fundamental for any developer": Not so important IMHO. If you have functional tests for every scenario, you don't need unit tests ;-)
* "unstable in support: very sensitive to code changes": Yeah, but that's the intended behavior ;-)

### DIFF
--- a/docs/03-AcceptanceTests.md
+++ b/docs/03-AcceptanceTests.md
@@ -25,6 +25,15 @@ $I->see('Welcome, Davert!');
 ```
 
 **This scenario can be performed either by a simple PHP Browser or by a browser with Selenium WebDriver**.
+
+| | PhpBrowser | WebDriver |
+| --- | --- | --- |
+| Browser Engine | Symfony BrowserKit | Firefox or Chrome |
+| JavaScript | No | Yes |
+| `see`/`seeElement` checks if… | …text is present in source code | …text is actually visible to the user |
+| Read HTTP response headers | Yes | No |
+| Speed | Fast | Slow |
+
 We will start writing our first acceptance tests with a PhpBrowser.
 
 ## PHP Browser
--- a/docs/03-AcceptanceTests.md
+++ b/docs/03-AcceptanceTests.md
@@ -28,7 +28,7 @@ $I->see('Welcome, Davert!');
 
 | | PhpBrowser | WebDriver |
 | --- | --- | --- |
-| Browser Engine | Guzzle + Symfony BrowserKit | Firefox or Chrome |
+| Browser Engine | Guzzle + Symfony BrowserKit | Chrome or Firefox |
 | JavaScript | No | Yes |
 | `see`/`seeElement` checks if… | …text is present in source code | …text is actually visible to the user |
 | Read HTTP response headers | Yes | No |
@@ -335,16 +335,17 @@ Selenium Server is a piece of software that "drives" (i.e. runs) your local brow
 
 Overview of components:
 ```
-Codeception -> facebook/php-webdriver -> Selenium Server -> Gecko/Chrome Driver -> Browser (Firefox or Chrome)
+Codeception -> facebook/php-webdriver -> Selenium Server -> Gecko/Chrome Driver -> Browser (Chrome or Firefox)
 ```
 
 Installation instructions:
 
-1. You need [Firefox](http://www.mozilla.com/) and/or [Chrome](https://www.google.com/chrome/browser/desktop/) :-)
+1. You need [Chrome](https://www.google.com/chrome/browser/desktop/) and/or [Firefox](http://www.mozilla.com/) :-)
 1. You need [Java](http://www.java.com/)
 1. Download [Selenium Standalone Server](http://www.seleniumhq.org/download/)
-1. Download [geckodriver](https://github.com/mozilla/geckodriver/releases) (for Firefox) and/or
-[ChromeDriver](https://sites.google.com/a/chromium.org/chromedriver/) and extract the executable to a folder where Selenium Server can find it
+1. Download [ChromeDriver](https://sites.google.com/a/chromium.org/chromedriver/)
+and/or [geckodriver](https://github.com/mozilla/geckodriver/releases) (for Firefox)
+and extract the executable to a folder where Selenium Server can find it
 (best guess: just drop it in the same folder where `selenium-server-standalone-xxx.jar` lives)
 1. Start Selenium Server with:
     ```bash
--- a/docs/03-AcceptanceTests.md
+++ b/docs/03-AcceptanceTests.md
@@ -28,7 +28,7 @@ $I->see('Welcome, Davert!');
 
 | | PhpBrowser | WebDriver |
 | --- | --- | --- |
-| Browser Engine | Symfony BrowserKit | Firefox or Chrome |
+| Browser Engine | Guzzle + Symfony BrowserKit | Firefox or Chrome |
 | JavaScript | No | Yes |
 | `see`/`seeElement` checks if… | …text is present in source code | …text is actually visible to the user |
 | Read HTTP response headers | Yes | No |
--- a/docs/01-Introduction.md
+++ b/docs/01-Introduction.md
@@ -16,7 +16,21 @@ To test the behavior of your application as a whole, you should write functional
 Codeception supports all three testing types.
 Out of the box you have tools for writing unit, functional, and acceptance tests in a unified framework.
 
-Let's review the listed testing paradigms in reverse order.
+| | Codeception Unit Tests | Codeception Functional Tests | Codeception Acceptance Tests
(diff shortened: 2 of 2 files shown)


ThomasLandauer:
The aspect that you can run acceptance tests with PhpBrowser is missing! Please tell me: What's the difference/advantage between an acceptance test with PhpBrowser and a functional test?

ThomasLandauer:
Installation instructions: Docker is missing. Please tell me: What is Docker?

docs/03-AcceptanceTests.md
```diff
@@ -25,6 +25,15 @@ $I->see('Welcome, Davert!');
 ```
 
 **This scenario can be performed either by a simple PHP Browser or by a browser with Selenium WebDriver**.
+
+| | PhpBrowser | WebDriver |
+| --- | --- | --- |
+| Browser Engine | Symfony BrowserKit | Firefox or Chrome |
```
DavertMik: Guzzle + Symfony BrowserKit

docs/03-AcceptanceTests.md
```diff
@@ -310,13 +319,65 @@ $I->seeInCurrentUrl('user/1');
...
+```
+Codeception -> facebook/php-webdriver -> Selenium Server -> Gecko/Chrome Driver -> Browser (Firefox or Chrome)
+```
+
+Installation instructions:
+
+1. You need [Firefox](http://www.mozilla.com/) and/or [Chrome](https://www.google.com/chrome/browser/desktop/) :-)
+1. You need [Java](http://www.java.com/)
+1. Download [Selenium Standalone Server](http://www.seleniumhq.org/download/)
+1. Download [geckodriver](https://github.com/mozilla/geckodriver/releases) (for Firefox) and/or
```
DavertMik: Please make Chrome to be more priority than Firefox
Current WebDriveri support in Firefox is not really good...

docs/03-AcceptanceTests.md
```diff
@@ -310,13 +319,65 @@ $I->seeInCurrentUrl('user/1');
...
+[ChromeDriver](https://sites.google.com/a/chromium.org/chromedriver/) and extract the executable to a folder where Selenium Server can find it
+(best guess: just drop it in the same folder where `selenium-server-standalone-xxx.jar` lives)
```
DavertMik: Better to drop it into $PATH
Otherwise you need to specify path to the executable like this: `-Dwebdriver.chrome.driver=./chromedriver`
Anyway, this documentation was written in [WebDriver module reference](http://codeception.com/docs/modules/WebDriver#Local-Testing)

Maybe we should not duplicate it and have one source?

docs/03-AcceptanceTests.md
```diff
@@ -310,13 +319,65 @@ $I->seeInCurrentUrl('user/1');
...
-## Selenium WebDriver
+## WebDriver: Selenium Server or PhantomJS
 
 A nice feature of Codeception is that most scenarios are similar, no matter of how they are executed.
 `PhpBrowser` was emulating browser requests but how to execute such test in a real browser like Chrome or Firefox? 
 Selenium WebDriver can drive them so in our acceptance tests we can automate scenarios we used to test manually.
 In such tests we should concentrate more on **testing the UI** than on testing functionality.
 
+"[Selenium WebDriver](http://www.seleniumhq.org/projects/webdriver/)" is the name of an API (specified by the Selenium project)
+to drive browsers automatically. Codeception supports two implementations of this API: Selenium Standalone Server and PhantomJS
```
DavertMik: To be honest, PhantomJS is deprecated, not maintained, and soon will be overtaken by headless Chrome.
You can connect to Chrome via ChromeDriver natively without Selenium Server, and facebook/webdriver supports this.
This is how Laravel Dusk works. However, people still use Phantom because it is simpler. We keep it in reference but better not to recommend it that much. If it is possible - try to run Codeception with [headless Chrome](https://developers.google.com/web/updates/2017/04/headless-chrome) (didn't have chance to try it yet)

docs/03-AcceptanceTests.md
ThomasLandauer: Done :-)

docs/03-AcceptanceTests.md
ThomasLandauer: Done :-)

docs/03-AcceptanceTests.md
ThomasLandauer: On Windows it works if it's in the same directory. Not so in Linux?  
(Idea: No reason to create a directory for a single file, which is **only** used by Selenium anyway)

WebDriver: Yeah, but I only noticed that afterwards ;-) However, I'm going to rewrite (i.e. mostly delete) this section in the Module.  
Reason: Installation is really a headache!!! I spent hours (if not days) to get all this shit (Selenium etc.) up and running. This very page would probably be the only resource on the entire web to explain **everything** (not just some ridiculous part). And here is the better place than in the Module, cause: If people can't get the entire package up and running in 15 minutes, they're frustrated and never going to look into specific modules. At that stage, they don't even know what "module" means....

docs/03-AcceptanceTests.md
ThomasLandauer: OK, didn't know that!  
But the solution can't be to conceal PhantomJS, since I bet there are 100 pages out there, claiming that Codeception works with PhantomJS....  
The solution is to write in the docs what you are writing here! => I'll do that. But some questions first:

1. Is there any (sort of) "official" source claiming that Phantom is dead?
1. Current version of Chrome (Windows) is 58 (Linux too?). Headless is included in 59 for Linux. Do you have an idea when this will be released? In days, weeks or months?
1. Do you have an idea when Headless for Windows will be released?

Naktibalda: 1. One maintainer stepped down after headless chrome announcement. I think that PhantomJS had more than one maintainer.

Naktibalda:
Seriously? Goggle it up.

On 29 May 2017 1:04 a.m., "Thomas Landauer" <notifications@github.com>
wrote:

Installation instructions: Docker is missing. Please tell me: What is
Docker?

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/Codeception/Codeception/pull/4267#issuecomment-304548820>,
or mute the thread

docs/03-AcceptanceTests.md
DavertMik: @Naktibalda I listened to interview with him. He was the **last** maintainer. The codebase is so complicated that no one could take it. He also suggests switching to Chrome... https://groups.google.com/forum/#!msg/phantomjs/9aI5d-LDuNE/5Z3SMZrqAQAJ

2. Not sure. Anyway, it won't be used widely less than in 6 months
3. No.

So I agree PhantomJS is a good solution at current state. But we build documentation for the future so we should at least know that things will change soon and our docs will be outdated

ThomasLandauer: OK, thanks! 2 things to PhantomJS - users' perspective:

* It's installed in 2 minutes. And the transition to Chrome is probably straightforward too - since no changes in the test files (Codeception syntax) are necessary.

* Noteworthy is the fact that is just doesn't make sense anymore to "debug" test code to get it running on Phantom. So the message is: If Phantom works for you, keep it for the moment. If it doesn't work, don't try to get it to work, but rather go with Selenium for now.

I'll include such a note. But please merge all pull requests first - I'm again losing track of what the status is and what needs to be done....

Naktibalda: Oh, OK,  I've looked at releases page, the last release was made in January 2016. 
Looks really dead to me.

DavertMik:
> So the message is: If Phantom works for you, keep it for the moment. If it doesn't work, don't try to get it to work, but rather go with Selenium for now.

Ok, I will add this notice to the page

PR #4272, opened 2017-05-29
Use WebDriver instead of PhpBrowser to allow usage of other Drivers

maciejmrozinski:


Naktibalda:
Please explain what are you trying to do here.

maciejmrozinski:
I want to use Facebook module, but with Chrome WebDriver and not PHPBrowser. When i declare my config as using WebDriver and that Facebook module depends on WebDriver i get an error that PHPBrowser was expected.

Facebook module uses common methods from WebDriver module, so no need to strict it to PHPBrowser only.

Naktibalda:
But your change restricts Facebook module to WebDriver, so it no longer works with PhpBrowser.

Please make it work with both.

src/Codeception/Module/Facebook.php
```diff
@@ -121,10 +121,10 @@ public function _requires()
 
     public function _depends()
     {
-        return ['Codeception\Module\PhpBrowser' => $this->dependencyMessage];
+        return ['Codeception\Module\WebDriver' => $this->dependencyMessage];
```
DavertMik: Please use `Codeception\Lib\Interfaces\Web`, as it will match both PhpBrowser and WebDriver

maciejmrozinski:
To cover all used methods, `Codeception\Lib\Interfaces\Web` and `Codeception\Lib\Interfaces\Remote` need to be checked. What do You think?

src/Codeception/Module/Facebook.php
```diff
@@ -121,11 +123,17 @@ public function _requires()
 
     public function _depends()
     {
-        return ['Codeception\Module\PhpBrowser' => $this->dependencyMessage];
+        return ['Codeception\Lib\Interfaces\Remote|Codeception\Lib\Interfaces\Remote' => $this->dependencyMessage];
```
DavertMik: I think it won't work this way :(
Our parser is not so smart to handle both interfaces... Especially two the same interface...
Let's ask for `Codeception\Lib\Interfaces\Web` interface here

src/Codeception/Module/Facebook.php
```diff
@@ -121,11 +123,17 @@ public function _requires()
...
     }
 
-    public function _inject(PhpBrowser $browserModule)
+    public function _inject($browserModule)
     {
+        if (!$browserModule instanceof Remote || !$browserModule instanceof Web) {
```
DavertMik: And check if passed a browserModule is an instance of `Remote`, if not throw extension as you do it here

Naktibalda:
Any updates?

Issue #4280, opened 2017-05-30
[AMQP] Infinity channel creation


Tekill:
#### What are you trying to achieve?
Remove infinity channel creation or make it optional. We have about 700+ tests with enabled AMQP module with `cleanup=true` option. When we cleanup all queues before test module send one request per queue. Every request create new channel and if you have a lot of queues (~20) it will create 14 000+ active channels in RAM memory. These channels die only after execution all test so we have huge memory overhead and CPU usage.

#### What do you get instead?
New channel per method call, 20 000+ active channels in RAM memory, huge CPU usage. When the RabbitMQ Server breaks memory threshold, raises a memory alarm and blocks all connections that are publishing messages.  https://www.rabbitmq.com/memory.html

### Details

* Codeception version: 2.2
* PHP Version: 7.1
* Operating System: Ubuntu 16.04, RabbitMQ 3.6.9
* Installation type: Composer
* List of installed packages (`composer show`)
* Suite configuration:

```yml
modules:
         enabled:
             - AMQP:
                 host: 'localhost'
                 port: '5672'
                 username: 'guest'
                 password: 'guest'
                 vhost: '/'
                 queues: 
                     - one
                     - two
                     ...
                     - twenty
```

Tekill:
Will be fixed in next release.

PR #4281, opened 2017-05-30
Added AMQP Module option to use only one channel

Tekill:
This PR add new option which fix this issue: 
#4280

### DIFF
--- a/src/Codeception/Module/AMQP.php
+++ b/src/Codeception/Module/AMQP.php
@@ -1,15 +1,15 @@
 <?php
 namespace Codeception\Module;
 
+use Codeception\Exception\ModuleException as ModuleException;
 use Codeception\Lib\Interfaces\RequiresPackage;
 use Codeception\Module as CodeceptionModule;
-use Codeception\Exception\ModuleException as ModuleException;
 use Codeception\TestInterface;
 use Exception;
 use PhpAmqpLib\Channel\AMQPChannel;
 use PhpAmqpLib\Connection\AMQPStreamConnection;
-use PhpAmqpLib\Message\AMQPMessage;
 use PhpAmqpLib\Exception\AMQPProtocolChannelException;
+use PhpAmqpLib\Message\AMQPMessage;
 
 /**
  * This module interacts with message broker software that implements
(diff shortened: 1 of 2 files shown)


Naktibalda:
Are there any disadvantages of using only one channel compared to using a new channel for each action?

Tekill:
Current code does not delete old created channels so we have to keep in memory what:
1. Channel count is limited and probably reachable (standard 65535);
1. Every time spends CPU time and available RAM for rabbitmq decreases. Low available memory will cause memory alarm.

In previous commit I made mistake because rabbitmq may sent signal "channel_close" (e.g when to purge non-existent queue). Fortunately php-amqplib has build in channel manager and we can delegate that headache to him. Now it stable and may handle wrong module usage.
I ran tests with this fix (600 tests cases, 20+ queues) and it reduced total test execution time (from 12m to 10m) with rabbitmq memory usage (from 350mb to 1mb). I see only advantages in this optimisation but made this feature disabled by default. Old code will work as worked even after updating.

Tekill:
Ideally, I would like this option to be enabled by default, but it takes more time to check any possible cases. Using a new channel for each action eliminate any side effects.

Naktibalda:
@Tekill Please rebase on 2.3 branch to get the tests passing.

Tekill:
@Naktibalda mission complete.

src/Codeception/Module/AMQP.php
```diff
@@ -47,12 +49,13 @@
...
-        'port'     => '5672',
-        'vhost'    => '/',
-        'cleanup'  => true,
+        'host'             => 'localhost',
+        'username'         => 'guest',
+        'password'         => 'guest',
+        'port'             => '5672',
+        'vhost'            => '/',
+        'cleanup'          => true,
+        'only_one_channel' => false
```
DavertMik: `single_channel` seems nicer to me.
Could you change it, please?

Tekill: Yep, il do it at evening.

Tekill:
@DavertMik option has been renamed.

DavertMik:
Thanks!

Issue #4277, opened 2017-05-30
Symfony/console 3.3.0 broke `--xml` and other parameters with optional default value


Naktibalda:
https://travis-ci.org/Codeception/Codeception/jobs/237467274#L2737

```
1) IncludedCest: Run included with xml output
 Test  tests/cli/IncludedCest.php:runIncludedWithXmlOutput
 Step  See file found "report.json","tests/_output"
 Fail  File "report.xml" not found at ""

Scenario Steps:
 4. $I->seeFileFound("report.xml") at tests/cli/IncludedCest.php:42
 3. $I->amInPath("_log") at tests/cli/IncludedCest.php:41
 2. $I->executeCommand("run --xml") at tests/cli/IncludedCest.php:40
 1. $I->amInPath("tests/data/included") at tests/cli/IncludedCest.php:16
```

https://github.com/symfony/console/blob/master/CHANGELOG.md#330

> [BC BREAK] Input::getOption() no longer returns the default value for options with value optional explicitly passed empty

DavertMik:
Ok, I will try to fix that today and release a new version quickly

magroski:
Any progress on this issue?

Naktibalda:
3.3.0-beta was released on the 1st of March,
3.3.0-rc1 - 2 weeks ago.
We need a build with unstable dependencies to get early warning about such changes in the future.

Naktibalda:
1st of May, not March.

acuthbert:
Just in case this issue is breaking deployments and have just skimmed the detail here - the fastest workaround is just to supply the default values to your switches explicitly - that'll get things moving again. For me this was changing --xml to --xml report.xml

PR #4285, opened 2017-05-31
Temporary constraint for symfony/console to stop builds breaking

Naktibalda:
#4277

A real fix is necessary, but this is good as a temporary measure.


Naktibalda:
@DavertMik Please remember to undo it after fixing commands.

DavertMik:
Ok

PR #4290, opened 2017-05-31
Don’t skip empty strings in headers.

### LINKED ISSUE #4287: Empty values for headers are skipped
I'm trying to set a `DNT` header to the value of `0`, as it is described in the specification:

    $I->haveHttpHeader('DNT', '0');

#### What do you get instead?
The `DNT` header is not set to the server as it is skipped.

### Details
The problem is in this line of code: https://github.com/Codeception/Codeception/blob/b54eaf4007484f36145c1dc8c64da1874adbc340/src/Codeception/Lib/InnerBrowser.php#L156-L158.

The string value `0` is evaluated as false and will not be send.
A possible solution would be to check if the value is a string, and if it is use it anyway. I could create a pull request for this, if it is an acceptable fix for you.

tobiastom:
This will close #4287.

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -153,7 +153,7 @@ class InnerBrowser extends Module implements Web, PageSourceSaver, ElementLocato
 
         foreach ($this->headers as $header => $val) { // moved from REST module
 
-            if (!$val) {
+            if ($val === null || $val === '') {
                 continue;
             }
 


Naktibalda:
I find it ironic that it still skips empty string ''.

tobiastom:
That's the intended behaviour, right?

For me I din't see why this check is there in the first place. Either I set a header, or I delete it. When I delete it, `deleteHeader` will unset it.

DavertMik:
Thanks!

tobiastom:
Thanks you the merge! :-)

Issue #4270, opened 2017-05-28
Suggestion: New chapter "IDE integration" in the docs


ThomasLandauer:
What about a tutorial on how to configure some popular IDE's for work with Codeception? I could provide NetBeans explanation. Probably an own "Guides" page would be the best place? Can I create one myself?

DavertMik:
Yes, please. That would be great. 
I think we can handle writing PHPStorm chapter as well.

ThomasLandauer:
OK. Can/should I create the page myself (where)? Or will you give me a blank file - something like `13-IDEIntegration.md`?

DavertMik:
Probably you can... But how about making this inside `Getting Started` chapter? I think it will be a good guide for newbies

ThomasLandauer:
Let's see the content first, and decide then where to place it :-)

I have some questions/problems with NetBeans:

1. It is possible to run *all* tests (Ctrl+F6). But Strg+F6 ("Test File") is not working as I would have expected. I get the error `Test file for the selected source file was not found.` From that I'm infering that the idea would be to select a source file (no test!) and then run its *associated* test file. I have no idea how this "association" is defined/infered. Do you? For acceptance and functional tests it's probably hopeless, since there's no 1:1 relation between source files to test files. But maybe for unit tests?

1. When running a single test method (right-click, then "Run Focused Test Method"), this command gets issued:
    ```bash
    "C:\...\bin\codecept.bat" "--ansi" "--no-interaction" "run" "src/AppBundle/Tests/functional/AbcCest.php:tryIndex" "--xml=C:\Users\...\Temp\nb-codeception-log.xml"
    ```
    which leads to this error:
    ```bash
    [RuntimeException] Suite 'src/AppBundle/Tests/functional/AbcCest.php:tryIndex' could not be found
    ```
    I guess the problem is that `-c src/AppBundle` is missing. Is this Symfony-specific? And is it a NetBeans or a Codecption bug?

1. Bug: `[ConnectionException]`s are not correctly reported to NetBeans. Details: When the Selenium Server is not started, in NetBeans "Test Results" window (which is the one that's comming up when you start the tests), 100% are green. You have to manually open the "Output" window to see that all acceptance tests gave an error.  
On the other hand, if Codeception throws a `[RuntimeException]` it is communicated correctly to NetBeans (resulting in the test file being marked as "Failed").

ThomasLandauer:
[...]

PR #4302, opened 2017-06-04
Added NetBeans section

ThomasLandauer:
See https://github.com/Codeception/Codeception/issues/4270

### DIFF
--- a/docs/13-IDEIntegration.md
+++ b/docs/13-IDEIntegration.md
@@ -1,0 +1,32 @@
+# IDE Integration
+
+Several major IDE's feature a Codeception integration.
+
+## NetBeans 8.2
+
+### What You'll Get
+
+* Run all tests (i.e. `codecept run`) from NetBeans
+* Run the currently focused test method. It is not possible to run just one suite (unit, functional, acceptance) or just one test file.
+* View the test results in NetBeans ("Test Results" window and "Output" window)
+* In "Projects" window, test files are displayed in a folder "Test Files" (i.e. moved outside of "Source Files")
+
+### Setup
+
+In "Projects" window, right-click on your project > `Properties` > `Testing`, then:
+
+1. `Add Folder...` and select all folders containing test files. Each of these will be (visually) moved into a separate "Test Files" folder in "Projets" window.
+1. Check `Codeception`
+1. `Codeception` > `Use Custom Codecept File`: Enter the path to your Codeception executable (`codecept`, `codecept.bar` or `codecept.phar`).  
+If you have Codeception installed globally on your machine, you may instead provide the path in `Tools` > `Options` > `PHP` > `Frameworks & Tools` > `Codeception` > `Codecept File`.
+1. `Codeception` > `Use Custom codeception.yml`: Enter the path to your `codeception.yml`
+1. `Codeception` > `Ask for Additional Parameters Before Running Tests`: If you check this, a dialogue box will appear every time you run the tests, allowing you to enter additional Codeception command-line parameters.
+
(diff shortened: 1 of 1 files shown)


DavertMik:
Hey, thanks for the PR and sorry for the delay for a respond. 
This documentation hits the specific version of one IDE and I think it is better to make it as a blogpost.

Also, I think we should include some animated GIFs showing the exact parts of this documentation. It is much easier to see it once than to follow the boring guides )

ThomasLandauer:
Blogpost: No, please, make it part of the docs! You initially wanted it on the "Getting Started" page, remember? https://github.com/Codeception/Codeception/issues/4270  
It covers the *curent* version of NetBeans, and the idea is to keep it updated (i.e. docs) with new NetBeans version, not frozen (i.e. Blogpost). Also the idea is to get other IDE's included as well; you said you could cover PhpStorm.

DavertMik:
I really like this document but I'm not sure about using the same format we have for all guides in docs: with brief intro, text, conclusion, etc. Also, there are some very specific parts, like 

> Caution: There's a bug in "Test Results" window: If you run acceptance tests and Selenium server is not started, the resulting ConnectionException isn't reported correctly. Result: All acceptance tests are silenty skipped and no error is shown, even though in fact no test was executed! See NetBeans Bug 270802 Codeception 'ConnectionException's not showing up in 'Test Results' window

...and pointing to the exact issue of NetBeans. This will likely be changed in future as maybe interface elements. Well, core documentation should be more stable to changes. Especially, external changes.

I think it should be a more basic document which can be easily referred to. 
What about a new format? Something similar to codeception.com/for/symfony pages?

Maybe we could create separate pages at codeception.com, like:

```
codeception.com/ide/Netbeans
codeception.com/ide/PhpStorm
```

Those pages will be referenced in Getting Started guide, so anyone could configure IDE of their own choice. 

```
Codeception is nicely played with:

* [NetBeans](http://codeception.com/ide/Netbeans)
* [PhpStorm](http://codeception.com/ide/PhpStorm)
```

It also can be listed at `Docs` section in the site menu.

Does it sound ok?

ThomasLandauer:
Too many details:  
Well, that's always the tradeoff of documentation ;-) If it's too precise, it might get outdated. And if it's too general, it's useless.  
I think: It's unlikely that they (NetBeans) will redo this feature *completely* (i.e. change more than some details). And if they do, well, we must hope that somebody will update this page too (that's why I want to have it in the repo).

codeception.com/ide/:  
I think it's not worth it for three reasons:
* Nobody is gonna decide on a testing framework based on the question of IDE integration. But they will base it on framework integration! That's why codeception.com/for/ is great!
* The page is not that long ;-) And there are no higher "concepts" you need to understand (as in codeception.com/for/). So I thought just add the other IDE's on the same page, and that's it....
* If there *is* already a page, it's liklier that somebody will add his IDE (Sublime, whatever,...), than if they need to create a very new page.

DavertMik:
> If there is already a page, it's liklier that somebody will add his IDE (Sublime, whatever,...), than if they need to create a very new page.

Agree on all points!

So let's make an IDE page inside `codeception.com` repo (with a link to edit it and add more IDEs).
And move the IDE documentation to that repository. I still think it should be a part of website, not part of a package for better maintainability.

Just resend this PR into `ide.md` page for https://github.com/Codeception/codeception.github.com
I will continue work by adding the PhpStorm into it and adding the page to the menu

ThomasLandauer:
Done :-) https://github.com/Codeception/codeception.github.com/pull/142

Issue #4306, opened 2017-06-06
Symfony 3.3 & REST module error


Korikulum:
After updating to Symfony 3.3, tests using the REST module started throwing the following error:
`PHP Warning:  Uncaught PHPUnit_Framework_Exception: Object of class Symfony\Component\VarDumper\Cloner\Stub could not be converted to string`.

I've created an example project: https://github.com/Korikulum/codeception-symfony-3.3-example

> The full output

```bash
Codeception PHP Testing Framework v2.3.3
Powered by PHPUnit 5.7.20 by Sebastian Bergmann and contributors.

Functional Tests (1) ----------------------------------------------------------------------------------------------------------------------------------
Modules: \Helper\Functional, Asserts, Symfony, REST
-------------------------------------------------------------------------------------------------------------------------------------------------------
Testing functional
DefaultControllerCest: Index action
Signature: tests\AppBundle\Controller\DefaultControllerCest:indexAction
Test: tests/functional/AppBundle/Controller/DefaultControllerCest.php:indexAction
Scenario --
 I send get "/"
  [Request] GET http://localhost:8000/
  [Request Headers] []
  [Page] http://localhost:8000/
  [Response] 200
  [Request Cookies] []
  [Response Headers] {"cache-control":["no-cache, private"],"date":["Tue, 06 Jun 2017 14:15:30 GMT"],"content-type":["text/html; charset=UTF-8"],"x-debug-token":["6e7460"]}
PHP Warning:  Uncaught PHPUnit_Framework_Exception: Object of class Symfony\Component\VarDumper\Cloner\Stub could not be converted to string

/var/www/symfony-3.3/tests/_support/_generated/FunctionalTesterActions.php:3163
/var/www/symfony-3.3/tests/functional/AppBundle/Controller/DefaultControllerCest.php:11

  thrown in /var/www/symfony-3.3/vendor/codeception/codeception/src/Codeception/Subscriber/ErrorHandler.php on line 75
PHP Stack trace:
PHP   1. {main}() /var/www/symfony-3.3/vendor/codeception/codeception/codecept:0
PHP   2. Codeception\Application->run() /var/www/symfony-3.3/vendor/codeception/codeception/codecept:36
PHP   3. Symfony\Component\Console\Application->run() /var/www/symfony-3.3/vendor/codeception/codeception/src/Codeception/Application.php:103
[...]

PR #4309, opened 2017-06-07
Symfony::extractRawRoles() failed on security collector from Symfony >= 3.3

Basster:
When extracting roles from security collector, you can simply utilize the getValue method of Data class, when it exists.

### DIFF
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -521,8 +521,7 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
      */
     private function extractRawRoles(Data $data)
     {
-        // prior to Symfony 3.3
-        if (method_exists($data, 'getValue')) {
+        if ($this->dataIsFromSymfony3_3($data)) {
             $roles = $data->getValue();
         } else {
             $raw = $data->getRawData();
@@ -578,5 +577,17 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
         if ($this->client) {
             $this->client->rebootKernel();
         }
+    }
+
+    /**
+     * Public API from Data changed from Symfony 3.2 to 3.3.
+     *
+     * @param \Symfony\Component\VarDumper\Cloner\Data $data
+     *
+     * @return bool
+     */
+    private function dataIsFromSymfony3_3(Data $data): bool
+    {
+        return method_exists($data, 'getValue');
     }
 }
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -521,7 +521,7 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
      */
     private function extractRawRoles(Data $data)
     {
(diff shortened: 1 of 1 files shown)


Naktibalda:
@Basster Does this patch fix #4306 or is it a different issue?

src/Codeception/Module/Symfony.php
```diff
@@ -521,9 +521,15 @@ protected function debugResponse($url)
      */
     private function extractRawRoles(Data $data)
     {
-        $raw = $data->getRawData();
+        // prior to Symfony 3.3
```
Naktibalda: Are you sure that this comment is correct?
`getRawData` method is deprecated since 3.3 and [removed in 4.0](https://github.com/symfony/var-dumper/commit/18e4a4c30f10f5879129cbcf73c4a5b9af745abc).

Basster:
```
diff --git a/src/Symfony/Component/VarDumper/Cloner/Data.php b/src/Symfony/Component/VarDumper/Cloner/Data.php
index 1c6dd0df86..5a6997cf25 100644
--- a/src/Symfony/Component/VarDumper/Cloner/Data.php
+++ b/src/Symfony/Component/VarDumper/Cloner/Data.php
@@ -16,7 +16,7 @@ use Symfony\Component\VarDumper\Caster\Caster;
 /**
  * @author Nicolas Grekas <p@tchwork.com>
  */
-class Data
+class Data implements \ArrayAccess, \Countable, \IteratorAggregate
 {
     private $data;
     private $position = 0;
@@ -34,10 +34,146 @@ class Data
     }
 
     /**
+     * @return string The type of the value.
+     */
+    public function getType()
+    {

```
`getType()` was added with 3.3. On the same release the deprecation warning ws added

Basster:
@Naktibalda I didn't submitted an issue for this PR but it looks like this may fix #4306 as well. Everytime a user is somehow authenticated (even with an AnonymousToken) this branch of the `debugResponse()` gets executed. So, yes - without further investigation, this will fix #4306 as well.

src/Codeception/Module/Symfony.php
```diff
@@ -573,4 +578,16 @@ public function rebootClientKernel()
...
     }
+
+    /**
+     * Public API from Data changed from Symfony 3.2 to 3.3.
+     *
+     * @param \Symfony\Component\VarDumper\Cloner\Data $data
+     *
+     * @return bool
+     */
+    private function dataIsFromSymfony33(Data $data)
```
Naktibalda: I don't like this name because it will be confusing when Symfony 4.0 comes out.

Korikulum:
@Naktibalda @Basster Confirmed, this fixes https://github.com/Codeception/Codeception/issues/4306 .

src/Codeception/Module/Symfony.php
Basster: I'm with you...
> There are only two hard things in Computer Science: cache invalidation and naming things. -- Phil Karlton

Made the name a little more generic

PR #4313, opened 2017-06-07
Grab multiple rows from a database

### LINKED ISSUE #3813: Grab multiple rows from a database
I'm trying to add some object with parameters which depends on some tables in a database. For example, let's call the object – 'car' and one of parameters call 'type'. The 'type' can't be managed from UI, it's already filled in the database. When I generate the 'car' I should take one of parameters from the database. Of course, I can look at the database manually and write my faker method to generate the 'type' but it's not very good solution because that data can be changed. In this case, I should take all available types from the database and select one of them.

Unfortunately, the Codeception don't have a method to get more than one row from a database. If I use `$I->grabFromDatabase('types', 'id');`, I'll always get the first row. I write my own method in new module for getting multiple rows, but I think it would be great to have this method in the Codeception 'from the box'.

If you need more abstract reason, I can tell the case by another words: when we should generate an object for test(s) but data of this object is depends from outside source (that can be changed) we must generate fake data using this source without duplicating this in the code.

RebOOter:
Closes #3813

### DIFF
--- a/src/Codeception/Module/Db.php
+++ b/src/Codeception/Module/Db.php
@@ -517,6 +517,19 @@ class Db extends CodeceptionModule implements DbInterface
         return $sth->fetchColumn();
     }
 
+    public function grabColumnFromDatabase($table, $column, array $criteria = [])
+    {
+        $query      = $this->driver->select($column, $table, $criteria);
+        $parameters = array_values($criteria);
+        $this->debugSection('Query', $query);
+        if (!empty($parameters)) {
+            $this->debugSection('Parameters', $parameters);
+        }
+        $sth = $this->driver->executeQuery($query, $parameters);
+        
+        return $sth->fetchAll(\PDO::FETCH_COLUMN, 0);
+    }
+
     public function grabFromDatabase($table, $column, $criteria = [])
     {
         return $this->proceedSeeInDatabase($table, $column, $criteria);
--- a/src/Codeception/Module/Db.php
+++ b/src/Codeception/Module/Db.php
@@ -517,6 +517,21 @@ class Db extends CodeceptionModule implements DbInterface
         return $sth->fetchColumn();
     }
 
+    /**
+     * Fetches all values from the column in database.
+     * Provide table name, desired column and criteria.
+     *
+     * ``` php
+     * <?php
+     * $mails = $I->grabColumnFromDatabase('users', 'name', array('name' => 'RebOOter'));
+     * ```
+     *
+     * @param string $table
+     * @param string $column
+     * @param array $criteria
+     *
+     * @return array
+     */
     public function grabColumnFromDatabase($table, $column, array $criteria = [])
     {
         $query      = $this->driver->select($column, $table, $criteria);
--- a/src/Codeception/Module/Db.php
+++ b/src/Codeception/Module/Db.php
@@ -523,7 +523,7 @@ class Db extends CodeceptionModule implements DbInterface
      *
      * ``` php
      * <?php
-     * $mails = $I->grabColumnFromDatabase('users', 'name', array('name' => 'RebOOter'));
+     * $mails = $I->grabColumnFromDatabase('users', 'email', array('name' => 'RebOOter'));
      * ```
      *
      * @param string $table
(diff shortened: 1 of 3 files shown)


RebOOter:
Tests failed by this error:
`[GuzzleHttp\Exception\ConnectException] cURL error 7: Failed to connect to localhost port 8010: Connection refused (see http://curl.haxx.se/libcurl/c/libcurl-errors.html)`
I don't think that the reason is in commit (

RebOOter:
@Naktibalda Please, could you give me a list of names or emails in the test database? I need to compare column in the test method

Naktibalda:
https://github.com/Codeception/Codeception/blob/2.3/tests/data/dumps/sqlite.sql

RebOOter:
@Naktibalda Sorry, but i don't understand why tests failed after last commit. In the previous commit (2b1fcfc) actual array contains another two emails: 
```
+    4 => 'user@example.org'
+    5 => 'user2@example.org'
```
I added this emails for assert but now in logs I see, that:
```
-    4 => 'user@example.org'
-    5 => 'user2@example.org'
``` 
Please, could you give me advice and tell how I can fix this?

Naktibalda:
It happens because postgresql dump inserts 2 additional users: https://github.com/Codeception/Codeception/blob/2.3.3/tests/data/dumps/postgres.sql#L412-L433

RebOOter:
@Naktibalda Yes, I know, this users (emails) in array for assert, I added them in this commit (2252d8c), but test doesn't work.
```
$this->assertEquals([
            'davert@mail.ua',
            'nick@mail.ua',
            'miles@davis.com',
            'charlie@parker.com',
            'user@example.org',
            'user2@example.org'
                                ],
            $emails);
```

Naktibalda:
PostgresqlTest passed, but SqliteTest failed.

RebOOter:
@Naktibalda Oh, you right, now I see, sorry. I think there are two solutions for make test:
1. Write switch/condition construction in TestsForDb.
2. Write different methods in MySqlDbTest, PostgreSqlDbTest and in SqliteDbTest.

Which solution do you prefer?

Naktibalda:
`2.` I think that testing with one driver should be enough, there is nothing driver specific in your method.

RebOOter:
@Naktibalda I think I'm done here)

tests/unit/Codeception/Module/Db/MySqlDbTest.php
```diff
@@ -68,4 +68,16 @@ public function testConnectionIsResetOnEveryTestWhenReconnectIsTrue()
         $this->assertNotEquals($connection3, $connection2);
     }
 
+    public function testGrabColumnFromDatabase()
+    {
+        $emails = $this->module->grabColumnFromDatabase('users', 'email');
+        $this->assertEquals(
+            ['davert@mail.ua',
```
Naktibalda: Please move the first element of array to the next line.

src/Codeception/Module/Db.php
```diff
@@ -517,6 +517,19 @@ protected function proceedSeeInDatabase($table, $column, $criteria)
         return $sth->fetchColumn();
     }
 
+    public function grabColumnFromDatabase($table, $column, array $criteria = [])
```
DavertMik: Please add documentation in docblock for this method.
The similar method `grabFromDatabase` is documented in its interface https://github.com/Codeception/Codeception/blob/2.3/src/Codeception/Lib/Interfaces/Db.php
Use its documentation as an example and write some docs here. Usage examples with code would be nice.

src/Codeception/Module/Db.php
```diff
@@ -517,6 +517,34 @@ protected function proceedSeeInDatabase($table, $column, $criteria)
         return $sth->fetchColumn();
     }
 
+    /**
+     * Fetches all values from the column in database.
+     * Provide table name, desired column and criteria.
+     *
+     * ``` php
+     * <?php
+     * $mails = $I->grabColumnFromDatabase('users', 'name', array('name' => 'RebOOter'));
```
DavertMik: Ahh, wrong example, It should be:

```php
$I->grabColumnFromDatabase('users', 'email', array('name' => 'RebOOter'));
```

RebOOter:
Sorry, my bad
Fixed

DavertMik:
thanks!

PR #4319, opened 2017-06-09
fixed typos

astepin:
- fixed some typos
- replaced traditional syntax array literals with short array syntax


DavertMik:
This is awesome! You almost reviewed all our codebase for that typos! Thank you for fixing them

PR #4329, opened 2017-06-15
Fixed error

ThomasLandauer:
Changed from `string` to `int`.
Array syntax as recommended at https://www.phpdoc.org/docs/latest/guides/types.html#arrays


DavertMik:
:+1:

PR #4333, opened 2017-06-16
Enable the "disabled" mode for symfony/phpunit-bridge

### LINKED ISSUE #4316: No-exit Option is ignored when deprecation warnings happen.
I ran codeception using jenkins, the tests pass AND I am using the --no-exit option, so the exit code should be 0. 
 
#### What do you get instead?
but the exit code is 1, leading to a failed build. Reason seems to be the deprecation notices - because if I remove this module from my acceptance.suite.yml
```yml
        - Doctrine2:
            depends: Symfony
            populate: true
            cleanup: true
```
deprecation notice and exit code get back to normal.
(Without this deprecation notices I get the exit code of 1 only when I omit the --no-exit option AND have at least one failing test. As it should be.)

This is the end of the output of the test run:
```bash
Time: 8.92 seconds, Memory: 32.00MB

OK (3 tests, 7 assertions)

Remaining deprecation notices (6)

Setting the "doctrine" pre-defined service is deprecated since Symfony 3.3 and won't be supported anymore in Symfony 4.0: 3x
    3x in Application::run from Codeception

Setting the "doctrine.orm.default_entity_manager" pre-defined service is deprecated since Symfony 3.3 and won't be supported anymore in Symfony 4.0: 3x
    3x in Application::run from Codeception
```

### Details

* Codeception version: 2.2.3
* PHP Version: 7.1.5
* Operating System: Windows 7
* Installation type: Composer
* Suite configuration:

```yml
class_name: AcceptanceTester
modules:
    enabled:
        - PhpBrowser:
            url: http://localhost/reactex/web/app.php
        - \Helper\Acceptance
        - Doctrine2:
            depends: Symfony
            populate: true
            cleanup: true
        - REST:
            url: /
            depends: PhpBrowser
            part: Json
    config:
        - Symfony:
            var_path: 'var'
            app_path: 'app'
```

sh41:
Deal with disabled mode for symfony/phpunit-bridge in the same way that symfony/phpunit-bridge bootstrap does as seen in https://github.com/symfony/symfony/commit/1b216477eec2dca22847460473f29144cdb02843#diff-81bfee6017752d99d3119f4ddb1a09edR27

PHPUnit-Bridge does not deal with disabled mode inside the bridge, it is actually disabled by the bootstrap script. Codeception does not use the bootstrap script so disabled mode is not available. This change replicates the logic from phpunit-bridge bootstrap.php in codeception. 

I believe this will fix #4316


DavertMik:
This fix looks good to me but could you to update your branch so tests would pass? 
Thanks.

Also, we are looking for Symfony module maintainer. Would you like to help with that?

Naktibalda:
wrong target, this PR should be directed to 2.3, not to master.

sh41:
Thanks @Naktibalda I thought it looked wrong! I will fix and resubmit.

sh41:
I think that is the correct target now with the updated branch, but the tests are still failing. 

The appveyor failure looks like something to do with TLS negotiation when downloading Composer. 

The Travis failure looks like a timeout connecting to something. is it possible to retry the build or is this something I can resolve another way?

DavertMik:
Ok, that CI things happen... No problems, I can merge it anyway
Thanks!

PR #4338, opened 2017-06-18
Added RunProcess extension

DavertMik:
This extension allows in a simple way to start and stop background processes. 
It is useful for launching Selenium, PhantomJS, etc.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -49,7 +49,7 @@
         "predis/predis": "^1.0",
         "squizlabs/php_codesniffer": "~2.0",
         "vlucas/phpdotenv": "^2.4.0",
-        "symfony/process": "^3.3"
+        "symfony/process": ">=2.7 <4.0"
     },
     "suggest": {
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -48,7 +48,8 @@
         "mongodb/mongodb": "^1.0",
         "predis/predis": "^1.0",
         "squizlabs/php_codesniffer": "~2.0",
-        "vlucas/phpdotenv": "^2.4.0"
+        "vlucas/phpdotenv": "^2.4.0",
+        "symfony/process": "^3.3"
```
Naktibalda: Other Symfony dependencies are supported down to version 2.7, I think that this dependency should have the same version constraint.

DavertMik: Yes, thank you for pointing. Actually we already have this dependency by facebook/webdriver so it should not add anything new to vendor or phar, but should be specified explicitly

PR #4339, opened 2017-06-18
Docblock updates

ThomasLandauer:


src/Codeception/Lib/Interfaces/Web.php
```diff
@@ -14,7 +14,7 @@
      * $I->amOnPage('/register');
      * ```
      *
-     * @param $page
+     * @param string $page
```
DavertMik: For Yii2 module this can also be an array...

src/Codeception/Module/Doctrine2.php
```diff
@@ -461,7 +461,7 @@ public function grabEntitiesFromRepository($entity, $params = [])
      * @version 1.1
      * @param $entity
      * @param array $params
-     * @return array
+     * @return object
```
DavertMik: thanks!

src/Codeception/Lib/Interfaces/Web.php
DavertMik: I think Yii2 module should override this method as well as the docblock (and documentation) to change this behavior.

ThomasLandauer:
OK, so what's the bottom line? Should I change anything?  
I just wanted to get rid of `param null` which doesn't make any sense at all.

DavertMik:
I think I should not forget to update the Yii2 module, that's all. Thanks for the update!

Issue #4326, opened 2017-06-12
Change Module Config per Test


ThomasLandauer:
Since the `cleanup` option of the Doctrine Module defaults to `true`, each test is always being wrapped in a transaction.  
I'm suggesting to start this transaction only when a *write* action is performed on the database - not for `SELECT` only.

My scenario / use case:  
1. I'm checking if the entity is not present in the database (`dontSeeInRepository()`)
1. I'm creating the entity through a command-line call (PHP's `exec()`)
1. I'm checking if the entity is present  (`seeInRepository()`)

With the `cleanup` option being absent (i.e. `true`), Codeception hangs at step 2 with this error message:
```
SQLSTATE[HY000]: General error: 5 database is locked
```
I'm guessing this is because of the transaction, since it works fine, as soon as I set `cleanup` to `false`

Naktibalda:
Your step 2. makes a non-reversible change to database content.

Naktibalda:
A problem with your suggestion is that it is impossible to start a transaction before write when that write is performed by command line command or by application code.

You are welcome to come-up with a different workaround, e.g. helper methods for starting and rolling back transactions, but please don't touch `cleanup` option.

ThomasLandauer:
Yeah, you're right, I certainly missed the fact that most write actions are not performed by Codeception but by the application, so Codeception cannot determine when it would be a good time to start the transaction. Sorry!

But still, the fact that I don't like is that `cleanup` is a *global* setting. What about some way to set (i.e. override) this on a per-test-base? Something like `$I->dontStartTransaction()`?

DavertMik:
@ThomasLandauer the problem is that

> $I->dontStartTransaction()

will be already already a transaction :upside_down_face: 

Doctrine2 module can be reconfigured inside a Helper. In `_before` method you can change configuration of a Doctrine2, so to disable transaction based on a name of a test. This helper should be registered before Doctrine2 so changes could be applied

ThomasLandauer:
@DavertMik OK, cool! :-) I think this should be explained somewhere! (like all "magic" things which happen under the hood) Maybe at http://codeception.com/docs/modules/Doctrine2 ?

So what are the steps?

1. Create a file in `_support/Helper` and call it something like `Transaction.php`
1. Have a `_before()` method in this file?
1. What is the method supposed to look like? How can you access the test name? And how can you modify the transaction setting?
[...]

PR #4343, opened 2017-06-19
Added @prepare annotation to run custom methods before the module

DavertMik:
Following the question from #4326 I discovered that there was no way to change the module configuration for a specific test (only for groups via GroupObject, a bit complicated way). 

This PR adds a new annotation `@prepare` to Cest and Test classes, which executes a method with a module instance passed into it. This allows preparing the module before the `_before` method is invoked.

```php
/*
 * @prepare disableTransactions
 */
function testDoctrine()
{
}

protected function disableTransactions(Doctrine2 $module)
{
   $module->_reconfigure(['cleanup' => false);
}
```

For instance, this can be used to:

* switch browsers in some acceptance tests
* disable transactions for some Doctrine2, Laravel, Yii2 tests
* start custom browser session for WebDriver (maybe in cloud service), esp with `start: false` option
* ...??? something else to think about.

This PR also makes similar result to be achieved with GroupObject:

```php
class Chrome extends \Codeception\GroupObject
{
    public static $group = 'chrome';

    public function _before()
    {
        $this->getModule('WebDriver')->_reconfigure(['browser' => 'chrome']);
    }
}
```

Also: 

* Refactored Metadata class to accept all possible annotations and store them in $params
* Created new Subscriber to invoke methods in @prepare
* Updated WebDriver to update capabilities on _reconfigure
* Changed Subscriber\Module to reset module config after the test, so it could be updated before
* Simplified Parser

---

Work In Progress:

* [x] tests
* [x] documentation

### Update

* renamed `@init` to `@prepare`
* refactored to use DI to auto-inject modules, helpers, and support classes
* simplified Subscriber


DavertMik:
Following conversation from #4326 
@ThomasLandauer 

> Two further ideas about that: Allow enabling/disabling of entire modules too.

No. Just no. It makes a very complicated setup then.

> Allow reconfigure of codeception.yml too.

It is out of the scope of a single test. 

>  What about noTransactions(Doctrine) instead of Doctrine noTransactions? This would show clearer that Doctrine is being passed as parameter to the noTransactions function. Plus it would make things possible (if not now, then maybe in the future) like: changeWhatever(Doctrine, Symfony, WebDriver).

Makes sense. I think I can even use autowriring to inject all kind of dependencies with DI!

> And finally I must say that I don't like annotations. I guess the con's are well-known (no syntax highlighting; looks like a comment; hard to handle if you have more than 1 sort of annotations

Yes, this can turn into a long discussion but to my opinion annotations is evil in production code but they fit very good for tests. They provide do what they intend to: they provide meta information for tests. This make test focused on the problem and not on implementation. However, it is good to keep this meta information inside a code as it highly connected with a test itself.

ThomasLandauer:
+1

Naktibalda:
I think that Travis push build is failing because @DavertMik forgot to pull before creating a new branch.

DavertMik:
Yep, that's right. I merged and added documentation

docs/06-ModulesAndHelpers.md
```diff
@@ -264,46 +264,6 @@ In this helper you can replace the parent's methods with your own implementation
 You can also replace the `_before` and `_after` hooks, which might be an option
 when you need to customize starting and stopping of a testing session.
 
-If some of the methods of the parent class should not be used in a child module, you can disable them.
```
DavertMik: This part is a bit magical and overcomplicated. I don't think it should be used anymore and probably we will drop this functionality in future releases.

PR #4345, opened 2017-06-19
Fix Travis CI

DavertMik:
prefer yii stable builds


Naktibalda:
It is a duplicate of #4342

DavertMik:
Ok, sorry, didn't notice thta

PR #4346, opened 2017-06-19
Clarified docblock of `seeInField`

ThomasLandauer:
Please double-check if this is true! And what does `dontSeeInField` do?: `!==` or `does not contain`?


DavertMik:
Yes, thanks. InnerBrowser and WebDriver checks for equality, not containing.

> And what does dontSeeInField do?: !== or does not contain?

Yes

ThomasLandauer:
Ahem, which one of the two?:
* `!==`

OR

* does not contain

=> Docs of `dontSeeInField` should be updated accordingly.

PR #4347, opened 2017-06-19
[DOC] Clean up doc example for haveHttpHeader

acorncom:
`setHeader` is an old api it looks like


DavertMik:
Yes, it is. At least `haveHttpHeader` is consistent between different modules.
But still we have discussion `set` vs `have`. https://github.com/Codeception/Codeception/issues/4268 Please share your opinion!

PR #4348, opened 2017-06-20
Updated Acceptance Testing documentation:

DavertMik:
* Cleanup for Selenium/WebDriver/Phantom sections
* Added ChromeDriver (this was the initial motivation for the change)
* Manual controlling of browser sessions (can be cool for advanced usage on cloud systems)
* Recommended RunProcess extension (in #4338)

@ThomasLandauer please review )

### DIFF
--- a/docs/03-AcceptanceTests.md
+++ b/docs/03-AcceptanceTests.md
@@ -555,7 +555,7 @@ To start a browser you will need to write corresponding methods in Acceptance [H
 WebDriver module provides advanced methods for the browser session, however, they can only be used from Helpers.
      
 * [_initializeSession](http://codeception.com/docs/modules/WebDriver#_initializeSession) - starts a new browser session
-* [_closeSession](http://codeception.com/docs/modules/WebDriver#_closeSession) - starts a new browser session
+* [_closeSession](http://codeception.com/docs/modules/WebDriver#_closeSession) - stops the browser session
 * [_restart](http://codeception.com/docs/modules/WebDriver#_restart) - updates configuration and restarts browser
 * [_capabilities](http://codeception.com/docs/modules/WebDriver#_capabilities) - set [desired capabilities](https://github.com/SeleniumHQ/selenium/wiki/DesiredCapabilities) programmatically.
 
--- a/docs/03-AcceptanceTests.md
+++ b/docs/03-AcceptanceTests.md
@@ -3,11 +3,11 @@
 Acceptance testing can be performed by a non-technical person. That person can be your tester, manager or even client.
 If you are developing a web-application (and you probably are) the tester needs nothing more than a web browser
 to check that your site works correctly. You can reproduce an acceptance tester's actions in scenarios
-and run them automatically. Codeception keeps tests clean and simple,
+and run them automatically. Codeception keeps tests clean and simple
 as if they were recorded from the words of an actual acceptance tester.
 
 It makes no difference what (if any) CMS or framework is used on the site. You can even test sites created with different
-languages, like Java, .NET, etc. It's always a good idea to add tests to your web site.
+languages, like Java, .NET, etc. It's always a good idea to add tests to your website.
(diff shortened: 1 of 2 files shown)


ThomasLandauer:
PR: Please merge it. I'll take a closer look then. I need to see the entire page - can't deal with GitHub's red/green lines ;-)

The most important question for me is: If you can connect to Chromedriver directly, what's the advantage of having Selenium at all? What do you mean by "limited support for WebDriver protocol": Is every other Codeception command not working? Or just some minor differences (I noticed the output on errors being less descriptive)?  
Cause at first sight, Chromedriver makes a very good impression to me! So is there any **real** drawback? If no, I would clearly recommend (in the docs) to start with Chromedriver! (Selenium installation is a mess, and documentation is a joke)

And what will be the difference to Headless Chrome? Is facebook/php-webdriver talking to Headless Chrome directly? Or still via Chromedriver?

And do you know of any plans for something comparable on Firefox? facebook/php-webdriver cannot connect with Geckodriver directly?

DavertMik:
> The most important question for me is: If you can connect to Chromedriver directly, what's the advantage of having Selenium at all? What do you mean by "limited support for WebDriver protocol": Is every other Codeception command not working? Or just some minor differences (I noticed the output on errors being less descriptive)?
Cause at first sight, Chromedriver makes a very good impression to me! So is there any real drawback? If no, I would clearly recommend (in the docs) to start with Chromedriver! (Selenium installation is a mess, and documentation is a joke)

Selenium is a standard de facto, that's why it is recommended. It can launch any browser, even Edge or Safari. So it's a comprehensive solution. 
I'm also not so sure about the state of support for WebDriver protocol in ChromeDriver. Not everything is yet possible with running ChromeDriver directly. For instance, it can't change window size, while Selenium can.

So to me, it is better to spend some time installing but use the most professional tool for the job.

> And what will be the difference to Headless Chrome? Is facebook/php-webdriver talking to Headless Chrome directly? Or still via Chromedriver?

http://phptest.club/t/how-to-run-headless-chrome-in-codeception/1544/1

> And do you know of any plans for something comparable on Firefox? facebook/php-webdriver cannot connect with Geckodriver directly?

Not sure, didn't have a chance to try Geckodriver by itself yet.

ThomasLandauer:
Fuck the window size! ;-) One of the coolest features of Codeception is its easy installation. And now it's getting still easier - supercool!

So what I'm gonna do is send a PR that moves the Chromedriver-alone option to the top of the docs! Reason:
* Chromedriver is a must for anybody, anyway.
* Those who know what Selenium is will still find it, even if it's mentioned further down.
* And those who don't know what Selenium is: Well, they don't have to find out anymore! (OK, more precise: ... not *immediately*)

DavertMik:
> So what I'm gonna do is send a PR that moves the Chromedriver-alone option to the top of the docs! 

Nope, I'm a bit against it. What you actually can do is create a new `InitTemplate` that will create a new project with ChromeDriver configured. It can even try download ChromeDriver for a user. It may also enable RunProcess extension to start/stop ChromeDriver.

But because it is simplest it should not be mentioned at top of documentation. 
PhantomJS is even simpler but it's not the best option, as well. We can make a badge like "hey, it is the simplest stuff here!", and that's all

ThomasLandauer:
Merge the PR, and we'll find a solution that makes both of us happy :-)

ThomasLandauer:
I thought this would be online at http://codeception.com/docs/03-AcceptanceTests right after you merged it. Can't see it there.... :-( Looks like I haven't understood the deployment process yet ;-)

Naktibalda:
Deployment process is in https://github.com/Codeception/Codeception/blob/2.3/RoboFile.php -
it checks out https://github.com/Codeception/codeception.github.com, makes changes and pushes them back to codeception.github.com

The site is updated when @DavertMik runs it manually.

PR #4366, opened 2017-06-27
Fix running a single test with included setup

zebraf1:
Fixed issue when trying to run a single test when there are included tests.

Example:
```yml
#codeception.yml
include:
  - jazz
  - jazz/pianist
  - shire
```

Calling "run jazz/pianist/tests/functional/PianistCest.php" fails with error "[RuntimeException] Suite 'jazz/pianist/tests/functional/PianistCest' could not be found".

**Workaround**:
It is still possible to run it by going into the sub-category jazz/pianist and running "run functional PianistCest", but this is not possible when running a single test from within PHPStorm IDE


zebraf1:
Weird that appveyor has some fails, any ideas why this occurs?

DavertMik:
Thanks. This is really impressive work. I know it is a very complicated feature, so thank you so much for working on it!

PR #4367, opened 2017-06-27
Added recursive access grant for output directory

kvarts:


src/Codeception/Configuration.php
```diff
@@ -520,8 +520,7 @@ public static function outputDir()
         }
 
         if (!is_writable($dir)) {
-            @mkdir($dir);
-            @chmod($dir, 0777);
```
Naktibalda: I don't like change. Original code wasn't perfect, but the new code introduces new issues.
Original code could chmod existing directory, the new one can't.
I think that it would be better to separate `!file_exists` and `!is_writable` checks and use `mkdir` in former and `chmod` in latter.

kvarts: Yes you are right. Thank you

DavertMik:
Thanks

PR #4369, opened 2017-06-28
[WebDriver] Smart wait for elements.

DavertMik:
This should dramatically improve the stability for tests.

Whenever an element is required and is not on a page, setting `wait: 5` will make Codeception wait for 5 seconds to find it. 

This should simplify tests, and reduce a number of waiters:

```php
$I->waitForElement('#click-me', 5);
$I->click('#click-me');
```
can be simplified to

```php
$I->click('#click-me');
```
with `wait: 5` in config.

*(use CSS ids or strict locators to make this work with click)*

In the same way `$I->seeElement()` will wait for an element to appear before failing.
This should not affect the performance as implicit waits are added only when element is requried to continue a test.

---

Technical note:
Currently `wait` option corresponds to turning on Selenium [Implicit waits](http://www.seleniumhq.org/docs/04_webdriver_advanced.jsp#implicit-waits).
This was not working great as it was waiting for elements everywhere, especially when guessing locators, or checking the absence of an element. It slowed down tests. Current patch also uses implicit waits but only for cases when exactly one element is required.

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -866,9 +866,9 @@ class WebDriver extends CodeceptionModule implements
         if (!$selector) {
             return $this->assertPageContains($text);
         }
-        $this->enableImplicitWait(true);
+        $this->enableImplicitWait();
         $nodes = $this->matchVisible($selector);
-        $this->enableImplicitWait(false);
+        $this->disableImplicitWait();
         $this->assertNodesContain($text, $nodes, $selector);
     }
 
@@ -1073,9 +1073,9 @@ class WebDriver extends CodeceptionModule implements
 
     public function seeLink($text, $url = null)
     {
-        $this->enableImplicitWait(true);
+        $this->enableImplicitWait();
         $nodes = $this->baseElement->findElements(WebDriverBy::partialLinkText($text));
-        $this->enableImplicitWait(false);
+        $this->disableImplicitWait();
         $currentUri = $this->_getCurrentUri();
 
         if (empty($nodes)) {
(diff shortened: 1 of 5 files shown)


Ocramius:
I don't fully understand the patch as I don't know selenium waits, but shouldn't this be converted to an element polling or equivalent, with a timeout?

src/Codeception/Module/WebDriver.php
```diff
@@ -3157,4 +3158,16 @@ protected function setBaseElement($element = null)
         }
         $this->baseElement = $this->matchFirstOrFail($this->webDriver, $element);
     }
+
+    protected function enableImplicitWait($enable)
```
Naktibalda: Using Boolean arguments is a bad practice, please split this method to `enableImplicitWait` and `disableImplicitWait`.

src/Codeception/Module/WebDriver.php
```diff
@@ -2318,7 +2316,8 @@ public function wait($timeout)
      */
     public function executeInSelenium(\Closure $function)
     {
-        return $function($this->webDriver);
+        $result = $function($this->webDriver);
+        return $result;
```
Naktibalda: Have you missed something here?
If not, please revert.

DavertMik:
@Ocramius implicit waits are polling elements for uppear at the Selenium Server side. So they are much faster than explicit waits, especially over the network. 

But I need to run tests to prove that)

src/Codeception/Module/WebDriver.php
DavertMik: Yes, I was trying to put implicit waits there but that was a bad idea )

DavertMik:
Some updates. Tried 2 tests, times are pretty the same:

```
✔ WebDriverTest: Wait for element (4.20s)
✔ WebDriverTest: Implicit wait (3.61s)
✔ WebDriverTest: Wait for element (3.88s)
✔ WebDriverTest: Implicit wait (3.47s)
```

PR #4374, opened 2017-07-02
Add assertStringStartsWith

guidocella:
This adds assertStringStartsWith and assertStringStartsNotWith to the available asserts.


DavertMik:
Thanks

PR #4376, opened 2017-07-03
Update WebDriver.php

Marciobds:
switchToPreviousTab had it's documentaton telling to use switchToNextTab method as well as was saying that it switches to the next browser tab.

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -3060,7 +3060,7 @@ class WebDriver extends CodeceptionModule implements
      * // switch to previous tab
      * $I->switchToPreviousTab();
      * // switch to 2nd previous tab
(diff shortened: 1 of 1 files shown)


src/Codeception/Module/WebDriver.php
```diff
@@ -3052,15 +3052,15 @@ public function switchToNextTab($offset = 1)
...
      * An offset can be specified.
      *
      * ```php
      * <?php
      * // switch to previous tab
-     * $I->switchToNextTab();
+     * $I->switchToPreviousTab();
      * // switch to 2nd previous tab
-     * $I->switchToNextTab(-2);
+     * $I->switchToPreviousTab(-2);
```
Naktibalda: I think that this example should be `$I->switchToPreviousTab(2);`

Naktibalda:
thanks

PR #4379, opened 2017-07-03
[Yii] Splitted `cleanup` and `transactional` configuration

leandrogehlen:


Naktibalda:
@leandrogehlen Could you make it backwards compatible and use `cleanup` option of `transaction` option is not set?

leandrogehlen:
@Naktibalda, i believe that is happening, because:
https://github.com/Codeception/Codeception/pull/4379/files#diff-938f756d763fc21aace78bb43ea26870R136

Naktibalda:
Situation: we have an existing Yii2 website with `cleanup: false` in functional.suite.yml.
After update, the tests will start failing because your new setting will use a default value `transaction: true`.

A backwards compatible way would be to set default value of `transaction` to `null` and use `cleanup` setting if `transaction` is null, because it isn't set in suite.yml.

leandrogehlen:
ok, you are rigth, i will do the change

DavertMik:
From my side, I'd like to keep the `cleanup` option for transactions.
This is compatible with other ORM modules, like Doctrine2 and Laravel. 
I know this name is not perfect but at least it is consistent and backwards compatible.
Current implementation is consistent but is not clear from the documentation that setting cleanup to false will disable transactions as well.

Maybe we can create a different option to cleanup fixtures?

`cleanupFixtures: true`

But let's wait for a review @samdark

leandrogehlen:
The transactional cleanup is not valid for functional tests. Does not make sense start a transanction in client side.

DavertMik:
> The transactional cleanup is not valid for functional tests. Does not make sense to start a transanction in client side.

Ok, makes sense

samdark:
@DavertMik looks good to me.

DavertMik:
Thanks

michaelarnauts:
This change caused issue #4417

PR #4380, opened 2017-07-03
Cleanup formatting, word case in Silex module docs

randomecho:


randomecho:
Redid it so that it is only fixing up the Silex.php file, not sure if I should edit the other parts of the documentation generated.

DavertMik:
Thanks, @randomecho

PR #4385, opened 2017-07-04
[Db] Add update in database

eXorus:
Hello,

For my acceptance I need sometimes to update record in the database, so I added a new function to be able to do that in the module Db:

```php
$I->updateInDatabase('users', ['banned' => 'true'], ['id' => '123']);
```

What do you think ? I know it's better to start test with a clean database but I have a complex database that is coming from the production (anonymized) and I test it like a blackblox from the web. I can't populate all my database with only testing data.

### DIFF
--- a/src/Codeception/Lib/Driver/Db.php
+++ b/src/Codeception/Lib/Driver/Db.php
@@ -299,7 +299,7 @@ class Db
         
         $set = [];
         foreach ($data as $column => $value) {
-            $set[] = $this->getQuotedName($column)." = '{$value}'";
+            $set[] = $this->getQuotedName($column)." = ?";
         }
 
         $where = $this->generateWhereClause($criteria);
--- a/src/Codeception/Lib/Driver/Db.php
+++ b/src/Codeception/Lib/Driver/Db.php
@@ -288,4 +288,17 @@ class Db
 
         return empty($this->primaryKeys);
     }
+    
+    public function update($table, array $data, array $criteria)
+    {
+
+        $set = [];
+        foreach ($data as $column => $value) {
+            $set[] = "{$column} = {$value}";
+        }
+
+        $where = $this->generateWhereClause($criteria);
+
+        return sprintf('UPDATE %s SET %s %s', $this->getQuotedName($table), implode(', ', $set), $where);
+    }
 }
--- a/src/Codeception/Lib/Driver/Db.php
+++ b/src/Codeception/Lib/Driver/Db.php
@@ -299,7 +299,7 @@ class Db
         
         $set = [];
         foreach ($data as $column => $value) {
-            $set[] = "{$column} = {$value}";
+            $set[] = $this->getQuotedName($column)." = '{$value}'";
         }
 
         $where = $this->generateWhereClause($criteria);
--- a/src/Codeception/Lib/Driver/Db.php
+++ b/src/Codeception/Lib/Driver/Db.php
@@ -291,7 +291,12 @@ class Db
     
     public function update($table, array $data, array $criteria)
     {
-
+        if (empty($data)) {
+            throw new \Exception(
+                "Query update can't be prepared without data."
+            );
+        }
+        
         $set = [];
         foreach ($data as $column => $value) {
             $set[] = "{$column} = {$value}";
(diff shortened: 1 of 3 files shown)


src/Codeception/Lib/Driver/Db.php
```diff
@@ -288,4 +288,17 @@ protected function flushPrimaryColumnCache()
 
         return empty($this->primaryKeys);
     }
+    
+    public function update($table, array $data, array $criteria)
+    {
+
+        $set = [];
+        foreach ($data as $column => $value) {
+            $set[] = "{$column} = {$value}";
```
Naktibalda: I think that it would be better to quote column names witrh  `$this->getQuotedName($column)`, because column name can be a reserved keyword or contain some illegal characters.

src/Codeception/Lib/Driver/Db.php
```diff
@@ -288,4 +288,17 @@ protected function flushPrimaryColumnCache()
 
         return empty($this->primaryKeys);
     }
+    
+    public function update($table, array $data, array $criteria)
+    {
+
+        $set = [];
```
Naktibalda: Please throw exception if $data is empty.

eXorus: Fixed

src/Codeception/Lib/Driver/Db.php
eXorus: Fixed

src/Codeception/Lib/Driver/Db.php
```diff
@@ -288,4 +288,22 @@ protected function flushPrimaryColumnCache()
...
+    {
+        if (empty($data)) {
+            throw new \Exception(
+                "Query update can't be prepared without data."
+            );
+        }
+        
+        $set = [];
+        foreach ($data as $column => $value) {
+            $set[] = $this->getQuotedName($column)." = '{$value}'";
```
Naktibalda: It would be better to use placeholders `?` instead of injecting `$value`.

src/Codeception/Module/Db.php
```diff
@@ -534,4 +534,16 @@ public function grabNumRecords($table, array $criteria = [])
     {
         return $this->countInDatabase($table, $criteria);
     }
+    
+    public function updateInDatabase($table, array $data, array $criteria = [])
+    {
+
+        $query = $this->driver->update($table, $data, $criteria);
+        $parameters = array_values($criteria);
```
Naktibalda: If placeholders are used in SET, $parameters will have to contain $data and $criteria.
```php
$parameters = array_merge(array_values($data), array_values($criteria));
```

DavertMik:
It could have a test at https://github.com/Codeception/Codeception/blob/2.3/tests/unit/Codeception/Module/Db/TestsForDb.php

but still looks good to me. @Naktibalda how do you think?

eXorus:
@DavertMik @Naktibalda 
Test added

src/Codeception/Module/Db.php
```diff
@@ -534,4 +534,15 @@ public function grabNumRecords($table, array $criteria = [])
     {
         return $this->countInDatabase($table, $criteria);
     }
+    
+    public function updateInDatabase($table, array $data, array $criteria = [])
+    {
+        $query = $this->driver->update($table, $data, $criteria);
+        $parameters = array_merge(array_values($data), array_values($criteria));
+        $this->debugSection('Query', $query);
+        if (!empty($parameters)) {
```
Naktibalda: This condition isn't necessary, because $data can't be empty.

src/Codeception/Lib/Driver/Db.php
```diff
@@ -288,4 +288,22 @@ protected function flushPrimaryColumnCache()
 
         return empty($this->primaryKeys);
     }
+    
+    public function update($table, array $data, array $criteria)
+    {
+        if (empty($data)) {
+            throw new \Exception(
```
Naktibalda: A more specific exception would be better. Codeception ModuleException is a good choice IMO.

DavertMik: Agree on `InvalidArgumentException`

DavertMik:
I will merge and update the requested changes.
Thanks @eXorus

PR #4389, opened 2017-07-07
Added docs for SmartWait

DavertMik:


DavertMik:
@ThomasLandauer could you provide any feedback? Thanks

ThomasLandauer:
First of all: Cool feature! :-)

Some questions:

1. Is this available for ChromeDriver (without Selenium) too? If no, are there plans that it will become available any time soon?  
This feature sounds like a game changer for me, since waiting for stuff is really a pain in the ass - costs me some stupid extra trial-and-error cycles each time....

1. The disctinction between "specific" and other locators: Is this the only feature where this is important, or are there other situations too, in which the locator "type" matters? Is there any other documentation somewhere where this disctinction is explained?  
Wording: There's also the term "fuzzy" locator, but this is used to describe the fact that a locator like `span` can refer to an HTML element, an input with that `name` attribute, a link with that text, etc.?
    * So is "specific" the opposite of "fuzzy" then? In the sense that any locator is either specific or fuzzy?
    * Or is it rather that there are fuzzy locators, then normal locators (what would that be for example?), and finally specific locators?

1. If the test fails anyway, I'm guessing (haven't seen it in action yet!) that there's no feedback which of the situations did occur:
    * The element really didn't show up
    * SmartWait didn't work at all, cause the wrong locator "type" was used

    => Any ideas how this could be communicated to the user?

1. Also cool would be anything like "SmartScroll". I took a first (small) step towards that by setting up a test which shows the problem: https://github.com/Codeception/Codeception/pull/4248 How do we move on from there?

DavertMik:
> Is this available for ChromeDriver (without Selenium) too? 

 Implicit wait is a feature specific to Selenium Server, it doesn't belong to WebDriver protocol. So it will work in Selenium only.

> The disctinction between "specific" and other locators: Is this the only feature where this is important, or are there other situations too, in which the locator "type" matters? I

Not sure it is documented but specific locators are always much faster. It doesn't really matter for PhpBrowser/WebDriver tests, but can be important in WebDriver tests. 

> Wording: There's also the term "fuzzy" locator

I think we should use it in docs. Probably in source code we use term "wide locators".

> So is "specific" the opposite of "fuzzy" then? In the sense that any locator is either specific or fuzzy?

Yes

> If the test fails anyway, I'm guessing (haven't seen it in action yet!) that there's no feedback which of the situations did occur:
> The element really didn't show up
>SmartWait didn't work at all, cause the wrong locator "type" was used

I think we can print the debug output once SmartWait mode is entered.

ThomasLandauer:
> It doesn't really matter for PhpBrowser/WebDriver tests, but can be important in WebDriver tests.

I don't understand that :-( When does it (not) matter?

Since you're saying that the selector type decision doesn't apply to SmartWait only, I think it should get an own section in AcceptanceTests (or anywhere else).  

Please release it, and update the docs, so I can see what the current status is (lost overview again ;-)

SmartScroll: Yeah, sure! Just wanted to know if you're expecting something from me in this matter.

DavertMik:
> I don't understand that :-( When does it (not) matter?

Sorry, I meant "PhpBrowser/Framework" tests :) Ups

 > Since you're saying that the selector type decision doesn't apply to SmartWait only, I think it should get an own section in AcceptanceTests 

Yes, it can be mentioned there. We have this section in WebDriver module: http://codeception.com/docs/modules/WebDriver#Locating-Elements

ThomasLandauer:
Um, this feature is not working for me :-( 
When I change
```php
$I->waitForText('abc', 5, ['css'=>'.foo']);
```
to
```php
$I->see('abc', ['css'=>'.foo']);
```
the test fails. In debug output I don't have any indication that SmartWait is active at all. In acceptance.suite.yml I have:
```yml
class_name: AcceptanceTester
modules:
    enabled:
        # ...
        - WebDriver:
            url: 'http://localhost/'
            browser: firefox
            window_size: 1400x1000
            wait: 5
```
Additional question: Should SmartWait also work for things like `waitForElementNotVisible`?

DavertMik:
> the test fails. In debug output I don't have any indication that SmartWait is active at all
Ok, we should print that in debug information...

```php
$I->waitForText('abc', 5, ['css'=>'.foo']);
```
Well... Are you waiting for text here or for `.foo`? 

With SmartWait
```php
$I->see('abc', ['css'=>'.foo']);
```
will wait for `.foo` to appear but it won't wait for text abc there. Because `abc` is not a locator
So SmartWait are better to be used with `seeElement`

ThomasLandauer:
> Ok, we should print that in debug information...

Yes, please!! Just a short notice like `SmartWait: active` or `SmartWait: 5 seconds`

So let's get the functionality straight:
```php
$I->see('abc', ['css'=>'.foo']);
```
1. Will wait for the `.foo` element to appear, and will then check if "abc" is in it?
1. But if the (empty) `.foo` is already there, it will not wait for "abc" to eventually appear, but fail the test right away?
1. So it *is* possible to use `see` with SmartWait, if you bear 1 and 2 in mind?
1. Will SmartWait also work with `dontSeeElement`, i.e. will it (if the element is there) wait until it disappears?

Issue #4402, opened 2017-07-13
Call stack arguments, that were passed by reference, get modified by grabDataFromResponseByJsonPath debug


alambe:
#### What are you trying to achieve?
Not have unexpected side effects when passing by reference in debug mode.

#### What do you get instead?
Some debug messages can change the value of call stack arguments that were passed by reference.

In the example case below when run with --debug, REST module's grabDataFromResponseByJsonPath has a side effect of stringifying the referenced value of an  argument passed by reference, higher up in the call stack.

This exposes two potential bugs (or at least two features used together to cause this issue):
1. grabDataFromResponseByJsonPath, when displaying debug messages, replaces meta step arg values with the stringified version.  This could be intentional for performance or convenience but seems sloppy and exposes the next bug.
2. meta steps are created from backtrace, but stack arguments are not dereferenced.  This allows bug 1 to change the value of the referenced variable.

I have a quick fix for bug#2, which I will submit as a pull request.
I don't think bug#1 is specific to grabDataFromResponseByJsonPath, but my fix for bug#2 resolves the overall issue.

Example case:
```php
class stackrefbugCest
{
    public function tryToTest(AcceptanceTester $I)
    {
        $r=[];
        var_dump($r);
        $I->sendGET('/ticket');//REST module, get some json response
        var_dump($r);//$r equals []
        $I->gjp('$.success',$r);//see below, AcceptanceTester wrapper for REST module's grabDataFromResponseByJsonPath, but with an extra argument
        var_dump($r);//bug: $r now equals '[]', if run with --debug
        $I->assertEquals([],$r);//fails, if run with --debug
    }
}

class AcceptanceTester extends \Codeception\Actor
{
    use _generated\AcceptanceTesterActions;
    function gjp($p,&$r){//grab json path
        return $this->grabDataFromResponseByJsonPath($p);
    }
}
```
### Details

* Codeception version: 2.3.4
* PHP Version: 7.0.20
* Operating System: Codeception docker image running on Mac OSX 
* Installation type: Composer
* List of installed packages (`composer show`)
behat/gherkin                      v4.4.5 Gherkin DSL parser for PHP 5.3
codeception/codeception            2.3.4  BDD-style testing framework
codeception/visualception          0.9.0  Visual regression tests for Codeception
doctrine/instantiator              1.0.5  A small, lightweight utility to instantiate objects in PHP without invoking their constructors
facebook/webdriver                 1.4.1  A PHP client for Selenium WebDriver
[...]

PR #4403, opened 2017-07-13
fix Step::addMetaStep by dereferencing stack arguments that may have …

alambe:
…been passed-by-reference, so that later modules like console do not clobber original referenced value by accident

Quick fix for issue #4402

### DIFF
--- a/src/Codeception/Step.php
+++ b/src/Codeception/Step.php
@@ -297,7 +297,11 @@ abstract class Step
 
             // in case arguments were passed by reference, copy args array to ensure dereference.  array_values() does not dereference values
             $args=[];
-            if(is_array($step['args'])) foreach($step['args'] as $a) $args[]=$a;
+            if (is_array($step['args'])) {
+                foreach($step['args'] as $a) {
+                    $args[]=$a;
+                }
+            }
 
             $this->metaStep = new Step\Meta($step['function'], $args);
             $this->metaStep->setTraceInfo($step['file'], $step['line']);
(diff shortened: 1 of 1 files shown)


alambe:
Is this being considered or are you waiting on me to do something?  I'm not sure why semaphore failed.  It passed before and the only changes since were nitpick formatting fixes.  I did push two nitpick fix commits in rapid succession (the second before the first's checks completed), and this could have caused an issue for semaphore.  Should I push an empty commit to trigger checks to run again?  Or, can someone with access re-run the semaphore check or direct me how to do that?

alambe:
Thank you.

DavertMik:
Just a guess, can this be solved with `array_map`? I would prefer it in this case

alambe:
Are you suggesting to just replace `array_values($step['args'])` with `array_map(null,$step['args']);`?

`array_map` without a callback does not dereference values.  Also, it maintains keys while `array_values` rekeys.  With my change, I was trying to maintain the rekeying behavior of array_values, though I don't know if rekeying is necessary.  If rekeying was not necessary, why was array_values used in the first place?

That said, if rekey is not necessary, then yes, the following should work:
`array_map(function($i){return $i;},$step['args'])`

If rekeying is necessary, the following will dereference and rekey:
`array_map(function($i){return $i;},array_values($step['args']))`

The later is functionally equivalent to the existing patch.  Do you want me to update the branch?

DavertMik:
Sorry for late response... @alambe but yes, I prefer this:

```php
array_map(function($i){return $i;},array_values($step['args']))
```

please!

alambe:
No problem.  I'm just glad to get this fix in upstream.

DavertMik:
Thanks very much! Sorry for the delay in responses and this strictness about the code )

PR #4411, opened 2017-07-20
[InnerBrowser] Submit default values of checkboxes

guidocella:
Currently if you use `submitForm` in functional tests passing a checkbox with the value of true and that checkbox doesn't have a value attribute in the HTML page, that field will have the value of an empty string in the resulting request because `setCheckboxBoolValues` just gets the value attribute of checkboxes. This pull request modifies it to check if the checkbox does have a value attribute, and if it doesn't it uses "on" as value.


DavertMik:
Thanks. Looks good to me

PR #4424, opened 2017-07-27
fix/postgres_cleanup alter postgres cleanup (recreate schema)

samusenkoiv:
This is a better way to do the cleanup for pgsql.


DavertMik:
> It looks like it should be working.

And all 3 CIs failed in the same time by different accidents )

DavertMik:
Ok, most of CIs passing. Merging then!
Thanks

PR #3905, opened 2017-01-10
add option text to seeInField

gimler:
* support option text in `seeInField` not only value
* fix bug match with and without whitespaces see also https://github.com/facebook/php-webdriver/pull/394
* fix bug `seeInField` not working after `selectOption`

PR #4425, opened 2017-07-28
[InnerBrowser] Make seeInField check options' texts

guidocella:
Makes InnerBrowser's seeInField check for options' texts and trimmed texts on top of checking their values. This makes it behave the same as WebDriver as requested in #3905.

When an option's text is also its desired value, this avoids having to add a value anyway just to make `seeInFormFields` pass.

I did not add a test for the non trimmed text in TestsForWeb and left WebDriverTest::testSeeInFieldSelect as it is since it seems that the check for non trimmed values should be skipped with Selenium.

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -634,7 +634,7 @@ class InnerBrowser extends Module implements Web, PageSourceSaver, ElementLocato
 
     protected function proceedSeeInField(Crawler $fields, $value)
     {
-        $testValues = $this->proceedGetValueFromField($fields, true);
+        $testValues = $this->getValueAndTextFromField($fields);
         if (!is_array($testValues)) {
             $testValues = [$testValues];
         }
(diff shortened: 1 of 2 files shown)


src/Codeception/Lib/InnerBrowser.php
```diff
@@ -1270,14 +1270,15 @@ public function grabValueFrom($field)
...
+        return $this->proceedGetValueFromField($nodes, false);
     }
 
     /**
      * @param Crawler $nodes
+     * @param bool $getOptionTexts
      * @return array|mixed|string
      */
-    protected function proceedGetValueFromField(Crawler $nodes)
+    protected function proceedGetValueFromField(Crawler $nodes, $getOptionTexts)
```
DavertMik: I think this should be refactored to offer more general behavior. Maybe to split into 2 methods:

one will return acutal values, second will return content values. 

right now this method does two things and uses hard-to-guess boolean param which changes the behavior.

guidocella:
I split proceedGetValueFromField.

PR #4430, opened 2017-08-02
Make wait() in Webdriver module accept fractional amount of seconds

gvlasov:
Use case: sometimes we want to wait for less than a second.


DavertMik:
To me too

PR #4434, opened 2017-08-03
_capabilities documentation fix

L0rdJ:
$capabilityFunction should return array, otherwise on second test run you got:
```
[PHPUnit_Framework_Exception] Argument 1 passed to Facebook\WebDriver\Remote\DesiredCapabilities::__construct() must be of the type array, object given
```


DavertMik:
Thanks

PR #4437, opened 2017-08-09
Symfony/fix_profiler_loading 

samusenkoiv:
- Problem:
When I have several request within one test, my services have different _entity_managers_, which results in bugs.

- Problem cause:
When _profiler_ service is being loaded, dozens of services are instantiated. Some of them needs _entity_manager_, which has not been instantiated by that time, so it is null, and so container creates new one. Then we set _persistent_services_ (which include _entity_manager_), so all the other services will get _entity_manager_ from _persistent_services_, not the one created during _profiler_ enabled.

- Solution:
Set the persistent services and **then** enable the profiler.


DavertMik:
Yep, makes sense for your case. Thank you

PR #4442, opened 2017-08-10
Mongo haveInCollection fix

samusenkoiv:
Method `$response->getInsertedId()` already returns string. In case there is not a string `_toString()` will be called


DavertMik:
Thanks

PR #4446, opened 2017-08-11
add line number to TestParseException exception message

gaainf:
add line number to debug easier

### DIFF
--- a/src/Codeception/Exception/TestParseException.php
+++ b/src/Codeception/Exception/TestParseException.php
@@ -3,9 +3,13 @@ namespace Codeception\Exception;
 
 class TestParseException extends \Exception
 {
-    public function __construct($fileName, $errors = null)
+    public function __construct($fileName, $errors = null, $line = null)
     {
-        $this->message = "Couldn't parse test '$fileName'";
+        if ($line) {
+            $this->message = "Couldn't parse test '$fileName' on line $line";
+        } else {
+            $this->message = "Couldn't parse test '$fileName'";
+        }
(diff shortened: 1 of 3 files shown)


Naktibalda:
Have you tested it? Does it work as expected?
I would expect getLine to return a line where TestParseException was [thrown](https://github.com/Codeception/Codeception/blob/b54eaf4007484f36145c1dc8c64da1874adbc340/src/Codeception/Test/Cept.php#L44) and not where the syntax error is.

gaainf:
You are right. I have to pass line number as separate parameter. Probably passing of $exception is more appropriate solution, but in some modules I saw throw new TestParseException($file, $some_string).

Fixed and tested with php:
```
$ php -l cli/DryRunCest.php
PHP Parse error:  syntax error, unexpected '$I' (T_VARIABLE) in cli/DryRunCest.php on line 12

Parse error: syntax error, unexpected '$I' (T_VARIABLE) in cli/DryRunCest.php on line 12

Errors parsing cli/DryRunCest.php
```
and with codecept:
```$ docker-compose run --rm codecept run cli DryRunCest
Codeception PHP Testing Framework v2.3.5
Powered by PHPUnit 6.2.4 by Sebastian Bergmann and contributors.


  [Codeception\Exception\TestParseException]
  Couldn't parse test '/repo/tests/cli/DryRunCest.php' on line 12
  syntax error, unexpected '$I' (T_VARIABLE)
```

DavertMik:
Thanks!

Issue #3514, opened 2016-09-14
Allow --retries parameter for tests 


darrenjennings:
Sometimes acceptance tests fail for unknown reasons, and would be nice to have it retry only the failed test scenario up to `n` number of times specified in the argument list, similar to how nightwatchjs does it.

http://nightwatchjs.org/guide

```
--retries | Retries failed or errored testcases up to the specified number of times. Retrying a testcase will also retry the beforeEach and afterEach hooks, if any.
```

Example command:

``` bash
php codecept run acceptance --retries 3
```

raistlin:
To me this leads to bad testing design.

You should not design a test that fail for an unknown reason, you should be able to debug them and discover why they fail. May be they hit an unknown edge case or bug in your code.

If you need an external service (db, files, webservice...) that it's unreliable, mock it.

oehlschl:
I agree with the sentiment above but also admit that we've implemented a similar feature above for our Codeception (acceptance/WebDriver) test platform. Tests run against an enterprise web app whose defects are beyond our control. It's an unfortunate reality, but rerunning failures does help to identify consistent issues from flakes.

The challenge in my opinion is implementing a solution within Codeception that meets the needs and expectations of different users--do you record failure if all or any fail, do you ignore leading failures, etc.

@dardrone, see also [\Codeception\Extension\RunFailed](http://codeception.com/addons#codeceptionextensionrunfailed). You might be able to achieve your desired behavior with this and some shell scripting.

DavertMik:
This feature can be useful for webdriver testing but I don't know how to make it consistent with unit and integration tests. Probably `retires` can be per suite option and should be stored inside suite config. Then by running in specific environment `--env ci` you can enable retries. 

So yep, this can be implemented

eXorus:
I will use this for acceptance testing and api testing. Sometimes there are curl error because the network is gone, it will be nice the morning to be sure that the error is a real error and not only a temporary environment break.

Naktibalda:
A problem with this suggestion is that it would make flaky tests harder to spot.
Some of them are legitimate issues.

dardrone:
@Naktibalda @raistlin I can't think of a good reason for unit tests to retry, but I've encountered a few scenarios where retry is beneficial, such as automated tests 'waking' the server on startup and then failing because the waitForElementVisible was timing out due to networking issues. Still cause for concern!

It would still be good to know that some tests have to retry more than others, and see those as warnings as part of this enhancement.

[...]

PR #4450, opened 2017-08-15
add get config method and tries from config

oshanrube:
i added a get config method so we can get the config values from module container
and i used the new settings value "reties" so the tests can retry on failure because selenium fails randomly cause of asset failures and selenium timeouts

### DIFF
--- a/src/Codeception/Lib/ModuleContainer.php
+++ b/src/Codeception/Lib/ModuleContainer.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace Codeception\Lib;
 
 use Codeception\Configuration;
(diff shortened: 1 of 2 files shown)


Naktibalda:
Thanks, but no.

There is no need for getConfig method in WebDriver module.
There is `_getConfig` method in `Codeception\Module` class which is accessible from helpers.
If you really need getConfig method in your test code, add it to Acceptance Helper.

```php
public function getConfig($key = null)
{
  return $this->getModule('WebDriver')->_getConfig($key);
}
```

Please raise a separate pull request for retries functionality, but please read through #3514 first.
Add your change to 2.3 branch, not to master.

* Next time please raise an issue before starting implementation, it saves your time.
* One change per pull request please.
* Please stick to existing conventions and don't reformat every file you touch. (Your IDE removes spaces around `.`, adds blank line before return and aligns assignments and comments, please disable all of that).

PR #4459, opened 2017-08-18
Set transaction also in backupConfig when initializing yii2 module

### LINKED ISSUE #4417: Transaction mode default behaviour for Yii2
Run all functional tests in transaction. I haven't specified transaction or cleanup params in config. As I understand they're both true by default.

#### What do you get instead?
Instead I get only first test run in transaction. All next tests are running not in transaction. And after whole run I have some useless rows in db that cause errors on the next run. Transactions are suppose to solve such problems.

### Details
I did some research. 
[This](https://github.com/Codeception/Codeception/blob/2.3/src/Codeception/Module/Yii2.php#L154) method works once at the beginning and sets transaction to true, then transaction starts in before and rollback in after methods respectively, but on the next test this method is not working(due to its purpose ofc) and transaction wil not start [here](https://github.com/Codeception/Codeception/blob/2.3/src/Codeception/Module/Yii2.php#L198) and will not rollback [here](https://github.com/Codeception/Codeception/blob/2.3/src/Codeception/Module/Yii2.php#L234) as transaction option get null value not true like in initialize method. Is this desirable behaviour or a bug? I waste quite a big time to figure it out. It's better be fixed or documented.

* Codeception version: 2.3.4
* PHP Version: 7.1.3
* Operating System: Centos 7
* Installation type: Composer

jorgdec:
Fixes #4417

### DIFF
--- a/src/Codeception/Module/Yii2.php
+++ b/src/Codeception/Module/Yii2.php
@@ -154,7 +154,7 @@ class Yii2 extends Framework implements ActiveRecord, PartedModule
     public function _initialize()
     {
         if ($this->config['transaction'] === null) {
-            $this->config['transaction'] = $this->config['cleanup'];
+            $this->config['transaction'] = $this->backupConfig['transaction'] = $this->config['cleanup'];
         }
 
         if (!is_file(Configuration::projectDir() . $this->config['configFile'])) {


samdark:
@DavertMik looks correct.

DavertMik:
Thanks! Builds are passing for few PHP versions so I think it's ok to merge it
Sorry for the delay. We had issues with recent Travis updates

PR #4464, opened 2017-08-21
Add doctrine2 to ZendExpressive Module

Rud5G:
Codeception\Module\ZendExpressive implements DoctrineProvider

### DIFF
--- a/src/Codeception/Module/ZendExpressive.php
+++ b/src/Codeception/Module/ZendExpressive.php
@@ -1,10 +1,12 @@
 <?php
+
 namespace Codeception\Module;
 
 use Codeception\Lib\Framework;
 use Codeception\TestInterface;
 use Codeception\Configuration;
 use Codeception\Lib\Connector\ZendExpressive as ZendExpressiveConnector;
+use Codeception\Lib\Interfaces\DoctrineProvider;
 
 /**
  * This module allows you to run tests inside Zend Expressive.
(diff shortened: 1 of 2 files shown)


Naktibalda:
Thanks,

Does it work when multiple requests are performed in a single test?
Other modules supporting Doctrine2 ([Symfony ](https://github.com/Codeception/Codeception/blob/2.3.5/src/Codeception/Lib/Connector/Symfony.php#L66-L80)and [ZF2](https://github.com/Codeception/Codeception/blob/2.3/src/Codeception/Lib/Connector/ZF2.php#L179-L181) with ServiceManager 2.*) implement service persistence by injecting services from the first instance of service manager to new instances created for later requests.

Rud5G:
@Naktibalda i will take a look into this. 
Are there any tests at the moment in the repository for this? i couldn't find them on the first glance.

Naktibalda:
Tests of this module are in https://github.com/Naktibalda/codeception-zend-expressive-tests
You can copy Doctrine tests from https://github.com/Naktibalda/codeception-zf2-tests/tree/2.2/tests/doctrine

src/Codeception/Module/ZendExpressive.php
```diff
@@ -48,8 +50,46 @@ class ZendExpressive extends Framework
      */
     public $application;
 
+
     protected $responseCollector;
 
+    /**
+     * Grabs a service from Interop container.
+     * Recommended to use for unit testing.
```
DavertMik: ...for integration testing

DavertMik:
Looks good to me. If this works for you, @Rud5G I think we can merge it 
Yes, better to have tests for Doctrine but It requires changes for external repositories so we can apply that later

Naktibalda:
It looks good, but it probably doesn't work.

PR #4468, opened 2017-08-24
generateWhereClause() now supports greater than and less than queries

susgo:


src/Codeception/Lib/Driver/Db.php
```diff
@@ -153,22 +153,47 @@ public function select($column, $table, array &$criteria)
...
-            } else {
-                $params[] = $this->getQuotedName($k) . " = ? ";
+            if ($v === null)
+            {
+              $params[] = $this->getQuotedName( $k ) . " IS NULL ";
+              unset( $criteria[ $k ] );
+              continue;
+            }
+
+            $bHasOperand = false;
```
Naktibalda: Please don't use Hungarian notation, `$hasOperand` would be a better name.

src/Codeception/Lib/Driver/Db.php
```diff
@@ -153,22 +153,47 @@ public function select($column, $table, array &$criteria)
         return sprintf($query, $column, $this->getQuotedName($table), $where);
     }
 
+    private function getSupportedOperators(){
+      return [
+        'like' => 'LIKE',
+        '<=' => '<=',
+        '>=' => '>=',
+        '<' => '<',
+        '>' => '>'
```
Naktibalda: Do you actually need key-value pairs here? Keys and values seem to be the same.

src/Codeception/Lib/Driver/Db.php
```diff
@@ -153,22 +153,47 @@ public function select($column, $table, array &$criteria)
...
+            if ($v === null)
+            {
+              $params[] = $this->getQuotedName( $k ) . " IS NULL ";
+              unset( $criteria[ $k ] );
+              continue;
+            }
+
+            $bHasOperand = false;
+            foreach ($operands as $operand => $sql){
+              if (!strpos(strtolower($k), " $operand") > 0){
```
Naktibalda: You could use stripos and str_ireplace to avoid case sensitivity issues with 'LIKE'.

src/Codeception/Lib/Driver/Db.php
```diff
@@ -153,22 +153,47 @@ public function select($column, $table, array &$criteria)
...
-                $params[] = $this->getQuotedName($k) . " = ? ";
+            if ($v === null)
+            {
+              $params[] = $this->getQuotedName( $k ) . " IS NULL ";
+              unset( $criteria[ $k ] );
+              continue;
+            }
+
+            $bHasOperand = false;
+            foreach ($operands as $operand => $sql){
```
Naktibalda: I don't like that value is called `$sql`, especially because it is usually equal to `$operand`.

src/Codeception/Lib/Driver/Db.php
DavertMik: this code is pretty complicated for the first sight to understand and debug. 
could you please add some comments to it

susgo:
@Naktibalda and @DavertMik Thanks for the review. Changes are done! :-)

DavertMik:
Thanks, @susgo 
Could you add a test for few of those operators?

* You will need to add a test to `tests/unit/Codeception/Module/Db/TestsForDb.php`
* Create a few records using `haveInDatabase` and check that you can get correct results using those operators.
* Run a test by executing  `tests/unit/Codeception/Module/Db/MySqlDbTest.php`
* It requires MySQL database `codeception_test`

susgo:
@DavertMik Yes, I will add the test - hopefully this week.

susgo:
@DavertMik I have added the tests but cannot run them. I allways get the message: permission denied. Database was created. Do you have an idea?

Naktibalda:
@susgo how do you run the tests?
Please show the full error output.

susgo:
@Naktibalda Start the tests with  `tests/unit/Codeception/Module/Db/MySqlDbTest.php` and get the following message: 
`zsh: permission denied: tests/unit/Codeception/Module/Db/MySqlDbTest.php`

Naktibalda:
Try `php codecept run tests/unit/Codeception/Module/Db/MySqlDbTest.php`

DavertMik:
Sorry, no idea about permissions but If you can't run tests locally, don't worry, make them run on CI
Right now I see that the database schema is not updated for tests

https://travis-ci.org/Codeception/Codeception/jobs/273827333#L3362

Please change the schema at https://github.com/Codeception/Codeception/tree/2.3/tests/data/dumps

Naktibalda:
Some drivers, e.g. SqlSrv, override generateWhereClause method, so they don't benefit from your change.

Naktibalda:
aargh, the tests were all wrong.

PR #4469, opened 2017-08-24
added imagick to travis.yml

DavertMik:


.travis.yml
```diff
@@ -35,6 +35,7 @@ services:
 
 install:
   - 'pecl install mongodb'
+  - yes '' | pecl install imagick
```
Naktibalda: why is `yes` necessary for imagic, but not for mongodb?

DavertMik: imagick has one interactive question...

DavertMik:
FINALLY!

PR #4471, opened 2017-08-28
Specify failed group name to save tests for Run Failed Ext

maxgorovenko:
If I run my tests parallel via Robo and some tests have failed, then RunFailed Ext tries to put failed tests to the same file 'failed' in _output dir. Each thread do this, so at the end I have list of failed tests only from one thread.

With this config I can manipulate output filename and aggregate all failed tests at the end of work.

### DIFF
--- a/codeception.yml
+++ b/codeception.yml
@@ -29,6 +29,9 @@ modules:
 
 extensions:
     enabled: [Codeception\Extension\RunFailed]
+    config:
+        Codeception\Extension\RunFailed:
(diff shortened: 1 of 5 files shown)


maxgorovenko:
@DavertMik Thanks. I'll fix that!

maxgorovenko:
@DavertMik what do you think about PR now? I've used '--override' as you asked.

DavertMik:
Thanks!

PR #4472, opened 2017-08-29
Add NTLM authentication support to REST module

Tenzian:
- Update composer.json to require Guzzle >=6.3.0 (supports NTLM auth option)
- Add amNTLMAuthenticated() function to Module\REST.php

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
         "phpunit/php-code-coverage": ">=2.2.4 <6.0",
         "phpunit/phpunit-mock-objects": ">2.3 <5.0",
         "facebook/webdriver": ">=1.1.3 <2.0",
-        "guzzlehttp/guzzle": ">=4.1.4 <7.0",
+        "guzzlehttp/guzzle": ">=6.3.0 <7.0",
         "guzzlehttp/psr7": "~1.0",
         "symfony/finder": ">=2.7 <4.0",
         "symfony/console": ">=2.7 <4.0",
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
         "phpunit/php-code-coverage": ">=2.2.4 <6.0",
         "phpunit/phpunit-mock-objects": ">2.3 <5.0",
         "facebook/webdriver": ">=1.1.3 <2.0",
-        "guzzlehttp/guzzle": ">=6.3.0 <7.0",
+        "guzzlehttp/guzzle": ">=4.1.4 <7.0",
         "guzzlehttp/psr7": "~1.0",
         "symfony/finder": ">=2.7 <4.0",
         "symfony/console": ">=2.7 <4.0",
--- a/docs/modules/REST.md
+++ b/docs/modules/REST.md
@@ -69,16 +69,6 @@ Adds HTTP authentication via username/password.
  * `[Part]` xml
 
 
-### amNTLMAuthenticated
-
-Adds Microsoft NTLM authentication via username/password.
-
- * `param` $username
- * `param` $password
- * `[Part]` json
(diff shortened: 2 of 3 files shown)


src/Codeception/Module/REST.php
```diff
@@ -301,6 +301,19 @@ public function amBearerAuthenticated($accessToken)
...
+     * Adds NTLM authentication via username/password.
+     *
+     * @param $username
+     * @param $password
+     * @part json
+     * @part xml
+     */
+    public function amNTLMAuthenticated($username, $password)
+    {
+        $this->client->setAuth($username, $password, 'ntlm');
```
Naktibalda: 1. Please copy `isFunctional` check from `amHttpAuthenticated` method. (throw ModuleException if it is a functional module unless you can implement this kind of auth for framework modules).
2. Implement some way to check a version of guzzle.
3. Add comment that this function only works with Guzzle >=6.3.0

docs/modules/REST.md
```diff
@@ -69,6 +69,16 @@ Adds HTTP authentication via username/password.
...
 
+### amNTLMAuthenticated
+
+Adds Microsoft NTLM authentication via username/password.
+
+ * `param` $username
+ * `param` $password
+ * `[Part]` json
+ * `[Part]` xml
+
```
Naktibalda: This file is auto-generated from PHP file before release, so there is no need to edit it manually.

composer.json
```diff
@@ -22,7 +22,7 @@
         "phpunit/php-code-coverage": ">=2.2.4 <6.0",
         "phpunit/phpunit-mock-objects": ">2.3 <5.0",
         "facebook/webdriver": ">=1.1.3 <2.0",
-        "guzzlehttp/guzzle": ">=4.1.4 <7.0",
+        "guzzlehttp/guzzle": ">=6.3.0 <7.0",
```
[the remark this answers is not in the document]
DavertMik: Just throw an exception if al older Guzzle is detected

src/Codeception/Module/REST.php
```diff
@@ -301,6 +301,28 @@ public function amBearerAuthenticated($accessToken)
...
+     *
+     * @param $username
+     * @param $password
+     * @throws ModuleException
+     * @part json
+     * @part xml
+     */
+    public function amNTLMAuthenticated($username, $password)
+    {
+        if (\GuzzleHttp\Client::VERSION < '6.2.1') {
```
Naktibalda: 6.2.1? It would be better to use [version_compare](http://php.net/version_compare) function, just in case if 6.20.0 is ever release.

Tenzian:
There's an issue with the release of Guzzle 6.3.0 reporting its version as 6.2.1:

It's fixed in source: https://github.com/guzzle/guzzle/commit/95f4deb816266e4ffac55beeb138caa4d42691fd

But at the moment 6.2.2 and 6.2.3 will pass the version check but fail to authenticate.

May want to hold this until a new release of guzzle then correct the version number check to use 6.3.0.

DavertMik:
> There's an issue with the release of Guzzle 6.3.0 reporting its version as 6.2.1:

😆

Ok, let's wait, I think they should release it pretty soon. I like the implementation so we can merge it once the Guzzle is ready.

Meanwhile, you can update the docblock to include an example of using this method:

```
```

Tenzian:
I'll leave the version being checked for as 6.2.1 for the moment (as that currently works for 6.3.0) and update it to 6.3.0 once Guzzle's fixed.

DavertMik:
Ok. Looks good to me. I think we can merge it

PR #4473, opened 2017-08-30
[Db] Update and Insert in database

eXorus:
Goals:

- Add doc for updateInDatabase()
- Add insertInDatabase() that insert a new record without deleted at the end of the test like haveInDatabase()

### DIFF
--- a/src/Codeception/Module/Db.php
+++ b/src/Codeception/Module/Db.php
@@ -395,7 +395,7 @@ class Db extends CodeceptionModule implements DbInterface
      */
     public function haveInDatabase($table, array $data)
     {
-        $lastInsertId = $this->insertInDatabase($table, $data);
+        $lastInsertId = $this->_insertInDatabase($table, $data);
 
         $this->addInsertedRow($table, $data, $lastInsertId);
 
--- a/src/Codeception/Module/Db.php
+++ b/src/Codeception/Module/Db.php
@@ -560,7 +560,20 @@ class Db extends CodeceptionModule implements DbInterface
     {
         return $this->countInDatabase($table, $criteria);
     }
-    
+
+    /**
+     * Update an SQL record into a database.
+     *
+     * ```php
+     * <?php
+     * $I->updateInDatabase('users', array('isAdmin' => true), array('email' => 'miles@davis.com'));
+     * ?>
+     * ```
+     *
+     * @param string $table
+     * @param array $data
+     * @param array $criteria
+     */
     public function updateInDatabase($table, array $data, array $criteria = [])
     {
         $query = $this->driver->update($table, $data, $criteria);
--- a/src/Codeception/Module/Db.php
+++ b/src/Codeception/Module/Db.php
@@ -395,6 +395,29 @@ class Db extends CodeceptionModule implements DbInterface
      */
     public function haveInDatabase($table, array $data)
     {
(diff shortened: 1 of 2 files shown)


src/Codeception/Module/Db.php
```diff
@@ -395,6 +395,29 @@ protected function loadDumpUsingDriver()
...
     {
+        $lastInsertId = $this->insertInDatabase($table, $data);
+
+        $this->addInsertedRow($table, $data, $lastInsertId);
+
+        return $lastInsertId;
+    }
+
+    /**
+     * Inserts an SQL record into a database.
```
DavertMik: how is this different from `haveInDatabase`?
I think having both methods are confusing

eXorus:
@DavertMik My usecase is for acceptance tests, I register a user, insert in the database some data to have it like an admin (that's way I don't need to login as another admin to put this user in admin).
And I use saveSessionSnapshot to save my user session to be able to use it in different test.

But currently with haveInDatabase, the user is not anymore an admin because the line inserted was removed.
So I need a way to keep the line inserted.

DavertMik:
Ok `insertInDatabase` could exist but it should be a hidden method so you could use it from your helpers only. For instance, in your case you should probably create an admin using inside `_beforeSuite` hook of a Helper

Naktibalda:
@eXorus How you are going to prevent a duplicated key error on the following run?

eXorus:
@DavertMik I did the changes to have insertInDatabase only accessible from Helper

@Naktibalda My user is a new registered so I know this user_id is unique and after I'm using loadSessionSnapshot to know if the user is already here or not.

eXorus:
@DavertMik I can't use _beforeSuite because I don't need this user on each suite. I really need a beforeCest :)

rebas82:
i have this...

/**
     * @param $sglQuery
     * @param $dbUsername
     * @param $dbPassword
     */
    public function doSqlUsingPDO($sqlQuery, $dbUsername, $dbPassword)
    {
        $I = $this;
        $serverName = $I->getDataFromEnv("db");
        if ($dbUsername == "") {
            $username = $I->getDataFromEnv("dbLogin");
        } else {
            $username = $dbUsername;
        }
        if ($dbPassword == "") {
            $password = $I->getDataFromEnv("dbPassword");
        } else {
            $password = $dbPassword;
        }
        $dbName = $I->getDataFromEnv("dbName");
        try {
            $conn = new PDO("mysql:host=$serverName;dbname=$dbName", $username, $password);
            // set the PDO error mode to exception
            $conn->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);
            // use exec() because no results are returned
            $conn->exec($sqlQuery);
            echo "SQL query was created successfully";
            $I->comment("SQL query was created successfully with this query:<br>" . $sqlQuery);
        } catch (PDOException $e) {
            $error = $e->getMessage();
            echo $sqlQuery . "<br>" . $error;
            $I->fail(
                "SQL query was not created successfully with this query\n" . $sqlQuery
            );
        }
        $conn = null;
    }

Naktibalda:
@rebas82 what has your code got to do with this pull request?
@eXorus thanks for updateInDatabase, I used it in my tests this week.

DavertMik:
Thanks!

PR #4480, opened 2017-08-31
fix generate stepobject

rebas82:
g:step not working, g:stepobject is correct


DavertMik:
Sorry, this class is genertaed. Change should be sent to `src/Codeception/Command/GenerateSepOject`

PR #4486, opened 2017-09-06
[Fixtures] Add exists method

eXorus:
The goal of this PR is to add an exists method to Fixtures.

```php
if (Fixtures::exists('user1'))
{
    return Fixtures::get('user1');
}
Fixtures::add('user1', 'new user');

```


DavertMik:
Thanks!

PR #4491, opened 2017-09-08
Inject exception file and line number frame into stack trace

### LINKED ISSUE #3200: The file and line where the exception was thrown is missing from the stack trace
With codeception 2.2 the following type of stack trace is shown:

```
  [Error] Call to a member function getFoo() on null

Scenario Steps:
... *scenario steps omitted for clarity* ...

#1  /app/src/Foo.php:280
#2  /app/src/Bar.php:150
...
```

The actual file and line where the exception was thrown (for instance ProblematicFile.php:410) is missing from the stack trace, which makes it necessary to harder to find the actual line that caused the error. 

This was previously reported as a comment in a separate issue: https://github.com/Codeception/Codeception/issues/1636#issuecomment-195749768

rhl-jfm:
This mirrors what PHPUnit does when filtering stack traces (since 2008).

It inject a "fake" frame into the top of the stack trace with the actual line where the unexpected exception was thrown. This makes the stack traces much more usable. Fixes #3200.

### DIFF
--- a/src/Codeception/PHPUnit/Overrides/Filter.php
+++ b/src/Codeception/PHPUnit/Overrides/Filter.php
@@ -19,6 +19,16 @@ class PHPUnit_Util_Filter
             $trace = $e->getSerializableTrace();
         }
 
+        $eFile = $e->getFile();
+        $eLine = $e->getLine();
+
+        if (!self::frameExists($trace, $eFile, $eLine)) {
+            array_unshift(
+                $trace,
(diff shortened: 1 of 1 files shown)


DavertMik:
Thanks for implementing this nice lifehack!

Naktibalda:
It isn't a lifehack, it is really important bugfix.
Thanks for it.

By the way, I just discovered that the most complete stacktrace is produced by `--debug` method.

PR #4493, opened 2017-09-09
unloadFixtures() should run after db transaction

devonliu02:


Naktibalda:
2.2 branch is no longer supported, do you want to make this change in 2.3?

devonliu02:
Sure,   
https://github.com/Codeception/Codeception/pull/4497

PR #4497, opened 2017-09-11
unloadFixtures() should run after db transaction

devonliu02:


DavertMik:
Makes sense... But makes a little sense to do both in transaction :)

devonliu02:
Fixtures were loaded before transaction (if `transaction` is on ), then if we unload fixtures before transaction rollback, the delete SQL in unloadFixture will be rollback too. so it is useless for cleanup configuration.

devonliu02:
Do you mean all fixtures operation should be done in transaction? @DavertMik

DavertMik:
Not sure. As for me - yes, inside transaction. But maybe Yii ocmmunity have another opinion

src/Codeception/Module/Yii2.php
```diff
@@ -223,17 +223,17 @@ public function _after(TestInterface $test)
...
         $_REQUEST = [];
 
+        if ($this->config['transaction'] && $this->transaction) {
+            $this->transaction->rollBack();
+            $this->debugSection('Database', 'Transaction cancelled; all changes reverted.');
+        }
+
         if ($this->config['cleanup']) {
             foreach ($this->loadedFixtures as $fixture) {
                 $fixture->unloadFixtures();
```
DavertMik: ok but fixtures are cleaned up by rolled back transaction...
and this call may throw an exception...

samdark:
I don't understand the change. Currently there are two options:

1. Wrap everything with a transaction so no cleanup actually needed: transaction reverts data from fixtures by itself.
2. Not using transaction but cleaning up manually with `unloadFixtures()`.

Why should we use both?

devonliu02:
the problem here is that fixtures were  **loaded before the transaction**, and transaction will not reverts them.

samdark:
Would this change fix loading fixtures before transaction?

devonliu02:
Maybe I didn't make that clear, or I misunderstood the configuration.

I want that fixtures will be unloaded after each test, and transaction on too.

Let's see the problem step by step,

1. before test (`_before`)
    1. load fixtures
    2. begin transaction
2. test run
3. after test (`_after`)
    1. unload fixtures
    2. rollback transaction

Now, we can see that fixtures are not really be unloaded because the transaction has been rollback.
In this case, the fixtures will be loaded `again` before next test run.

samdark:
As far as I remember, sequence was meant to be:

1. before test (`_before`)
    1. begin transaction
    2. load fixtures
2. test run
3. after test (`_after`)
    3. rollback transaction
    4. unload fixtures in case transactions aren't enabled

Is is different?

devonliu02:
https://github.com/Codeception/Codeception/blob/2.3/src/Codeception/Module/Yii2.php#L187

It doesn't look like that to me 😂

samdark:
Indeed. @DavertMik need your input on original intentions...

DavertMik:
Sorry, you are right. I missed the execution order in _before. Looks like it changed after I looked into it last time. Yes, we need to merge it

PR #4501, opened 2017-09-12
fillField without clearing it bevor

susgo:
Needed for textareas where user have to insert a line break with the enter-key

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -1617,10 +1617,12 @@ class WebDriver extends CodeceptionModule implements
         $field->click();
     }
 
-    public function fillField($field, $value)
+    public function fillField($field, $value, $clearField = true)
     {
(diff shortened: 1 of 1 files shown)


Naktibalda:
Using Boolean arguments is a bad practice.
There is an existing [appendField](http://codeception.com/docs/modules/WebDriver#appendField) method which does what you want.

susgo:
Thanks! We have not found this function during our research. Works for us. Why it is bad practice using booleans as arguments?

Naktibalda:
> Boolean arguments loudly declare that the function does more than one thing. They are confusing and [should be eliminated](http://www.informit.com/articles/article.aspx?p=1392524).

https://www.martinfowler.com/bliki/FlagArgument.html
https://blog.codinghorror.com/avoiding-booleans/
https://www.google.co.uk/search?q=boolean+arguments+are+bad

PR #4503, opened 2017-09-12
Fix @dataProvider example

FallDi:
* annotation is case-sensitive
* dataProvider must be `public`

### DIFF
--- a/docs/07-AdvancedUsage.md
+++ b/docs/07-AdvancedUsage.md
@@ -249,7 +249,7 @@ You can also use the `@dataprovider` annotation for creating dynamic examples, u
     /**
      * @return array
      */
-    public function pageProvider() // alternatively, if you want the function to be public, be sure to prefix it with `_`
+    public function pageProvider()
     {
         return [
             ['url'=>"/", 'title'=>"Welcome"],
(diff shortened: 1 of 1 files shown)


docs/07-AdvancedUsage.md
```diff
@@ -249,7 +249,7 @@ You can also use the `@dataprovider` annotation for creating dynamic examples, u
     /**
      * @return array
      */
-    protected function pageProvider() // alternatively, if you want the function to be public, be sure to prefix it with `_`
+    public function pageProvider() // alternatively, if you want the function to be public, be sure to prefix it with `_`
```
Naktibalda: Your change conflicts with comment.

FallDi: @Naktibalda i'm not sure what really comment means. Remove it as outdated

Naktibalda: If you make method public in Cest file, Codeception will run it as a test.

FallDi: Wow, I see.
My mistake is create `Test` file instead of `Cest`.

Sorry for any inconvenience

Issue #4509, opened 2017-09-13
`see` assertion fails when source code contains `<` in JS code


tobias-kuendig:
When asserting `$I->canSee('XYZ');` in an acceptance test using PhpBrowser the assertion returns `false` even though the page contains `XYZ`. 

The problem in my case was that my page contained code simillar to the following:

```html
<script>
    if (a <= b) {
        // ...
    }
</script>
<p>XYZ</p>
```

When the `getNormalizedResponseContent` gets called, [all HTML tags are removed with `strip_tags`](https://github.com/Codeception/Codeception/blob/758dd0d30fa7dd08b3eec6e444a162985b09c708/src/Codeception/Lib/InnerBrowser.php#L1880). `strip_tags` interprets the `<=` itself as a start tag and removes everything that follows the operator. Because of this problem the `XYZ` text cannot be found since it is removed from the response content.

I suggest removing the `script` tags manually before using `strip_tags`. For possible solutions see https://stackoverflow.com/a/22782081.

DavertMik:
Interesting issue and looks like a nice solution. Could you prepare a patch and a test for this case?

tobias-kuendig:
Sure, see https://github.com/Codeception/Codeception/pull/4510

PR #4510, opened 2017-09-13
Fix `see` when source code contains `<=` JS operator

tobias-kuendig:
This PR fixes https://github.com/Codeception/Codeception/issues/4509

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -1877,6 +1877,10 @@ class InnerBrowser extends Module implements Web, PageSourceSaver, ElementLocato
     protected function getNormalizedResponseContent()
     {
         $content = $this->_getResponseContent();
+        // Since strip_tags has problems with JS code that contains
+        // an <= operator the script tags have to be removed manually first.
+        $content = preg_replace('#<script(.*?)>(.*?)</script>#is', '', $content);
+        
         $content = strip_tags($content);
         $content = html_entity_decode($content, ENT_QUOTES);
         $content = str_replace("\n", ' ', $content);
--- a/tests/data/app/view/info.php
+++ b/tests/data/app/view/info.php
@@ -44,5 +44,14 @@
     <a id="third-link">Third</a>
 </div>
 
+<script>
+    var a = 2;
+    var b = 3;
+    if (a <= b) {
+        console.log('a is less than b!');
+    }
+</script>
+<p>Text behind JS comparision</p>
+
 </body>
 </html>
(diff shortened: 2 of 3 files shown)


tobias-kuendig:
My test is still passing but the closing `')` remains in the content. Is this a problem?

```html
<div>Kill & Destroy</div>

<div style="display: none" class="hidden">
    Invisible text
</div>

<div id="grab-multiple">
    <a id="first-link">First</a>
    <a id="second-link">Second</a>
    <a id="third-link">Third</a>
</div>

<script>
    var a = 2;
    var b = 3;
    if (a <= b) {
        console.log('a is less than b!');
    }
</script>
<script>console.log('</script>')</script>
<p>Text behind JS comparision</p>
```

becomes...

image

Naktibalda:
Another attempt at breaking your code:

```html

<script>
    var a = 2;
    var b = 3;
    console.log('</script>')
    if (a <= b) {
        console.log('a is less than b!');
    }
</script>
<p>Text behind JS comparision</p>
```

DOM based solution would be more robust.

tobias-kuendig:
I've implemented a new solution using `DOMDocument` that survives your code example.

Naktibalda:
The last change broke a bunch of other tests that are testing `see` and `dontSee` methods.

tobias-kuendig:
Looks like the `DOMDocument` solution brings along a lot of other problems with encoding and error handling with invalid html input.

I have tried another solution that simply replaces all `<=` with `==` inside the `script` tags to make sure `strip_tags` works as expected.

What do you think?

src/Codeception/Lib/InnerBrowser.php
```diff
@@ -1877,6 +1877,10 @@ protected function setCookiesFromOptions()
     protected function getNormalizedResponseContent()
     {
         $content = $this->_getResponseContent();
+        // Since strip_tags has problems with JS code that contains
+        // <= operators, these operators are replaced manually
+        // first to make sure strip_tags works as expected.
+        $content = preg_replace("/(<script.*>.*)(<=)(.*<\/script>)/is", "$1==$3", $content);
```
Naktibalda: This pattern replaces only one occurrence  of `<=` between the first `<script>` and the last `</script> tags.
Even if `.*` was changed to `.*?` it would replace only one occurrence  of `<=` in each script block.
What if there are more such conditions?

Naktibalda:
I think that the first solution was the best even if it hasn't covered all edge cases.

Solution on app side is to move JavaScript code to .js files.

src/Codeception/Lib/InnerBrowser.php
tobias-kuendig: That's true. And I just tested it with multiple `<script>` sections and it fails as well. I too think the first solution fixed the problem at hand, just not the edge cases.

tobias-kuendig:
I have reverted the code to the first solution.

DavertMik:
Please update Pull Request then

tobias-kuendig:
Sorry I'm not sure what you mean. The PR looks up-to-date to me?

PR #4511, opened 2017-09-13
Simplified DiffFactoryTest

Naktibalda:
In order to be compatible with different versions of sebastian/diff


Naktibalda:
@DavertMik My first impression after skimming changes of `sebastian/diff` was that it is a just another Sebastian style bump in required version of PHP and some PHP 7 sprinkles, but a deeper look shows that is a quite significant rewrite (ChangeLog entry doesn't cover it all).

In diff 1.x, constructor of Differ class had parameter `$showNonDiffLines = true`,
In diff 2 there is no such parameter.
So our test failed because new version doesn't return 2 lines that hasn't changed after the difference.

PR #4520, opened 2017-09-18
Fix Underscore Bug in HTTP Headers

bnpatel1990:
Codeception was converting all the underscores in the headers to Hyphen causing the headers to be malformed. 

We utilize custom headers in our development with underscores - example - "Client_id"
When the header is appended with HTTP and later removed in the guzzle utility the underscores are converted to hyphens leading to a malformed header. 

I am just resetting the header to original status once the http tag is stripped.

### DIFF
--- a/src/Codeception/Lib/Connector/Guzzle6.php
+++ b/src/Codeception/Lib/Connector/Guzzle6.php
@@ -216,7 +216,7 @@ class Guzzle6 extends Client
         foreach ($server as $header => $val) {
             $header = implode('-', array_map('ucfirst', explode('-', strtolower(str_replace('_', '-', $header)))));
             if (strpos($header, 'Http-') === 0) {
-                $headers[substr($header, 5)] = $val;
+                $headers[str_replace('-', '_', substr($header, 5))] = $val;
             } elseif (isset($contentHeaders[$header])) {
                 $headers[$header] = $val;
             }


src/Codeception/Lib/Connector/Guzzle6.php
```diff
@@ -216,7 +216,7 @@ protected function extractHeaders(BrowserKitRequest $request)
         foreach ($server as $header => $val) {
             $header = implode('-', array_map('ucfirst', explode('-', strtolower(str_replace('_', '-', $header)))));
             if (strpos($header, 'Http-') === 0) {
-                $headers[substr($header, 5)] = $val;
+                $headers[str_replace('-', '_', substr($header, 5))] = $val;
```
DavertMik: > We utilize custom headers in our development with underscores - example - "Client_id"

if so, why do you perform the change for general headers, not custom?

DavertMik:
I don't see a point in this change.

src/Codeception/Lib/Connector/Guzzle6.php
bnpatel1990: I did forget about the other general header types. I agree they should not be messed around with. 

I am pretty sure we cannot control the Client_id / Client_secret header, it is not something that is custom just to our development process. Here is a [link](https://docs.mulesoft.com/api-manager/configuring-an-api-gateway) to better shed the light on it. Thus we have those specific headers.

Now I did notice in innerbrowser.php (line 162) that headers are transformed. I am not sure if this is by design or not. So I am wondering why transform the headers here to underscore and revert it back in individual clients?? Can you please help understand the process around here please? 

```
  protected function clientRequest($method, $uri, array $parameters = [], array $files = [], array $server = [], $content = null, $changeHistory = true)
    {
               .
               .
               .
            $header = str_replace('-', '_', strtoupper($header));
            $server["HTTP_$header"] = $val;
              .
              .
              .

```

I think I will just have to extend out this functionality locally so I can handle those header types.

bnpatel1990: @DavertMik @Naktibalda 

Hi Guys adding more to the convo - I even dug up the [http standards](https://tools.ietf.org/html/rfc7230#section-3.2) - mentions underscore is a valid character that can be expected as part of header. Will this be addressed in future releases?? 

I am still trying to figure out why transform the headers?? Any help would be appreciated Thanks.

bnpatel1990: okay Nvm - I just saw this issue with [symfony](https://tools.ietf.org/html/rfc7230#section-3.2) I get why the code is converting those to underscores.

PR #4523, opened 2017-09-19
[Travis] Force installation of mongodb

Naktibalda:
https://github.com/travis-ci/travis-ci/issues/8399#issuecomment-330540303


Naktibalda:
It looks like this change is no longer necessary.

PR #4525, opened 2017-09-21
Cest generator: suite namespaces

pohnean:
Hi I've added support to allow for suites to have different namespaces from the main configuration namespace, while the actors still use the main namespace.

The specific use case for me was that I wanted to have a different namespace for each of my test suites, but still have the Actors be in the /tests/_support folder.

Sample suite configuration:
```
suite_namespace: 'tests\unit'
class_name: 'UnitTester'
```
Sample global configuration:
```
namespace: 'tests'
```

This pull request solves that by adding a new suite_namespace setting in the suite. This should not cause any breaking changes.

### DIFF
--- a/src/Codeception/Lib/Generator/Cest.php
+++ b/src/Codeception/Lib/Generator/Cest.php
@@ -50,14 +50,18 @@ EOF;
 
         $namespace = rtrim($this->settings['namespace'], '\\');
         $ns = $this->getNamespaceHeader($namespace.'\\'.$this->name);
-        if ($ns) {
+
+        $actorHasNamespace = strpos($actor, '\\') !== false;
+		if ($actorHasNamespace) {
+			$ns .= "use ".ltrim($actor, '\\').";";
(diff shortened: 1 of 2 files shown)


pohnean:
ok, please ignore this patch for now... I just realised this doesn't work yet because codeception won't run with a namespaced actor. Running it will Error: `\tests\UnitTester class doesn't exist in suite folder.`

DavertMik:
I like the idea, I like the implementation but I'm not so sure about the name.

`suite_namespace` is so general term, I'm not so sure it is very descriptive.

`tests_namespace` may sound a bit better. As we set a namespace for test classes...

but I'm not sure, anyway

DavertMik:
I didn't get better ideas so far so I'm about to merge this as it is

DavertMik:
Ok, I will update the documentation. Thanks, @pohnean

pohnean:
Sweet! 👍

PR #4541, opened 2017-09-27
Header Enhancements

bnpatel1990:
Updates to [Fix Underscore Bug in HTTP Headers #4520](https://github.com/Codeception/Codeception/pull/4520)

So this is an enhancement to accommodate special chars in the headers like underscore. 

I can now pass in a header with html encoded character example`client&#95;id` and later during header extraction decode the html character `client_id`.

The html decoding will only work if encoded chars are passed in else the functionality is unaffected.

### DIFF
--- a/src/Codeception/Lib/Connector/Guzzle.php
+++ b/src/Codeception/Lib/Connector/Guzzle.php
@@ -216,7 +216,7 @@ class Guzzle extends Client
 
         $contentHeaders = ['Content-Length' => true, 'Content-Md5' => true, 'Content-Type' => true];
         foreach ($server as $header => $val) {
-            $header = implode('-', array_map('ucfirst', explode('-', strtolower(str_replace('_', '-', $header)))));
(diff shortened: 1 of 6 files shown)


Naktibalda:
This workaround should be documented somewhere.

bnpatel1990:
I can update the required documents. And make another commit.

DavertMik:
I'm not so sure where to put it so it would be visible.
Is that feature used with `haveHttpHeader`? If so we should put it into the documentation section of InnerBrowser class for `haveHttpHeader` method http://codeception.com/docs/modules/PhpBrowser#haveHttpHeader

Naktibalda:
@DavertMik good idea.
@bnpatel1990 Please document this feature in the docblock of [InnerBrowser::haveHttpHeader](https://github.com/Codeception/Codeception/blob/2.3.5/src/Codeception/Lib/InnerBrowser.php#L286-L306).

bnpatel1990:
Yeah i use it in inner browser `havehttpheader`. Will do in awhile and push a commit

bnpatel1990:
Docs Updated @Naktibalda @DavertMik

DavertMik:
Thanks!

PR #4544, opened 2017-09-29
Improvements to Config Templates section in docs

unt01d:
Thought I'd have a stab at improving the text here, mainly from a spelling/grammar/style point of view.  Also corrected the example name format for suite config dist files, from `suitename.dist.yml` to `suitename.suite.dist.yml`.


Naktibalda:
Thanks

PR #4548, opened 2017-09-30
Db generate where clause tests

Naktibalda:
I merged pull request with failing tests #4468

Here are better tests.


DavertMik:
Nice to have them. Thank you

PR #4562, opened 2017-10-12
Added the ability to export the code coverage data in PHPUnit xml format

tobiasstadler:


DavertMik:
So this PR adds another XML exporting format. Default XML is JUnit while this PR adds proposal to support PHPUnit-flavoured XML, right?

tobiasstadler:
Yes, but the XML format is for exporting code coverage data only. So it should rather be compared to the Clover XML than JUnit.

DavertMik:
Yes, sorry missed that. What's the difference comparing to clover format?
Which software accepts PHPUnit coverage format?

tobiasstadler:
For each covered line of code PHPUnit format lists the tests which covered it.

It is used by e.g. https://github.com/infection/infection.

The test fails because https://github.com/Codeception/c3/pull/33 is not yet merged.

DavertMik:
> The test fails because Codeception/c3#33 is not yet merged.

Thanks, merged and released

tobiasstadler:
Thanks, so what is the next step here?

DavertMik:
c3 should be updated in this branch so tests could pass on Travis

tobiasstadler:
Should I change the require-dev in [composer.json](https://github.com/Codeception/Codeception/blob/master/tests/data/claypit/composer.json) to point to the 2.0 branch of c3 or should I merge my changes to the master branch of c3?

tobiasstadler:
@DavertMik the tests are finally passing. Can you please have a look if my changes are ready to merge?

tests/coverage/RemoteServerWithPHPUnitCept.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+if (!class_exists('\PHP_CodeCoverage_Report_PHPUnit')) {
```
Naktibalda: It would be good to add such checks to places where this class is used and throw an exception with message that this report is only available with certain version of PHPUnit (please specify requirements).

tobiasstadler:
Hi @Naktibalda,
I think I addressed your concern

tobiasstadler:
Any chance for getting this merged?

borNfreee:
Would it awesome to get it merged and [continue integrating Codeception with Infection](https://github.com/infection/infection/issues/45)
ping @DavertMik

DavertMik:
Oh, wow! I just saw Inflection inside a twitter, opened a website, started clicking links and now I'm here! 
Why not merge it then 😁

DavertMik:
Sorry for the delay!

DavertMik:
@tobiasstadler could you write a few paragraphs for https://github.com/Codeception/Codeception/blob/2.3/docs/05-UnitTests.md on how to use mutation testing  with Codeception? That should be part of our testing guides too )

tobiasstadler:
@DavertMik Yes, will do that when my infection changes are merged

DavertMik:
By that time I have some plans to dramatically improve unit testing documentation

DavertMik:
OK, I updated the documentation. Please add a section on Mutation Testing in the Unit Testing chapter.

CJDennis:
Are you planning to support Mutation testing only on PHP>=7 since Infection is is for PHP>=7? My work project (which I inherited without any tests at all) is stuck on PHP 5.5.9 until I can get enough code coverage, and not being able to use the tool that would allow me to upgrade it with confidence kind of defeats the purpose. I'd love to add Mutation testing to increase confidence that breaking changes would be detected.

PR #4566, opened 2017-10-16
Fixed setExpectedException() default value for PHPUnit 5.7.23

MilesChou:
There has some update on [PHPUnit 5.7.23][]. It's solve the exception with empty message but adjust default behavior, so that we should fix default value on `setExpectedException()` method for this update.

[PHPUnit 5.7.23]: https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-5.7.md#5723---2017-10-15


DavertMik:
Thanks!

PR #4577, opened 2017-10-19
coverage: removed unexpected page load

marc-mabe:
The line `$this->module->amOnPage('/');` in `LocalServer::startCoverageCollection()` is not needed but introduces an unexpected page load.

This is not only for performance reasons but my tests fail with coverage enables as I'm using an own written module to break if something happens to a specific error log and I don't have a route for `/` so I get an error log entry with message `ERR (3): error-router-no-match` on each test if coverage is enabled and my tests will fail :(

### DIFF
--- a/src/Codeception/Coverage/Subscriber/LocalServer.php
+++ b/src/Codeception/Coverage/Subscriber/LocalServer.php
@@ -156,7 +156,11 @@ class LocalServer extends SuiteSubscriber
         ];
         $value = json_encode($value);
 
-        $this->module->setCookie(self::COVERAGE_COOKIE, $value);
+        $c3Url = parse_url($this->settings['c3_url'] ? $this->settings['c3_url'] : $this->module->_getUrl());
+        $c3Host = isset($c3Url['host']) ? $c3Url['host'] : 'localhost';
+        $c3Host .= isset($c3Url['port']) ? ':' . $c3Url['port'] : '';
+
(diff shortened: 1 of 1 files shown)


Naktibalda:
It is necessary for setCookie to work correctly, isn't it?

marc-mabe:
> It is necessary for setCookie to work correctly, isn't it?

No it's not - `amOnPage()` will immediately make a request and `setCookie()` will define a cookie to send for the following requests.

At least for me everything works fine without that line.

Naktibalda:
setCookie doesn't know what domain to use before the first request is made. Request will only send cookies having a matching domain name. So your change could be working for localhost (not sure), but it will fail if a different hostname is used.

marc-mabe:
@Naktibalda Now I tested it using a different host by changing configuration to `google.com` and add `127.0.0.1  google.com` in `/etc/hosts`. -> It still works fine.

In case you encounter problems you can also force the domain in `setCookie()`. Something linke that:
```php
$c3Url = parse_url($this->settings['c3_url'] ? $this->settings['c3_url'] : $this->module->_getUrl());
$c3Host = isset($c3Url['host']) ? $c3Url['host'] : 'localhost';
$c3Host .= isset($c3Url['port']) ? ':' . $c3Url['port'] : '';
$this->module->setCookie(self::COVERAGE_COOKIE, $value, array('domain' => $c3Host));
```

Naktibalda:
Please set domain in setCookie line and I will approve this pull request.
It is still a subject to review by @DavertMik

PR #4582, opened 2017-10-22
[PhpBrowser] User-Agent can be set in config

### LINKED ISSUE #4576: user-agent header gets overwritten
Only by changing to User-Agentt header appears inside printed by php headers list.
```
modules:
    config:
        PhpBrowser:
          headers:
            User-Agentt: Mozilla/5.0 (iPhone; CPU iPhone OS 9_1 like Mac OS X) AppleWebKit/601.1.46 (KHTML, like Gecko) Version/9.0 Mobile/13B143 Safari/601.1
```
PHP received headers inside acceptance test:
```
[HTTP_USER_AGENTT] => Mozilla/5.0 (iPhone; CPU iPhone OS 9_1 like Mac OS X) AppleWebKit/601.1.46 (KHTML, like Gecko) Version/9.0 Mobile/13B143 Safari/601.1

[HTTP_USER_AGENT] => Symfony2 BrowserKit
```

This is not working:
```
modules:
    config:
        PhpBrowser:
          headers:
            User-Agent: Mozilla/5.0 (iPhone; CPU iPhone OS 9_1 like Mac OS X) AppleWebKit/601.1.46 (KHTML, like Gecko) Version/9.0 Mobile/13B143 Safari/601.1
```

Naktibalda:
Headers option is processed by the module instead of passing it to constructor of Guzzle Client

Fixes #4576

### DIFF
--- a/src/Codeception/Module/PhpBrowser.php
+++ b/src/Codeception/Module/PhpBrowser.php
@@ -23,17 +23,15 @@ use GuzzleHttp\Client as GuzzleClient;
  * * Maintainer: **davert**
  * * Stability: **stable**
  * * Contact: codeception@codeception.com
- * * Works with [Guzzle](http://guzzlephp.org/)
  *
- * *Please review the code of non-stable modules and provide patches if you have issues.*
  *
  * ## Configuration
  *
  * * url *required* - start url of your app
+ * * headers - default headers are set before each test.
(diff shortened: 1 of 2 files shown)


DavertMik:
Looks good. Is this compatible with old configuration style?

Naktibalda:
Yes, it takes key-value pairs in the same way as Guzzle: http://docs.guzzlephp.org/en/latest/request-options.html#headers

DavertMik:
Ok

PR #4584, opened 2017-10-23
allow domain_regex in symfony module

karser:
`getInternalDomains` dumps all the app's routes in order to find out which domains the app uses, it is too expensive and awkward. I find useful ability to specify the domain to test against
```
#tests/functional.suite.yml
modules:
    enabled:
        - Symfony:
            domain_regex:
                - '/.+\.domain\.com/'
```


src/Codeception/Module/Symfony.php
```diff
@@ -554,20 +555,7 @@ private function extractRawRoles(Data $data)
...
-            if (!is_null($route->getHost())) {
-                $compiled = $route->compile();
-                if (!is_null($compiled->getHostRegex())) {
-                    $internalDomains[] = $compiled->getHostRegex();
-                }
-            }
-        }
-
-        return array_unique($internalDomains);
+        return $this->config['domain_regex'];
```
Naktibalda: Your change is not backwards compatible.

Naktibalda:
How long does getInternalDomains take for your app?

karser:
In our app, there are more than 600 routes, so it takes:
coldstart: 11,6s
hotstart: 0,39s
```
time sf debug:router --env=test
```
But why even need to dump all the routes in order to determine which domains the app uses?
I know it's [possible](https://symfony.com/doc/current/routing/hostname_pattern.html) to specify the host in symfony routes, so am I supposed to override my routes for the test environment and specify the domain I'm going to test my app against? Ability to explicitly specify domain_regex list seems more convenient

karser:
@Naktibalda please review again, I added backward compatibility
https://github.com/survos/Codeception/commit/bdac84868d7df2caf5289786745c17277b4f05ed

Naktibalda:
> But why even need to dump all the routes in order to determine which domains the app uses?

This feature was implemented to prevent odd case scenario when external link is processed as an internal one.
For example `click` on `<a href="http://google.com/">`  happily returned your main page, because it matched `home` route `/`.

karser:
@Naktibalda do you think it is backward compatible now?

Naktibalda:
Yes, it is backwards compatible.

I don't like the name of the new setting. It describes what it contains, not what it does.

Also maybe it would be more user friendly if values were regular strings and not regex.

It would be good to add this feature to all modules, but I can do that later.

karser:
How would it be possible to specify the wildcard (`*.domain.com`) without regex? Now it's:
```
domain_regex:
    - '/.+\.domain\.com/'
```
The function `getInternalDomains` itself expects the result to be an array of regex, so it fits into its logic.

Regarding the option name, is `allowed_domains` better?

Naktibalda:
I apologize for forgetting your pull request.

How about naming your new parameter `domains`?
```yaml
domains:
  - localhost
  - admin.site.org
```

It is very unlikely that anyone actually needs a wildcard matching for domains in the test code.

Documentation could look like this:
```
* domains - List of domains that the app is tested against (also a good optimization if your app uses a lot of routes).
```

Domains could be converted to regexp using this code `'/' . preg_quote($domain) . '/'`

DavertMik:
I agree with idea to use just `domains` config value. Domain regex is too complex for using.

Issue #4588, opened 2017-10-24
Why is dontSeeInpopUp Gone?


kpascal:
#### What are you trying to achieve?

I want dontSeeInPopUp back in source code so I can use it 

#### What do you get instead?
no such function found

### test steps

>  $I->dontSeeInPopup("Your email will be re-sent shortly.");
>   $I->acceptPopup();
>   $I->wait(1);

#### Console Output

> Time: 59.31 seconds, Memory: 12.75MB
> 
> There was 1 error:
> 
> ---------
> 1) AccountViewEmailHistoryTabCest: Test verify the resend link is not displaying as the customersetting
>  Test  tests\cc\sdfdfCest.php:AccountView
> 
>   [RuntimeException] Call to undefined method CcTester::dontSeeInPopup
> 

```
#### Source Code

There is no source code in vendor, however I have this working in one of my local helpers


```

> public function dontSeeInPopup($text) {
>         $alert = $this->getModule('WebDriver')->webDriver->switchTo()->alert();
>         try {
>             $this->assertNotContains($text, $alert->getText());
>         } catch (\PHPUnit_Framework_AssertionFailedError $e) {
>             $alert->dismiss();
>             throw $e;
>         }
>     }

```


### Details

* Codeception version:  2.3.6
* PHP Version: 5.6.29
* Operating System: windows server 2008

DavertMik:
well, I assume it was in old Selenium module and was not reimplemented in WebDriver module.

yeah, I like your implementation. Could you send Pull Request?

kpascal:
Created the PR

https://github.com/Codeception/Codeception/pull/4589

PR #4589, opened 2017-10-24
added dontSeeInPopUp function

kpascal:
this is in reference to issue https://github.com/Codeception/Codeception/issues/4588

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -1907,6 +1907,28 @@ class WebDriver extends CodeceptionModule implements
     }
 
     /**
+     * Checks that the active JavaScript popup,
+     * as created by `window.alert`|`window.confirm`|`window.prompt`, does NOT contain the given string.
+     *
+     * @param $text
+     *
+     * @throws \Codeception\Exception\ModuleException
+     */
+    public function dontSeeInPopup($text)
+    {
+        if ($this->isPhantom()) {
+            throw new ModuleException($this, 'PhantomJS does not support working with popups');
+        }
+        $alert = $this->webDriver->switchTo()->alert();
+        try {
+            $this->assertNotContains($text, $alert->getText());
+        } catch (\PHPUnit_Framework_AssertionFailedError $e) {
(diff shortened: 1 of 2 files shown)


DavertMik:
Please fix that failing test with quotes:

> tests/web/WebDriverTest.php:testFailedDontSeeInPopup
Failed asserting that exception message 'Failed asserting that 'Really?' does not contain "Really?".' contains 'Failed asserting that 'Really?' contains "Really?"'.

and we are done :)

kpascal:
I believe it was failing because of the $alert->dismiss(); let me know if that really needs to be there?

DavertMik:
According to the message the reason is different.

> I believe it was failing because of the $alert->dismiss();

No it's really important to do something with alert. Consider a situation when test finishes with an open popup. A next test won't be able to handle new browser session as popup will be visible

nshelms:
I'm trying to help @kpascal with this PR but i'm having trouble running the tests locally.  We've added the function into WebDriver module, but when running `codecept run web WebDriverTest:testDontSeeInPopup` I'll get an error saying that Webdriver::dontSeeInPopup is undefined when in fact it is defined inside WebDriver module. I've already ran `codecept build` manually but it does not add any additional methods into the generated helpers. Even manually adding the references inside WebGuyActions isn't working. Any thoughts?

kpascal:
in the event the text is seen while using dontSeeInPopUp, the test will automatically die because its prefixed with 'dont' therefore not allowing the test to continue, killing the session and killing the popup

DavertMik:
> in the event the text is seen while using dontSeeInPopUp, the test will automatically die because its prefixed with 'dont' therefore not allowing the test to continue, killing the session and killing the popup

A test will but process won't. Browser also won't be closed. Esp if you reuse browser session.
Also, this method can be used as `cantSeeInPopup` so it won't stop on failure. 
You see, lots of cases

DavertMik:
Hey, looks like you are "First-time GitHub contributor" 
Congratulations on first merged pull request! :balloon:

PR #4599, opened 2017-11-02
Fail hard if test has impossible dependency

### LINKED ISSUE #4598: [Cest] last test in file is skipped if some test has invalid dependency
Example code:
```php
class DebugIssueCest
{
    /**
     * @depends test5
     */
    public function test1(\FunctionalTester $I){}

    public function test2(\FunctionalTester $I){}

    public function test3(\FunctionalTester $I){}
}
```

test3 has no dependencies, so it is expected that running it alone should execute the test,
but it is skipped.

```sh
$ codecept run tests/functional/DebugIssueCest.php:test3 -vv
....
Functional Tests (1) ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Modules: \Helper\Functional, REST, PhpBrowser
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Testing functional
DebugIssueCest: Test3
Signature: DebugIssueCest:test3
Test: tests/functional/DebugIssueCest.php:test3
Scenario --
 SKIPPED: Dependent test for test5 not found
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

Naktibalda:
Fixes #4598

Instead of skipping the last test Codeception will throw exception:
```
  [Exception]
  Dependent test test5 for DebugIssueCest:test1 not found
```

### DIFF
--- a/src/Codeception/Suite.php
+++ b/src/Codeception/Suite.php
@@ -27,14 +27,16 @@ class Suite extends \PHPUnit_Framework_TestSuite
         }
         $this->tests = $queue;
     }
+
     protected function getDependencies($test)
     {
         if (!$test instanceof Dependent) {
             return [$test];
         }
+        $forTest = Descriptor::getTestSignature($test);
         $tests = [];
         foreach ($test->getDependencies() as $requiredTestName) {
-            $required = $this->findMatchedTest($requiredTestName);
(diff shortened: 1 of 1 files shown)


Naktibalda:
This is too harsh - cross-file dependencies are not found when a single file is executed, skipping the right test should be enough.

PR #4604, opened 2017-11-03
Adding Logic In Case Of Exit

Fenikkusu:
I have been working on getting my testing setup within a CI. I was looking at the output of one such run when I realized that the test did not appear to be finishing, but was not giving any errors. After digging into it, I learned that one of my Junior Developers created a new class following some of the legacy code in the system. They put includes inside a class file (rather than relying on the PSR-4 Autoloader). One of these includes was a script to check a user's login/session and redirecting them to the homepage if they weren't logged in. Because a call to exit without params is equivalent to exit(0), my CI system thought that the test completed succesfully when it didn't actual do so.

This change is an attempt to catch such instances in the future. I use a shutdown function defined specifically for the execute to throw an exception in case the shutdown occurs.

Please note that I've not studied the Codeception Codebase so there may be a better way to detect this situation.

### DIFF
--- a/src/Codeception/Subscriber/ErrorHandler.php
+++ b/src/Codeception/Subscriber/ErrorHandler.php
@@ -11,7 +11,8 @@ class ErrorHandler implements EventSubscriberInterface
     use Shared\StaticEvents;
 
     public static $events = [
-        Events::SUITE_BEFORE => 'handle'
+        Events::SUITE_BEFORE => 'handle',
+        Events::SUITE_AFTER  => 'onFinish'
     ];
 
     /**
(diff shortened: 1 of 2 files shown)


Fenikkusu:
After reviewing the code a bit more, I moved the logic into SuiteManager::run instead of Run::execute. It made more since and fixed the issues with testing.

DavertMik:
Thanks. Sorry for taking that long to respond. 
But we already have the shutdown function, it's located in https://github.com/Codeception/Codeception/blob/2.3/src/Codeception/Subscriber/ErrorHandler.php

Could you update that class in order to keep the core classes decoupled from error listeners

Fenikkusu:
I've applied the updates. I did have to update another one of the tests that was creating an instance of the ErrorHandler and causing a Did Not Finish Error.

DavertMik:
Looks ok to me!

DavertMik:
Thanks!

PR #4611, opened 2017-11-09
Allow .tgz extensions to pass validation

Lukazar:
.tgz is a valid extension.

### DIFF
--- a/src/Codeception/Module/MongoDb.php
+++ b/src/Codeception/Module/MongoDb.php
@@ -158,7 +158,7 @@ class MongoDb extends CodeceptionModule implements RequiresPackage
                         "Tar gunzip archives are not supported for Windows systems"
                     );
                 }
-                if (strlen($this->dumpFile) <= 7 || substr($this->dumpFile, -7) !== '.tar.gz') {
+                if (!preg_match('/(\.tar\.gz|\.tgz)$/', $this->dumpFile)) {
                     throw new ModuleConfigException(
                         __CLASS__,
                         "Dump file must be a valid tar gunzip archive.\n


DavertMik:
Thanks. Good addition.

This line was not really nice but with this patch it will look really ugly :cry: 

```php
 if (strlen($this->dumpFile) <= 7 || (substr($this->dumpFile, -7) !== '.tar.gz' && substr($this->dumpFile, -4) !== '.tgz')) {
```
Could you think how to simplify this condition check so it would look nicer?

src/Codeception/Module/MongoDb.php
```diff
@@ -158,7 +158,7 @@ private function validateDump()
                         "Tar gunzip archives are not supported for Windows systems"
                     );
                 }
-                if (strlen($this->dumpFile) <= 7 || (substr($this->dumpFile, -7) !== '.tar.gz' && substr($this->dumpFile, -4) !== '.tgz')) {
+                if (!preg_match('/^.+(\.tar\.gz|\.tgz)$/', $this->dumpFile)) {
```
DavertMik: `^.` what about space characters?
I think regex can be simplified to just `/(\.tar\.gz|\.tgz)$/`

Lukazar: You bet.

DavertMik: Yes, sorry, space character is included

Lukazar: What you have should be fine so long you do not care about files that are named just ".tgz".  To be honest the check isn't really needed.  The system will puke if you try to untar something thats not a valid archive... I think... I didn't test it explicitly.

DavertMik: Yeah, got it, thanks. Once build finishes I'm merging

PR #4615, opened 2017-11-10
added support for dotenv version 1.1

sasmitung:
dotenv in [version 1.1](https://github.com/vlucas/phpdotenv/blob/1.1/src/Dotenv.php)
 doesn't have namespace, but dotenv 1.1 is still used in Laravel 5.1 (which is LTS with support until June 2018). 

This PR won't break backward compatibility.


Naktibalda:
Please make this change to 2.3 branch instead.

PR #4616, opened 2017-11-10
added support for dotenv version 1.1

sasmitung:
dotenv in [version 1.1](https://github.com/vlucas/phpdotenv/blob/1.1/src/Dotenv.php) doesn't have namespace, but dotenv 1.1 is still used in Laravel 5.1 (which is LTS with support until June 2018). 

This PR won't break backward compatibility.

### DIFF
--- a/src/Codeception/Lib/ParamsLoader.php
+++ b/src/Codeception/Lib/ParamsLoader.php
@@ -76,14 +76,20 @@ class ParamsLoader
 
     protected function loadDotEnvFile()
     {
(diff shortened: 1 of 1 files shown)


DavertMik:
Sorry, but no. If you would come with this PR a bit later I would probably accept it. But it looks like you are trying to push some code with already is legacy. Let's stick to new things.

There is a workaround. `\DotEnv` can be loaded inside a `_bootstrtap` by using `class_alias`:

```php
class_alias('DotEnv\DotEnv', 'DotEnv');
```

Issue #4602, opened 2017-11-03
Db-Module: MySQL SSL connection


bangertz:
I am missing the SSL option for the connection with the Db-Module. Is it planned or already possible? I rewrote the constructor in "Codeception\Lib\Driver\Db" for a test. The connection worked perfectly.

```php
public function __construct( $dsn, $user, $password )
{
    $this->dbh = new \PDO( $dsn, $user, $password, [
        \PDO::MYSQL_ATTR_SSL_KEY    => '/path/to/client-key.pem',
        \PDO::MYSQL_ATTR_SSL_CERT   => '/path/to/client-cert.pem',
        \PDO::MYSQL_ATTR_SSL_CA     => '/path/to/ca-cert.pem',
    ] );
    
    $this->dbh->setAttribute(\PDO::ATTR_ERRMODE, \PDO::ERRMODE_EXCEPTION);

    $this->dsn = $dsn;
    $this->user = $user;
    $this->password = $password;
}
```

But no parameter is provided for the additional options. I did not succeed in extending the module/driver. Has anyone a tip or an idea to solve the problem?

DavertMik:
> I am missing the SSL option for the connection with the Db-Module. Is it planned or already possible?

No. Not planned. Not possible. Yet.

Please send a pull request with the update to Db module.
`ssl_key`, `ssl_cert`, `ssl_ca` should be added as config parameters to Db.

PR #4622, opened 2017-11-13
Dbdriver ssloptions

kossi84:
Issue: Db-Module: MySQL SSL connection #4602

### DIFF
--- a/src/Codeception/Module/Db.php
+++ b/src/Codeception/Module/Db.php
@@ -274,19 +274,23 @@ class Db extends CodeceptionModule implements DbInterface
     {
         $options = [];
  
-        if( array_key_exists('ssl_key', $this->config)
-            AND !empty($this->config['ssl_key'])) {
-                $options[\PDO::MYSQL_ATTR_SSL_KEY] = $this->config['ssl_key'];
+        /**
+         * @see http://php.net/manual/en/pdo.construct.php
+         * @see http://php.net/manual/de/ref.pdo-mysql.php#pdo-mysql.constants
+         */
+        if (array_key_exists('ssl_key', $this->config) && !empty($this->config['ssl_key']))
+        {
+            $options[\PDO::MYSQL_ATTR_SSL_KEY] = $this->config['ssl_key'];
         }
  
-        if( array_key_exists('ssl_cert', $this->config)
-            AND !empty($this->config['ssl_cert'])) {
-                $options[\PDO::MYSQL_ATTR_SSL_CERT] = $this->config['ssl_cert'];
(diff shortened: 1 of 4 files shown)


src/Codeception/Module/Db.php
```diff
@@ -272,8 +272,25 @@ private function readSql()
 
     private function connect()
     {
+        $options = [];
+ 
+        if( array_key_exists('ssl_key', $this->config)
+            AND !empty($this->config['ssl_key'])) {
```
DavertMik: `AND` ?? No one writes like that 😊
Please follow PSR and replace it with `&&`.

Naktibalda: Also if statement should be formatted as `if (array_key_exists`. (space before opening parenthesis).

kossi84:
@Naktibalda can you please review again, so the changes can be pull by you? Thanks in advance.

kossi84:
@Naktibalda can you please review again, so the changes can be pulled by you? Thanks in advance.

src/Codeception/Module/Db.php
```diff
@@ -45,6 +45,9 @@
  * * populate: false - whether the the dump should be loaded before the test suite is started
  * * cleanup: false - whether the dump should be reloaded before each test
  * * reconnect: false - whether the module should reconnect to the database before each test
+ * * ssl_key - path to the SSL key
+ * * ssl_cert - path to the SSL certificate
+ * * ssl_ca - path to the SSL certificate authority
```
Naktibalda: I would like to see something like `(mysql specific option)` at the end of all three lines.

src/Codeception/Module/Db.php
```diff
@@ -58,6 +61,9 @@
  *              populate: true
  *              cleanup: true
  *              reconnect: true
+ *              ssl_key: '/path/to/client-key.pem'
+ *              ssl_cert: '/path/to/client-cert.pem'
+ *              ssl_ca: '/path/to/ca-cert.pem'
```
Naktibalda: You are using absolute paths, right?
I wonder what should we do with relative paths.
@DavertMik should the relative paths be relative to project dir or data dir?

bangertz: Now, relative paths from the project directory are working. Absolute paths should also be supported in the future.

Naktibalda: Actually they are relative to current working directory, which isn't always the same as project dir (running Codeception with `-c path` and included test suites are two cases when they don't match). But this can be changed later if necessary.

DavertMik:
Thanks!

PR #4623, opened 2017-11-14
AWS Authorization for REST requests

bnpatel1990:
Testing API using REST module in AWS Cloud required AWS authorization. 

Creating a AWS authorization wrapper in dorequest method for guzzle.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,6 +16,7 @@
 
     "require": {
         "php": ">=5.4.0 <8.0",
+        "aws/aws-sdk-php": "^3.8",
         "ext-json": "*",
         "ext-mbstring": "*",
         "phpunit/phpunit": ">4.8.20 <7.0",
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,6 @@
 
     "require": {
         "php": ">=5.4.0 <8.0",
-        "aws/aws-sdk-php": "^3.8",
         "ext-json": "*",
         "ext-mbstring": "*",
         "phpunit/phpunit": ">4.8.20 <7.0",
@@ -52,6 +51,7 @@
         "symfony/process": ">=2.7 <4.0"
     },
     "suggest": {
+        "aws/aws-sdk-php": "For using AWS Auth in REST module",
         "codeception/specify": "BDD-style code blocks",
         "codeception/verify": "BDD-style assertions",
         "flow/jsonpath": "For using JSONPath in REST module",
--- a/docs/modules/REST.md
+++ b/docs/modules/REST.md
@@ -89,19 +89,6 @@ $I->amNTLMAuthenticated('jon_snow', 'targaryen');
  * `[Part]` xml
 
 
-### amAWSAuthenticated
-Allows to send REST request using AWS Authorization
-```php
-<?php
-$I->amAWSAuthenticated('awsKey', 'awsSecret', 'awsService', 'awsRegion');
-?>
-```
- * `param` $key
- * `param` $secret
- * `param` $service
- * `param` $region
-
-
 ### deleteHeader
  
 Deletes the header with the passed name.  Subsequent requests
--- a/docs/modules/REST.md
+++ b/docs/modules/REST.md
@@ -89,6 +89,19 @@ $I->amNTLMAuthenticated('jon_snow', 'targaryen');
  * `[Part]` xml
 
 
+### amAWSAuthenticated
+Allows to send REST request using AWS Authorization
+```php
+<?php
+$I->amAWSAuthenticated('awsKey', 'awsSecret', 'awsService', 'awsRegion');
+?>
+```
+ * `param` $key
+ * `param` $secret
+ * `param` $service
+ * `param` $region
+
+
 ### deleteHeader
  
 Deletes the header with the passed name.  Subsequent requests
--- a/src/Codeception/Lib/Connector/Guzzle6.php
+++ b/src/Codeception/Lib/Connector/Guzzle6.php
@@ -355,9 +355,9 @@ class Guzzle6 extends Client
         try {
             $this->awsCredentials = new Credentials($config['key'], $config['secret']);
             $this->awsSignature = new SignatureV4($config['service'], $config['region']);
-        } catch ( CredentialsException $e) {
+        } catch (CredentialsException $e) {
             throw $e;
-        } catch ( UnresolvedSignatureException $e) {
+        } catch (UnresolvedSignatureException $e) {
             throw $e;
         }
     }
--- a/src/Codeception/Lib/Connector/Guzzle6.php
+++ b/src/Codeception/Lib/Connector/Guzzle6.php
@@ -2,8 +2,6 @@
 namespace Codeception\Lib\Connector;
 
 use Aws\Credentials\Credentials;
-use Aws\Exception\CredentialsException;
-use Aws\Exception\UnresolvedSignatureException;
 use Aws\Signature\SignatureV4;
 use Codeception\Util\Uri;
 use GuzzleHttp\Client as GuzzleClient;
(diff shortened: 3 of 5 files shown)


composer.json
```diff
@@ -16,6 +16,7 @@
 
     "require": {
         "php": ">=5.4.0 <8.0",
+        "aws/aws-sdk-php": "^3.8",
```
Naktibalda: This is a very big dependency for a very small, very specific functionality. Also it raises required version of PHP to 5.5.
Please move this dependency to `suggest` section and document it similarly to jsonpath dependency in http://codeception.com/docs/modules/REST#seeResponseJsonMatchesJsonPath

src/Codeception/Module/REST.php
```diff
@@ -333,6 +333,18 @@ public function amNTLMAuthenticated($username, $password)
...
     /**
+     * Adds AWS Authentication via key, secret, service and region
+     * @param $key
+     * @param $secret
+     * @param $service
+     * @param $region
+     */
+    public function amAWSAuthenticated($key, $secret, $service, $region)
+    {
+        $this->client->setAwsAuth($key, $secret, $service, $region);
```
Naktibalda: Please check if the setAwsAuth method exists before calling it. Guzzle isn't the only client used by REST module.

docs/modules/REST.md
```diff
@@ -89,6 +89,19 @@ $I->amNTLMAuthenticated('jon_snow', 'targaryen');
...
+<?php
+$I->amAWSAuthenticated('awsKey', 'awsSecret', 'awsService', 'awsRegion');
+?>
+```
+ * `param` $key
+ * `param` $secret
+ * `param` $service
+ * `param` $region
+
+
```
Naktibalda: This documentation file is autogenerated from REST.php, so this change will be lost.
Please move your code example to REST.php

src/Codeception/Module/REST.php
DavertMik: I don't think that `key` and especially `secret` should be visible and hardcoded in tests.
In the current implementation, you will need to pass them directly. Which is not a very secure way. 
I think this should be moved to configuration section, so this values could be loaded from environment variables or external config files.

I'd add to REST config options like:

```
aws:
    key:
    secret: 
    service:
    region:
```
any of those options could be overridden inside `amAWSAuthenticated`. So if you need to pass requests to a different region or different service you could set it there.

```php
public function amAWSAuthenticated($addiotionalAWSConfig = [])
{
    $config = array_merge($this->config['aws'], $addiotionalAWSConfig);
}
```

bnpatel1990: Haha I haD initially coded it to be passed from Config files, but then saw none of other authorizations were following that. So decided to hard code it. Will be pushing the fix in few hours.

bnpatel1990: Done

bnpatel1990: Cleaned it up

docs/modules/REST.md
bnpatel1990: Removed from Docs

composer.json
bnpatel1990: Done moved to suggested section

src/Codeception/Module/REST.php
```diff
@@ -333,15 +334,35 @@ public function amNTLMAuthenticated($username, $password)
...
+     * ?>
+     * ```
+     * @param array $additionalAWSConfig
+     * @throws ModuleException
      */
-    public function amAWSAuthenticated($key, $secret, $service, $region)
+    public function amAWSAuthenticated($additionalAWSConfig = [])
     {
-        $this->client->setAwsAuth($key, $secret, $service, $region);
+        if (!defined('\GuzzleHttp\Client::VERSION')) {
```
Naktibalda: This condition is wrong. Guzzle is a mandatory dependency - so the only case when this exception could be thrown is if Guzzle 5 or 4 is installed.
`method_exists($this->client, 'setAwsAuth')` or `$this->client instanceof \Codeception\Lib\Connector\Guzzle6` are better ways.

Also I don't like `Out of scope` in exception message.

src/Codeception/Module/REST.php
DavertMik: > Haha I haD initially coded it to be passed from Config files, but then saw none of other authorizations were following that. So decided to hard code it. Will be pushing the fix in few hours.

yeah, probably the difference here is that other Auth methods can be executed locally. In this case you don't have strong security restrictions. However, for AWS you are providing your AWS account credentials, and they should never appear in tests (and test log)

src/Codeception/Module/REST.php
bnpatel1990: Done

DavertMik:
Looks good to me

src/Codeception/Lib/Connector/Guzzle6.php
```diff
@@ -338,4 +349,16 @@ public static function createHandler($handler)
...
     }
+
+    public function setAwsAuth($config)
+    {
+        try {
+            $this->awsCredentials = new Credentials($config['key'], $config['secret']);
+            $this->awsSignature = new SignatureV4($config['service'], $config['region']);
+        } catch (CredentialsException $e) {
+            throw $e;
+        } catch (UnresolvedSignatureException $e) {
```
Naktibalda: Why do you catch these 2 exception if you do nothing with them?

src/Codeception/Module/REST.php
```diff
@@ -333,6 +334,37 @@ public function amNTLMAuthenticated($username, $password)
     }
 
     /**
+     * Allows to send REST request using AWS Authorization
+     * Requires client to by Guzzle
```
Naktibalda: It's `be`, not `by`.
Also that we are using Guzzle HTTP Client is an internal detail that some users may be unaware of.
I think that it would be better to write `Only works with PhpBrowser`.

src/Codeception/Module/REST.php
```diff
@@ -333,6 +334,37 @@ public function amNTLMAuthenticated($username, $password)
...
+     * <?php
+     * $I->amAWSAuthenticated();
+     * ?>
+     * ```
+     * @param array $additionalAWSConfig
+     * @throws ModuleException
+     */
+    public function amAWSAuthenticated($additionalAWSConfig = [])
+    {
+        if (method_exists($this->client, 'setAwsAuth')) {
```
Naktibalda: I'm not sure what is the best behaviour when client has no `setAwsAuth` method.
One way is to throw an exception, but doing nothing is probably good as well.

src/Codeception/Lib/Connector/Guzzle6.php
```diff
@@ -197,7 +204,11 @@ protected function doRequest($request)
         }
 
         try {
-            $response = $this->client->send($guzzleRequest, $options);
+            if (null !== $this->awsCredentials) {
+                $response = $this->client->send($this->awsSignature->signRequest($guzzleRequest, $this->awsCredentials), $options);
+            } else {
+                $response = $this->client->send($guzzleRequest, $options);
+            }
```
Naktibalda: I think that it would be good to copy this functionality to https://github.com/Codeception/Codeception/blob/2.3/src/Codeception/Lib/Connector/Guzzle.php
Guzzle connector is used when Guzzle 5 is installed (on PHP 5.4 with composer and on PHP 5.4-5.6 with phar file).

bnpatel1990: Copied over to Guzzle v5

src/Codeception/Lib/Connector/Guzzle6.php
bnpatel1990: so the idea was I wanted to catch if a key was not set. This was not the right location to do it. And yes just catching and throwing does not help.

src/Codeception/Module/REST.php
bnpatel1990: Updated the Docblock

DavertMik:
Thanks!

PR #4630, opened 2017-11-16
Add Iteration Count for Dataprovider Examples to Cest Test Signatures

### LINKED ISSUE #4110: HTML report shows example data set errors under wrong test run
Hi!

I would like to see correct html report of failures for a test with example data sets:

```php <?php
class WelcomeCest
{

	/**
	 * @example {"run": 1}
	 * @example {"run": 2}
	 */
	public function tryToTest(AcceptanceTester $I, \Codeception\Example $example)
	{
		$I->amOnPage('http://non-existent' . $example['run']);
	}
}
```
 

#### What do you get instead?
Error for example set 1 is shown also in example set 2 report:
html_report

### Details

```
$ ./composer.phar show
behat/gherkin                      v4.4.5 Gherkin DSL parser for PHP 5.3
codeception/codeception            2.2.10 BDD-style testing framework
doctrine/instantiator              1.0.5  A small, lightweight utility to instantiate objects in PHP without invoking their constru...
facebook/webdriver                 1.4.0  A PHP client for Selenium WebDriver
guzzlehttp/guzzle                  6.2.3  Guzzle is a PHP HTTP client library
guzzlehttp/promises                v1.3.1 Guzzle promises library
guzzlehttp/psr7                    1.4.2  PSR-7 message implementation that also provides common utility methods
phpdocumentor/reflection-common    1.0    Common reflection classes used by phpdocumentor to reflect the code structure
phpdocumentor/reflection-docblock  3.1.1  With this component, a library can provide support for annotations via DocBlocks or other...
phpdocumentor/type-resolver        0.2.1 
phpspec/prophecy                   v1.7.0 Highly opinionated mocking framework for PHP 5.3+
phpunit/php-code-coverage          2.2.4  Library that provides collection, processing, and rendering functionality for PHP code co...
phpunit/php-file-iterator          1.4.2  FilterIterator implementation that filters files based on a list of suffixes.
phpunit/php-text-template          1.2.1  Simple template engine.
phpunit/php-timer                  1.0.9  Utility class for timing
phpunit/php-token-stream           1.4.11 Wrapper around PHP's tokenizer extension.
phpunit/phpunit                    4.8.35 The PHP Unit Testing framework.
phpunit/phpunit-mock-objects       2.3.8  Mock Object library for PHPUnit
psr/http-message                   1.0.1  Common interface for HTTP messages
psr/log                            1.0.2  Common interface for logging libraries
sebastian/comparator               1.2.4  Provides the functionality to compare PHP values for equality
sebastian/diff                     1.4.1  Diff implementation
sebastian/environment              1.3.8  Provides functionality to handle HHVM/PHP environments
sebastian/exporter                 1.2.2  Provides the functionality to export PHP variables for visualization
sebastian/global-state             1.1.1  Snapshotting of global state
sebastian/recursion-context        1.0.5  Provides functionality to recursively process PHP variables
sebastian/version                  1.0.6  Library that helps with managing the version number of Git-hosted PHP projects
stecman/symfony-console-completion 0.7.0  Automatic BASH completion for Symfony Console Component based applications.
symfony/browser-kit                v3.2.7 Symfony BrowserKit Component
symfony/console                    v3.2.7 Symfony Console Component
symfony/css-selector               v3.2.7 Symfony CssSelector Component
symfony/debug                      v3.2.7 Symfony Debug Component
symfony/dom-crawler                v3.2.7 Symfony DomCrawler Component
symfony/event-dispatcher           v3.2.7 Symfony EventDispatcher Component
symfony/finder                     v3.2.7 Symfony Finder Component
symfony/polyfill-mbstring          v1.3.0 Symfony polyfill for the Mbstring extension
symfony/process                    v3.2.7 Symfony Process Component
symfony/yaml                       v3.2.7 Symfony Yaml Component
webmozart/assert                   1.2.0  Assertions to validate method input/output with nice error messages.
```

Tenzian:
## Issue
When iterating through dataprovider examples for a Cest method, the test signature remains the same for each example.

This causes any test failures picked up by `addFailure()` in `Codeception\PHPUnit\ResultPrinter\HTML` for all examples provided to the Cest method to accumulate in the same `$failures[Descriptor::getTestSignature($test)]` array.

The effect of this is the test results for each example in the generated HTML report also contain any failures from all previous examples provided by that dataprovider.

Fixes #4110

## Solution
An example iteration count is added to the test metadata in `Codeception\Test\Loader\Cest` which is then appended to the test signature in `Codeception\Test\Cest` so that each example gets its own key in the `$failures` array.

### DIFF
--- a/src/Codeception/Subscriber/Dependencies.php
+++ b/src/Codeception/Subscriber/Dependencies.php
@@ -13,11 +13,14 @@ class Dependencies implements EventSubscriberInterface
     use Shared\StaticEvents;
 
     static $events = [
-        Events::TEST_START => 'testStart',
-        Events::TEST_SUCCESS => 'testSuccess'
+        Events::TEST_START      => 'testStart',
+        Events::TEST_FAIL       => 'testUnuccessful',
+        Events::TEST_ERROR      => 'testUnuccessful',
+        Events::TEST_INCOMPLETE => 'testUnuccessful',
+        Events::TEST_SKIPPED    => 'testUnuccessful'
     ];
 
-    protected $successfulTests = [];
+    protected $unsuccessfulTests = [];
 
     public function testStart(TestEvent $event)
     {
@@ -28,20 +31,21 @@ class Dependencies implements EventSubscriberInterface
 
         $testSignatures = $test->getDependencies();
         foreach ($testSignatures as $signature) {
-            $matches = preg_grep('/' . preg_quote($signature) . '(?::.+)?/', $this->successfulTests);
-            if (empty($matches)) {
-                $test->getMetadata()->setSkip("This test depends on $signature to pass");
+            $matches = preg_grep('/' . preg_quote($signature) . '(?::.+)?/', $this->unsuccessfulTests);
+            if (!empty($matches)) {
+                $failures = implode(', ', array_unique($matches));
+                $test->getMetadata()->setSkip("This test depends on $failures to pass");
                 return;
             }
         }
     }
 
-    public function testSuccess(TestEvent $event)
+    public function testUnuccessful(TestEvent $event)
     {
         $test = $event->getTest();
         if (!$test instanceof TestInterface) {
             return;
         }
-        $this->successfulTests[] = Descriptor::getTestSignature($test);
+        $this->unsuccessfulTests[] = Descriptor::getTestSignature($test);
     }
 }
--- a/src/Codeception/Subscriber/Dependencies.php
+++ b/src/Codeception/Subscriber/Dependencies.php
@@ -28,7 +28,7 @@ class Dependencies implements EventSubscriberInterface
 
         $testSignatures = $test->getDependencies();
         foreach ($testSignatures as $signature) {
-            $matches = preg_grep('/'.addslashes($signature).'(?::\d+)?/', $this->successfulTests);
+            $matches = preg_grep('/' . preg_quote($signature) . '(?::\d+)?/', $this->successfulTests);
             if (empty($matches)) {
                 $test->getMetadata()->setSkip("This test depends on $signature to pass");
                 return;
--- a/src/Codeception/Subscriber/Dependencies.php
+++ b/src/Codeception/Subscriber/Dependencies.php
@@ -28,7 +28,8 @@ class Dependencies implements EventSubscriberInterface
 
         $testSignatures = $test->getDependencies();
         foreach ($testSignatures as $signature) {
-            if (!in_array($signature, $this->successfulTests)) {
+            $matches = preg_grep('/'.addslashes($signature).'(?::\d+)?/', $this->successfulTests);
+            if (empty($matches)) {
                 $test->getMetadata()->setSkip("This test depends on $signature to pass");
                 return;
             }
--- a/src/Codeception/Subscriber/Dependencies.php
+++ b/src/Codeception/Subscriber/Dependencies.php
@@ -14,12 +14,14 @@ class Dependencies implements EventSubscriberInterface
 
     static $events = [
         Events::TEST_START      => 'testStart',
-        Events::TEST_FAIL       => 'testUnuccessful',
-        Events::TEST_ERROR      => 'testUnuccessful',
-        Events::TEST_INCOMPLETE => 'testUnuccessful',
-        Events::TEST_SKIPPED    => 'testUnuccessful'
+        Events::TEST_SUCCESS    => 'testSuccessful',
+        Events::TEST_FAIL       => 'testUnsuccessful',
+        Events::TEST_ERROR      => 'testUnsuccessful',
+        Events::TEST_INCOMPLETE => 'testUnsuccessful',
+        Events::TEST_SKIPPED    => 'testUnsuccessful'
     ];
 
+    protected $successfulTests = [];
     protected $unsuccessfulTests = [];
 
     public function testStart(TestEvent $event)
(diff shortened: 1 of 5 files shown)


Tenzian:
Should address #4110

src/Codeception/Test/Cest.php
```diff
@@ -155,7 +155,11 @@ public function toString()
 
     public function getSignature()
     {
-        return get_class($this->getTestClass()) . ":" . $this->getTestMethod();
+        $signature = get_class($this->getTestClass()) . ":" . $this->getTestMethod();
+        if (!empty($this->getMetadata()->getCurrent('iteration'))) {
```
Naktibalda: `!empty(expr)` does not work on PHP 5.4,  please assign the result to temporary variable.

src/Codeception/Test/Loader/Cest.php
```diff
@@ -87,6 +88,7 @@ function ($v) {
                         }
                         $test = new CestFormat($unit, $method, $file);
                         $test->getMetadata()->setCurrent(['example' => $example]);
+                        $test->getMetadata()->setCurrent(['iteration' => ++$iteration]);
```
Naktibalda: There is no need to introduce a new variable, $k should be fit for this purpose.

tests/data/claypit/tests/scenario/DataProviderCest.php
```diff
@@ -47,7 +47,9 @@ public function testDependsWithDataProvider(ScenarioGuy $I, Example $example)
 
        /**
         * @group dataprovider
-        * @depends DataProviderCest:testDependsWithDataProvider
+        * @depends DataProviderCest:testDependsWithDataProvider:1
```
Naktibalda: You fixed our test, but your change will break tests for everyone else who are using test dependencies with data providers.
As such it can't be merged to 2.3 branch and I am not sure if it could be merged to master.

A better solution would be to make dependency checker ignore iteration suffix (I don't know where it is, it could be in PhpUnit codebase).
Another alternative would be to fix the original issue without making modifications to test signature.

Tenzian: Yeah I realised that as soon as I committed it and am looking for the dependency checker at the moment. Any pointers on where to look?

Naktibalda:
Is this issue specific to Cest format or does it affect Unit format too?

tests/data/claypit/tests/scenario/DataProviderCest.php
Naktibalda: https://github.com/Codeception/Codeception/blob/b54eaf4007484f36145c1dc8c64da1874adbc340/src/Codeception/Subscriber/Dependencies.php#L32

Naktibalda:
Could you add a test to RunCest to demonstrate the issue that you are trying to fix?

Tenzian:
I've gone with using `preg_grep` in Subscriber\Dependences.

It'll allow both `@depends DataProviderCest:testDependsWithDataProvider` to match all iterations and `@depends DataProviderCest:testDependsWithDataProvider:1` to match just one.

src/Codeception/Subscriber/Dependencies.php
```diff
@@ -28,7 +28,8 @@ public function testStart(TestEvent $event)
 
         $testSignatures = $test->getDependencies();
         foreach ($testSignatures as $signature) {
-            if (!in_array($signature, $this->successfulTests)) {
+            $matches = preg_grep('/'.addslashes($signature).'(?::\d+)?/', $this->successfulTests);
```
Naktibalda: Please use preg_quote instead of addslashes and put spaces on both sides of concatenation operator.
Also I think that there is no guarantee that iteration is numeric,
dataProvider function can return named examples.
```php
return [
  'a' => ['param1', 'param2'],
  'b' => ['param1', 'param2'],
];
```

Tenzian:
The issue doesn't actually affect the running or pass/fail conditions of the tests themselves.
 
The problem is how the failures  are recorded in the HTML report:
https://github.com/Codeception/Codeception/blob/835e05c968d0c8f05d39f7cd5fba6bdf3e95cb22/src/Codeception/PHPUnit/ResultPrinter/HTML.php#L249-L253

Without the iteration suffix all failures for methods using examples are collated under the same signature and the HTML report doesn't makes sense: each subsequent fail has the data from all previous ones.

There's a screenshot in issue #4110

Tenzian:
Just noticed this introduces a logic error in the way that dependencies are deemed to have been met.

Instead of `@depends DataProviderCest:testDependsWithDataProvider` meaning "ALL iterations were successful" it now means "ANY iteration was successful`

Will fix.

Tenzian:
- Specifying a `@depends` without a iteration suffix now fails unless all iterations pass.
- Adding a suffix to the `@depends` for an iteration that passes meets the dependency even if other iterations fail.
- Tests added to confirm.

Tenzian:
Strange... seems the automated tests count a larger number of assertions than I get locally. 🤷‍♂️ 
Removed assertions from `$I->seeInShellOutput()`.

Naktibalda:
This is getting worse, please change it back.
How are you going to handle the case when \@depends refers to non-existing test?

Tenzian:
Yikes. Sorry. That was negligent. 😅 

Is de16789 better or would you prefer to reset to fe7ded9?

I think having `@depends testWithDataProvider` mean "ALL iterations were successful" is preferable.
I've added a test to make sure `@depends` for a test that doesn't exist causes a skip.

Naktibalda:
A better way to fix original issue would be to clear failures in HTML Result Printer between test runs.

I think that settting `$this->failures[$name] = [];` somewhere around here:
https://github.com/Codeception/Codeception/blob/835e05c968d0c8f05d39f7cd5fba6bdf3e95cb22/src/Codeception/PHPUnit/ResultPrinter/HTML.php#L136

should fix it.

Tenzian:
Yes. That does provide a fix fro the original issue.

It doesn't give you any additional benefit of being able to treat each iteration as a discrete test, though this is probably a bit over-complex now.

Would you prefer this one closed?

Naktibalda:
Yes, lets keep it simple.

Issue #4636, opened 2017-11-17
AmazonSQS driver needs version configuration option


gitis:
#### What are you trying to achieve?
Tried to set-up amazon SQS for the Queue module and localstack.

#### What do you get instead?
```bash
 [InvalidArgumentException] Missing required client configuration options:

version: (string)

  A "version" configuration value is required. Specifying a version constraint
  ensures that your code will not be affected by a breaking change made to the
  service. For example, when using Amazon S3, you can lock your API version to
  "2006-03-01".

  Your build of the SDK has the following version(s) of "sqs": * "2012-11-05"

  You may provide "latest" to the "version" configuration value to utilize the
  most recent available API version that your client's API provider can find.
  Note: Using 'latest' in a production application is not recommended.

  A list of available API versions can be found on each client's API documentation
  page: http://docs.aws.amazon.com/aws-sdk-php/v3/api/index.html. If you are
  unable to load a specific API version, then you may need to update your copy of
  the SDK.

#1  /home/docker/ebetting/vendor/aws/aws-sdk-php/src/ClientResolver.php:374
#2  /home/docker/ebetting/vendor/aws/aws-sdk-php/src/ClientResolver.php:270
#3  /home/docker/ebetting/vendor/aws/aws-sdk-php/src/AwsClient.php:158
#4  /home/docker/ebetting/vendor/aws/aws-sdk-php/src/Sqs/SqsClient.php:59
#5  /home/docker/ebetting/vendor/aws/aws-sdk-php/src/AwsClient.php:326
#6  Codeception\Subscriber\Module->before
#7  /home/docker/ebetting/vendor/symfony/symfony/src/Symfony/Component/EventDispatcher/EventDispatcher.php:212
#8  /home/docker/ebetting/vendor/symfony/symfony/src/Symfony/Component/EventDispatcher/EventDispatcher.php:44

ERRORS!
Tests: 1, Assertions: 0, Errors: 1.
```

### Details

* Codeception version: 2.3.6
* PHP Version: 7.1.10
* Operating System: Debian GNU/Linux 8.9 (jessie)
* Installation type: Composer
* List of related installed packages
```yml
aws/aws-sdk-php                             3.38.4             AWS SDK for PHP - Use Amazon Web Services in your PHP project
codeception/codeception                     2.3.6              BDD-style testing framework
```
* Suite configuration:

```yml
actor: FunctionalTester
modules:
    enabled:
        - Symfony:
            app_path: 'app'
[...]

PR #4637, opened 2017-11-17
Fixed AmazonSQS client init configuration. Added version. #4636

gitis:
https://github.com/Codeception/Codeception/issues/4636

### DIFF
--- a/docs/modules/Queue.md
+++ b/docs/modules/Queue.md
@@ -41,6 +41,7 @@ service.
 * token - Iron.io access token.
 * project - Iron.io project ID.
 * key - AWS access key ID.
+* version - AWS version (e.g. latest)
 * secret - AWS secret access key.
      Warning:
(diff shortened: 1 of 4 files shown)


docs/modules/Queue.md
```diff
@@ -78,40 +79,41 @@ service.
        enabled: [Queue]
        config:
           Queue:
-             'type' => 'iron',
-             'host' => 'mq-aws-us-east-1.iron.io',
-             'token' => 'your-token',
-             'project' => 'your-project-id'
+             'type': 'iron',
+             'host': 'mq-aws-us-east-1.iron.io',
```
DavertMik: Oh, just one thing. This file is generated and will be overridden. 
Please update the corresponding docblock in `Queue` module so the documentation was updated.

gitis: Thanks! Updated.

DavertMik:
Thanks!

Issue #4645, opened 2017-11-24
[Queue] Configuration parameter endpoint is missing in AmazonSQS lib


gitis:
#### What are you trying to achieve?
Trying to pass custom endpoint to AmazonSQS lib in order to be able to assert queue using [localstack](https://github.com/localstack/localstack)
#### What do you get instead?
There is no such option.

Working configuration.
```yml
actor: FunctionalTester
modules:
    enabled:
        - Queue:
            type: 'aws_sqs'
            key: 'foobar'
            secret: 'foobar'
            region: 'eu-central-1'
            version: 'latest'
```

Needed configuration
```yml
actor: FunctionalTester
modules:
    enabled:
        - Queue:
            type: 'aws_sqs'
            key: 'foobar'
            secret: 'foobar'
            region: 'eu-central-1'
            version: 'latest'
            endpoint: 'http://localstack:4576'
```

gitis:
Will open a PR later today.

PR #4648, opened 2017-11-27
[Queue] Configuration parameter endpoint is missing in AmazonSQS

gitis:
issue #4645


gitis:
@DavertMik do you have any idea why is the Travis CI build failing?

PR #4659, opened 2017-12-01
Fix code snippets

josephzidell:
Make subtle corrections to code snippets


DavertMik:
Thanks @josephzidell but could you patch this file instead?
 https://github.com/Codeception/Codeception/blob/2.3/src/Codeception/Util/Stub.php
Docs are auto generated from it

josephzidell:
@DavertMik @Naktibalda yes

josephzidell:
Opened #4670

PR #4671, opened 2017-12-08
Refactoring tests

carusogabriel:
I've refactored some tests, using:
- `assertCount` instead of `count` function;
- `assertObjectNotHasAttribute`, `assertObjectHasAttribute` and `assertArrayHasKey` instead of `isset` function;
- `assertContains` instead of `array_search` function.


DavertMik:
Wow, thanks for better tests! :+1:

PR #4680, opened 2017-12-12
[Db] Database in DSN regex matching fix

yesdevnull:
If your DSN for MS SQL (sqlsrv) specifies the `Database` param as the last option then the regex for the Database name find it due to the semicolon not being optional.


DavertMik:
Thanks!

Issue #4696, opened 2017-12-15
PostgreSQL driver should not be hard-coded to public schema


Watkinsong:
#### What are you trying to achieve?
In src/Codeception/Lib/Driver/PostgreSql.php, the schema is hard-coded to "public" in method cleanup. In my setup, where the schema used for testing is not public, this causes tests to fail with a ModuleException:
` [ModuleException] Db: SQLSTATE[42501]: Insufficient privilege: 7 ERROR:  must be owner of schema public`
The schema is taken into account in "load" method, and this should also be brought into cleaning up the test database.

#### What do you get instead?

> Provide console output if related. Use `-vvv` mode for more details.

```bash
# paste output here
```
> Provide test source code if related

```php
// paste test
```
### Details

* Codeception version: 2.26
* PHP Version: 7.0.22
* Operating System: Ubuntu
* Installation type: Phar
* List of installed packages (`composer show`)

Naktibalda:
Can you fix it?

PR #4697, opened 2017-12-15
Method cleanup considers the schema.

Watkinsong:
#4696


DavertMik:
Unfortunately PostgreSQL tests has failed. See CI logs and try to reproduce them locally to fix it

PR #4700, opened 2017-12-17
class level event should be consistent between different test cases

lubobill1990:
Refer: https://github.com/Codeception/Codeception/pull/4621

### DIFF
--- a/src/Codeception/Lib/Connector/Yii2.php
+++ b/src/Codeception/Lib/Connector/Yii2.php
@@ -83,9 +83,6 @@ class Yii2 extends Client
         static::$db = null;
         static::$mailer = null;
         \yii\web\UploadedFile::reset();
-        if (method_exists(\yii\base\Event::className(), 'offAll')) {
-            \yii\base\Event::offAll();
-        }
     }
 
     /**


glendemon:
I have same problem, resetPersistentVars works bad, need other implementation, may be some need some config for that.

DavertMik:
@samdark looks like we are walking in circles here. I don't understand why some Yii patches are replacing the previous patches and what solution should be consistent.

glendemon:
We have some business logic on events system, old patch breaks down this logic, when it delete old events, which are added in boostrap file. I think previous patch was mistake.

lubobill1990:
@samdark 
I have a question, where can we subscribe class-level event? According to this issue, should all class-level events be subscribed in config.php? Or somewhere like that where the app is initialized?
```
'on beforeRequest' => function () {
        \yii\base\Event::on(\yii\base\Controller::class, \yii\base\Controller::EVENT_BEFORE_ACTION, function () {})
```

It's better to add a paragraph to the document http://www.yiiframework.com/doc-2.0/guide-concept-events.html at Class-Level Event section to advise developers where to put Event::on() code.

PR #4704, opened 2017-12-19
Clean elses

carusogabriel:
I've cleaned the `else`s when we have already returned something :put_litter_in_its_place:


DavertMik:
Thanks, I really like to do that as well :+1:

PR #4710, opened 2017-12-23
[Laravel5] Fix memory leak

guidocella:
See  laravel/framework#21015

Running 369 functional tests
Before: 1014 MB used
After: 708 MB

### DIFF
--- a/src/Codeception/Module/Laravel5.php
+++ b/src/Codeception/Module/Laravel5.php
@@ -217,6 +217,8 @@ class Laravel5 extends Framework implements ActiveRecord, PartedModule
                 }
             }
         }
+
+        $this->app->flush();
     }
 
     /**


guidocella:
Actually never mind, I found a test that breaks with this change.
If you call `$I->amOnroute` with a model as the second argument which has a variable appended to its array form whose getter fetches a related model, `Step::stringifyArgument()` causes the error `Method App\MyModel::__toString() must not throw an exception, caught ReflectionException: Class config does not exist`, even though you can see that the related model is fetched correctly with `dd`.
It only happens if I run that test through PHPStorm but not through the CLI. Really weird. I guess I'll just apply this fix locally, since the error is easily circumvented by passing the model's id to `amOnRoute`.

PR #4712, opened 2017-12-26
Updated paths section in Configuration docs

Naktibalda:
Renamed logs directory to output because it was renamed in 2.3.0


DavertMik:
Thanks!

PR #4713, opened 2017-12-26
[DataFactory] Removed redundant dependency

### LINKED ISSUE #4646: Add DataFactory Module support for factory-muffin-faker: ^2.0 or drop it requirements
https://github.com/Codeception/Codeception/blob/2.3/src/Codeception/Module/DataFactory.php#L154
As fact "league/factory-muffin-faker" - is simple wrapper for "frazinotto/faker" and it versions strictly depends on faker versions.  And also league/factory-muffin can be used without factory-muffin-faker, so is not neccessary to check. The contents of this module does not depend on this library.

Naktibalda:
This is a clean pull request.

It replaces #4647
and fixes #4646


DavertMik:
Thanks!

PR #4718, opened 2017-12-30
Compatibility with phpunit-mock-objects 5.*

### LINKED ISSUE #4715: PHPUnit 6.5 support

Utilize latest PHPUnit version in my project together with Codeception.

#### What do you get instead?

I cannot install latest PHPUnit version because it requires `phpunit/phpunit-mock-objects: ^5.0.4` when Codeception supports only `<5.0`.

### Details

* Codeception version: v2.3.6
* PHP Version: 7.2
* Operating System: Ubuntu 16.0.4
* Installation type: Composer
* Suite configuration: (not relevant for this case)

Naktibalda:
Add class aliases used by Codeception\Util\Stub

Allows to install PHPUnit 6.5
Fixes #4715

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "ext-mbstring": "*",
         "phpunit/phpunit": ">=4.8.28 <5.0.0 || >=5.6.3 <7.0",
         "phpunit/php-code-coverage": ">=2.2.4 <6.0",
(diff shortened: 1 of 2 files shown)


DavertMik:
Thank you!

PR #4681, opened 2017-12-12
CI: added Symfony 4.0.0 environment and symfony-demo app included

DavertMik:

PR #4721, opened 2017-12-31
Removed 2 environments from travis build matrix

Naktibalda:
Because they were redundant


DavertMik:
This conflicts with #4681

DavertMik:
Ok, I resolved the conflict. I think PHP 7.2 and Symfony 3.x can be safely removed

PR #4722, opened 2018-01-01
Fallback to Symfony Dotenv to parse .env

ebuildy:
Symfony have its own Dotenv component (not sure a good idea but ...), so this PR will use Symfony Dotenv if Dotenv is not found.

https://symfony.com/doc/current/components/dotenv.html


DavertMik:
Lol they have pretty the same API :)
Thanks for this patch

PR #4723, opened 2018-01-02
[BUGFIX] Adjust group names from p* to paracept_* 

egobude:
Based on the examples in the documentation i adjusted the group names from p* to paracept_*. Otherwise the examples wont work.


DavertMik:
Thanks. Looks like you are the first who noticed this inconsistency. :+1:

PR #4728, opened 2018-01-04
Fix typo

ptrkcsk:


ptrkcsk:
Sorry for missing that section in the contributing guide.

Do I need to run Robo or is that taken care of automatically?

Naktibalda:
Robo script is executed as a part of release process, you don't need to run it.

PR #4731, opened 2018-01-06
fix(Lumen): Undefined property: Codeception\Module\Lumen::$module

ppetpadriew:
# Problem
Got this message when bootstrap file can't be found. The message does not point to the direct problem.


ppetpadriew:
Not sure why travis is failing. Seems like the failure does not relate to my fix

DavertMik:
Yes, sure, the build failure was unrelated. Thanks for this patch!

PR #4732, opened 2018-01-08
Add Doctrine2 integration to ZendExpressive module

artmnv:
Previous PR #4464 was rejected because of missing tests. I've made two tests with single and multiple POST requests that creates entities in database. Now it is in this repository: https://github.com/artmnv/codeception-zend-expressive-tests/tree/doctrine-tests
I don't send PR in source repository because there is no 2.3 branch.

Feel free to ask additional help.


artmnv:
@Naktibalda I'll do my best. Is this [test](https://github.com/artmnv/codeception-zend-expressive-tests/blob/doctrine-tests/tests/doctrine/createTwoUsersAndGetListCept.php) covers your fears that it is not working with multiple requests in one test?

Naktibalda:
Yes, I think it does.

DavertMik:
Thank you!

PR #4735, opened 2018-01-09
Fix: fail to run single test with include

### LINKED ISSUE #4733: Can't run single test from the main project when "include" is present in codeception config
Hi,

I have the main project with codeception tests, but besides that, I have some packages inside vendor dir with their own tests too. I'd like to run some tests from the main project by specifying their path.
When I run all tests, everything works. Tests from main projects are run before those from the vendor dir. But when I just want to run a single test from the main project I'm getting an error below:
```
 Suite 'tests/unit/ExampleUnitTest.php' could not be found 
``` 
Here is the exact command:
```
/var/www$ ./vendor/bin/codecept run tests/unit/ExampleUnitTest.php 
```
To fix this issues I need to remove include section from my codeception config.
I tried to debug why this happening and I found that when I have "includes" in my cfg it doesn't run this part https://github.com/Codeception/Codeception/blob/2.3/src/Codeception/Command/Run.php#L325

### Details

* Codeception version: 2.3.7
* PHP Version: 7.1
* Suite configuration:

```yml
paths:
    tests: tests
    output: tests/_output
    data: tests/_data
    support: tests/_support
    envs: tests/_envs
actor_suffix: Tester
coverage:
    enabled: true
    remote: false
    include:
        - app/Http/*
        - app/User.php
    exclude:
        - app/Http/Kernel.php
#include:
#    - vendor/xxx/yyyy  # it works when I comment those out
#    - vendor/xxx/zzz    
settings:
    bootstrap: _bootstrap.php
    lint: true
    shuffle: true
    colors: true
    memory_limit: 1024M
extensions:
    enabled:

```

ppetpadriew:
closes #4733

### DIFF
--- a/tests/cli/RunSingleTestWithIncludeCest.php
+++ b/tests/cli/RunSingleTestWithIncludeCest.php
@@ -6,11 +6,11 @@ class RunSingleTestWithIncludeCest
     {
         $I->amInPath('tests/data/single_test_with_include');
         $I->wantTo('execute one test with include in config');
-        
+
         $I->executeCommand('run unit/ExampleCest.php');
 
         $I->seeResultCodeIs(0);
-        +$I->dontSeeInShellOutput('RuntimeException');
-        +$I->dontSeeInShellOutput('could not be found');
+        $I->dontSeeInShellOutput('RuntimeException');
+        $I->dontSeeInShellOutput('could not be found');
     }
 }
--- a/tests/cli/RunSingleTestWithIncludeCest.php
+++ b/tests/cli/RunSingleTestWithIncludeCest.php
@@ -1,0 +1,16 @@
+<?php
+
+class RunSingleTestWithIncludeCest
+{
+    public function run(\CliGuy $I)
+    {
+        $I->amInPath('tests/data/single_test_with_include');
+        $I->wantTo('execute one test with include in config');
+        
+        $I->executeCommand('run unit/ExampleCest.php');
+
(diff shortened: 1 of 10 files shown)


ppetpadriew:
Can someone rebuild the Appveyor ? It failed when it was trying to download dependencies. I can't even see if the tests run through.

tests/cli/RunSingleTestWithIncludeCest.php
```diff
@@ -0,0 +1,16 @@
...
+{
+    public function run(\CliGuy $I)
+    {
+        $I->amInPath('tests/data/single_test_with_include');
+        $I->wantTo('execute one test with include in config');
+        
+        $I->executeCommand('run unit/ExampleCest.php');
+
+        $I->seeResultCodeIs(0);
+        +$I->dontSeeInShellOutput('RuntimeException');
```
DavertMik: Why do you have `+` here?

ppetpadriew: Oops. Typo.

ppetpadriew:
Not sure why this time **semaphoreci**  failed. I just removed `+` char from my test code and didn't touch the code under test. It was pass with the same code. Btw it passed on my local machine.
image

ppetpadriew:
I rebuilt each CI by amend the commit message and then they all pass. Should be ok now.

DavertMik:
Thanks! It's ok! :+1:

PR #4737, opened 2018-01-10
Refactored Codeception\Util\Stub to use standalone package

DavertMik:
Codeception\Util\Stub moved to the standalone package.


composer.json
```diff
@@ -33,7 +33,8 @@
         "symfony/dom-crawler": ">=2.7 <5.0",
         "behat/gherkin": "~4.4.0",
         "sebastian/comparator": ">1.1 <3.0",
-        "sebastian/diff": ">=1.4 <3.0"
+        "sebastian/diff": ">=1.4 <3.0",
```
Naktibalda: I think that `"phpunit/phpunit-mock-objects": ">2.3 <6.0",` can be removed, because it actually is a dependency of Stub.

DavertMik:
Ready to merge

Naktibalda:
Please remove MockObject class aliases from shim.php

PR #4745, opened 2018-01-12
Please set getResponseStatusCode as public

FanchTheSystem:
Because I have in some of my Helper the same code as in this method witch I can't call as it is protected

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -1244,7 +1244,12 @@ class InnerBrowser extends Module implements Web, PageSourceSaver, ElementLocato
         $this->debugSection('Response Headers', $this->getRunningClient()->getInternalResponse()->getHeaders());
     }
 
-    public function getResponseStatusCode()
+    public function _getResponseStatusCode()
+    {
+        return $this->getResponseStatusCode();
+    }
+
+    protected function getResponseStatusCode()
     {
         // depending on Symfony version
         $response = $this->getRunningClient()->getInternalResponse();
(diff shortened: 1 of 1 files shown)


FanchTheSystem:
Example method I already have in my helper:  

``` php
public function getStatusCode()
    {
        $response = $this->getModule('PhpBrowser')->client->getInternalResponse();
        if (method_exists($response, 'getStatus')) {
            return $response->getStatus();
        }
        if (method_exists($response, 'getStatusCode')) {
            return $response->getStatusCode();
        }
        return null;
    }
```


Example method I would like to have in my helper:
``` php
public function getStatusCode()
    {
        return = $this->getModule('PhpBrowser')->getResponseStatusCode();
    }
```

Naktibalda:
Internal response of PhpBrowser always returns an object with getStatus method.
You only need to use getStatusCode with Symfony module (probably Laravel, Silex and Lumen too).

https://github.com/symfony/browser-kit/blob/4.0/Response.php#L87
https://github.com/symfony/http-foundation/blob/4.0/Response.php#L475

src/Codeception/Lib/InnerBrowser.php
```diff
@@ -1244,7 +1244,7 @@ protected function debugResponse($url)
         $this->debugSection('Response Headers', $this->getRunningClient()->getInternalResponse()->getHeaders());
     }
 
-    protected function getResponseStatusCode()
+    public function getResponseStatusCode()
```
DavertMik: In this case method name should be prefixed with `_` so it won't go to a tester class and would be accessible only from helpers.

FanchTheSystem: Like https://github.com/Codeception/Codeception/pull/4745/commits/8f4a04970dbcd46bd1a8d99d4197deea57b5302a ?

DavertMik:
Thanks!

PR #4748, opened 2018-01-12
add forceLoadFixtures functional

bscheshirwork:
In some case we need insert data from fixture to database udnependence of check `empty($this->loadedFixtures)`. 
For example for second test with same fixtures. + change database from another Codeception module (Db cleaning) 
without this PR option the db data will be lost but fixtures data still alive


samdark:
OK. Looks valid.

DavertMik:
Sorry, I'm not Yii developer so I can be wrong but this looks like a pretty rare case to me. 
This introduces additional complexity and a new parameter which may not be used so widely. 
I'd consider to use a standalone helper instead of patching the module itself. 
But if you feel that this might be useful for wide variety of Yii developers I can merge it

bscheshirwork:
@DavertMik 
> use a standalone helper instead of patching the module

without path of module - same situation at #4745

https://github.com/bscheshirwork/Codeception/blob/e6e0985c06ad7fa36c409fb68a764366117ffdcd/src/Codeception/Module/Yii2.php#L208
is a "private". I think no way modify it in helper easily (without reflection &co)

force load can be use for any out-of-yii2-module-change-db situations.

bscheshirwork:
Also IMHO is a good idea - make this method "public"

DavertMik:
> Also IMHO is a good idea - make this method "public"

Ok, I will do that

bscheshirwork:
I can see so many changes in `loadFixtures`
fresh on 2.4 (ambiguous and non-simple-redefined)
```php
    /**
     * load fixtures before db transaction
     *
     * @param mixed $test instance of test class
     */
    private function loadFixtures($test)
    {
        $this->debugSection('Fixtures', 'Loading fixtures');
        /** @var Connection[] $connections */
        $connections = [];
        // Register event handler.
        Event::on(Connection::class, Connection::EVENT_AFTER_OPEN, function (Event $event) use (&$connections) {
            $this->debugSection('Fixtures', 'Opened database connection: ' . $event->sender->dsn);
            $connections[] = $event->sender;
        });
        if (empty($this->loadedFixtures)
            && method_exists($test, $this->_getConfig('fixturesMethod'))
        ) {
            $this->haveFixtures(call_user_func([$test, $this->_getConfig('fixturesMethod')]));
        }

        Event::offAll();
        // Close all connections so they get properly reopened after the transaction handler has been attached.
        foreach ($connections as $connection) {
            $this->debugSection('Fixtures', 'Closing database connection: ' . $connection->dsn);
            $connection->close();
        }
        $this->debugSection('Fixtures', 'Done');
    }
```

old (tiny and cute)
```php
    private function loadFixtures($test)
    {
        if (empty($this->loadedFixtures)
            && method_exists($test, self::TEST_FIXTURES_METHOD)
        ) {
            $this->haveFixtures(call_user_func([$test, self::TEST_FIXTURES_METHOD]));
        }
    }
```

so... I strongly against to idea "make this method is public" now (after 2.4 changes)

And `yii2` module now have many news options. We can add another one without any problem in design...

I try to merge 2.4 into this branch...

bscheshirwork:
So... I want to merge this into 2.4 now

PR #4749, opened 2018-01-12
added PHP 7.2 support, xdebug added as 2.6.0beta1

### LINKED ISSUE #4742: Update Docker to use PHP 7.2

Use the official Docker image to test code written with PHP 7.2

https://hub.docker.com/r/codeception/codeception/

#### What do you get instead?

various errors, e.g:

`  [ParseError] syntax error, unexpected '?'` (Docker image uses PHP 7.0, which does not support nullable return types)

#### Notes
* PHP 7.2 does not yet have support from XDebug, so if the Docker image for Codeception requires XDebug, then PHP 7.1 is the latest version which the Docker image can be updated to.
* XDebug will support PHP 7.2 in it's upcoming 2.6 release - https://xdebug.org
* In any case, PHP 7.0 (which the Docker image currently uses) is no longer "actively supported" by the PHP project: http://php.net/supported-versions.php

schmunk42:
fixes #4742 - but xdebug is only available as a beta version for now(!)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM php:7.0-cli
+FROM php:7.2-cli
 
 MAINTAINER Tobias Munk tobias@diemeisterei.de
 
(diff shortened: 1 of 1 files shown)


DavertMik:
Thanks!

PR #4767, opened 2018-01-19
Laravel5 multi databases

### LINKED ISSUE #4761: Laravel5 module - testing cross multiple database connections
Trying to make test cross database connections using the Laravel5 module. It only starts transactions for the default connection ATM.

#### What do you get instead?
Did not want to copy our massive non-trivial output - so will show it with a quick fx. 

Testing on a non-default database connect is not rolled back

```php
class Product extend Illuminate\Database\Eloquent\Model {
       $connection = 'default'
}

class Payment extend Illuminate\Database\Eloquent\Model {
       $connection = 'payment'
}

class Test {
    public function testProduct(){
          factory(Product::class)->create();
          $this->tester->seeNumRecords(1, Product::class);
    }

    public function testProductWorksWithTransactions(){
          factory(Product::class)->create();
          $this->tester->seeNumRecords(1, Product::class);
    }

    public function testPayment(){
          factory(Payment::class)->create();
          $this->tester->seeNumRecords(1, Payment::class);
    }

    public function testPaymentFailesSeesToRecords(){
          factory(Payment::class)->create();
          $this->tester->seeNumRecords(1, Payment::class);
    }
}
```

### Details

* Codeception version: 2.3.7
* PHP Version: 7.0
* Operating System: Linux
* Installation type: Composer
* Suite configuration:

```yml
actor: GraphQLTester
modules:
    enabled:
        - \App\Administration\TestExamples\Tests\Helper\GraphQL
        - Asserts
        - Laravel5:
            root: '../../../'
```

#### Suggestions
I have thought of a solution :
Make a new configuration variable 'database_transactions' which specific what database connections you are testing. The configuration could also bee additonal database. But it makes it more flexible if you have the explicitly options.
So if no 'database_transactions' just start the default database connection

```yml
actor: GraphQLTester
modules:
    enabled:
        - \App\Administration\TestExamples\Tests\Helper\GraphQL
        - Asserts
        - Laravel5:
            root: '../../../'
            database_transactions: 
                - default
                - payment
```

sazo:
fixes #4761

### DIFF
--- a/docs/modules/Laravel5.md
+++ b/docs/modules/Laravel5.md
@@ -15,6 +15,7 @@ See the Acceptance tests section below for more details.
 * database_migrations_path: `string`, default `null` - path to the database migrations, relative to the root of the application.
 * run_database_seeder: `boolean`, default `false` - run database seeder before each test.
 * database_seeder_class: `string`, default `` - database seeder class name.
+* databases: `string`, default `` - database connection(s) used in test fx "default,testing".
 * environment_file: `string`, default `.env` - the environment file to load for the tests.
 * bootstrap: `string`, default `bootstrap/app.php` - relative path to app.php config file.
 * root: `string`, default `` - root path of the application.
--- a/src/Codeception/Module/Laravel5.php
+++ b/src/Codeception/Module/Laravel5.php
@@ -127,6 +127,7 @@ class Laravel5 extends Framework implements ActiveRecord, PartedModule
                 'database_migrations_path' => null,
                 'run_database_seeder' => false,
                 'database_seeder_class' => '',
+                'databases' => '',
                 'environment_file' => '.env',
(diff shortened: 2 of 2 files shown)


DavertMik:
The PR looks good from the technical side. But I'm not sure the majority of developers would benefit from it.
There are not so much Laravel applications with multiple databases. Adding support for them adds a complexity to code and I'm not sure we'd be able to support it.

The solution looks pretty good for multi-tenant apps but it should not be coupled to Laravel5 module.
I'd suggest to create MultiTenantLaravel module and publish it as an Addon http://codeception.com/addons

You can extend the current module or connect to it using `$this->getModule()`, anyway, there are lots of options how to extend current functionality without touching the core. 

So for presonal use, I'd recommend to put that in Helper, and if you see this as valuable solution, publish it as a package and suggest to add this to addons section. 

Thanks

Issue #4765, opened 2018-01-18
coverage problem:  [InvalidArgumentException] Cookie domain "nginx:8081" should not contain a port


bscheshirwork:
#### What are you trying to achieve?
codecoverage still work

#### What do you get instead?

```bash
docker-compose -f ~/projects/docker-yii2-app-advanced-rbac/docker-codeception-run/docker-compose.yml run --rm codecept run --coverage --coverage-html
```
```
 AdminCreateItemCest: Ensure that create route work
 Test  ../backend/tests/acceptance/AdminCreateItemCest.php:routePage
                                                                                   
  [InvalidArgumentException] Cookie domain "nginx:8081" should not contain a port  
                                                                                   
#1  /repo/vendor/facebook/webdriver/lib/Cookie.php:122
#2  /repo/vendor/facebook/webdriver/lib/Cookie.php:65
#3  /repo/vendor/facebook/webdriver/lib/WebDriverOptions.php:47
#4  /repo/src/Codeception/Module/WebDriver.php:807
#5  /repo/vendor/symfony/event-dispatcher/EventDispatcher.php:212
#6  /repo/vendor/symfony/event-dispatcher/EventDispatcher.php:44
#7  /repo/codecept:42
```

### Details

* Codeception version:2.3
* PHP Version:7.2.1
* Operating System:alpine
* Installation type: Composer
* List of installed packages (`composer show`)
* Suite configuration:

```yml
# paste suite config here
namespace: backend\tests
actor_suffix: Tester
paths:
    tests: tests
    output: tests/_output
    data: tests/_data
    support: tests/_support
    envs: tests/_envs
settings:
    bootstrap: _bootstrap.php
    colors: true
    memory_limit: 1024M
extensions:
    enabled:
        - Codeception\Extension\RunFailed
modules:
    config:
        Yii2:
            configFile: 'config/test-local.php'
coverage:
    # from nginx-conf-test/nginx.conf
    c3_url: http://nginx:8081/index-test.php
    # redefine `php-code/codeception.yml` for `c3.php`. Relative to `c3.php` dir:
    remote_config: backend/codeception.yml
    enabled: true
    include:
        - models/*
        - controllers/*
    exclude:
        - assets/*
        - config/*
        - runtime/*
        - views/*
        - web/*
        - tests/*
```

```yml
suite_namespace: backend\tests\acceptance
actor: AcceptanceTester
modules:
    enabled:
[...]

PR #4768, opened 2018-01-19
Remove :port for cookie domain; $this->module->amOnPage('/'); 

bscheshirwork:
#4765


DavertMik:
Thanks! Looks like a good patch

PR #4779, opened 2018-01-24
Fix REST module formatFilesArray function

zebraf1:
Fixed an issue when passing a file to sendPOST() method without name, size or type parameter.
Reproduced by:
```php
$I->sendPOST('url', [], [
  'uploadedFile' => [
    'tmp_name' => 'path to existing file',
  ],
]);
```

PR #4789, opened 2018-01-28
Fix REST module formatFilesArray function

zebraf1:
Fixed an issue when passing a file to sendPOST() method without name, size or type parameter.
Reproduced by:
```
$I->sendPOST('url', [], [
  'uploadedFile' => [
    'tmp_name' => 'path to existing file',
  ],
]);
```
Note: clone of #4779 but forked from 2.3

### DIFF
--- a/src/Codeception/Module/REST.php
+++ b/src/Codeception/Module/REST.php
@@ -583,13 +583,13 @@ EOF;
                 if (isset($value['tmp_name'])) {
                     $this->checkFileBeforeUpload($value['tmp_name']);
                     if (!isset($value['name'])) {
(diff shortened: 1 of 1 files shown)


zebraf1:
I don't understand what's wrong for these test runners to fail, I branched off 2.3 and my change could not have broken these...
Any idea?

Naktibalda:
If Travis passed, your change is good.
We have no tests covering the code you changed.

zebraf1:
Ok, cool. Travis passed :)

DavertMik:
OK!

PR #4790, opened 2018-01-29
2.3 - fix class not found error in Facebook Module

drull755:


src/Codeception/Module/Facebook.php
```diff
@@ -6,6 +6,7 @@
 use Codeception\Lib\Driver\Facebook as FacebookDriver;
 use Codeception\Lib\Interfaces\DependsOnModule;
 use Codeception\Lib\Interfaces\RequiresPackage;
+use Codeception\Module\Notification;
```
Naktibalda: It is `Codeception\Lib\Notification` - https://github.com/Codeception/Codeception/blob/2.3/src/Codeception/Lib/Notification.php

PR #4798, opened 2018-02-01
Allow to specify EntityManager container key

svycka:
sometimes the hardcoded name is not what you want :) also this will allow using different EntityManager's not just default one when needed

### DIFF
--- a/src/Codeception/Module/ZendExpressive.php
+++ b/src/Codeception/Module/ZendExpressive.php
@@ -20,6 +20,7 @@ use Codeception\Lib\Interfaces\DoctrineProvider;
  * ## Config
  *
  * * container: relative path to file which returns Container (default: `config/container.php`)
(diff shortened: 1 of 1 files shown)


Naktibalda:
Please document this new configuration setting in the docblock above the class.

svycka:
I am not sure how to describe it :D if someone has better idea I am open for suggestions

DavertMik:
This patch looks good but I'd rather call the config option `service_name`

svycka:
I can rename but don't you think service is too generic? maybe then
ormservice or something.

On Feb 17, 2018 4:53 AM, "Michael Bodnarchuk" <notifications@github.com>
wrote:

This patch looks good but I'd rather call the config option service_name

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/Codeception/Codeception/pull/4798#issuecomment-366410792>,
or mute the thread

DavertMik:
ok, please make it `orm_service`
Sorry for delay in reply )

DavertMik:
Thanks for reminding me of this PR! 
I really missed it was updated

PR #4799, opened 2018-02-02
PHPUnit 7 support

DavertMik:
* PHPUnit 7 
* Phar build improvements

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ composer.phar
 composer.lock
 vendor
 package/codecept.phar
+package/codecept5.phar
 package/php54/codecept.phar
 tests/support/_generated
 tests/data/app/db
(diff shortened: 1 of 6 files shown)


riotCode:
@DavertMik on appveyor/pr the error is abrupt, no context for it unless it's the "Json type matches" test but that looks like it's successful. Only thing I could find related to this error: https://help.appveyor.com/discussions/problems/4873-command-exited-with-code-1073741819

DavertMik:
Yeah, that was temporary Appveyor issue. Let me merge it before it started failing again 😊

dhensby:
I've tried to get this working locally but composer is unable to resolve the dependencies because `phpunit/phpunit:^7` requires `sebastian/diff:^3` but this library requires `sebastian/diff:>=1.4 <3.0`

It looks like the `sebastian/diff` dependency needs a bump too...

dhensby:
Ah, seems like this is a known issue: https://github.com/Codeception/Codeception/pull/4816

PR #4803, opened 2018-02-06
Changed domain name used in redirect test

Naktibalda:
Because tests failed on Travis with:
[GuzzleHttp\Exception\ConnectException] cURL error 35: gnutls_handshake() failed: Handshake failed

PR #4807, opened 2018-02-09
Appveyor.yml to specify php 7.1 and PhpBrowserTest fixes

riotCode:
Pull request #4799 fails due to AppVeyor and Chocolatey installing PHP 7.2 and not constrained to any version. This entire file assumes PHP 7.1, so I added what I believe is the constraint for Chocolatey/cist. 

Disclaimer: I am not familiar with Chocolatey or AppVeyor so I'm kind of taking a stab in the dark, well an educated guess have you.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ matrix:
     - php: 5.4 # lowest versions of all dependencies
       env: SYMFONY=2.7.19 SYMFONY_DEPRECATIONS_HELPER=weak # latest version of 2.7.*
     - php: 5.5
-      env: SYMFONY=2.7.5 dependencies=lowest SYMFONY_DEPRECATIONS_HELPER=weak #lowest version of 2.7
+      env: SYMFONY=2.7.19 SYMFONY_DEPRECATIONS_HELPER=weak #latest version of 2.7.*
     - php: 5.6
       env: SYMFONY=2.8.12 SYMFONY_DEPRECATIONS_HELPER=weak # latest version of 2.8.*
     - php: 7.0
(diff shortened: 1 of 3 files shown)


riotCode:
Added fix for PhpBrowserTest.php method. I originally added this to what I thought was fixing the phpunit7 branch but I think this might be something that should be rebased to 2.8 or master. If that's the case I'm fine with doing that.

riotCode:
Duplicate of #4803, I'm stuck in the same spot with TravisCI. However, hope this can be of help somewhat.

Also, I'm more in favor of using the project domain https://codeception.com than an external domain, even if it is http://example.com which would seem to be there forever, but who knows, just my two cents.

DavertMik:
Thanks. Fix works for Appveyor but for Travis there is something more to be done
https://travis-ci.org/Codeception/Codeception/jobs/339335079#L2323

DavertMik:
Maybe we should stick with #4803 fore redirection test?

riotCode:
My fix works for everything else, just looks like TravisCI is having php-curl/ssl issues. https://github.com/travis-ci/travis-ci/issues/6339
So I'm thinking #4803 is the way to go for the redirection test in that case.
In this case, we just need to resolve the composer memory issue that is occuring in #4803.

riotCode:
I hope I wasn't being presumptuous, but seeing as that TravisCI matrix was the only one failing in #4803, and considering minor versions are bug/security fixes I did not think it made sense to force this constraint. Anyone who is using Symfony 2.7.x should be keeping their packages up to date, imo.

That aside @DavertMik let me know if there's anything I should clean up and I'm guessing if your pull request #4799 is satisfied by this then it could be applied to 2.3 that way. Otherwise I'm willing to apply this against #4803 or directly to 2.3 if needed.

DavertMik:
Thanks for the fix. Let's have it merged!

riotCode:
Awesome 👍

Issue #4812, opened 2018-02-16
Cest _after hook not called on ERROR and FAIL


fffilimonov:
Cest _after hook not called on ERROR and FAIL.
In doc https://codeception.com/docs/07-AdvancedUsage there is no information about this case.
But we need to run cleaup after every test (PASSED or FAILED). We can copy paste to _failed hook from _after.

https://github.com/Codeception/Codeception/blob/bd6ad8af1fa7cd12684f0287374b9fe71a8c958a/src/Codeception/Test/Cest.php#L86

There is only failed hook call after exception catch. Is it correct?
Tear down in phpunit  allow to use it in all cases (PASSED or FAILED)

fffilimonov:
Now we have a workaround until PR not submitted. after hook in module called in all cases. We have all our cleanups now in a module.

PR #4813, opened 2018-02-16
call after hook for Cest after every test (PASSED or FAILED)

fffilimonov:
https://github.com/Codeception/Codeception/issues/4812


DavertMik:
what if failure happens in `after` hook?
will it be executed twice?

yes, there is some inconsistency here but this patch may produce some unpredictable issues

PR #4816, opened 2018-02-18
Allow sebastian/diff version 3.0

slavcodev:
This version is required by PhpUnit 7.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         "symfony/dom-crawler": ">=2.7 <5.0",
         "behat/gherkin": "~4.4.0",
         "sebastian/comparator": ">1.1 <3.0",
-        "sebastian/diff": ">=1.4 <3.0 || ^3.0",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -32,7 +32,7 @@
         "symfony/dom-crawler": ">=2.7 <5.0",
         "behat/gherkin": "~4.4.0",
         "sebastian/comparator": ">1.1 <3.0",
-        "sebastian/diff": ">=1.4 <3.0",
+        "sebastian/diff": ">=1.4 <3.0 || ^3.0",
```
DavertMik: Better ">=1.4 <4.0

Naktibalda:
No changes in API?

DavertMik:
No changes are huge and I missed them in my previous PR.
We need to move out all PHPUnit deps in separate package to maintain 6.0.x and 7.x branches separately
See #4817

PR #4822, opened 2018-02-20
Missing bracket in DataFactory.md

ibpavlov:


Naktibalda:
docs/modules/DataFactory.md files is generated from [src/Codeception/Module/DataFactory.php](https://github.com/Codeception/Codeception/blob/2.3.8/src/Codeception/Module/DataFactory.php#L33-L42)
so your change will be lost during the next release and will never make to the website.

I think that adding space between `*` and `]` should fix this issue.
(Please make that change in 2.3 branch, not in master)

ibpavlov:
ok, thank you

PR #4824, opened 2018-02-20
Add `make` method to DataFactory

ibpavlov:
- add `make` method to DataFactory that will return an instance without creating a record in the DB.
- add missing type definitions in DataFactory's DocBlocks


DavertMik:
Thanks!

ibpavlov:
No problem. I'm glad I can help.

PR #4825, opened 2018-02-20
[Yii1] Add missing SERVER_NAME and SERVER_PORT variables

freezy-sk:
CHttpRequest rely on existing SERVER_NAME and SERVER_PORT


DavertMik:
Wow. A patch for Codeception 1.8! 
Please don't use and switch to Codeception 2.3. It still can be used with new PHP versions
1.8 is not maintained

PR #4833, opened 2018-02-27
Fixed Cookies debug section for WebDriver

yesdevnull:
When running Codeception in debug mode and setting cookies, the debug section shows an array of cookies with no data in them.  For example:

```
 [Cookies] [{}]
```

This is what I was seeing after setting a cookie.   Because the method `$this->webDriver->manage()->getCookies()` returns an array of `\Facebook\WebDriver\Cookie`s that have no public properties (or a `__toString()` method), `json_encode` will return an empty object for each cookie - not so ideal.

My suggestion is to perform an `array_map` on the aforementioned `getCookies()` method and call the `toArray()` method on each Cookie object.  Initially I had it as just `$cookie->getName();` but personally I found `$cookie->toArray();` to be more beneficial.  I now get the following when doing `setCookie` in my tests:

```
 [Cookies] [{"name":"PHPSESSID","value":"random123456","path":"/","domain":"localhost","expiry":1521004033,"secure":false,"httpOnly":false}]
```

:tada: much more useful!

**Codeception version:** 2.4.0 (also occurred in 2.3.8)
**WebDriver browser:** chrome

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -774,7 +774,7 @@ class WebDriver extends CodeceptionModule implements
             },
             $cookies
         );
-        $this->debugSection('Cookies', json_encode($this->webDriver->manage()->getCookies()));
+        $this->debugSection('Cookies', json_encode($this->getCookiesAsArray($this->webDriver->manage()->getCookies())));
         $this->assertContains($cookie, $cookies);
     }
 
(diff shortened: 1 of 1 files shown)


DavertMik:
Thanks, looks nice but somehow it broke the test https://semaphoreci.com/codeception/codeception/branches/pull-request-4833/builds/1

This test is about using cookies for session snapshot so probably it's related. Any ideas what may go wrong?

src/Codeception/Module/WebDriver.php
```diff
@@ -774,7 +774,7 @@ function ($c) {
             },
             $cookies
         );
-        $this->debugSection('Cookies', json_encode($this->webDriver->manage()->getCookies()));
+        $this->debugSection('Cookies', json_encode($this->getCookiesAsArray($this->webDriver->manage()->getCookies())));
```
DavertMik: Suggestion: how about moving this into a `debugCookies` method which will do excatly the same: print cookies.
THis way you can reuse it accross different methods without code duplication

yesdevnull:
Unfortunately I haven't had the time to investigate why these tests are failing.  Closing for now, but will re-open in the (hopefully) not too distant future to resolve this.

Issue #4819, opened 2018-02-19
dataProvider  enhancement


fffilimonov:
PhpUnit has useful option - named data in dataprovider.
Example:
https://phpunit.de/manual/current/en/textui.html#textui.examples.filter-shortcuts

Is it possible to run tests with particular data from dataprovider with codeception?
It's needed to make parallel run tests with dataprovider.

fffilimonov:
I made some changes to support running tests with particular data set index from dataset.
Example:
codecept run tests/acceptance/FirstCest.php:frontpageWorks#0
codecept run tests/acceptance/FirstCest.php:frontpageWorks#1

So, it will be possible to split tests by groups with dataproviders.
There is another one PR for this issue:
https://github.com/Codeception/phpunit-wrapper/pull/3

KacerCZ:
It would be also great to print name of dataset in console output.

marabesi:
@fffilimonov I believe that using #0 is for unamed dataproviders, what should we do with named dataproviders? Just replace the 0 witht he name ?

fffilimonov:
@KacerCZ values are printed in console output.
@marabesi we are using named dataproviders and run tests by index (it's enough for us).
There is some problem with name of dataset - codeception replace it with "example".
https://github.com/Codeception/Codeception/blob/1d4bb38d1a5ac852ddb8a35f7aa88c18180176df/src/Codeception/Test/Loader/Cest.php#L90
So dataset name not passed to phpunit (it's always "example").
If you need it - you can submit PR

alecgeatches:
> It would be also great to print name of dataset in console output.

@KacerCZ This worked for me. Here's what the output originally looked like:

```php
public function objectTypeProvider() {
    return [
        ['item'],
        ['term'],
        ['user'],
    ];
}

/**
* @dataProvider objectTypeProvider
*/
public function test_someTest($objectType) {
    // ...
}
```

Output:

```bash
✔ ComponentTest: some test | #0
✔ ComponentTest: some test | #1
✔ ComponentTest: some test | #2
```

The output was more useful when the data provider function used an associative array:

```php
public function objectTypeProvider() {
    return [
        'item' => ['item'],
        'term' => ['term'],
        'user' => ['user'],
    ];
}

/**
* @dataProvider objectTypeProvider
*/
public function test_someTest($objectType) {
    // ...
}
```

Output:

```bash
✔ ComponentTest: some test | "item"
✔ ComponentTest: some test | "term"
[...]

PR #4835, opened 2018-02-28
Running tests with particular data set index

fffilimonov:
https://github.com/Codeception/Codeception/issues/4819


DavertMik:
Thanks!

Issue #4838, opened 2018-03-01
Codeception can't find Kernel using Symfony module with Symfony 4


n1ks2n:
#### Got exception `Kernel class was not found in /srv/dsp-blacklister/src/Kernel.php. Specify directory where file with Kernel class for your application is located with `app_path` parameter.`

``` bash
Codeception PHP Testing Framework v2.4.0
Powered by PHPUnit 7.0.2 by Sebastian Bergmann and contributors.

In Symfony.php line 300:

  [Codeception\Exception\ModuleRequireException]
  [Symfony] module requirements not met --

  Kernel class was not found in /srv/dsp-blacklister/src/Kernel.php. Specify directory where file with Kernel class for your application is located with `app_path` parameter.


Exception trace:
 Codeception\Module\Symfony->getKernelClass() at /srv/dsp-blacklister/vendor/codeception/codeception/src/Codeception/Module/Symfony.php:153
 Codeception\Module\Symfony->_initialize() at /srv/dsp-blacklister/vendor/codeception/codeception/src/Codeception/SuiteManager.php:80
 Codeception\SuiteManager->initialize() at /srv/dsp-blacklister/vendor/codeception/codeception/src/Codeception/Codecept.php:187
 Codeception\Codecept->runSuite() at /srv/dsp-blacklister/vendor/codeception/codeception/src/Codeception/Codecept.php:158
 Codeception\Codecept->run() at /srv/dsp-blacklister/vendor/codeception/codeception/src/Codeception/Command/Run.php:466
 Codeception\Command\Run->runSuites() at /srv/dsp-blacklister/vendor/codeception/codeception/src/Codeception/Command/Run.php:361
 Codeception\Command\Run->execute() at /srv/dsp-blacklister/vendor/symfony/console/Command/Command.php:252
 Symfony\Component\Console\Command\Command->run() at /srv/dsp-blacklister/vendor/symfony/console/Application.php:865
 Symfony\Component\Console\Application->doRunCommand() at /srv/dsp-blacklister/vendor/symfony/console/Application.php:241
 Symfony\Component\Console\Application->doRun() at /srv/dsp-blacklister/vendor/symfony/console/Application.php:143
 Symfony\Component\Console\Application->run() at /srv/dsp-blacklister/vendor/codeception/codeception/src/Codeception/Application.php:108
 Codeception\Application->run() at /srv/dsp-blacklister/vendor/codeception/codeception/codecept:42

[...]

Issue #4695, opened 2017-12-15
AppKernel name only


Mark1Z:
```php
# codeception/codeception/src/Codeception/Module/Symfony.php:282

        $possibleKernelClasses = [
            'AppKernel', // Symfony Standard
            'App\Kernel', // Symfony Flex
        ];
        foreach ($possibleKernelClasses as $class) {
            if (class_exists($class)) {
                $refClass = new \ReflectionClass($class);
                if ($refClass->getFileName() === $file->getRealpath()) {
                    return $class;
                }
            }
        }

        throw new ModuleRequireException(
            __CLASS__,
            "Kernel class was not found in $file. "
            . "Specify directory where file with Kernel class for your application is located with `app_path` parameter."
        );
```
### Details
Code listed above disables possibility to run tests with kernels named different than AppKernel.

* Codeception version: 2.3.7
* PHP Version: 7.1.9
* Operating System: Linux Ubuntu
* Installation type: Phar || Composer
* List of installed packages (`composer show`)
* Suite configuration:

VolCh:
What kernel name causes a problem?

DavertMik:
I think if it causes a problem kernel name should be added to the config.

Mark1Z:
I'm using app/Kernels/Frontend/FrontendKernel and yml config:
```yml
class_name: FunctionalTester
modules:
    enabled:
        - Symfony:
            app_path: '../../app/Kernels/Frontend'
            var_path: '../../var'
            environment: 'test'
        - Doctrine2:
            depends: Symfony
        - \frontend\Helper\Functional
```

and got error from codeception:

Kernel class was not found in /var/www/html/tests/white_label/../../app/Kernels/WhiteLabel/WhiteLabelKernel.php. Specify directory where file with Kernel class for your application is located with `app_path` parameter.

If I add FrontendKernel here:
```php
$possibleKernelClasses = [
            'FrontendKernel',
            'AppKernel', // Symfony Standard
            'App\Kernel', // Symfony Flex
        ];
```
all seems to be fine. As I see, this code was added for symfony 4 in one of last commits, because it was working for me before.

DavertMik:
Yes, that's right, we updated code for symfony 4. Sure as in all major update there could have happened some breaking things unnoticed. I think we should add this option to config

VolCh:
I have two options for simple resolving the issue:
- path with optional filename (according PSR-0/PSR-4 conventions for classname)  in the existing `app_path` option
[...]

PR #4839, opened 2018-03-01
Added support for namespace option for Symfony module

n1ks2n:
With this fix one can add `app_kernel` option to Symfony module config to specify full qualified Application Kernel Class Name (with all namespace chaining) of Symfony 4 Kernel.

### DIFF
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -299,7 +299,7 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
         // mixing default kernel classes and class with changed namespace
 
         if ($this->config['app_kernel']) {
-            $this->kernelClasses = array_merge($this->kernelClasses, [$this->config['app_kernel']]);
+            $this->kernelClasses = [$this->config['app_kernel']];
         }
 
         foreach ($this->kernelClasses as $class) {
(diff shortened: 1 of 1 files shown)


n1ks2n:
This pool request corresponds to #4838  and #4695

n1ks2n:
After checking some conversations on this issue replace `namespace` variable to `app_kernel` . If used `app_kernel` should be a full qualified Kernel Class Name

src/Codeception/Module/Symfony.php
```diff
@@ -284,11 +296,13 @@ protected function getKernelClass()
...
 
-        $possibleKernelClasses = [
-            'AppKernel', // Symfony Standard
-            'App\Kernel', // Symfony Flex
-        ];
-        foreach ($possibleKernelClasses as $class) {
+        // mixing default kernel classes and class with changed namespace
+
+        if ($this->config['app_kernel']) {
+            $this->kernelClasses = array_merge($this->kernelClasses, [$this->config['app_kernel']]);
```
DavertMik: If we already declared class name explicitly name in config why would we need to look for other kernels?
Why not just `$this->kernelClasses = [$this->config['app_kernel']`

n1ks2n: Ok it seems reasonable, i've mixed it just in case the "fool's protection". If u're insisting on that change as it can have an impact on speed it's not problems to fix it =) I also thought of some refactoring for this module cause it looks pretty dirty. Does the module maintainer is actual or u're open to add some maintainers there?

n1ks2n:
Fixed according comment

PapsOu:
Any news about this PR ?

Naktibalda:
This feature was implemented in #4949

Issue #2759, opened 2016-02-01
UTF-8


Nemmo:
Ohhh, that infinite problems with charsets!
I have one another issue about cyrilic charset when I use the Db module. The code below not working cause of hieroglyhs. PHP file in utf8, database too. Codeception version 2.0.16, Yii 2

``` php
$I->canSeeInDatabase('member', [
    'name' => 'Светлана'
]);
```

The log:

``` console
* I see in database "member",{"name":"��������"}
��"}

  [Query] select count(*) from `member` where `name` = ?
 FAIL
```

Naktibalda:
I will try to reproduce it later.

Naktibalda:
I blame encoding of your test file.
Your output clearly has one byte per character.

I created the cept file in utf-8 and got this output:
`* I see in database "users",{"name":"Светлана"}`

Nemmo:
```mysql
SHOW FULL COLUMNS FROM member;
```
![default](https://cloud.githubusercontent.com/assets/5058015/12738533/b1335d54-c98b-11e5-9a7c-5aae8f592bdd.PNG)

Cest file:
![1](https://cloud.githubusercontent.com/assets/5058015/12738535/b3bafa78-c98b-11e5-85e9-913f3286ae33.PNG)

Naktibalda:
Your test code.

Nemmo:
```php
public function myTest(FunctionalTester $I) {
    $I->canSeeInDatabase('member', [
        'name' => 'Светлана'
    ]);
}
```

Naktibalda:
Can your terminal display UTF-8 characters?

Test with `$I->wantTo("see Светлана");`

Yesterday I tested on Ubuntu in Konsole.

Now I tried to run above test via Putty and got this crap:

`1) Failed to see ð¡ð²ðµñð»ð°ð½ð° `

Nemmo:
Yes, of course. If I print Debug::debug('Светлана') I see the name clear. One more thing: I am using Vagrant as server environment.

Naktibalda:
This has nothing to do with Db module, "I see in database" output is printed in the same way as debug output.

@DavertMik you should know more about cyrilic charset issues.

Nemmo:
Maybe the problem is when we encode and decode objects with json inside the db module?

InVeX:
[...]

PR #4850, opened 2018-03-07
Fixed humanize for utf8 strings

nicholascus:
Hi DavertMik

I’ve fixed Bug #2759 in the branch named “Issue-2759” in my fork. The problem wasn't DB module specific and was affecting all step description of parameters processed by Codeception\Step::humanize().

e.g. https://github.com/allure-framework/allure-codeception/issues/41  

Thanks!

### DIFF
--- a/src/Codeception/Step.php
+++ b/src/Codeception/Step.php
@@ -244,7 +244,11 @@ abstract class Step
         $text = preg_replace('/([A-Z]+)([A-Z][a-z])/', '\\1 \\2', $text);
         $text = preg_replace('/([a-z\d])([A-Z])/', '\\1 \\2', $text);
         $text = preg_replace('~\bdont\b~', 'don\'t', $text);
-        return strtolower($text);
+        if (!preg_match('!!u', $text)) {
+            return strtolower($text);
+        } else {
+            return mb_strtolower($text);
+        }
     }
 
     public function run(ModuleContainer $container = null)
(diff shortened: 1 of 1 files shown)


src/Codeception/Step.php
```diff
@@ -244,7 +244,11 @@ protected function humanize($text)
         $text = preg_replace('/([A-Z]+)([A-Z][a-z])/', '\\1 \\2', $text);
         $text = preg_replace('/([a-z\d])([A-Z])/', '\\1 \\2', $text);
         $text = preg_replace('~\bdont\b~', 'don\'t', $text);
-        return strtolower($text);
+        if (!preg_match('!!u', $text)) {
```
Naktibalda: This is an empty pattern and it matches any string I pass to it, so this code block is equivalent to unconditional `return mb_strtolower($text);`

nicholascus: It matches any utf8 string. To get a negative result you need to give it non utf8 string.

Naktibalda: OK, confirmed.
But Codeception only supports UTF-8 (it is hardcoded it many places):
https://github.com/Codeception/Codeception/blob/c50789a9a62cc0eefc0252e88a5f04f8c47b55f4/src/Codeception/Lib/Console/Message.php#L110-L115
https://github.com/Codeception/Codeception/blob/c50789a9a62cc0eefc0252e88a5f04f8c47b55f4/src/Codeception/Scenario.php#L117-L128
https://github.com/Codeception/Codeception/blob/c50789a9a62cc0eefc0252e88a5f04f8c47b55f4/src/Codeception/Actor.php#L38
https://github.com/Codeception/Codeception/blob/c50789a9a62cc0eefc0252e88a5f04f8c47b55f4/src/Codeception/Step.php#L126

So unconditional mb_strtolower with hardcoded 'utf-8' charset would be better, unless you want to update all other places too.

nicholascus: Yes, sure, I just though that non mb lowercase function could be used there for a reason so I was trying to do the change in a backward compatible way.

DavertMik:
Thanks. After @Naktibalda's suggestions, this looks pretty clean :+1:

PR #4857, opened 2018-03-08
[Done - waiting review] Multi Database

eXorus:
Hello,

This is a work in progress but I need your help to continue. Do you think it will be useful ?

## My current issue

> My application have 2 databases that I need to use in my test. I saw the extension https://github.com/iamdevice/codeception-multidb but I can't use it because it is not up-to-date with the all the last modification I did in the official Db module. I don't want to do the job two times and I think we don't need to have an extension for that we can include it in the official Db module.

## How does it work ?

This PR is already working with this simple example:

**multidbtest.suite.yml**
```yml
actor: MultidbtestTester
modules:
    enabled:
      - Db
          dsn: pgsql:host=localhost;dbname=db1
          user: user
          password: password
          connections:
            db2:
              dsn: pgsql:host=localhost;dbname=db2
              user: user
              password: password
```

```php
class MultiDbFirstCest
{
    public function tryToTest(MultidbtestTester $I)
    {
        $I->seeInDatabase('user', array('id' => 1));
        $I->amConnectedToDb('db2');
        $I->seeInDatabase('projects', array('id' => 1));
    }

    public function tryToTest2(MultidbtestTester $I)
    {
        $I->seeInDatabase('user', array('id' => 1));
    }
}
```

and the good news is that it does not break anything for the existing.
**dbtest.suite.yml**
```yml
actor: DbtestTester
modules:
    enabled:
      - Db
          dsn: pgsql:host=localhost;dbname=db2
          user: user
          password: password
```
👍 

## What next ?
- throw an exception when the connection name doesn't exist
- add test
- manage transaction, dump, ...
- be able to execute one or more query in a scope like this
```php
        $I->amConnectedToDb('claire', function() {
            $I->seeInDatabase('projects', array('id' => 1));
        });
```

My PR is only 7 lines currently, I think it's a clean way to do it.


quantum-x:
@eXorus Great idea.
I have been also looking for a means to have more control on multiple database configurations.
It would be very useful to be able to:
 - repopulate the database when needed
 - change the dump file location easily from a test context

Could be useful additions to your patch?

DavertMik:
Did you see the MultiDb helper https://github.com/iamdevice/codeception-multidb ?

eXorus:
@DavertMik Yes I saw it like I said in the part my current issue but it's not working well because it's based on an old version of db (for example there is no updateInDatabase function).

DavertMik:
Ok, looks like I can't oppose this feature anymore. Too many people tried to implement this...

But I still have some concerns about database cleanup and repopulation. How do you see it? Is that possible to populate a database for all connections?

eXorus:
@DavertMik I will try.
I dont use the feature db cleanup and repopulation.

eXorus:
This is not so easy but I did something with cleanup and repopulation.

I don't like how I did this, I think there is a better way to do it.

Let me know if you have any comment.

DavertMik:
Strange but I like it. Really. The only issue here that we need tests for multiple databases...

eXorus:
@DavertMik I started a refacto and I will add tests for multiple databases.

DavertMik:
Great! Thanks
If you will have questions please join our Slack https://join.slack.com/t/codeception/shared_invite/enQtMzE2MzgxNzM1OTUzLTcwMmMxZjMxYTdkMzljMzNjNmNiNWQ0NGNjNTY0MjJlNWNjOTI0ZWU0Mjg4YmE5NTI0MmMwNDZmNzRhOTZmNWE

eXorus:
@DavertMik This is a big refacto on how Db module works, it was necessary. I didn't touch the existing test to be sure that I did not break anything.

- I added test for multidatabase on Sqlite because I don't know how to have 2 databases in MySql and PostgreSql I think I need to edit docker files.
- I added the anonymous function to scope some query in another database without changing the current database
- I added cleanup and repopulation
- I throw an exception if the database doesn't exist
- I will work on addInsertedRow/removeInserted for multi database


@quantum-x Nice idea but I don't know how to test annotation

eXorus:
@DavertMik I added one test for addInsertedRow/removeInserset and implement it.

I always need to add test for MySQL et PostgreSql and clean some code but the main part is done. Could you review it ?

src/Codeception/Module/Db.php
```diff
@@ -205,156 +194,271 @@ class Db extends CodeceptionModule implements DbInterface
...
+    public $databasesPopulated = [];
+    public $databasesSql = [];
     protected $insertedRows = [];
+    public $currentDatabase = self::DEFAULT_DATABASE;
 
-    /**
-     * @var array
-     */
-    protected $requiredFields = ['dsn', 'user', 'password'];
+    public function getDatabases()
```
DavertMik: should not be available inside `$I->`.
So this should be either prefixed with `_` or made protected

src/Codeception/Module/Db.php
```diff
@@ -205,156 +194,271 @@ class Db extends CodeceptionModule implements DbInterface
...
+                    'cleanup' => false,
+                    'reconnect' => false,
+                    'dump' => null,
+                    'populator' => null,
+                ], $databaseConfig);
+            }
+        }
+        return $databases;
+    }
+    public function connectToDatabases()
```
DavertMik: should not be available inside $I->.
So this should be either prefixed with _ or made protected

src/Codeception/Module/Db.php
```diff
@@ -205,156 +194,271 @@ class Db extends CodeceptionModule implements DbInterface
...
+    {
+        foreach ($this->getDatabases() as $databaseKey => $databaseConfig) {
+            $this->connect($databaseKey, $databaseConfig);
+        }
+    }
+    public function cleanUpDatabases()
```
DavertMik: should not be available inside $I->.
So this should be either prefixed with _ or made protected

src/Codeception/Module/Db.php
```diff
@@ -205,156 +194,271 @@ class Db extends CodeceptionModule implements DbInterface
...
+    {
+        foreach ($this->getDatabases() as $databaseKey => $databaseConfig) {
+            $this->_cleanup($databaseKey, $databaseConfig);
+        }
+    }
+    public function populateDatabases($configKey)
```
DavertMik: should not be available inside $I->.
So this should be either prefixed with _ or made protected

src/Codeception/Module/Db.php
```diff
@@ -205,156 +194,271 @@ class Db extends CodeceptionModule implements DbInterface
...
+    {
+        foreach ($this->getDatabases() as $databaseKey => $databaseConfig) {
+            if ($databaseConfig[$configKey]) {
+                $this->_loadDump($databaseKey, $databaseConfig);
+            }
+        }
+    }
+    public function readSqlForDatabases()
```
DavertMik: should not be available inside $I->.
So this should be either prefixed with _ or made protected

src/Codeception/Module/Db.php
```diff
@@ -205,156 +194,271 @@ class Db extends CodeceptionModule implements DbInterface
...
+    {
+        foreach ($this->getDatabases() as $databaseKey => $databaseConfig) {
+            $this->readSql($databaseKey, $databaseConfig);
+        }
+    }
+    public function removeInsertedForDatabases()
```
DavertMik: should not be available inside $I->.
So this should be either prefixed with _ or made protected

src/Codeception/Module/Db.php
```diff
@@ -205,156 +194,271 @@ class Db extends CodeceptionModule implements DbInterface
...
+    {
+        foreach ($this->getDatabases() as $databaseKey => $databaseConfig) {
+            $this->amConnectedToDatabase($databaseKey);
+            $this->removeInserted($databaseKey, $databaseConfig);
+        }
+    }
+    public function disconnectDatabases()
```
DavertMik: should not be available inside $I->.
So this should be either prefixed with _ or made protected

src/Codeception/Module/Db.php
```diff
@@ -205,156 +194,271 @@ class Db extends CodeceptionModule implements DbInterface
...
+    {
+        foreach ($this->getDatabases() as $databaseKey => $databaseConfig) {
+            $this->disconnect($databaseKey, $databaseConfig);
+        }
+    }
+    public function reconnectDatabases()
```
DavertMik: should not be available inside $I->.
So this should be either prefixed with _ or made protected

src/Codeception/Module/Db.php
```diff
@@ -205,156 +194,271 @@ class Db extends CodeceptionModule implements DbInterface
...
+    {
+        foreach ($this->getDatabases() as $databaseKey => $databaseConfig) {
+            if ($databaseConfig['reconnect']) {
+                $this->disconnect($databaseKey, $databaseConfig);
+                $this->connect($databaseKey, $databaseConfig);
+            }
+        }
+    }
+
+    public function __get($name)
```
DavertMik: Ok. We don't need this hook to be used on long-term
Please add a deprecation warning here. 
`Codeception\Lib\Notification::deprecate`

src/Codeception/Module/Db.php
```diff
@@ -205,156 +194,271 @@ class Db extends CodeceptionModule implements DbInterface
...
+    {
+        if ($name == 'driver') {
+            return $this->getDriver();
+        }
+        if ($name == 'dbh') {
+            return $this->getDbh();
+        }
+    }
+
+    public function getDriver()
```
DavertMik: Add `_` prefix

src/Codeception/Module/Db.php
```diff
@@ -205,156 +194,271 @@ class Db extends CodeceptionModule implements DbInterface
...
+    {
+        return $this->drivers[$this->currentDatabase];
+    }
+    public function getDbh()
```
DavertMik: Add `_` prefix

src/Codeception/Module/Db.php
```diff
@@ -205,156 +194,271 @@ class Db extends CodeceptionModule implements DbInterface
...
+    {
+        return $this->dbhs[$this->currentDatabase];
+    }
+
+    public function amConnectedToDatabase($databaseKey, $callback = null)
```
DavertMik: Ok, it's not that clear what does the callback do.
Please add a docblock here with documentation. Inside documentation please add code examples of usage

src/Codeception/Module/Db.php
eXorus: Fixed with protected

src/Codeception/Module/Db.php
eXorus: Fixed with protected

src/Codeception/Module/Db.php
eXorus: Fixed

src/Codeception/Module/Db.php
eXorus: Fixed

src/Codeception/Module/Db.php
eXorus: Fixed

src/Codeception/Module/Db.php
eXorus: Fixed

src/Codeception/Module/Db.php
eXorus: Fixed

src/Codeception/Module/Db.php
eXorus: Fixed

src/Codeception/Module/Db.php
eXorus: Fixed

src/Codeception/Module/Db.php
eXorus: Fixed

src/Codeception/Module/Db.php
eXorus: Fixed

src/Codeception/Module/Db.php
eXorus: Fixed

eXorus:
@DavertMik I fixed all your comments except for breaking change on driver and dbh I don't understand what you want. I need to add again the public properties and let the magic function __get do the job or you want that I manage the value of driver and dbh with the current database ?

eXorus:
@DavertMik Any news about this PR ?

eXorus:
@DavertMik Sorry to insist I know it's difficult to accept this PR.
I'm using this new module DB in my project since 3 months without issue.

src/Codeception/Module/Db.php
```diff
@@ -205,156 +194,271 @@ class Db extends CodeceptionModule implements DbInterface
...
-     */
-    public $driver;
-
     /**
      * @var array
      */
+    protected $requiredFields = ['dsn', 'user', 'password'];
+    const DEFAULT_DATABASE = 'default';
+
+    public $drivers = [];
```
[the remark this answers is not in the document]
eXorus: See the comment bellow

src/Codeception/Module/Db.php
```diff
@@ -205,156 +194,271 @@ class Db extends CodeceptionModule implements DbInterface
...
+    public $dbhs = [];
```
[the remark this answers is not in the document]
eXorus: See the comment bellow

eXorus:
@DavertMik Sorry to disturb you again, but I really need to know what you want ?

If you are not anymore agree with this PR, just tell me like it is, maybe I could provide another extension to do the same think ...

DavertMik:
See PR #5140

PR #4860, opened 2018-03-09
Laravel 5.4+ (5.1+ backward compatible) support for callArtisan metho…

mohamed-aiman:
**Added Laravel 5.4+ (5.1+ backward compatible) support for callArtisan method in Laravel5 module.**

Laravel 5.4 Laravel enables a third parameter can be passed to the call method (https://laravel.com/api/5.4/Illuminate/Foundation/Console/Kernel.html#method_call)  of console Kernel. Which enables the developer to control the output while using commands within commands etc.

For example here is a sample command which will benefit from the change.
**What does it do?**
When the command is executed it calls an Artisan command which from inside calls to other Artisan commands.
If the proposed change is made the output interface (OutputInterface $output) that is being by default by Codeception execute method can be passed on to the Laravel artisan commands, which then will let the outputs be printed to the console as one would expect, else it won't be displayed.

```php

<?php
namespace Tests;

use Codeception\Lib\Di;
use Codeception\Lib\ModuleContainer;
use Codeception\Module\Laravel5;
use \Codeception\CustomCommandInterface;
use \Symfony\Component\Console\Command\Command;
use \Symfony\Component\Console\Input\InputInterface;
use \Symfony\Component\Console\Output\OutputInterface;

class SampleCommand extends Command implements CustomCommandInterface
{
    public static function getCommandName()
    {
        return "platform:init";
    }

    protected function configure()
    {
        parent::configure();
    }


    private function getModuleConfig($moduleName)
    {
        $config = isset($this->config['modules']['config'][$moduleName])
            ? $this->config['modules']['config'][$moduleName]
            : [];

        if (!isset($this->config['modules']['enabled'])) {
            return $config;
        }

        if (!is_array($this->config['modules']['enabled'])) {
            return $config;
        }

        foreach ($this->config['modules']['enabled'] as $enabledModuleConfig) {
            if (!is_array($enabledModuleConfig)) {
                continue;
            }

            $enabledModuleName = key($enabledModuleConfig);
            if ($enabledModuleName === $moduleName) {
                return Configuration::mergeConfigs(reset($enabledModuleConfig), $config);
            }
        }

        return $config;
    }

    public function getDescription()
    {
        return "As the database is huge, this command does the initial migrations, seeding,
                    create initial users and attach roles, permissions ";
    }

    protected function execute(InputInterface $input, OutputInterface $output)
    {
        $this->config = Configuration::suiteSettings('acceptance', Configuration::config());
    	$mc = $this->moduleContainer = new ModuleContainer(new Di(), $this->config);
    	$laravel = new \Helper\Laravel($mc, $this->getModuleConfig('Helper\Laravel'));
    	new \Codeception\Lib\Connector\Laravel5($laravel);
    	$laravel->callArtisan('multipleartisan:commandThatHasOtherCommandsInside', [], $output);
    }

}

```

### DIFF
--- a/src/Codeception/Module/Laravel5.php
+++ b/src/Codeception/Module/Laravel5.php
@@ -14,6 +14,7 @@ use Codeception\Util\ReflectionHelper;
 use Illuminate\Contracts\Auth\Authenticatable;
 use Illuminate\Database\Eloquent\Model as EloquentModel;
 use Illuminate\Support\Collection;
+use Symfony\Component\Console\Output\OutputInterface;
 
 /**
  *
@@ -423,13 +424,18 @@ class Laravel5 extends Framework implements ActiveRecord, PartedModule
 
      * @param string $command
      * @param array $parameters
+     * @param OutputInterface $output
      */
-    public function callArtisan($command, $parameters = [])
+    public function callArtisan($command, $parameters = [], OutputInterface $output = null)
     {
         $console = $this->app->make('Illuminate\Contracts\Console\Kernel');
-        $console->call($command, $parameters);
-
-        return trim($console->output());
+        if (!$output) {
+            $console->call($command, $parameters);
+            return trim($console->output());
+        }
+        
+        $console->call($command, $parameters, $output);
+            
     }
 
     /**


DavertMik:
Thanks!

PR #4866, opened 2018-03-14
added app url option for laravel

mohamed-aiman:
When the initialize method of Codeception\Lib\Connector\Laravel5 is called first .env has never been loaded. 

Lets say  you have the following on  
**config/app.php** 
```php
'url' => env('APP_URL', 'http://wrongurl'),
```
**.env.testing**
```
APP_URL=http://testing
```
**.env**

```
APP_URL=http://production
```

When initialize method is called for the first time even though http://testing has to be the value http://wrongurl is taken as the `url`.

here is the actual code and what happens when the method is run for the first time

```php

    private function initialize($request = null)
    {

        ........

        $this->app = $this->kernel = $this->loadApplication();

        // lets say request is null
        if (is_null($request)) {
            $appConfig = require $this->module->config['project_dir'] . 'config/app.php';
            // dump($_ENV['APP_URL']); returns undifined index, expected http://testing
            // so the value of $appConfig['url] is http://wrongurl
            // eventhough http://testing is expected
            $request = SymfonyRequest::create($appConfig['url']);
        }

        ........

       // dump($_ENV['APP_URL']); returns undifined index, expected http://testing
        $this->app->make('Illuminate\Contracts\Http\Kernel')->bootstrap();
       // dump($_ENV['APP_URL']); returns http://testing, expected http://testing

        .........
```

So what I did is to get url from module's config.

```YML
modules:
    enabled:
        - Asserts
        - REST:
            depends: Laravel
        - Laravel:
            url: http://testing
```

This fix will optimize by getting the already loaded config value from module instead of loading whole config/app.php file to just get the url. Additionally, the previous option is set as the fallback.

### DIFF
--- a/src/Codeception/Lib/Connector/Laravel5.php
+++ b/src/Codeception/Lib/Connector/Laravel5.php
@@ -171,8 +171,7 @@ class Laravel5 extends Client
 
         // Set the request instance for the application,
         if (is_null($request)) {
-            $appConfig = require $this->module->config['project_dir'] . 'config/app.php';
-            $request = SymfonyRequest::create($appConfig['url']);
+            $request = SymfonyRequest::create($this->getUrl());
         }
         $this->app->instance('request', Request::createFromBase($request));
 
@@ -221,6 +220,28 @@ class Laravel5 extends Client
         }
 
         $this->module->setApplication($this->app);
+    }
+
+    /**
+     * Get the Laravel app url.
+     * @return string
+     * @throws ModuleConfig
+     */
+    private function getUrl()
+    {
+        return (isset($this->module->config['url'])) ? $this->module->config['url'] :  $this->getUrlFromApp();
+    }
+    
+    /**
+     * Get the Laravel app url from app config file.
+     * @return string
+     * @throws ModuleConfig
+     */
+    private function getUrlFromApp()
+    {
+        $appConfig = require $this->module->config['project_dir'] . 'config/app.php';
+        
+        return $appConfig['url'];
     }
 
     /**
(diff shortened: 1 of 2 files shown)


DavertMik:
I'm really sorry to answer after 2 weeks. I had a hope that some other guys with better Laravel expertise will comment something but looks like no... 
So I have a question to you: why `url` param should be so special and in which cases it should be taken to documentation. For most cases I don't think the URL should be somehow important in functional testing...

I'd also ask you to document the `url` parameter inside the docblock of Laravel5 module class.

src/Codeception/Lib/Connector/Laravel5.php
```diff
@@ -223,6 +222,28 @@ private function initialize($request = null)
...
     }
 
+    /**
+     * Get the Laravel app url.
+     * @return string
+     * @throws ModuleConfig
+     */
+    private function getUrl()
+    {
+        return (isset($this->module->config['url'])) ? $this->module->config['url'] :  $this->getUrlFromApp();
```
DavertMik: please add `url` to list of config options inside `Laravel5` helper with a default value.
All config options should have default values, so `isset` is not needed

edno:
I agree with @DavertMik. I cannot see a valid reason for having a dedicated parameter `url` for dealing with application url, `APP_URL` should be enough.

@mohamed-aiman, did you try to use the[ dynamic configuration with parameters](https://codeception.com/docs/06-ModulesAndHelpers#Dynamic-Configuration-With-Parameters) (`params` configuration in `codeception.yml`)?

DavertMik:
No response. Closing

PR #4901, opened 2018-03-31
Improve support of PHPUnit warning status

edno:
This PR requires following PRs for **phpunit-wrapper** :
- codeception/phpunit-wrapper#21
- codeception/phpunit-wrapper#22

Better support of PHPUnit warning status:
- add new event `test.warning`
- display 'W' instead of success for warning test cases

### DIFF
--- a/src/Codeception/Events.php
+++ b/src/Codeception/Events.php
@@ -93,6 +93,13 @@ final class Events
      */
     const TEST_SKIPPED = 'test.skipped';
 
+
+    /**
+     * The event listener method receives a {@link Codeception\Event\FailEvent} instance.
+     */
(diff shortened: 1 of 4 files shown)


DavertMik:
Thanks!

Naktibalda:
This test fails with PHPUnit 9.5 because invalid data provider is no longer a warning, but an error.

The only remaining cases when PHPUnit throws a warning are when test class has no methods and when constructor of test class isn't public: https://github.com/sebastianbergmann/phpunit/blob/9.5.0/src/Framework/TestSuite.php#L211-L251
I failed to get these warnings in Codeception.

Issue #4836, opened 2018-02-28
Errors using `finally` with PHP ~ 5.4


ricpelo:
Commit 7c2776d introduces a `finally` keyword which can be used only with PHP >= 5.5, but Codeception 2.4.0 requires PHP >= 5.4.0:

https://github.com/Codeception/Codeception/blob/1c3bf80b631c58a920008dcde87965b2078641ca/composer.json#L18

### Details

* Codeception version: 2.4.0
* PHP Version: 5.4
* Operating System: Linux
* Installation type: Composer

Issue #4837, opened 2018-03-01
Codeception 2.4 will be installed to PHP 5.4


nojimage:
#### What are you trying to achieve?

I try codeception install on PHP 5.4

```bash
composer require --dev codeception/codeception
```

#### What do you get instead?

In PHP 5.4 it will be installed despite being unable to use Codeception 2.4.

I thinks should be change php version in composer.json, like `"php": ">=5.6.0 <8.0",`

DavertMik:
Yes. Thanks looks like I forgot to update composer.json

PR #4905, opened 2018-04-01
Bump PHP min version to 5.6

edno:
Fix issues #4836 and #4837


DavertMik:
Thanks. How could I forget about it!

Issue #4869, opened 2018-03-15
No way to delete a sesion snapshot when you logout


vi4o:
#### What are you trying to achieve?
My system has two logins - frontend and backend. 
My test tries to:
1) login in the frontend and place an order
2) logout from the frontend
3) login in the backend
4) accept the order
5) logout from the backend
6) login in the frontend again
7) check if the order is accepted

#### What do you get instead?
It fails on step 6, because there is an existing session snapshot called 'frontend' and it assumes it is logged in.

My login is written like this:
```
public function loginToFrontend(
        $name = Credentials::FRONTEND_USERNAME,
        $password = Credentials::FRONTEND_PASSWORD,        
    ) {
        $I = $this->tester;

        if ($I->loadSessionSnapshot('frontend')) {
            return;
        }
        $I->amOnPage(self::$URL);
        $I->fillField(self::$usernameField, $name);
        $I->fillField(self::$passwordField, $password);
        $I->click(self::$loginButton);
        $I->wait(2);
        $I->see(self::$frontendTitle);
        $I->saveSessionSnapshot('frontend');
```
The login to the backend is almost the same, but it saves to a different snapshot.

Currently there is no way to delete a snapshot when i need to logout.

Naktibalda:
Snapshot deletion was implemented by linked pull requests.

PR #4909, opened 2018-04-02
Add session snapshot deletion

vi4o:
Simple session snapshot deletion.
Helps you to delete session snapshot if you need to logout.
Fixes issue #4869


DavertMik:
Thanks!

PR #4912, opened 2018-04-03
Added safety check for runtimes emulating php 7.1+ but not supporting pcntl_async_signals()

etki:
Hi!

Latest available HHVM doesn't have `pcntl_async_signals()` on board, but mimics PHP 7.1, which was assumed to have. I added basic function_exists() check to prevent function-does-not-exist errors. This probably should be ported to older minor releases.


DavertMik:
> Latest available HHVM 

wow you are still using it!
If tests are passing than it looks good to me

etki:
@DavertMik 

> wow you are still using it!

i have a habit of dragging myself into weird situations

DavertMik:
Nice. That means we are keeping HHVM support for some reason. Thanks!

PR #4935, opened 2018-04-12
Allow custom steps to be Comment

Slamdunk:


Slamdunk:
The failing build isn't related to this PR changes

DavertMik:
Yeah, thanks for the idea but I'm not so sure it's actually usable. 
`expect` prefix can be used for some real steps not just comments, like `expectException`. So magically declaring all them comments is not a great idea. If that would be something we supported from the very beginning it could be possible. But nowadays there are possible lots of `expect*` methods in projects which use Codeception.

Slamdunk:
Hi, I agree `expect` is not suitable.
The target of this PR is to enable `Comment` step to be used also by the user: how would you suggest to get to this goal?

Slamdunk:
@DavertMik any suggestion on how to do this in the userland?

Issue #1342, opened 2014-08-22
Add option to Yii2 module to enable log targets


Ragazzo:
I find it useful to check some debug request data with `yii2 - debug` module that is collected from logs when tests are run. So we can add `enableLogs` option to `Yii2` module to enable this log [targets](https://github.com/Codeception/Codeception/blob/2.0/src/Codeception/Lib/Connector/Yii2.php#L72)

samdark:
Won't be implemented till more people will request it. Closing.

PR #4943, opened 2018-04-17
[Yii2] make logger replaceable and allow enabled log targets

yurii-github:
make logger replaceable and allow enabled log targets

my use case: I currently use logger for logging activity, data change etc via target and I found out that it is impossible to test logger at all.

similar issues
#1342

### DIFF
--- a/src/Codeception/Lib/Connector/Yii2.php
+++ b/src/Codeception/Lib/Connector/Yii2.php
@@ -48,7 +48,11 @@ class Yii2 extends Client
      */
     const CLEAN_MANUAL = 'manual';
 
-
+    /**
+     * @var bool keep log targets enabled or force disable. Set 'false' to force disable
+     */
+    public $logTargetsEnabled = false;
+    
     /**
      * @var string application config file
      */
@@ -157,12 +161,11 @@ class Yii2 extends Client
         $app = $this->getApplication();
 
         // disabling logging. Logs are slowing test execution down
-        foreach ($app->log->targets as $target) {
-            $target->enabled = false;
+        if (!$this->logTargetsEnabled) {
+            foreach ($app->log->targets as $target) {
+                $target->enabled = false;
+            }
         }
-
-
-
 
         $yiiRequest = $app->getRequest();
         if ($request->getContent() !== null) {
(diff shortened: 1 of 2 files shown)


yurii-github:
sorry, it will break some stuff! I will reconsider better use

yurii-github:
but still it disables log targets in doRequest() by default and somehow targets are got available in test as
\Yii::getLogger()->dispatcher->targets

yurii-github:
somehow dispatcher is not available in test as
**\Yii::getLogger()->dispatcher**
but available at 
**\Yii::$app->log** 

maybe this is the issue of the framework in general?

yurii-github:
issue with \Yii::setLogger() - it doesn't set dispatcher for logger

samdark:
That would result in writing app logs in Codeception runtime dir before getter is called...

yurii-github:
hm.. I see, I will need to properly checkout fork then  to test it as issue went deeper than expected.
It looks like logger was never planned as replaceable.

Maybe somebody with more knowledge can help too.

regards

src/Codeception/Lib/Connector/Yii2.php
```diff
@@ -107,12 +111,13 @@ public function startApp()
         if (!isset($config['class'])) {
             $config['class'] = 'yii\web\Application';
         }
-
+        if (!isset($config['container']['definitions']['yii\log\Logger']['class'])) {
+            $config['container']['definitions']['yii\log\Logger']['class'] = 'Codeception\Lib\Connector\Yii2\Logger';
```
DavertMik: what if `$config['container']['definitions']['yii\log\Logger']` is not set?
this goes to break the stuff. Also, I don't like how hardcoded it is here...

src/Codeception/Module/Yii2.php
```diff
@@ -34,6 +34,7 @@
  * * `configFile` *required* - the path to the application config file. File should be configured for test environment and return configuration array.
  * * `entryUrl` - initial application url (default: http://localhost/index-test.php).
  * * `entryScript` - front script title (like: index-test.php). If not set - taken from entryUrl.
+ * * `logTargetsEnabled` - keep log targets enabled or set 'false' to force disable. Default is 'false'
```
DavertMik: I'm not so sure that from user's perspective it's clear what are logTargets. We need a better user-firendly name and description. Focus on use case and not on implementation

DavertMik:
No feedback for a month. Closing this. Feel free to reopen it

PR #4945, opened 2018-04-18
Documentation, grabFromDatabase, example error

rwese:
Example for grabFromDatabase was incomplete.


Naktibalda:
Module documentation is generated from source, please update https://github.com/Codeception/Codeception/blob/2.4/src/Codeception/Module/Db.php instead.

DavertMik:
Thanks!

PR #4946, opened 2018-04-18
Makes JsonArray convert all decoded non-arrays to array

### LINKED ISSUE #4944: Util\JsonArray expects input to be JSON array and causes issues when it gets bare JSON string
When using `REST\seeResponseMatchesJsonType()` it internally uses Util\\[JsonArray](https://github.com/Codeception/Codeception/blob/2.4/src/Codeception/Util/JsonArray.php) object to represent (convert) the request data to an actual array and work with that. It works as expected for most inputs - if JsonArray can't decode the string to json, for example, it correctly throws an exception.

However, when it gets input like this: `"I am a JSON string"` it is able to decode it (because it is a correct JSON). But it is not an *array*, and thus you get a JsonArray object that is, in fact, not an array at all.

When this is passed to `seeResponseMatchesJsonType`'s internal workings, it eventually fails on some `array_key_exists` call.

It doesn't really matter how you deal with this - when you don't get an array in JsonArray's constructor either throw an exception or just convert whatever you got an array, but something should probably be done about it. [JsonArrayTest](https://github.com/Codeception/Codeception/blob/2.4/tests/unit/Codeception/Util/JsonArrayTest.php) should also probably have tests for the constructor, including one where the input is a bare JSON string.

Amunak:
Also adds a unit test covering this case.

Fixes #4944

### DIFF
--- a/src/Codeception/Util/JsonArray.php
+++ b/src/Codeception/Util/JsonArray.php
@@ -23,7 +23,13 @@ class JsonArray
             throw new InvalidArgumentException('$jsonString param must be a string.');
         }
 
-        $this->jsonArray = json_decode($jsonString, true);
+        $jsonDecode = json_decode($jsonString, true);
+
+        if (!is_array($jsonDecode)) {
+            $jsonDecode = [$jsonDecode];
+        }
+
+        $this->jsonArray = $jsonDecode;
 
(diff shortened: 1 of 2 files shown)


DavertMik:
Thanks!

PR #4949, opened 2018-04-22
Added support for custom kernel names for Symfony

omnilight:
Symfony really has not requirement to place kernel class into App\Kernel or AppKernel namespaces. So this PR gives ability to configure it's placement


src/Codeception/Module/Symfony.php
```diff
@@ -104,6 +108,11 @@
  */
 class Symfony extends Framework implements DoctrineProvider, PartedModule
 {
+    const POSSILE_KERNEL_CLASSES = [
```
Naktibalda: Mistype in the name of constant.

src/Codeception/Module/Symfony.php
```diff
@@ -284,10 +294,12 @@ protected function getKernelClass()
...
         require_once $file;
 
-        $possibleKernelClasses = [
-            'AppKernel', // Symfony Standard
-            'App\Kernel', // Symfony Flex
-        ];
+        if (empty($this->config['kernel_class'])) {
+            $possibleKernelClasses = static::POSSILE_KERNEL_CLASSES;
+        } else {
+            $possibleKernelClasses = $this->config['kernel_class'];
```
Naktibalda: `$possibleKernelClasses` is an array, `kernel_class is a string` and it would look odd as an array in the config, so I think that this line would be better writter as 
```php
$possibleKernelClasses = [ $this->config['kernel_class'] ];
```

Also it would be good to do some typechecking and throw ModuleException if kernel class is not a string. (unless you want to accept arrays, then typechech that too).

src/Codeception/Module/Symfony.php
omnilight: Fixed

src/Codeception/Module/Symfony.php
omnilight: Done

PR #4951, opened 2018-04-24
Fix broken links to code in customisation docs

JackRobbers:


JackRobbers:
I'm not sure if there is a way to link files of 'default' rather than a specific branch. If it is possible you wouldn't need to change the urls every release.

DavertMik:
Yeah, I'm pretty sure I won't update it every time 😕
There should be a way to do it automatically... Permalinks or docs parsing...
Anyway, let's merge it for now and think about more constant solution

PR #4952, opened 2018-04-25
Fix issue regarding 'The xxx service is already initialized, you cannot replace it'

DzeryCZ:
Every time I clean /test cache and run API suite the error appears for the first test. Otherwise, it's OK.

The issue was reported here: #4806

### DIFF
--- a/src/Codeception/Lib/Connector/Symfony.php
+++ b/src/Codeception/Lib/Connector/Symfony.php
@@ -76,7 +76,9 @@ class Symfony extends \Symfony\Component\HttpKernel\Client
         $this->container = $this->kernel->getContainer();
 
         foreach ($this->persistentServices as $serviceName => $service) {
-            $this->container->set($serviceName, $service);
+            if (!$this->container->has($serviceName)) {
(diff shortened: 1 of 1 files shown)


Naktibalda:
In my opinion "Cannot replace" breaks persistent service functionality, so it is worth at least to log a warning with codecept_debug if service cannot be replaced.

Naktibalda:
On a second thought, replacing works with older versions of Symfony, so it would be better to try to set and catch exception if it failed.

DzeryCZ:
Hello,

Thank you for such fast response.
Yep, It seems to be not a correct solution.

Do you think to put the catch inside of `src/Codeception/Lib/Connector/Symfony.php::rebootKernel`?

Naktibalda:
Yes, I think that try catch should be inside foreach loop, around `$this->container->set($serviceName, $service);` statement.

DzeryCZ:
In the end, removing Doctrine configuration from `Functional.suite.yaml` solved the issue for me.

```
- Doctrine2:
            depends: Symfony
            cleanup: true
```

So this change is not needed.

PR #4956, opened 2018-04-26
Skip cleanup when ORM specifies no cleanup

yesdevnull:
If you don't specify `cleanup: false` on the `DataFactory` module configuration and instead rely on the `cleanup` value for the ORM module (e.g. `Doctrine2`) then the `DataFactory` should correctly follow that value.

Correct me if I'm wrong but the `DataFactory` module should skip the cleanup if the `ORM` module (of which it depends on) has `cleanup: false`, right?

### DIFF
--- a/src/Codeception/Module/DataFactory.php
+++ b/src/Codeception/Module/DataFactory.php
@@ -187,7 +187,7 @@ EOF;
     public function _after(TestInterface $test)
     {
         $skipCleanup = array_key_exists('cleanup', $this->config) && $this->config['cleanup'] === false;
-        if ($skipCleanup || $this->ormModule->_getConfig('cleanup')) {
+        if ($skipCleanup || $this->ormModule->_getConfig('cleanup') === false) {
             return;
         }
         $this->factoryMuffin->deleteSaved();


yesdevnull:
Error from SemaphoreCI:

```
There was 1 error:

---------
1) WebDriverTest: Changing capabilities
 Test  tests/web/WebDriverTest.php:testChangingCapabilities
                                                                                                                                                                                               
  [TypeError] Argument 2 passed to PHPUnit\Framework\Assert::assertTrue() must be of the type string, null given, called in /home/runner/Codeception/tests/web/WebDriverTest.php on line 1119  
                                                                                                                                                                                               
#1  /home/runner/Codeception/tests/web/WebDriverTest.php:1119
#2  /home/runner/Codeception/codecept:42
```

[Appears to be completely unrelated to my change](https://github.com/Codeception/Codeception/blob/2.4/tests/web/WebDriverTest.php#L1119).

DavertMik:
> Correct me if I'm wrong but the DataFactory module should skip the cleanup if the ORM module (of which it depends on) has cleanup: false, right?

No. 

Here is the case: ORM doesn't wrap test in a transaction (cleanup: false) but we want DataFactory to remove its values it inserted for a test (cleanup: true). So if we want to specify that DataFactory values should be cleaned no matter what - here is our option.

yesdevnull:
Gotcha - that makes sense.  Thanks @DavertMik!

PR #4960, opened 2018-05-03
Handle absolute paths in ParamsLoader properly

SilverFire:


DavertMik:
We will need to write a use case for this feature. 
So I'm merging and could you post your case to Cookbook http://phptest.club/c/codeception/cookbook

PR #4961, opened 2018-05-03
Respect absolute paths in Yii2 module

SilverFire:


Naktibalda:
I wonder if this is a bugfix or a breaking change. It could be the latter.

@Slamdunk what do you think about this change?

SilverFire:
It's partially a breaking change because it will (obviously) break paths that start with a `/`, but it was never advised to use absolute paths as relative.

samdark:
Yes. That sounds fine to me.

DavertMik:
This also won't work for Windows absolute paths...

SilverFire:
> This also won't work for Windows absolute paths...

Good point. Maybe we should introduce a function on top of `codecept_root_dir()` that will check whether the path is absolute?

```php
function codecept_resolve_path($path) {
    return $path[0] === DIRECTORY_SEPARATOR ? $path : codecept_root_dir($path);
}
```

DavertMik:
@SilverFire good idea

SilverFire:
Added function and changelog. Ready for review.

DavertMik:
Looks good!

bscheshirwork:
```php
function codecept_resolve_path($path) {
    return $path[0] === DIRECTORY_SEPARATOR ? $path : codecept_root_dir($path);
}
```
This function return a different result in cases (in yii2-app-advanced for example)
* global `codeception.yml` within includes (
```yml
include:
    - common
    - frontend
    - backend
```
)
* from frontend/backend/common folder directly

for relative path. And this function proposed a way to avoid it. I am right?

SilverFire:
Yes, If I have correctly understood your question.

Issue #3849, opened 2016-12-19
[Laravel 5.3] Can't use factory states with codeception


vesol:
#### What are you trying to achieve?

There is no possibility to use factory states: [example](https://laravel.com/docs/5.3/database-testing#factory-states) with $i->have function. I requesting for such functionality

torkiljohnsen:
👍

AdrianSkierniewski:
Hi,

I'm using factories with Codeception quite some time. I'm just using them in my helpers (example **[here](https://github.com/GrupaZero/cms/blob/master/tests/_support/Helper/Functional.php#L57)**), but I encountered a problem with some memory leaks from Faker library. 

If we want to support factories we'd need to re-bind Faker & EloquentFactory in `_after` hook, to remove all references to faker between tests, so garbage collector can do his job.

Something like that:
```
            $this->app->singleton(FakerGenerator::class, function ($app) {
                return FakerFactory::create($app['config']->get('app.faker_locale', 'en_US'));
            });

            $this->app->singleton(EloquentFactory::class, function ($app) {
                return EloquentFactory::construct(
                    $app->make(FakerGenerator::class), $this->app->databasePath('factories')
                );
            });
```

I know that this isn't perfect but Laravel doesn't have build in method to run it on application. They're doing this **[here](https://github.com/laravel/framework/blob/5.6/src/Illuminate/Database/DatabaseServiceProvider.php#L39)**

PR #4971, opened 2018-05-15
Fix memory leak when using Laravel factories inside Codeception

AdrianSkierniewski:
Hi,

Recently I started noticing out of memory problem during Codeception run in my Laravel 5 project. After some profiling & investigation, I found out that Faker library used in Laravel factories causing some of those problems.
For now Laravel module doesn't support factories yet, but if we plan to do that we'd probably need to take care of this memory leak. Here is a ticket for it - #3849 

I expect this PR to be low risk because I'm just removing something from Service Container that is initialized one more time in next request.

I made some manual tests in my project on 96 tests, most of them are using factories:

Before:
```
Time: 12.31 seconds, Memory: 404.00MB

OK (96 tests, 380 assertions)
```
After:
```
Time: 13.43 seconds, Memory: 134.00MB

OK (96 tests, 380 assertions)
```

### DIFF
--- a/src/Codeception/Module/Laravel5.php
+++ b/src/Codeception/Module/Laravel5.php
@@ -217,6 +217,10 @@ class Laravel5 extends Framework implements ActiveRecord, PartedModule
                     $connection->disconnect();
                 }
             }
+
+            // Remove references to Faker in factories to prevent memory leak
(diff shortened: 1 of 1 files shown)


DavertMik:
Thanks, memory leaks may happen in functional tests. Thanks for detecting and fixing.

PR #4981, opened 2018-05-18
Stub 2.0

Naktibalda:
I released Stub 2.0, because the change is incompatible with Codeception 2.3


DavertMik:
Builds are passing. Merging!

PR #4984, opened 2018-05-21
Features: Allow PRESETs for codeception and suite config files

johannesschobel:
Hey @DavertMik , @Naktibalda and whoever it may concern,

### TL;DR
This PR adds a `presetfile` config tag to the `codeception.yml` as well as the `*.suites.yml` files. This tag specifies a file that holds basic configurations

### What is a PRESET File?
In my application, I have a lot of `modules`. I successfully used the `one runner for multiple applications` approach described in the docs. However, I came upon the issue, that i need to specify the same things over and over again (e.g., use laravel, define the main uri, and so on).

I thought it would be a good idea to just have some kind of `preset`, that would allow you to "include" yml snippets over and over again.. 

So, here it is ;)

#### Codeception.yml File
In your `codeception.yml` file of one single module you can use it like follows:
```yml
namespace: App\Modules\X\Tests\
presetfile: ../../../tests/_shared/preset.codeception.yml
```
As you can see, the whole `paths` stuff is not present any more, however, it is defined in the `preset.codeception.yml` file.. Respective file, in turn, looks like this:
```yml
paths:
    tests: tests
    output: tests/_output
    data: tests/_data
    support: tests/_support
    envs: tests/_envs
actor_suffix: Tester
extensions:
    enabled:
```

#### x.suite.yml Files
The same applies to the `x.suite.yml` files.. For example, i needed to constantly define the same modules to be enabled.. A `functional.suite.yml` file, in turn, now looks like this:
```yml
actor: FunctionalTester
presetfile: ../../../../tests/_shared/preset.functional.suite.yml
modules:
    enabled:
        - \App\Modules\X\Tests\Helper\Functional
```

Again, the linked `preset.functional.suite.yml` file is shown below and can be "included" in any other file:
```yml
modules:
    enabled:
      - Asserts
      - Laravel5:
            root: ../../../../
            part: ORM
            run_database_migrations: true
            database_seeder_class: DatabaseSeeder
      - REST:
            depends: Laravel5
```

As you can see, the `preset` files are written in `yml` and may be easily merged.. In fact, this is already provided by the PR.. For example, the `functional.suite.yml` that is provided for the testing environment, contains the enabled modules `Asserts`, `Laravel5`, `REST` **and** `App\Modules\X\Tests\Helper\Functional`

If you have any questions, please message me!
Otherwise, i would be happy to see this PR merged ;)

Kind regards and cheers
Johannes


johannesschobel:
Additional Information:
@SilverFire and me already started discussing this issue in #4868 ... This PR will also affect other issues, i believe..

DavertMik:
Wow, looks awesome! I might start using it soon.

The only concern I have is the name: `presetfile` is the new term which adds some complexity to understanding. Why not to call this config options just: `extends`?

```
actor: FunctionalTester
extends: ../../../../tests/_shared/preset.functional.suite.yml
modules:
    enabled:
        - \App\Modules\X\Tests\Helper\Functional
```

johannesschobel:
@DavertMik done.. checks need to run, but then you can merge :)

DavertMik:
Awesome, thank you!

johannesschobel:
Thank you very much for merging.. :+1: I strongly believe that this PR works towards making this framework even more awesome :)

PR #4986, opened 2018-05-23
Fix error on single file test: headers already sent

mikbox74:
Notice: Undefined index: tests in Codeception\Command\Run.php on line 389


DavertMik:
Thank you!

Issue #4994, opened 2018-05-28
Configure the data provider's maximum output length


mlambley:
https://github.com/Codeception/Codeception/blob/bc8433e3da75bc28c47df6f3afc22b9d4f65c0b4/src/Codeception/Test/Cest.php#L53

I would like to be able to print out the entire data provider arguments, regardless of size. But that line of code is hardcoding it to 100. Could this be customised via one of the settings files?

Thank you.

Codeception: 2.4

PR #4995, opened 2018-05-28
Configurable output limit for data provider

mlambley:
Closes https://github.com/Codeception/Codeception/issues/4994

Should not break anything - it will use the current default value if the new settings are not specified.

Any feedback will be appreciated. Thank you.

### DIFF
--- a/docs/reference/Configuration.md
+++ b/docs/reference/Configuration.md
@@ -134,6 +134,7 @@ modules:
 
 ```
 
+* `data_provider_output_max`: maximum string size when printing data provider information when running cests. Default is 100.
 * `extends`: allows you to specify a file (relative to the `*.suite.yml` file) that holds some already pre-defined values. This can be used to always use the same configuration for modules or whatever.
 * `namespace`: default namespace of actor, support classes and tests.
 * `suite_namespace`: default namespace for new tests of this suite (ignores `namespace` option)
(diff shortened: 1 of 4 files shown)


DavertMik:
Sorry I have some concerns about it.

1. You add a new config value which most of the users won't be aware of...
2. You change the signature to core classes.

Maybe there should be some sane defaults instead of configuration option which no one would reconfigure?

mlambley:
Putty's line length is around 200 but cmd.exe is around 100. I don't think that there can be a sane default.

DavertMik:
Can this be done via environment variable then?

mlambley:
I'm not certain how your environment variables work, but looking at:

https://codeception.com/docs/06-ModulesAndHelpers#Dynamic-Configuration-With-Parameters
https://codeception.com/docs/07-AdvancedUsage#Environments

I see that they allow you to group different configurations and specify which to use, either in the CLI or in the code. However, in both instances they need to be configuration options in the first place.

Could you give me more information regarding your suggestion that the line length be changed via an environment variable? Thanks.

edno:
Maybe configuration as part of the [parameters for `run` command line](https://codeception.com/docs/reference/Commands#Run) .
Something like: `--output-limit=200`

Naktibalda:
This PR died from the old age.

PR #5011, opened 2018-06-08
WIP: Adding Ability To Specify More Than One Dump File For The Db Module

Fenikkusu:
I still need to update the testing on this. Posting for any comments. 

Currently the Db module only allows you to specify one file for the dump file. I personally was running across two issues with this. The first is that the software requires a base set of records to work. If These are put in the default dump file, then it has to be re added anytime someone updates the dump file. Very tedious. The second issue was that some of the base set of records have changes that need to be made to run additional checks/testing. With one dump file, this isn't possible without specifying a new copy of the entire db to change one row. This of course is overkill.

This PR adds the functionality to specify more than one file as the dump file:

```yaml
modules:
    enabled:
        - Db:
            dsn: '%DSN%'
            host: '%HOST%'
            dump:
                - ./tests/_data/schema.sql
                - ./tests/_data/dumps/base.sql
```

This change can also be used within environments:
```yaml
env:
    owner:
        modules:
            config:
                Db:
                    dump:
                        - ./tests/_data/schema.sql
                        - ./tests/_data/dumps/base.sql
                        - ./tests/_data/dumps/owner.sql
    employee:
        modules:
            config:
                Db:
                    dump:
                        - ./tests/_data/schema.sql
                        - ./tests/_data/dumps/base.sql
                        - ./tests/_data/dumps/employee.sql
```

I also updated the DbPopulator to be able to handle the multi dumping. If you have more than one dump file, it will call the popular command for each dump file. The idea of course being that you use the '$dump' syntax in the populator.

### DIFF
--- a/src/Codeception/Lib/DbPopulator.php
+++ b/src/Codeception/Lib/DbPopulator.php
@@ -8,13 +8,6 @@ namespace Codeception\Lib;
 class DbPopulator
 {
     /**
-     * The command to be executed.
-     *
-     * @var string
-     */
-    private $builtCommand;
-
-    /**
      * @var array
      */
     protected $config;
@@ -29,8 +22,6 @@ class DbPopulator
     public function __construct($config)
     {
         $this->config = $config;
-        $command = $this->config['populator'];
-        $this->builtCommand = $this->buildCommand((string) $command);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


DavertMik:
I like the idea. Please proceed with implementation.
Could you fix current tests and if you have time maybe you could add new tests for this feature?

Fenikkusu:
> Could you fix current tests and if you have time maybe you could add new tests for this feature?

I plan to. I just haven't yet. I did this on the fly for the project I needed it on and haven't had the chance to work on updating the testing.

PR #5013, opened 2018-06-09
Update shim.php

josephzidell:
Fix indentation [skip ci]

### DIFF
--- a/shim.php
+++ b/shim.php
@@ -6,7 +6,7 @@ namespace {
 }
 
 namespace Symfony\Component\CssSelector {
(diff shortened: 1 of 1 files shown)


ddinchev:
Imho if we are going to modify indentations, these should be imposed and checked by some coding standards tool like CodeSniffer.

josephzidell:
I agree

DavertMik:
Yeah, shim.php outstands from our main code base so we don't have any strict policy about it's code. 
But thanks for improving it.

josephzidell:
Thanks @DavertMik

Issue #5014, opened 2018-06-10
Add new option to Recorder extension to disable recording of specified steps


sspat:
#### What are you trying to achieve?

Exclude some steps from the Recorder extension generated pages.
For example, when using the DataFactory module, the test record gets polluted by a lot of `have` and `haveMultiple` steps that do not produce any change on the tested page and their according screenshots are all the same.

#### What do you get instead?

Cannot do this currently by simply configuring the Recorder extension.

#### Proposal

Add a configuration option to Recorder extension to allow ignoring specified steps by their name.

```yaml
extensions:
    enabled:
        - Codeception\Extension\Recorder:
            module: WebDriver
            ignore_steps:
              - have
              - haveMultiple
```

reinholdfuereder:
@DavertMik or @Naktibalda Can this issue actually be resolved by now? (Cf. "Merged")

PR #5015, opened 2018-06-10
Added new option to Recorder extension to disable recording of specified steps

sspat:
Implementation for proposed feature described in #5014


DavertMik:
Nice! How about adding wildcard ignores? LIke 

```
ignore_steps:
  - grab*
  - wait*
  - see*
  - have*
```

which would be very useful as well

DavertMik:
The implementation looks nice. My bad I didn't include it in the last release 😖. But I hope the next one will go pretty soon.

PR #5022, opened 2018-06-13
Bug fix with multiple kernels

alefcastelo:
It's not possible to set a Kernel different of the AppKernel, and I have
two kernels an Admin and an Api, and I need use ApiKernel to run tests.

This commit makes a checking if kernel defined exists in the path.
Because the previous version it gets the first kernel in my case it gets
AdminKernel and always returns Kernel not found, because the Kernel defined is different of the loaded kernel 

http://symfony.com/doc/3.4/configuration/multiple_kernels.html


DavertMik:
Thanks!

PR #5029, opened 2018-06-20
adding new assert methods

johannesschobel:
Adding new `assertXXX` methods..
The methods check for `NotTrue / NotFalse`, e.g., if you want to check, if a value is `not true`.. A simple `assertFalse()` may not suitable in this case, because the value also may be `null` or whatever..

The new `assert` methods from PHPUnit allow for the latter!

I also added Tests for the new assertions

### DIFF
--- a/src/Codeception/Module/Asserts.php
+++ b/src/Codeception/Module/Asserts.php
@@ -264,6 +264,17 @@ class Asserts extends CodeceptionModule
     }
 
     /**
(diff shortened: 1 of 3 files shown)


johannesschobel:
@DavertMik pretty sure both build failures are not related to my PR :disappointed:

DavertMik:
Yes, thanks

johannesschobel:
You are awesome, thanks

PR #5030, opened 2018-06-20
new REST / InnerBrowser Asserts

johannesschobel:
This PR adds new assertions to check, if a `HttpStatusCode` is between a certain range (e.g., `4xx`) and provides "wrappers" for common blocks..

For example:
- `seeStatusCodeIsSuccessful()` checks for `2xx`
- `seeStatusCodeIsClientError()` checks for `4xx`
and so on..

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -1577,6 +1577,24 @@ class InnerBrowser extends Module implements Web, PageSourceSaver, ElementLocato
     }
 
     /**
+     * Checks that response code is between a certain range. Between actually means [from <= CODE <= to]
+     *
+     * @param $from
+     * @param $to
+     */
+    public function seeResponseCodeIsBetween($from, $to)
+    {
+        $failureMessage = sprintf(
+            'Expected HTTP Status Code between %s and %s. Actual Status Code: %s',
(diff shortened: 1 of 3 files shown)


DavertMik:
Thanks. Sorry for CI, it is failing for external project so it should not bother you right now.
But can you implement a test for those methods? That should be useful

johannesschobel:
Dear @DavertMik ,

i would love to add a test, however, I was not sure where / how to implement it. I found a `RestTest` (or something like this) class that contains tests for the `REST` module. However, I could not get it to work locally on my dev machine. Further, the endpoints most likely only return `200` or `404` ;)

Do you have an idea how to properly test it? I found a page that returns specific http status codes ( https://httpstat.us/ ) . Maybe i can use this to "fake" some codes and properly test them!? What do you think?

johannesschobel:
ok, sorry.. actually i was quite dumb.. added some tests for the methods as requested..

DavertMik:
Awesome! Thank you!

johannesschobel:
Hey @DavertMik , thanks a lot for merging..
Just a quick question: Are you interested in additional `assert` functions for specific modules? In this case i could / would love to add additional features.. Just drop me a line if you are interested..

All the best,
Johannes

Issue #5012, opened 2018-06-09
Unit tests without DB


josephzidell:
#### What are you trying to achieve?
Better performance for certain unit tests that do not need the database. 

#### What do you get instead?
The Db module either cleans up every time or never, depending on the configuration.

#### PR
The idea is simple and requires a one-line PR that I'd be happy to make.

Test:
```php
	/**
	 * @disableDbCleanup
	 *
	 * @return void
	 */
	public function testTextInputBasic() {...}
```

src/Codeception/Module/Db.php
```diff
     public function _before(TestInterface $test)
     {
         if ($this->config['reconnect']) {
             $this->connect();
         }
-        if ($this->config['cleanup'] && !$this->populated) {
+        if ($this->config['cleanup'] && !$this->populated && empty($test->getMetadata()->getParam('disableDbCleanup'))) {
             $this->_cleanup();
             $this->_loadDump();
         }
         parent::_before($test);
     }
```

If people like this idea, we can extend it by allowing it on the class with another one-liner.

josephzidell:
Perhaps a clearer annotation would be `@skipDbCleanup`

johannesschobel:
can you provide a PR for this feature? I think, this may be useful in certain situations..
I would, however, go for a more cleaner / meaningful annotation, as you already suggested..

The only issue i see, however, is, that this feature **highly** depends on the ORDER how tests are executed. I can imagine, that you would apply this annotation in order to better structure your tests (and make them incremental). However, if you "insert" a test inbetween (or the execution of the runner changes!!!!) this may become a **huge** problem..

What is your opinion on this?!

josephzidell:
@johannesschobel I'll get the PR.
My understanding of `_before()` is that it's run for each test method, thereby evaluating this condition anew. Am I wrong?

johannesschobel:
Yes.. However if you don't clean the DB before the test, your test may be wrong..

Consider the following example..
You have a test for registering a new user, that passes.. next you execute a test that returns all users. This test is annotated with the introduced `skipDB`. The `count` of the returned users would be `n+1` (because of the create it test that was executed before)..

If you revert the test order, the `count` would be `n`..

josephzidell:
I may have explained the intended usage wrong. The purpose is for tests that don't use the DB. For example, how would you test pure PHP code that doesn't use SQL?

johannesschobel:
Aaaaaah.. now I get it..
Then, however, I think, the annotation name is misleading..

[...]

PR #5033, opened 2018-06-20
Add annotation `@skipDbCleanup`

josephzidell:
Add optional annotation `@skipDbCleanup` on a test method to hand-pick certain tests that don't care about DB cleanup. This allows most tests to use cleanup. This would allow a per-test override of config `cleanup: true`

Implements #5012


johannesschobel:
Can you add tests to show that it actually works?!

josephzidell:
I could try. Never written tests for codeception before

johannesschobel:
Then now is the time I guess.. :D the is a "how to" in the wiki!

DavertMik:
Very good initiative. The only concern here is how to make it documented. 
Also the same effect can be achieved via https://codeception.com/docs/06-ModulesAndHelpers#Runtime-Configuration-of-a-Test

(so yeah, documentation is a thing. If it's hard to find you will never know how to implement things)

josephzidell:
@DavertMik Your approach is much better and more generic, and will solve many more use-cases
@sevastyanovio thanks for implementation.

Let me know whether I can be of assistance here.

josephzidell:
Can annotations be placed at the class level? I think both class- and method- level would be useful.

josephzidell:
@sevastyanovio Were you able to take a crack at an implementation?

Naktibalda:
This PR died from the old age after the target branch was deleted.
If you want to try again, please make a new pull request to 4.0 branch (or newer if things moved on again,  changes affecting Db module must be made to https://github.com/Codeception/module-db

PR #5035, opened 2018-06-21
minor doc updates

johannesschobel:
improving docs


DavertMik:
Thanks!
The only thing that module docs are generated so you should have updated docblocks of corresponding classes. But everything else is nice so I'm merging

PR #5037, opened 2018-06-22
Unicode

lovelock:
remove useless json_encode and add JSON_UNESCAPE_UNICODE as an option

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name":"codeception/codeception",
+    "name":"lovelock/codeception",
     "description":"BDD-style testing framework",
     "keywords":["BDD", "acceptance testing", "functional testing", "unit testing", "tdd"],
     "homepage":"http://codeception.com/",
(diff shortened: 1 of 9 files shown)


DavertMik:
Looks ok but I'm pretty afraid to accept this as it can break some output for some systems...
Removing json_encode can change the debug output which is not automated tested.

lovelock:
yeah, I know it may cause some chaos in systems that do not support unicode,but in most cases, it would be better to output non-escaped unicode characters as they don't have to decode specific words manually. I would believe the absense of json_decode options is because the project was started before PHP5.3 and options were not available at that time.

On the other hand, removing json_encode is because the 2nd argument is encoded already and handled in both ways.

Anyway thanks for your reply

DavertMik:
Looks like you updated the PR for your own fork. 
Also I'm not sure if you are actually using SOAP, Laravel, Phalcon to test your changes...
So I'd rather close it for now

PR #5040, opened 2018-06-25
New REST / InnerBrowser Helper

johannesschobel:
Hi @DavertMik ,

This PR adds a new "helper" that allows loading custom `Stub`s from the `_data` directory of the current suite. This may be useful in the context of RESTful API testing (e.g., you can define the JSON documents to be sent to the API).

Further, you can define "placeholders" in the stub files and automatically replace them..

Consider the following example:
**STUB FILE:**
``` json 
{
  "data" : {
    "title" : "{{book-title}}",
    "author" : "{{book-author}}",
    "year" : "{{book-year}}"
  }
}
```

In the `REST` module you can call it like so:
``` php
$replacements = [
    'book-author' => 'JRR Tolkien',
    'book-title' => 'Lord of the Rings',
    'book-year' => 1954,
];

// $I is an Instance of an Tester that includes the REST module!
$stub = $I->loadAndPrepareStub('path/to/stub.json', $replacements);
```

The content of `$stub`, in turn, will then be:
``` json
{
  "data" : {
    "title" : "Lord of the Rings",
    "author" : "JRR Tolkien",
    "year" : "1954"
  }
}
```

I think, this will help to further improve API testing by separating.. Can you give me feedback on this?!
Will love to contribute again, if you agree :laughing: 

All the best

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -210,6 +210,24 @@ class InnerBrowser extends Module implements Web, PageSourceSaver, ElementLocato
     }
 
     /**
+     * Load a given file from the "data" directory defined in the suite
+     *
+     * @param $filename
+     *
+     * @return string
+     */
+    public function _loadFileFromDataDirectory($filename)
+    {
+        // get the absolute path on disk for this codeception instance
+        $filePath = codecept_data_dir() . $filename;
+
+        $this->checkIfFileExists($filePath);
+
+        $stub = file_get_contents($filePath);
+        return $stub;
+    }
+
+    /**
      * Opens a page with arbitrary request parameters.
      * Useful for testing multi-step forms on a specific step.
      *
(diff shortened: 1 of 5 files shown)


DavertMik:
Sorry, it seems that this update adds an additional logic to InnerBrowser which goes beyond its scope. 
That's a breaking of single responsibility principle. How about making a new mudle for that purpose and publishing it as standalone package?

johannesschobel:
Hey @DavertMik , thank you very much for your response on this..

I initially added it to the `InnerBrowser` as this module already had a `attachFile()` method..

I have 3 suggestions how to proceed. Which one do you prefer?!
1. Create a `FileLoaderTrait`, add the latter to the `InnerBrowser` and move the functionality to this Trait
2. Move the functionality to the `Filesystem` module
3. Create a dedicated `FileLoader` module and move it there.. 

What do you think would be best?
Cheers and thanks for you time and effort

DavertMik:
I think the best option would be to create a standalone helper inside a custom package. 
Currently, I don't think this functionality belongs to any of core modules.

johannesschobel:
Hey man, thanks again for your input on this..
Just to be sure, I got your point correct:

You would do something like this:
``` php
namespace Codeception\Module;

class FileLoader extends CodeceptionModule implements X, Y, Z {
// ...
}
```
right?

If you like, I can develop such a new module. What do you think?

Cheers

PR #5046, opened 2018-06-28
New StubWorker Module

johannesschobel:
This PR enhances the prior PR #5040 .
More precisely, it introduces a new `StubWorker` module that holds this specific functionality

Maybe @DavertMik can give feedback on this?
All the best!
Johannes 

PS: wow, that was very easy to write a new module.. props to you, guys.. this framework is awesome!!!

### DIFF
--- a/src/Codeception/Module/StubWorker.php
+++ b/src/Codeception/Module/StubWorker.php
@@ -1,0 +1,148 @@
+<?php
+namespace Codeception\Module;
+
+use Codeception\Lib\Interfaces\ConflictsWithModule;
+use Codeception\Module as CodeceptionModule;
+use Codeception\Lib\Interfaces\DependsOnModule;
+use Codeception\Lib\Interfaces\PartedModule;
+
+/**
+ * Module for loading and preparing Stubs and other files
+ *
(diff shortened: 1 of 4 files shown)


DavertMik:
Thanks but unfortunately I can't accept this PR.

My only concern is: I don't think we can take this module to the core. I'd like not to take responsibility to maintain and support it as I don't personally see it very useful. But we have a directory of community-driven modules: https://codeception.com/addons. And it's really easy to add one to the list! 

Also, the name is not very good. I was thinking this is module is aimed to stub WebWorkers or ServiceWorkers but no... So FileDataLoader or smth like this can be a better name... But anyway, that should be published as standalone package on packagist.

Also, please join our Slack https://join.slack.com/t/codeception/shared_invite/enQtMzE2MzgxNzM1OTUzLTcwMmMxZjMxYTdkMzljMzNjNmNiNWQ0NGNjNTY0MjJlNWNjOTI0ZWU0Mjg4YmE5NTI0MmMwNDZmNzRhOTZmNWE
Ping me there to ask questions and receive responses faster.

johannesschobel:
thank you very much - i will do so :+1: 
I joined the slack channel, so maybe we could have a chat there, if i have some questions / ideas in mind.. 

All the best

PR #5059, opened 2018-07-05
Symfony test email count

### LINKED ISSUE #5058: Testing email in Symfony
I would like to have more fine grained controls over testing sent emails in the Symfony module.

#### What do you get instead?
We can only test if any email was sent. I would like to be able to check:
* if no email was sent
* number of emails sent

mbohal:
This should resolve #5058.

I did not modify nor test nor docs.
The test for this I could not find. The [contributing page](https://github.com/Codeception/Codeception/blob/2.4/CONTRIBUTING.md) says that docs are autogenerated from dockblocks.

### DIFF
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -446,7 +446,13 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
 
     /**
      * Checks if the desired number of emails was sent.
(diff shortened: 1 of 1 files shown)


DavertMik:
Thanks! You did everything right.

To test this method you need to update another repo https://github.com/Codeception/symfony-demo 
So you will need to introduce emails to symfony app and add a test for it. This is an optional task but it can improve stability of Symfony module

PR #5060, opened 2018-07-05
Typo fix in phpdoc

sasha-ch:
Wasn't work as-is.

+ May be add snippet 
  ``` yaml
  extensions:
      enabled:
          - Codeception\Extension\Recorder:
              module: AngularJS # enable for Angular
              delete_successful: false # keep screenshots of successful tests
  ```

to some file in /docs/ ?


DavertMik:
> to some file in /docs/ ?

Yes, docs will be auto-updated from your patch. Thanks!

PR #5065, opened 2018-07-09
match title in input fields

gimler:
Make it possible to click on a input by title. For example when you have a form input with only a font icon.

```
<input type="submit" class="btn btn-success" value="&#xf011;" title="Login" />
```

<a class="jsbin-embed" href="http://jsbin.com/bevotayuga/1/embed?html,css,output">JS Bin on jsbin.com</a>

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -989,7 +989,7 @@ class WebDriver extends CodeceptionModule implements
      *
      * ```
      * @api
-     * @param $page WebDriver instance or an element to search within
+     * @param RemoteWebDriver $page WebDriver instance or an element to search within
      * @param $link a link text or locator to click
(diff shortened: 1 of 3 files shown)


gimler:
@DavertMik we cannot change the application for that ;(

title is a global attribute see
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title

DavertMik:
> title is a global attribute see

Ah, ok. Then could you update the same XPath for InnerBrowser for internal compatibility?
Then I'm merging.

gimler:
@DavertMik i am not shure where to do this. i think `InnerBrowser.php::369` is the correct place but `selectButton()`is a method from symfony/dom-crawler.

DavertMik:
Oh, yes. Looks like it's impossible then (
Ok, let's have it this way.

PR #5068, opened 2018-07-10
always disconnect before connect when reconnect is set

ashnazg:
(edit: removed the 'is already connected' check behavior)

### DIFF
--- a/src/Codeception/Module/Db.php
+++ b/src/Codeception/Module/Db.php
@@ -278,8 +278,12 @@ class Db extends CodeceptionModule implements DbInterface
 
     private function connect()
     {
+        if ($this->dbh instanceof \PDO) {
+            return;
+        }
+
         $options = [];
- 
+
         /**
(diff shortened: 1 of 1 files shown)


Naktibalda:
What issue does this change solve?

ashnazg:
Small runtime optimization.  While working on #5069, I saw more runtime connect calls than I expected.  

Since I reasoned that the target db doesn't change during an execution (unlike browser changes via `env` customizations), it seemed unnecessary to connect again when a connection was already made. Again, this is when the `reconnect` option is *not* enabled.

However, **if** there is a viable use case of targeting multiple databases in the same config via `env`, then this change should *not* be accepted.

DavertMik:
Makes sense to me. 
However, a test may run for a long time and connection can be lost. Maybe some healthcheck queries should be sent instead?

DavertMik:
Thanks!

PR #5069, opened 2018-07-10
add config option for session lock wait setting

ashnazg:
(and Oracle implementation)

The point here is to avoid false negative test error-failures that might occur due to database object locks.  In particular, I'm seeing occasional `ORA-00054: resource busy and acquire with NOWAIT specified or timeout expired` occurrences during the before-test dump-populate step, when running a `--group failed` follow-up execution after a previous execution recorded errors.

In my case, I'm hoping having a 10-30sec lock wait (rather than the default no-wait of 0) might avoid these `ORA-00054`s.


ashnazg:
If others are familiar enough with other database drivers and can tell me the right statement to execute for those sessions, I can easily add the other implementations to this PR.

DavertMik:
Thanks. I'm not familiar with waitlock at all but I'm ok to merge PR. Maybe when this goes to release more people will be interested in implementing this for other drivers

PR #5072, opened 2018-07-13
More explicit PDO closing upon destruction

ashnazg:
For more graceful DB closing upon exceptions.


DavertMik:
Not sure where transactions are used in Db module...

ashnazg:
`inTransaction` check is a habit I've learned over the years for clean disconnects vs an exception during the disconnect attempt.  Maybe `Db` doesn't do any currently, but someone using the module directly might.

DavertMik:
Ok. Let's do it then!

PR #5074, opened 2018-07-13
Fixes a missing _ for _getResponseContent() in seeXmlResponseEquals

Voziv:
Simple fix I think. I'm 99.9% sure this was a typo when the phpunit wrapper was added.


DavertMik:
Thank you!

Issue #5081, opened 2018-07-18
file upload tests with Lumen module


retnek:
#### What are you trying to achieve?

I want to send a file to my endpoint in a functional tests through REST module (depends to Lumen).

#### What do you get instead?

I've got an unexpected validation error (Lumen's common file upload validation error)

```php
// in the target controller method
$this->validate($request, [
    'title' => 'required',
    // this file upload validation will be failed
    'list' => 'required|mimes:csv,txt'
]);
```

```php
$I->sendPOST('/any/endpoint', ['title' => 'anystring'], [
    'list' => new \Illuminate\Http\UploadedFile(
        codecept_data_dir('file.csv'),
        'file.csv',
        'text/csv',
        filesize(codecept_data_dir('file.csv')),
        UPLOAD_ERR_OK,
        true
    )
]);
// it will be 422 instead of 200
$I->seeResponseCodeIs(200);
```

### Details

* Codeception version: 2.4.4
* PHP Version: 7.1.5
* Operating System: Mac OS
* Installation type: Composer
* List of installed packages (`composer show`)
* Suite configuration:

```yml
actor: FunctionalTester
modules:
    enabled:
        - Lumen
        - REST:
            depends: Lumen
        - Filesystem
        - Asserts
```

I've sucessfully investigate this issue. The Codeception\Lib\Connector\Lumen does not contain that functions which are Codeception\Lib\Connector\Laravel5 has, which are required to sucessfully validate file uploads without HTTP. If I copy `filterFiles` and `convertToTestFiles` functions from Laravel5 connector into Lumen, the test will be success.

Naktibalda:
> If I copy filterFiles and convertToTestFiles functions from Laravel5 connector into Lumen, the test will be success.

Please make a pull request.

PR #5082, opened 2018-07-18
fix Lumen file upload via REST module

retnek:
for #5081

(there's no Lumen tests)

### DIFF
--- a/src/Codeception/Lib/Connector/Lumen.php
+++ b/src/Codeception/Lib/Connector/Lumen.php
@@ -9,6 +9,7 @@ use Illuminate\Support\Facades\Facade;
 use Symfony\Component\HttpFoundation\Request as SymfonyRequest;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Client;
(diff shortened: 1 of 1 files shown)


src/Codeception/Lib/Connector/Lumen.php
```diff
@@ -124,4 +124,43 @@ private function initialize($request = null)
...
+     */
+    private function convertToTestFiles(array $files)
+    {
+        $filtered = [];
+
+        foreach ($files as $key => $value) {
+            if (is_array($value)) {
+                $filtered[$key] = $this->convertToTestFiles($value);
+            } else {
+                $filtered[$key] = UploadedFile::createFromBase($value, true);
```
Naktibalda: It looks like copy-paste from Laravel5 module, but you missed `use Illuminate\Http\UploadedFile;` line at the top.
https://github.com/Codeception/Codeception/blob/2.4.4/src/Codeception/Lib/Connector/Laravel5.php#L10

retnek:
tested with my project and it's working

DavertMik:
Thanks!

Issue #4913, opened 2018-04-03
Function getMetadata() may not exist, results in fatal error


marcovtwout:
#### What are you trying to achieve?

Running test suite with html report.

#### What do you get instead?

```bash
PS S:\www\test\tests> ../vendor/bin/codecept run unit --html -vvv
Codeception PHP Testing Framework v2.4.1
Powered by PHPUnit 5.7.27 by Sebastian Bergmann and contributors.

Unit Tests (17) -----------------------------------------------------------------------------------------------------------------------------------
Modules: Asserts, \Helper\Unit
---------------------------------------------------------------------------------------------------------------------------------------------------

  Printing HTML report into report.html

In CDbTestCase.php line 78:

  [Exception]
  Unknown method 'getMetadata' for class 'UserIdentityTest'.


Exception trace:
 CDbTestCase->__call() at S:\www\test\vendor\codeception\codeception\src\Codeception\Test\Descriptor.php:38
 Codeception\Test\Descriptor::getTestSignatureUnique() at S:\www\test\vendor\codeception\phpunit-wrapper\src\ResultPrinter\HTML.php:263
 Codeception\PHPUnit\ResultPrinter\HTML->startTest() at S:\www\test\vendor\phpunit\phpunit\src\Framework\TestResult.php:386
 PHPUnit_Framework_TestResult->startTest() at S:\www\test\vendor\phpunit\phpunit\src\Framework\TestResult.php:628
 PHPUnit_Framework_TestResult->run() at S:\www\test\vendor\phpunit\phpunit\src\Framework\TestCase.php:868
 PHPUnit_Framework_TestCase->run() at S:\www\test\vendor\phpunit\phpunit\src\Framework\TestSuite.php:733
 PHPUnit_Framework_TestSuite->run() at S:\www\test\vendor\codeception\phpunit-wrapper\src\Runner.php:106
 Codeception\PHPUnit\Runner->doEnhancedRun() at S:\www\test\vendor\codeception\codeception\src\Codeception\SuiteManager.php:157
 Codeception\SuiteManager->run() at S:\www\test\vendor\codeception\codeception\src\Codeception\Codecept.php:189
 Codeception\Codecept->runSuite() at S:\www\test\vendor\codeception\codeception\src\Codeception\Codecept.php:158
 Codeception\Codecept->run() at S:\www\test\vendor\codeception\codeception\src\Codeception\Command\Run.php:466
[...]

Issue #4672, opened 2017-12-09
HTML Report: Failure snapshots are overwritten when using Cest with examples


Tenzian:
#### What are you trying to achieve?

Create a test using a Cest with examples to iterate the same test over several scenarios and generate an HTML report with meaningful snapshots of the failed pages.

#### What do you get instead?

Snapshot files for the failures in \_output are overwritten for each failing iteration of the examples. 

All the _HTML snapshot_ links in the report end up pointing to the same `FailsCest.staticPages.fail.html` file generated for the final failure.

#### Example Screenshots

HTML Report:
![screen shot 2017-12-09 at 00 30 56](https://user-images.githubusercontent.com/13963952/33790726-9dcbfbc8-dc7a-11e7-8fbc-1a460b734dea.png)
Snapshot:
![screen shot 2017-12-09 at 00 32 01](https://user-images.githubusercontent.com/13963952/33790728-a1f32960-dc7a-11e7-9a8f-d4da981559e8.png)

---

#### Suspected Cause

In several places the code assumes that each test has a unique signature.

However, when examples are used, all iterations of the Cest method have the same test signature which causes the snapshots for each failure to have the same filename.

You can see in `_failed` the filename for the page source is being created using `Descriptor::getTestSignature($test)` in the following examples:

https://github.com/Codeception/Codeception/blob/75a1f48f291323083caebed7d0244de96ff8c008/src/Codeception/Lib/InnerBrowser.php#L59-L68

https://github.com/Codeception/Codeception/blob/75a1f48f291323083caebed7d0244de96ff8c008/src/Codeception/Module/WebDriver.php#L499-L509

---

> Provide console output if related. Use `-vvv` mode for more details.

```bash
$ codecept run acceptance --html
Codeception PHP Testing Framework v2.3.6
Powered by PHPUnit 6.4.4 by Sebastian Bergmann and contributors.

Acceptance Tests (3) ---------------------------------------------------------------------------------------------------------------
✖ FailsCest: Static pages | "http://www.google.com","Giraffe" (2.43s)
✖ FailsCest: Static pages | "http://www.php.net","Elephant" (1.09s)
✖ FailsCest: Static pages | "http://codeception.com","Zebra" (1.23s)
------------------------------------------------------------------------------------------------------------------------------------


Time: 4.82 seconds, Memory: 10.00MB

There were 3 failures:

---------
1) FailsCest: Static pages | "http://www.google.com","Giraffe"
[...]

PR #5086, opened 2018-07-19
Fixes https://github.com/Codeception/Codeception/issues/4913

marcovtwout:
is_callable() cannot be used if $testCase implements __call().
instanceof TestInterface should be the correct check,

### DIFF
--- a/src/Codeception/Test/Descriptor.php
+++ b/src/Codeception/Test/Descriptor.php
@@ -34,7 +34,7 @@ class Descriptor
     {
         $example = null;
 
-        if ($testCase instanceof TestInterface
+        if (method_exists($testCase, 'getMetaData')
             && $example = $testCase->getMetadata()->getCurrent('example')
         ) {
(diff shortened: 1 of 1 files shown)


marcovtwout:
This change needs verification - I don't know for sure getMetaData() is only called on classes that implement TestInterface.

Tenzian:
This causes #4672 to start happening again.

Tenzian:
81c2269 fixes my issues with this change. 👍

DavertMik:
I think it's ok. Thank you

axboory:
Hello. These issues #4672 and Codeception/phpunit-wrapper#4 are still happening to me. I have to run 10 tests but for like 30 times. And in the report.html, the first 2 times I can see some Passed green tests but the next 28 times are all red even if they Passed and with the same snapshot.  Does anyone have other idea?

Naktibalda:
Please make comment in the issue and I will reopen it.

Issue #5049, opened 2018-06-29
Synchronous processes in the RunProcess extension


aashmelev:
I would like to improve the `RunProcess` extension.

In this extension, processes are started in the asynchronous mode, but there are times when it is necessary to start a process in synchronous mode. For example, we want to build files using webpack before running the tests.

I propose for this case to add some option that will indicate which processes should start in synchronous mode.

```
extensions:
    enabled:
        - Codeception\Extension\RunProcess:

            #this is a synchronous process
            0: yarn run build-dev

            #these processes will start after the previous process is completed
            1: chromedriver.exe --url-base=/wd/hub
            2: php -S 127.0.0.1:8888 -t tests/_app/public
            
            #synchronous process keys will be listed here
            sync: [0]
```

What do you think about it?

DavertMik:
I think that should go to another extension so we could simplify the overall configuration. 
Something like: `\Codeception\Extension\RunBefore`
If you make it as pull request I will merge it

PR #5089, opened 2018-07-20
Adding an extension to run processes before tests

aashmelev:
For [https://github.com/Codeception/Codeception/issues/5049](https://github.com/Codeception/Codeception/issues/5049)


DavertMik:
Thanks!

Issue #5092, opened 2018-07-23
Incompatible with PHPUnit ^7.2


Slamdunk:
#### What are you trying to achieve?

Installing Codeception 2.4 together with PHPUnit 7.2.

`Codeception/Stub` requires `phpunit/phpunit-mock-objects`, which is now merged in PHPUnit.

Pull request already opened in Stub: https://github.com/Codeception/Stub/pull/9

```
$ composer.phar require phpunit/phpunit:^7.2 codeception/codeception:^2.4
```

#### What do you get instead?

<details>
<summary>Click to show composer output</summary>
<pre><code>
./composer.json has been created
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Conclusion: don't install phpunit/phpunit 7.2.7
    - Conclusion: don't install phpunit/phpunit 7.2.6
    - Conclusion: don't install phpunit/phpunit 7.2.5
    - Conclusion: don't install phpunit/phpunit 7.2.4
    - Conclusion: don't install phpunit/phpunit 7.2.3
    - Conclusion: don't install phpunit/phpunit 7.2.2
    - Conclusion: don't install phpunit/phpunit 7.2.1
    - codeception/codeception 2.4.0 requires codeception/stub ^1.0 -> satisfiable by codeception/stub[1.0.0, 1.0.1, 1.0.2, 1.0.3, 1.0.4].
    - codeception/codeception 2.4.1 requires codeception/stub ^1.0 -> satisfiable by codeception/stub[1.0.0, 1.0.1, 1.0.2, 1.0.3, 1.0.4].
    - codeception/codeception 2.4.2 requires codeception/stub ^2.0 -> satisfiable by codeception/stub[2.0.0, 2.0.1].
    - codeception/codeception 2.4.3 requires codeception/stub ^2.0 -> satisfiable by codeception/stub[2.0.0, 2.0.1].
    - codeception/codeception 2.4.4 requires codeception/stub ^2.0 -> satisfiable by codeception/stub[2.0.0, 2.0.1].
[...]

PR #5094, opened 2018-07-24
Fixed fatal errors with PHP 7.2 and PHPUnit 7.2

SilverFire:
This pull request fixed methods declaration incompatibilites that lead to Fatal error with PHP 7.2 and PHPUnit 7.2

### DIFF
--- a/src/Codeception/Test/Unit.php
+++ b/src/Codeception/Test/Unit.php
@@ -130,7 +130,7 @@ class Unit extends \PHPUnit\Framework\TestCase implements
         ];
     }
 
-    public function getDependencies()
+    public function getDependencies(): array
     {
         $names = [];
         foreach ($this->getMetadata()->getDependencies() as $required) {
--- a/src/Codeception/Test/Cept.php
+++ b/src/Codeception/Test/Cept.php
@@ -50,7 +50,7 @@ class Cept extends Test implements Interfaces\Plain, Interfaces\ScenarioDriven, 
         return $this->getMetadata()->getName() . 'Cept';
     }
 
-    public function toString()
+    public function toString(): string
     {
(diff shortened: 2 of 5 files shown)


Naktibalda:
Please make it compatible with PHP 5.6

src/Codeception/Test/Unit.php
```diff
@@ -130,7 +130,7 @@ public function getReportFields()
         ];
     }
 
-    public function getDependencies()
```
Naktibalda: getDependencies is a more interesting case. Conflict wasn't cause by a change of signature.
Codeception had this method first, PhpUnit introduced it in 7.2.0, so fixing signature could cause interesting bugs.
https://github.com/sebastianbergmann/phpunit/commit/66a0765da30140eb9618b50cbeddd6223eccdf4a#diff-511c90242d5589e941582006e0485184R964

ddinchev: One solution could be to constrain codeception's phpunit version up to 7.1 until we drop the support of php `5.x`.

Naktibalda:
@DavertMik tried to fix phpunit signature compatibility issues by moving PhpUnit related code to [phpunit-wrapper](https://github.com/Codeception/phpunit-wrapper/).

It looks like we have to move test format classes to phpunit-wrapper too.
Version without result types goes to 6.0 branch, version with result types - to 7.0 branch.

Oh great, this change will make phpunit-wrapper incompatible with older versions of Codeception, so major releases will be necessary (8.0 and 9.0? 🤣 ).

Naktibalda:
Related issue: #5092

Slamdunk:
I think a BC break release isn't necessary: #5097

PR #5097, opened 2018-07-26
Add PHPUnit 7.2 compatibility

Slamdunk:
Fixes https://github.com/Codeception/Codeception/issues/5092 https://github.com/Codeception/Codeception/pull/5094

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
         "symfony/dom-crawler": ">=2.7 <5.0",
         "behat/gherkin": "^4.4.0",
         "codeception/phpunit-wrapper": "^6.0.9|^7.0.6",
-        "codeception/stub": "^2.0"
+        "codeception/stub": "dev-master"
     },
     "require-dev": {
         "monolog/monolog": "~1.8",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -30,7 +30,7 @@
         "symfony/dom-crawler": ">=2.7 <5.0",
         "behat/gherkin": "^4.4.0",
         "codeception/phpunit-wrapper": "^6.0.9|^7.0.6",
-        "codeception/stub": "^2.0"
+        "codeception/stub": "dev-master"
```
Slamdunk: This line can be dropped before the merge, but need to stay here for now to prove the goodness of the PR in Travis

Slamdunk:
The other signature changes are not required to fix the compatibility with PHPUnit 7.2, nor with PHP 7.2, and should only be addresses in the next major Codeception version.

Slamdunk:
The only Travis failure is a timeout one, not related to the PR.

DavertMik:
Looks like this PR is better to be pushed to phpunit-wrapper library which was specially moved out from Codeception to maintain compatibility with phpunit.

DavertMik:
Oh, I see, that can't be done via phpunit wrapper alone. But looks like we need just to change the method name in interface, because we didn't use the method of the same name from PHPUnit.

So this can be improved just by renaming the method

DavertMik:
Thanks! I replaced with my own PR.

PR #5101, opened 2018-07-27
Fixing edge-case issues with the recording extension

### LINKED ISSUE #5093: Recorder extension throws an exception if the first step in suite errors
## Description
I've configured my recorder extension to delete successful reports, and given there's an exception being thrown (in my case this exception (ImageDiviationException). That is coming from the externally installed VisualCeption module (which was a valid failure), in my case all the setup happens using my own home backed module (using _before), so the failing dontSeeVisualChanges assertion is the first step both in the Cest and in that particular Suite.

To recap; there's an error being thrown at the first "step" of the suite.

In this scenario, the recorder module would still try to generate a report for the Cest, and would fail to do so. It seems that "$this->dir" would always be hydrated with the previous suites' details, and upon the next step in the recorder module it would try to file_put_contents the report into that location, which was already being deleted as that suite has already finished hence the following exception would be thrown;

## Error
```In ErrorHandler.php line 83                                                                                                                                                          
  file_put_contents(/project/tests/_output/record_5b574feedf045_AuthenticationCest__Verify_that_the_test_user_can_logout/index.html): failed to open stream: No such file or directory                                                                                                                                                                               
```
This is being thrown at: https://github.com/Codeception/Codeception/blob/ba6bd27873392d15034a52d3ec64a19f82220cb5/ext/Recorder.php#L333

## Environment
* Version: Official (latest) docker image (2.4.4) sha256:0a5e7a0e23f014f7567e839ab803a9f41b247de76221fc15d15e99935c72accf

```yml
extensions:
    enabled:
        - Codeception\Extension\Recorder
    config:
        Codeception\Extension\Recorder:
            delete_successful: true
```

## Reproduce
Create two suites and enable the recorder extension with the delete_successful: true being configured.
Make the first suites cest/s to pass, and throw an exception on the first step of the next suite.

OneEyedSpaceFish:
Fixes #5093

### DIFF
--- a/ext/Recorder.php
+++ b/ext/Recorder.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace Codeception\Extension;
 
 use Codeception\Event\StepEvent;
@@ -55,7 +56,7 @@ class Recorder extends \Codeception\Extension
         'module'            => 'WebDriver',
         'template'          => null,
         'animate_slides'    => true,
-        'ignore_steps'      => []
+        'ignore_steps'      => [],
     ];
 
     protected $template = <<<EOF
@@ -232,6 +233,8 @@ EOF;
     protected $stepNum = 0;
     protected $seed;
     protected $recordedTests = [];
+    protected $errors = [];
+    protected $errorMessages = [];
 
     public function beforeSuite()
     {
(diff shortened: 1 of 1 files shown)


OneEyedSpaceFish:
Failed to take a screenshot, possibly the webdriver is unresponsive. A little explanation; it seems that if something in one of the _before hooks before the Recorders one throws an exception, then have a file_put_contents would throw an exception as it would try to save the report using the previous suites recording dir.

This should fix that, and it would attempt to generate a screenshot in this scenario called error.png which then would be saved in the corresponding recorder dir.

Also while I was debugging my changes, I noticed that sometimes if my webdriver dies, then the screenshot would not be taken, and the recorder extension would silently output this as a debug line. The issue with this is that the webdriver would also then output it's verbose logs so that debug line would be missed in a sea of webdriver logs.

This has been changed, so if the webdriver dies and if the extension can't take any screenshots then for each step the extension would output this in afterSuite.

And last - but not least; if there are no screenshots taken due to webdriver flakyness, then it won't generate any reports, and would mention this in afterSuite.

If you have any issues/suggestions, just let me know!

OneEyedSpaceFish:
Just a couple of things to sort out:
What is the intended behavior when we fail to capture a screenshot? Do we like to save this as a step, even to the recorder extension failed to record anything for it? What about if we fail to capture ANY screenshots for a test?

Just as a side-note, we might want to catch/handle exceptions from file_put_contents and mkdir as well (although it's not entirely related to this issue).

OneEyedSpaceFish:
Okay, so I've been thinking about this extension, I've noticed a couple of possible improvements that could be done if someone from the Codeception team could have a look if those are desired that would be appreciated. I could have a look at them in a week or two, and make a commit in the same fork to implement them:

1. handle file operation exceptions / log them without throwing exceptions
2. the ability to automatically purge old reports (from previous runs)
3. the ability to configure recorder not to generate pages if the screenshot fails
4. the ability to access main codeception command line arguments (to support 5 and 6)
5. the ability not to display any Unicode characters if ANSI only output is requested
6. the ability not to display any colors in output if no-colors is requested
7. the ability to set the output verbosity for the extension (with my current change it's outputting any errors it encounters)

DavertMik:
Thanks. So far I'm pretty happy with Recorder extension but it can be improved.

> handle file operation exceptions / log them without throwing exceptions

:+1: 

> the ability to automatically purge old reports (from previous runs)

:+1: 

> the ability to configure recorder not to generate pages if the screenshot fails

I think that should be an error page instead...

> the ability to access main codeception command line arguments (to support 5 and 6)

please look how it works for other extensions like SimpleReporter

> the ability not to display any Unicode characters if ANSI only output is requested

:+1: 

> the ability not to display any colors in output if no-colors is requested

:+1: 

> the ability to set the output verbosity for the extension (with my current change it's outputting any errors it encounters)

I think it's minor. So I wouldn't implement this.

Thanks for the current PR. I will include it into the next release

Naktibalda:
@OneEyedSpaceFish It looks like your change caused #5128

Could you help to investigate it?

PR #5104, opened 2018-07-31
fixed compat with phpunit 7.2

DavertMik:
Replaces #5097 

We have our own methods for dependencies collection so it's easier to rename ours than maintain compatibility with constantly changing PHPUnit signatures.

### DIFF
--- a/src/Codeception/Codecept.php
+++ b/src/Codeception/Codecept.php
@@ -7,7 +7,7 @@ use Symfony\Component\EventDispatcher\EventDispatcher;
 
 class Codecept
 {
-    const VERSION = "2.4.4";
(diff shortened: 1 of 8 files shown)


src/Codeception/Codecept.php
```diff
@@ -7,7 +7,7 @@
 
 class Codecept
 {
-    const VERSION = "2.4.4";
+    const VERSION = "2.4.5";
```
Slamdunk: Is it ok to change the version without tagging it?

DavertMik: I should have released it yesterday but I was very tired to do that. 
So I plan to make it right now

PR #5108, opened 2018-08-02
Clarify `grabFromDatabase` and `grabColumnFromDatabase`

bimlas:
The methods does not fetching "all values", but a single column.

It seems that `grabFromDatabase` and `grabColumnFromDatabase` doing the same:
I recomend to keep `grabFromDatabase` method but rename it to
`grabColumnFromDatabase` because it's more verbose.

I'm not sure that it's accepting comparison expressions or not, but if it
isn't, then the documentation about this should be affected in another PR (see
src/Codeception/Lib/Interfaces/Db.php:67).

### DIFF
--- a/docs/modules/Db.md
+++ b/docs/modules/Db.md
@@ -208,7 +208,7 @@ Supported operators: `<`, `>`, `>=`, `<=`, `!=`, `like`.
 
 ### grabColumnFromDatabase
  
-Fetches all values from the column in database.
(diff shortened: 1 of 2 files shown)


bimlas:
o_O OMG... *snap* Sorry, I'm a bit tired and I misread the original documentation... `(-_-')`

BTW both methods are doing the same, so code duplication should be removed.

PR #5110, opened 2018-08-02
Fixing recorder extension issues caused by phpunit 7.2.7 update

### LINKED ISSUE #5109: Recorder extension issue after phpunit 7.2.7 update
@DavertMik 
 
Since my new fix of #5093 I've been using a patch in a step of our docker build, and I just noticed that the new release is present, so I've removed this aforementioned patch and updated my codeception image.

But I've noticed that there's a strange issue occurring since the new release:
The following value will contain any uppercase characters that are present in the wantTo()
https://github.com/Codeception/Codeception/blob/0af1030fa35299241812301683d00d86d4712def/ext/Recorder.php#L303

And here only has the first letter is capitalized.
https://github.com/Codeception/Codeception/blob/0af1030fa35299241812301683d00d86d4712def/ext/Recorder.php#L330

I can only assume without diving into this right now, is that since the new PHPUnit 7.2.7 update, there must be some sort of string transformation to convert the characters to lowercase after the `_before()` hook is called but before the persist method is executed. Unfortunately, my check would then trip, and think that the first test step is not executed yet, and trigger the error handling mechanism.

One could argue that we should fix this inconsistency in the first place, but right now I wanted to fix this asap before any other Codeception users would encounter this issue themselves, so as a 'fix' for this issue, I'll just make my string comparison case-insensitive.

Docker image version: `codeception/codeception:2.4.5@sha256:0fae0163045541ce4d11e45e38395501ac7fe78ddbc62b09b1d6a0e43dcb63df`

My apologies for the incomplete report, I wanted to fix this issue asap, since this is live now.
I'm not too sure if it's worth creating another issue to fix the underlying bug in the first place.
I assume that it's worth keeping an eye out of any other strange behavior since the update, just in case some other mechanism trips on this issue just as my mechanism did.

OneEyedSpaceFish:
Fixes #5109

### DIFF
--- a/ext/Recorder.php
+++ b/ext/Recorder.php
@@ -245,6 +245,8 @@ EOF;
         }
         $this->seed = uniqid();
         $this->webDriverModule = $this->getModule($this->config['module']);
+        $this->errors = [];
+        $this->errorMessages = [];
         if (!$this->webDriverModule instanceof ScreenshotSaver) {
             throw new ExtensionException(
                 $this,
@@ -260,7 +262,7 @@ EOF;
 
     public function afterSuite()
     {
-        if (!$this->webDriverModule or !$this->dir) {
+        if (!$this->webDriverModule) {
             return;
         }
         $links = '';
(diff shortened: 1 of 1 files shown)


DavertMik:
Thanks!

PR #5124, opened 2018-08-17
Cleanup fixture by name

soupli:


DavertMik:
I think one day we will deprecate Fixtures as it's just a global storage.
But currently this improvement looks ok

src/Codeception/Util/Fixtures.php
```diff
@@ -32,9 +32,15 @@ public static function get($name)
...
+    public static function cleanup($name = null)
     {
-        self::$fixtures = [];
+        if ($name === null) {
+            self::$fixtures = [];
+        } elseif (!self::exists($name)) {
+            throw new \RuntimeException("$name not found in fixtures");
+        }
+
+        unset(self::$fixtures[$name]);
```
DavertMik: so this unsets value even when `self::$fixtures = [];` ? 
please use `return`
Also I don't think the exception should be thrown if a key is not there. It means that it already empty, so cleaned up

soupli: Sorry, late response. I'll update my merge request by the following:

```php
public static function cleanup($name = null) {
    if (self::exists($name)) {
        unset(self::$fixtures[$name]);
        return;
    }

     self::$fixtures = [];
}
```

Naktibalda:
The code was never updated, please make a new pull request to 4.0 branch.

soupli:
> The code was never updated, please make a new pull request to 4.0 branch.

Correct, i did not get any response about my "code review" so it fell into my oblivion
I will make an PR to 4.0, thanks

src/Codeception/Util/Fixtures.php
soupli: https://github.com/Codeception/Codeception/pull/5124#issuecomment-565353860

PR #5125, opened 2018-08-17
Fix errors in parallelRun() codeblock (docs)

bertramakers:
The opt() methods should be option(), and there was a syntax error (`unexpected ;` on line 290).


DavertMik:
Thanks! I think API was updated in some point and I didn't update the docs

PR #5136, opened 2018-08-23
Prevent NPE in Yii2 module

SilverFire:
There are some cases when `_after` method can be called without calling `_before`, that leads to null-pointer exception. For example, `_after` can be called from error handler during unhandled exception proceeding.


samdark:
Merged. Thanks!

samdark:
Merged. Thanks!

gijsstegehuis:
When will this be released?

PR #5137, opened 2018-08-24
[WebDriver] Keep coverage cookies in LoadSessionSnapshot

rajras:
As same in saveSessionSnapshot, at the moment of load the snapshot, I think we need to avoid the deletion of all cookies. we need to delete all of them except those related to COVERAGE.

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -3066,7 +3066,14 @@ class WebDriver extends CodeceptionModule implements
         if (!isset($this->sessionSnapshots[$name])) {
             return false;
         }
-        $this->webDriver->manage()->deleteAllCookies();
+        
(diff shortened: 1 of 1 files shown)


Naktibalda:
Has loadSessionSnapshot caused any issues to you? or is this fix just for consistency?

rajras:
Hi, yes, i was having troubles with the coverage when i use the save and
load session snapshot...

For example i have some test in cest format. In _before method I have a
call to a loadsessionsnapshot and if I run the test with coverage, only the
first one is recording the executed code. This is because when was calling
"deleteallcookies", the coverage cookie was loose...

I could provide the test code if you need further information...

Let me know..
Kind regards, and by the way, excellent work...

DavertMik:
Thanks. This makes sense and the fix is very clean

PR #5156, opened 2018-09-04
Update Asserts.php

Hexoplon:
assertEquals and assertNotEquals had the arguments mixed up in the examples. It should be $expected, $actual. It was $actual, $expected


Naktibalda:
thank you

PR #5160, opened 2018-09-05
New Extension for a "sticky" failed Group

mneuhaus:
Saves failed tests into tests/log/failed-sticky in order to rerun failed tests.
Tests will stay logged in this file until the tests passes.
This makes it possible to "collect" all failed tests in a Suite and fix them one by one with --fail-fast

### DIFF
--- a/ext/RunFailedStickyExtension
+++ b/ext/RunFailedStickyExtension
@@ -1,0 +1,89 @@
+<?php
+namespace Codeception\Extension;
+
+use Codeception\Event\PrintResultEvent;
+use Codeception\Events;
+use Codeception\Extension;
+use Codeception\Test\Descriptor;
+
+/**
+ * Saves failed tests into tests/log/failed-sticky in order to rerun failed tests.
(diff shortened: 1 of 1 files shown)


ext/RunFailedStickyExtension
```diff
@@ -0,0 +1,89 @@
...
+ * php codecept run -g failed-sticky
+ * ```
+ *
+ * To change failed group name add:
+ * ```
+ * --override "extensions: config: Codeception\Extension\RunFailedSticky: fail-group: another_group1"
+ * ```
+ * Remember: if you run tests and they generated custom-named fail group, to run this group, you should add override too
+ *
+ * Starting from Codeception 2.1 **this extension is enabled by default**.
```
DavertMik: Copy-paste detected :)

DavertMik:
Thanks. I like this improvement. But I think this should not be another extension, please just update RunFailed extension instead

Naktibalda:
Please update RunFailed extension instead and make a new pull request to 4.0 branch.

PR #5184, opened 2018-09-20
[REST] Add response code assertions to json and xml parts

### LINKED ISSUE #5183: Why is the `seeResponseCodeIsSuccessful()` method not allowed after setting the `part` parameter?
In the `REST` module, when I set the `part` parameter, I found that some methods like `seeResponseCodeIsSuccessful()` are not generated in Trait `ApiTesterActions`.
After reading the source code of the `REST` module, I know that it is through `Reflection` to limit the generation of a particular method, but I think `seeResponseCodeIsSuccessful()` is similar to `dontSeeResponseCodeIs($code)`. Why not use it in the above situation?
### Details

* Codeception version: 2.4.4
* PHP Version: 7.1.16
* Operating System: Centos 7
* Installation type: Yii2 default
* Suite configuration:

```yml
actor: ApiTester
modules:
    enabled:
        - REST:
            # url:
            depends: Yii2
            part: Json
    config:
        - Yii2
```

Naktibalda:
Fixes #5183

### DIFF
--- a/src/Codeception/Module/REST.php
+++ b/src/Codeception/Module/REST.php
@@ -1163,6 +1163,9 @@ EOF;
 
     /**
      * Checks that the response code is 2xx
+     *
+     * @part json
+     * @part xml
      */
     public function seeResponseCodeIsSuccessful()
     {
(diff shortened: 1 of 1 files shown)


DavertMik:
Thanks!

PR #5204, opened 2018-09-26
Fixes PHPStan issue

josephzidell:
When using namespaces in tests, and building the `*TesterActions` classes, PHPStan throws an error that it cannot find `<MyNamespace>\TestInterface` as well as `<MyNamespace>\actions`.

This fixes those issues.


DavertMik:
Thanks

PR #5207, opened 2018-09-29
correct regex to root url

gimler:
The regex to match the root url is not correct.

`$` stands for end in regex
`^` is beginning in regex


DavertMik:
Awesome, you noticed it!
Thank you!

PR #5213, opened 2018-10-03
Asserts: Add expectThrowable() method

burned42:
I thought it would be quite useful to be also able to test for [Errors](http://php.net/manual/de/class.error.php) and (possibly in future) other [Throwables](http://php.net/manual/de/class.throwable.php), too instead of just [Exceptions](http://php.net/manual/de/class.exception.php).

So this pull request would add a new method `expectThrowable()` and refactor `expectException()` to just call this method. I also suggest marking `expectException()` as deprecated since I think the old method could be removed some time later (maybe with a new major release or such?). I also added some tests and updated the documentation. Please tell me what you think about this and if there are any additional steps required that I didn't know of.


tests/unit/Codeception/Module/AssertsTest.php
```diff
@@ -66,8 +71,60 @@ public function testExceptionFails()
...
+        $this->module->expectThrowable(new Exception('here'), function () {
+            throw new Exception('here');
+        });
+        $this->module->expectThrowable(new Exception('here', 200), function () {
+            throw new Exception('here', 200);
+        });
+    }
+
+    /**
+     * @expectedException PHPUnit\Framework\AssertionFailedError
```
DavertMik: Hey, could you update the code to not use this annotation? 
Because of https://github.com/sebastianbergmann/phpunit/issues/3332

tests/unit/Codeception/Module/AssertsTest.php
```diff
@@ -66,8 +71,60 @@ public function testExceptionFails()
...
+     */
+    public function testExpectThrowableFailOnDifferentClass()
+    {
+        $this->module->expectThrowable(new RuntimeException(), function () {
+            throw new Exception();
+        });
+    }
+
+    /**
+     * @expectedException PHPUnit\Framework\AssertionFailedError
```
DavertMik: Please don't use annotations https://github.com/sebastianbergmann/phpunit/issues/3332

tests/unit/Codeception/Module/AssertsTest.php
```diff
@@ -66,8 +71,60 @@ public function testExceptionFails()
...
+     */
+    public function testExpectThrowableFailOnDifferentMessage()
+    {
+        $this->module->expectThrowable(new Exception('foo', 200), function () {
+            throw new Exception('bar', 200);
+        });
+    }
+
+    /**
+     * @expectedException PHPUnit\Framework\AssertionFailedError
```
DavertMik: Please don't use annotations https://github.com/sebastianbergmann/phpunit/issues/3332

tests/unit/Codeception/Module/AssertsTest.php
```diff
@@ -66,8 +71,60 @@ public function testExceptionFails()
...
+    public function testExpectThrowableFailOnDifferentCode()
+    {
+        $this->module->expectThrowable(new Exception('foobar', 200), function () {
+            throw new Exception('foobar', 2);
+        });
+    }
+
+    /**
+     * @expectedException PHPUnit\Framework\AssertionFailedError
+     * @expectedExceptionMessageRegExp /RuntimeException/
```
DavertMik: Please don't use annotations https://github.com/sebastianbergmann/phpunit/issues/3332

tests/unit/Codeception/Module/AssertsTest.php
burned42: Yes, sure, thanks for pointing that out, didn't know that! :)

Fixed in f8a2b5289cc907bb095ee0be889d1fa86ef086af

tests/unit/Codeception/Module/AssertsTest.php
burned42: Fixed in f8a2b5289cc907bb095ee0be889d1fa86ef086af

tests/unit/Codeception/Module/AssertsTest.php
burned42: Fixed in f8a2b5289cc907bb095ee0be889d1fa86ef086af

tests/unit/Codeception/Module/AssertsTest.php
burned42: Fixed in f8a2b5289cc907bb095ee0be889d1fa86ef086af

burned42:
> > I also suggest marking expectException() as deprecated since I think the old method could be removed some time later
> 
> I think it's ok how it is. I got used to it )

Alright, so should I then just remove those lines?
https://github.com/Codeception/Codeception/pull/5213/files#diff-214a6d3c36d25915108ad90033f96177R319
https://github.com/Codeception/Codeception/pull/5213/files#diff-b3d3622a11b4688b45ee6999ce435bc0R459 ?

> 
> > I also added some tests and updated the documentation. Please tell me what you think about this and if there are any additional steps required that I didn't know of.
> 
> Thanks. Everything looks fine, I just didn't have enough time to review it before

Thanks for reviewing :)

DavertMik:
Thanks. Let's do that!

PR #5218, opened 2018-10-08
use param files .yaml like .yml

kardagan:
Since version 4, Symfony the default app skeleton uses .yaml files


DavertMik:
Ok, accepted

PR #5220, opened 2018-10-10
Adding Ability To Specify More Than One Dump File

Fenikkusu:
Note: This is a re-open of PR #5011

Currently the Db module only allows you to specify one file for the dump file. I personally was running across two issues with this. The first is that the software requires a base set of records to work. If These are put in the default dump file, then it has to be re added anytime someone updates the dump file. Very tedious. The second issue was that some of the base set of records have changes that need to be made to run additional checks/testing. With one dump file, this isn't possible without specifying a new copy of the entire db to change one row. This of course is overkill.

This PR adds the functionality to specify more than one file as the dump file:

```yaml
modules:
    enabled:
        - Db:
            dsn: '%DSN%'
            host: '%HOST%'
            dump:
                - ./tests/_data/schema.sql
                - ./tests/_data/dumps/base.sql
```

This change can also be used within environments:
```yaml
env:
    owner:
        modules:
            config:
                Db:
                    dump:
                        - ./tests/_data/schema.sql
                        - ./tests/_data/dumps/base.sql
                        - ./tests/_data/dumps/owner.sql
    employee:
        modules:
            config:
                Db:
                    dump:
                        - ./tests/_data/schema.sql
                        - ./tests/_data/dumps/base.sql
                        - ./tests/_data/dumps/employee.sql
```

I also updated the DbPopulator to be able to handle the multi dumping. If you have more than one dump file, it will call the popular command for each dump file. The idea of course being that you use the '$dump' syntax in the populator.


DavertMik:
Please fix the failures on CI https://travis-ci.org/Codeception/Codeception/jobs/439456643#L2930

Fenikkusu:
Found a bug when specifying multiple databases. Went ahead and applied a fix here.

Fenikkusu:
Updated Documentation And Rebased/Squashed Changes

DavertMik:
Thanks for working on this. Let me take another look.

docs/modules/Db.md
```diff
@@ -60,6 +60,18 @@ if you run into problems loading dumps and cleaning databases.
              ssl_ca: '/path/to/ca-cert.pem'
              ssl_verify_server_cert: false
              ssl_cipher: 'AES256-SHA'
+             
+## Example 2
```
DavertMik: Ok, please move this example to `Codeception\Module\Db.php` into the docblock. This file is generated from it

DavertMik:
@Naktibalda what do you think of this PR?

src/Codeception/Lib/DbPopulator.php
```diff
@@ -98,8 +120,35 @@ public function run()
         return true;
     }
 
+    private function buildCommands()
+    {
+        if ($this->commands !== null) {
+            return $this->commands;
+        } else if (!isset($this->config['dump'])) {
+            return $this->buildCommand($this->config['populator']);
```
Naktibalda: I think that this is a bug, `run` method expects to get array from `buildCommands` but it returns string.
Wrapping it to array, would allow to simpify `run` by removing the first condition.

DavertMik:
@Naktibalda how do you find it now? Looks good to me

DavertMik:
Yeah!
We can make a release

PR #5224, opened 2018-10-12
Adding Finally To Ensure Suite After Always Runs, Even If Exception

Fenikkusu:
I noticed that the 'Command Did Not Finish Properly' feature I added previously was having adverse effects when an exception was thrown in the beforeSuite or afterSuite methods of a module. This was occurring even though the exception occurred. In my mind, an Exception should disable the Command Did Not Finish  Properly as it is to be handled by the normal error trapping. I've wrapped the Suite Event that fires the tests and the after-tests with a try/finally so that the after suite will still get called if an exception occurs. This will prevent the display of the original exception as well as the following 'Command Did Not Finish Properly'


DavertMik:
Thanks. I'm not sure if it won't break anything. But let's try

PR #5225, opened 2018-10-14
Adding Ability To Translate Remote Paths To Local (IE: Docker)

Fenikkusu:
This PR adds the ability to use coverage locally when utilizing docker or other forms of mounted/shared directories. It otherwise gives you the ability to run coverage without having to resort to remote coverage, provided you can use some form of shared mount when testing.

I use docker for most of my stuff. This includes both on my local computer and for my GitLab Pipelines (CI). When running coverage, if you are using docker, you have no choice but to set coverage up as a remote server. This of course means you will not get merged coverage. The only way you can use local coverage with a docker setup is to either run codeception through a docker container (And map everything to be the same paths) or to update the working directories of docker to be the same locally and within the container. Personally, I found both of these to be overkill. 

While trying to debug why I was getting no coverage on my acceptance testing, I found that the only reason was that the PHP_Codecoverage system is storing the absolute path to the file. As a proof of concept in my project, I just replaced the absolute path from the docker container with the absolute path in my actual computer. This worked perfectly, so I threw this together.

Adding the `work_dir` setting to the coverage will update the path of all covered files to match the project root. 

I haven't gotten any testing in yet and am open to suggestions on how to add testing for this, outside of just doing a unit test with reflection.

```yaml
coverage:
  enabled: true
  work_dir: /var/www
```

### DIFF
--- a/src/Codeception/Coverage/Subscriber/LocalServer.php
+++ b/src/Codeception/Coverage/Subscriber/LocalServer.php
@@ -9,6 +9,7 @@ use Codeception\Event\TestEvent;
 use Codeception\Events;
 use Codeception\Exception\ModuleException;
 use Codeception\Exception\RemoteException;
+use PHP_CodeCoverage;
 
 /**
  * When collecting code coverage data from local server HTTP requests are sent to c3.php file.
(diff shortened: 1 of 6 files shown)


Fenikkusu:
I've applied a couple updates to this. The first is some extra debugging just so you can see what gets replaced with what. The second is actual testing (Note: this is more testing the code runs than actually testing the full replacement...not sure the best way to test actual replacement). The third, item is a slight enhancement I used for debugging. I added the ability to pass php options to the executeCommand method. This allowed me to initialize debugging breakpoints on the underlying tests of my test.

src/Codeception/Coverage/Subscriber/LocalServer.php
```diff
@@ -9,6 +9,7 @@
 use Codeception\Events;
 use Codeception\Exception\ModuleException;
 use Codeception\Exception\RemoteException;
+use PHP_CodeCoverage;
```
DavertMik:
```suggestion
```

Old style PHPUnit class name. Please remove it

src/Codeception/Coverage/Subscriber/LocalServer.php
```diff
@@ -122,7 +123,41 @@ public function afterSuite(SuiteEvent $e)
...
+        //Only Process If Work Directory Set
+        if ($this->settings['work_dir'] === null) {
+            return $this;
+        }
+
+        $workDir    = rtrim($this->settings['work_dir'], '/\\') . DIRECTORY_SEPARATOR;
+        $projectDir = Configuration::projectDir();
+        $data       = $coverage->getData(true); //We only want covered files, not all whitelisted ones.
+
+        codecept_debug("Replacing All Instances Of {$workDir} with {$projectDir}");
```
DavertMik: Why is this capitalzied?

DavertMik:
This PR is also helpul for remote executed code coverage via c3. 
Before this we couldn't map remote files to local

Fenikkusu:
Updated Documentation And Rebased/Squashed Changes

DavertMik:
Thanks, @Fenikkusu 
Looks very deep and cool feature. Are in our Slack? I'd like to invite you to our development channel.

Fenikkusu:
@DavertMik, I am not in it that I recall.

Fenikkusu:
Turns out I am. Same ID.

PR #5232, opened 2018-10-17
Updating Phalcon Module To Be Able To Override MemorySession Class

Fenikkusu:
The Phalcon Module uses the supplied MemorySession class for the session object. The problem I've found with this is that I tend to extend session and add helper methods to the extensions. These end up breaking in testing because the methods aren't there. To get around this, I have to add extra module code to override the session after init.

This PR adds a new config to the module to allow setting the class name to use for the session. It also updates it so that Phalcon users can use the built in service override of Phalcon's DIC.


DavertMik:
Thanks, we use this repo https://github.com/Codeception/phalcon-demo
For internal Phalcon testing. Could you move your test code into it?

Fenikkusu:
Removed Testing and Added it https://github.com/Codeception/phalcon-demo/pull/12.

DavertMik:
Thanks!

Issue #5177, opened 2018-09-19
Recorder ext improvements


OneEyedSpaceFish:
**This is not a real issue, however, I promised that I'd make further improvements to the recorder extension based on the needs of my project.** 

Since this can be used by everyone, and since that it's a general improvement, **I'd like to raise this to the attention of the general CC team, and I'd like to ask for feedback on the current state of the changes.**

Even though this is a work in progress, a lot of things had been improved, and we are currently deployed this on our production system, where it generates a fairly large report every 10 minutes. We encountered different edge cases already, and so far it seems to be stable, including those edge cases where the previous version of this extension seemed to have issues.

There were a lot of edge cases why the previous extension didn't meet our expectations; for example, there were some issues where a lengthy wantTo caused the generated file name to exceed the allowed length causing an exception being thrown while it tried to flush the file to the disk.

**Quality of life improvements**
For example, the generated lists are now structured as follows:
- Suite Name 1 (Environment 1)
  - Cest Name 1
  - Cest Name 2 
  - Cest Name 3
- Suite Name 2 (Environment 1)
  - Cest Name 1
  - Cest Name 2 
  - Cest Name 3
- Suite Name 1 (Environment 2)
  - Cest Name 1
  - Cest Name 2 
  - Cest Name 3
- Suite Name 2 (Environment 2)
  - Cest Name 1
  - Cest Name 2
  - Cest Name 3

And the color of the list `-` represents if the tests had passed or failed.
Just an FYI, failures used to be red in my copy, while passes are marked with a green. This was the case until I discovered that one of the developers in my team had a green - red color blindness and he couldn't differentiate between the two at all. Then I changed the failures to be yellow instead. In the final release, I plan to add support for setting those colors in the config for color blind people and to revert the default color for failures to red. This is an upcoming feature, so that's why currently the failures are just marked in yellow. I also plan to display errors as black by default, which are currently not supported/saved yet.

I also made the title in the recorded steps page to be a link back to the main page, as I noticed that previously people were just clicking the back button in their browser of choice.

**Structural improvements**
All edge case issues we had encountered before had been fixed with file operations, and now the failures are handled more gracefully. I also updated the dependencies on jquery and bootstrap to be the latest and obviously fixed all issues caused by the update.

Just to keep any new readers on this saga updated with the previous events and conversations, most of the upcoming or implemented improvements had been discussed in my previous issue: #5101 

**Example**
[...]

PR #5235, opened 2018-10-17
Failing gracefully if webDriver is not hydrated in Recorder extension

OneEyedSpaceFish:
See https://github.com/Codeception/Codeception/pull/5221#issuecomment-430548993
Parent issue: #5177

### DIFF
--- a/ext/Recorder.php
+++ b/ext/Recorder.php
@@ -449,6 +449,10 @@ EOF;
             $status = 'error';
 
             try {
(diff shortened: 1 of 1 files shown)


OneEyedSpaceFish:
Correct me if I'm wrong, but that build failure is not related to the change.

DavertMik:
Thanks! The fix looks good and tests pass

OneEyedSpaceFish:
Thanks for you effort looking into this massive saga of changes as well :+1:

PR #5237, opened 2018-10-18
Added error_level to the global configuration

boan2010:
According to the documentation (https://codeception.com/docs/04-FunctionalTests#Error-Reporting), it should be possible to configure the error_level globally in the `codeception.yml`. But the config was not passed to the suite configurations.

> error_level can also be set globally in codeception.yml file

### DIFF
--- a/src/Codeception/Configuration.php
+++ b/src/Codeception/Configuration.php
@@ -309,7 +309,7 @@ class Configuration
 
         // load global config
         $globalConf = $config['settings'];
-        foreach (['modules', 'coverage', 'namespace', 'groups', 'env', 'gherkin', 'extensions'] as $key) {
+        foreach (['modules', 'coverage', 'namespace', 'groups', 'env', 'gherkin', 'extensions', 'error_level'] as $key) {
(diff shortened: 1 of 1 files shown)


boan2010:
> I think you can set error_reporting inside `settings` section.

It is working within `settings`, can you set all the suite configurations globally within settings?
If so, this may be an important part to mention in the documentation here:

https://codeception.com/docs/reference/Configuration#Global-Configuration

>settings: provide additional options for test runner. They may dramatically change the way Codeception is executed. For instance, take a note of shuffle option which allows to randomize tests execution order and lint option that toggles parsing a test file (using php -l) before loading it.

And the documentation here: https://codeception.com/docs/04-FunctionalTests#Error-Reporting, could also add that it is within the `settings` section.

image

Thank you :-)

DavertMik:
Nice. If it worked for you, could you send a PR to update the documentation?
This would be useful for everyone

alexkuc:
@DavertMik I have created a PR for this https://github.com/Codeception/codeception.github.com/pull/431

Issue #5107, opened 2018-08-02
"The EntityManager is closed" with Symfony and Doctrine2 modules


benr77:
#### What are you trying to achieve?

Use the Symfony and Doctrine2 modules together, to allow "cleanup" by reversing transactions between tests, in functional tests against an SQLite database.

#### What do you get instead?

Many tests are simply failing with

```bash
[Doctrine\ORM\ORMException] The EntityManager is closed.
```
If I do not enable the Doctrine2 module, the tests all succeed. But without it enabled, the transactions are not implemented and reversed between tests.

functional.suite.yml is as follows

```yml
class_name: FunctionalTester
modules:
  enabled:
    - \Helper\Functional
    - Asserts
    - Cli
    - Filesystem
    - REST
    - Symfony
        environment: test
        var_path: 'var'
    - Doctrine2
        depends: Symfony
        cleanup: true
```
### Details

* Codeception version: 2.4.5
* PHP Version: 7.2
* Operating System: Linux Mint
* Installation type: Composer

freiondrej:
Hello @benr77,

the problem can be caused by a \Throwable thrown during a transaction. When you look into the code of method `UnitOfWork#commit()`, it does an automatic rollback after a \Throwable is thrown **and** it closes EntityManager. When trying to work with EntityManager after this, an exception is thrown because EntityManager is already closed.
I hope this will be of any help to you :)

Amunak:
@freiondrej, did you find a workaround? I have the same issue, and besides writing some dirty patch for Codeception I found no solution.

Also would be great to hear from the devs (@DavertMik?)  whether they plan to fix this...

Amunak:
Well apparently this fixes the issue by reopening the entity manager:
```patch
--- /dev/null
+++ src/Codeception/Module/Symfony.php
@@ -231,9 +231,15 @@
         if ($this->kernel === null) {
             $this->fail('Symfony2 platform module is not loaded');
         }
-        if (!isset($this->permanentServices[$this->config['em_service']])) {
+        if (!isset($this->permanentServices[$this->config['em_service']]) || !$this->permanentServices[$this->config['em_service']]->isOpen()) {
             // try to persist configured EM
             $this->persistService($this->config['em_service'], true);
+
+            // based on https://github.com/symfony/symfony/issues/5339
+            if (!$this->permanentServices[$this->config['em_service']]->isOpen()) {
+                $this->grabService('doctrine')->resetManager();
[...]

PR #5243, opened 2018-10-24
Rollback all nested transactions created within test (#5107)

Dukecz:
This fixes tests in which other transactions are created but not closed for various reasons.


DavertMik:
Thanks

PR #5248, opened 2018-10-31
Changed raw 404 numbers to http constant

h311ion:


DavertMik:
Sorry but I don't think this is somehow valueble...

PR #5250, opened 2018-11-01
[Doctrine2] Making cleanup property work after on reconfigure

joelmedeiros:
After last release the cleanup property on Doctrine2 module is not working when you try to reconfigure the module.


joelmedeiros:
@DavertMik could you check this PR?

PR #5262, opened 2018-11-08
Fix "already initialized service", "reboot kernel issue"

gdmfx:
I have an issue. When i work with DataFactory and Doctrine2 modules, and when my testing services uses EntityManager - the instances of entity managers are different. Therefore DataFactory EntityManager knows about created entities through have() method. But my EntityManager is not. Also Doctrine2 cleanup service is not working, because all of them (DataFactory, Doctrine2, my service) use different connections and transactions. Because of this i have alot of bugs during testing.
I have investigated the issue. The reason is - codeception Symfony module rebooted the kernel and on first request services that was in kernel's container are lost. But this kernel's container still uses for App services.
There is code for persistent services, but, this is not works for first request. Because of Symfony module $this->container is empty, and kernel's container services are not saving. I'm suggest initialize $this->container variable from kernel first, and then perform rebootKernel method. Thus all persistent services from initial kernel can be saved and re-seted. This patch works and all bugs are gone. I spend a lot of time on these bugs. Now works DataFactory and Doctrine2 "cleanup" options together perfect.
Also i had another one bug. This is described in https://github.com/Codeception/Codeception/issues/4806. I'm suggest check that service is already initialized by perform "initialize" method. It works perfect. 
Also can i know why need reboot kernel on first request in Symfony module contructor?

### DIFF
--- a/src/Codeception/Lib/Connector/Symfony.php
+++ b/src/Codeception/Lib/Connector/Symfony.php
@@ -77,7 +77,7 @@ class Symfony extends \Symfony\Component\HttpKernel\Client
         $this->container = $this->kernel->getContainer();
 
         foreach ($this->persistentServices as $serviceName => $service) {
-            if(!$this->container->initialized($serviceName)) {
+            if (!$this->container->initialized($serviceName)) {
                 $this->container->set($serviceName, $service);
             }
         }
--- a/src/Codeception/Lib/Connector/Symfony.php
+++ b/src/Codeception/Lib/Connector/Symfony.php
@@ -36,6 +36,7 @@ class Symfony extends \Symfony\Component\HttpKernel\Client
         $this->followRedirects(true);
         $this->rebootable = (boolean)$rebootable;
         $this->persistentServices = $services;
+        $this->container = $this->kernel->getContainer();
         $this->rebootKernel();
     }
 
@@ -76,7 +77,9 @@ class Symfony extends \Symfony\Component\HttpKernel\Client
         $this->container = $this->kernel->getContainer();
 
         foreach ($this->persistentServices as $serviceName => $service) {
-            $this->container->set($serviceName, $service);
+            if(!$this->container->initialized($serviceName)) {
+                $this->container->set($serviceName, $service);
+            }
         }
 
         if ($this->container->has('profiler')) {


gdmfx:
@DavertMik can you please review. It closes also https://github.com/Codeception/Codeception/issues/4806

ewgRa:
After this change, Symfony 3.4 persistent services functionality is broken.

We have service that subscribed to doctrine.event_listener. It never returns false for initialized check and neven persistent.

Naktibalda:
Could you make this code work with 3.4 and 4.x?

I think that detecting a version of Symfony and only checking `!initialized` on 4.x should fix it.

Naktibalda:
Would it be better to attempt to set the service and suppress exception if it fails?

Naktibalda:
I'm sorry, but it is on you to implement that functionality.
We have no regular maintainer of Symfony so it is up to people who care to implement it.

This was the fourth pull request with this fix for "already initialized service" issue, so it looked important enough to be merged.

Naktibalda:
Please upgrade to Codeception 2.5.3 and check if it works correctly.

dfostercb911:
Still happening in more recent versions!

**Error**
```
The "App\PATH\TO\MY\MockApiClass" service is already initialized, you cannot replace it.
 /var/www/drm-case/vendor/symfony/dependency-injection/Container.php:166
 /var/www/drm-case/tests/_support/Helper/Functional.php:63
 /var/www/drm-case/tests/_support/_generated/FunctionalTesterActions.php:7244
```

**Specifications:**
PHP: `v8.1.17`
Codeception Modules: 
```
    "codeception/assert-throws": "^1.0",
    "codeception/codeception": "^4.1.22",
    "codeception/module-asserts": "^2.0",
    "codeception/module-cli": "^2.0",
    "codeception/module-db": "^1.0.0",
    "codeception/module-doctrine2": "^2.0",
    "codeception/module-queue": "^1.0.0",
    "codeception/module-rest": "^1.0.0",
    "codeception/module-symfony": "^2.1.1",
```

Codeception Config: 

```
paths:
    tests: tests
    output: tests/_output
    data: tests/_data
    support: tests/_support
    envs: tests/_envs
actor_suffix: Tester
bootstrap: _bootstrap.php
groups:
    dispute_outcome_suite: [tests/api/some/path,tests/functional/some/path,tests/unit/some/path]
    rule:
      - tests/api/Rule
      - tests/unit/Rule
      - tests/functional/Rule
extensions:
    enabled:
        - BypassFinalsExtension
settings:
    shuffle: false
    memory_limit: 4096M
params:
    - .env
    - .env.test
modules:
    enabled:
        - Queue
        - Asserts
        - Symfony:
              app_path: 'src'
              environment: 'test'
        - Doctrine2:
              depends: Symfony
              cleanup: true
        - \Helper\Functional
        - \Helper\Api
        - \Helper\ArrayHelper
        - \Helper\DoctrineHelper
        - \Helper\UserHelper
    config:
        Queue:
            type: 'aws'
            host: 'my-localstack'
            port: 4576
            key: 'not-a-real-key'
            secret: 'not-a-real-secret'
            region: 'us-east-1'
            version: 'latest'
            endpoint: 'http://my-localstack:4576'
```

Inside of `tests/_support/Helper/Functional.php`
```
    /**
     * @throws ModuleException
     */
    public function replaceServiceWithMock(string $serviceClassName, $mockedService)
    {
        /** @var Symfony $symfony */
        $symfony = $this->getModule('Symfony');
        $this->originalServices[$serviceClassName] = $symfony->grabService($serviceClassName);
        $symfony->kernel->getContainer()->set($serviceClassName, $mockedService);
        $symfony->persistService($serviceClassName);
    }

    /**
     * @throws ModuleException
     */
    public function restoreMockedService(string $serviceClassName): void
    {
        /** @var Symfony $symfony */
        $symfony = $this->getModule('Symfony');
        $symfony->unpersistService($serviceClassName);
        $symfony->kernel->getContainer()->set($serviceClassName, $this->originalServices[$serviceClassName]);
        $symfony->persistService($serviceClassName);
    }
```

**Am I missing something here or is this just not something that is possible during functional tests anymore with symfony??**

PR #5265, opened 2018-11-13
Correct Windows entrypoint (codecept.bat)

sergeyklay:
I simplified the `codecept.bat` file and removed `@var_name@` variables. The `@var_name@` vars should be replaced during a PHAR build. I think that `@var_name@` vars are not universal, and I am also not sure that they make sense (or even used) right now.

It's using `%PHP_PEAR_PHP_BIN%` because that's a global variable which may be set by PEAR. I just figure, if it's available anyway, why not use it; and in the other case, it will just try php, assuming most Windows users will have added that to their `PATH`.

These changes should be covered by ` appveyor.yml:test_script` (I hope).

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,7 +4,7 @@ platform: 'x86'
 branches:
   except:
     - gh-pages
-    
+
 cache:
   - c:\tools\php71 -> appveyor.yml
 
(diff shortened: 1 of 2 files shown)


sergeyklay:
Okay, I've found what was causing `OpenSSL.Light` package installation to fail. Changed Windows update service start type.

DavertMik:
Looks good to me

DavertMik:
I think we no longer use .bat but let's keep that patch :)

PR #5273, opened 2018-11-20
JsonType: Allow to use : in regex filter

### LINKED ISSUE #2501: \Codeception\Util\JsonType filter delimiter ":" does not allow to use ":" in regex filter

ellisgl:
Fixes #2501 
Fixes regex filter handling in Util\JsonType

### DIFF
--- a/src/Codeception/Util/JsonType.php
+++ b/src/Codeception/Util/JsonType.php
@@ -138,12 +138,12 @@ class JsonType
             preg_match_all($regexMatcher, $type, $regexes);
 
             // Do the same match as above, but replace the the 'any character' + delimiter with a place holder ($${count}).
-            $type = preg_replace_callback($regexMatcher, function () {
+            $filterType = preg_replace_callback($regexMatcher, function () {
                 static $count = 0;
                 return ':regex($$' . $count++ . ')';
             }, $type);
 
-            $matchTypes  = preg_split("#(?![^]\(]*\))\|#", $type);
+            $matchTypes  = preg_split("#(?![^]\(]*\))\|#", $filterType);
             $matched     = false;
             $currentType = strtolower(gettype($data[$key]));
 
@@ -168,13 +168,6 @@ class JsonType
 
                         return $regexes[1][$pos];
                     }, $filter);
-
-                    // Fill regex pattern back into the type.
-                    $type = preg_replace_callback('/\$\$\d+/', function ($m) use ($regexes) {
-                        $pos = (int)substr($m[0], 2);
-
-                        return $regexes[1][$pos];
-                    }, $type);
 
                     $matched = $matched && $this->matchFilter($filter, $data[$key]);
                 }
--- a/src/Codeception/Util/JsonType.php
+++ b/src/Codeception/Util/JsonType.php
@@ -153,7 +153,7 @@ class JsonType
 
             foreach ($matchTypes as $matchType) {
                 $filters      = preg_split("#(?![^]\(]*\))\:#", $matchType);
-                $expectedType = trim(strtolower(array_shift($filters)));
+                $expectedType = strtolower(trim(array_shift($filters)));
 
                 if ($expectedType !== $currentType) {
                     continue;
@@ -164,14 +164,14 @@ class JsonType
                 foreach ($filters as $filter) {
                     // Fill regex pattern back in to the filter.
                     $filter = preg_replace_callback('/\$\$\d+/', function ($m) use ($regexes) {
-                        $pos = (int)(substr($m[0], 2));
+                        $pos = (int)substr($m[0], 2);
 
                         return $regexes[1][$pos];
                     }, $filter);
 
                     // Fill regex pattern back in to the type.
                     $type = preg_replace_callback('/\$\$\d+/', function ($m) use ($regexes) {
-                        $pos = (int)(substr($m[0], 2));
+                        $pos = (int)substr($m[0], 2);
 
                         return $regexes[1][$pos];
                     }, $type);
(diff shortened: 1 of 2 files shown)


src/Codeception/Util/JsonType.php
```diff
@@ -126,14 +126,25 @@ protected function typeComparison($data, $jsonType)
...
                 continue;
             }
-            $matchTypes = preg_split("#(?![^]\(]*\))\|#", $type);
-            $matched = false;
+
+            $regexes = [];
+
+            preg_match_all('/:regex\((.*?[!"#$%&\'*+,.\/:;=?@^_`|~-])\)/', $type, $regexes);
+
+            $type = preg_replace_callback('/:regex\((.*?[!"#$%&\'*+,.\/:;=?@^_`|~-])\)/', function ($m) {
```
DavertMik: could you explain a bit of this magic?

DavertMik:
Looks good to me but I don't understand what's happening there :stuck_out_tongue: 
Could you add some inline comments to your code?
I know regex is complicated by itself but some minor comments would help

src/Codeception/Util/JsonType.php
ellisgl: So what's happening here's it's matching for everything up to the regex closing delimiter (This should cover 99.999% of the delimiter usage - There are some strange ones like '<'...'>' and '{'...'}' which I've only come across a couple times in my lifetime.). What was happening before was that stuff like ':' was causing the match to match prematurely, so adding the closing delimiter match after wildcard and before the ')' character, it shouldn't get tripped up.

ellisgl:
Looks like wercker/build stuff needs to fixed.

src/Codeception/Util/JsonType.php
Naktibalda: Just use the first character as a delimiter, that's what preg_match does.

ellisgl: Well the offending code was preg_split, which was looking for ':', which would break if ':' was in the regex filter definition. So this is what I first thought of. There is the (?R) recursive pattern, but I would have to look over that again, since it's probably been over 10 years since I've used that.

ellisgl: @DavertMik @Naktibalda I've updated my explanation on the how and why.

tests/unit/Codeception/Util/JsonTypeTest.php
```diff
@@ -82,6 +82,12 @@ public function testRegexFilter()
 
         $jsonType = new \Codeception\Util\JsonType(['time' => '21:00']);
         $this->assertTrue($jsonType->matches(['time' => 'string:regex(~^([0-1]\d|2[0-3]):[0-5]\d$~)']));
+        
+        $jsonType = new \Codeception\Util\JsonType(['text' => '21@:00']);
+        $this->assertTrue($jsonType->matches(['text' => 'string:regex(~^(\d\d@):\dd$~)']));
+        
+        $jsonType = new \Codeception\Util\JsonType(['text' => '21@:aa']);
+        $this->assertFalse($jsonType->matches(['text' => 'string:regex(~^(\d\d@):\dd$~)']));
```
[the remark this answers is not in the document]
Naktibalda: Sorry for that, let's see if fixing it helps.

Naktibalda:
I tried to match time with `'time' => 'string:regex(~\d\d:\d\d:\d\d~)',` in my project,
so I copy-pasted your patch to see if it works.
Matching works correctly, but one thing it messes up is error message.

Here is the output I got:
```
 Fail  `time: '2018-12-21'` is of type `string:regex($$0)`
```
I would like to see the actual pattern instead of `$$0`.

This issue was visible in my failed attempt to create failing test: https://travis-ci.org/Codeception/Codeception/jobs/467890679
```
Failed asserting that '`text: '21@:00'` is of type `string:regex($$0):\dd$~)`' is true.
```

Naktibalda:
It should be enough to use a different variable instead of `$type` to store result of `preg_replace_callback` to avoid overriding original value of `$type`.

ellisgl:
OK - so there actually was an issue with the regex I was using which prematurely stopped on that example you gave (Which I had a pattern I wrote during the last commit that I thought was overkill, but I guess it wasn't). I also had to just replace the $$\d+ in the type. Reason for this was because of the split on | and the split on : would have still caused some issues if I had just used a second variable for the regexes. I also did some fixes around PHP Inspections (EA Extended) suggestions.

src/Codeception/Util/JsonType.php
```diff
@@ -119,55 +119,71 @@ protected function typeComparison($data, $jsonType)
...
             $regexes = [];
 
             // Match the string ':regex(' and any characters until a regex delimiter (matches 99.999% use cases) followed by character ')'
             // Place the 'any character' + delimiter matches in to an array.
-            preg_match_all('/:regex\((.*?[!"#$%&\'*+,.\/:;=?@^_`|~-])\)/', $type, $regexes);
+            preg_match_all($regexMatcher, $type, $regexes);
 
             // Do the same match as above, but replace the the 'any character' + delimiter with a place holder ($${count}).
-            $type = preg_replace_callback('/:regex\((.*?[!"#$%&\'*+,.\/:;=?@^_`|~-])\)/', function ($m) {
+            $type = preg_replace_callback($regexMatcher, function ($m) {
```
Naktibalda: Using different variable name instead of `$type` here and in the `preg_split` below would avoid need to refill at the bottom.

ellisgl: Ah!

PR #5276, opened 2018-11-21
Changes for upcoming Travis' infra migration

sergeyklay:
See: https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

Travis is deprecating the `sudo` keyword and moves everything to the same infrastructure (sudo really selects between two infrastructures).


DavertMik:
Thanks

PR #5283, opened 2018-11-27
add method to count and test the number of messages in a queue

kardagan:


src/Codeception/Module/AMQP.php
```diff
@@ -305,6 +305,55 @@ public function seeMessageInQueueContainsText($queue, $text)
...
     }
 
+    /**
+     * Count messages in queue.
+     *
+     * @param string $queue
+     *
+     * @return int
+     */
+    public function countMessageInQueue($queue)
```
Naktibalda: I think that this method shouldn't be a part of public API.
Leaving it public but prepending undescore to the name (`_countMessageInQueue`) would hide it in tests but allow modules ands helpers to use it.

src/Codeception/Module/AMQP.php
```diff
@@ -305,6 +305,55 @@ public function seeMessageInQueueContainsText($queue, $text)
...
+     * $I->pushToQueue('queue.emails', 'Hello, davert');
+     * $I->seeMessageInQueue('queue.emails');
+     * $I->seeMessageInQueue('queue.emails',1);
+     * ?>
+     * ```
+     *
+     * @param string $queue
+     * @param int $expected
+     */
+    public function seeMessageInQueue($queue, $expected = null)
```
Naktibalda: I really don't like the name of this method - if I saw it in the test, I would assume that it is checking if specific message is in the queue.
A better name matching Codeception conventions would be `seeQueueIsNotEmpty`
and a more appropriate name for `dontSeeMessageInQueue` is `seeQueueIsEmpty`.

On another hand, it would probably be good to match method names of [Queue module](https://codeception.com/docs/modules/Queue#seeEmptyQueue) - `seeEmptyQueue`, `dontSeeEmptyQueue`, `seeQueueHasCurrentCount`.

kardagan: ok to match Queue module, but if someone have enable `AMQP` and `Queue` there is conflict, or we assume that it will never append ?

Naktibalda: You are right, using different method names to avoid conflict is better.

kardagan: I've made the changes

src/Codeception/Module/AMQP.php
kardagan: I've made the changes

kardagan:
I'm not sure that the errors are made by me, can someone look please ?

Naktibalda:
Errors weren't caused by you.
There was a failure of build for comment change earlier.

I think that PhpBrowserRestTest failures could be caused by some dependency.
Tests that are making requests to localhost with `Host: www.example.com` header, are receiving responses from actual www.example.com host.
I will try to fix it later.

kardagan:
@Naktibalda are the changes ok for you now ?

src/Codeception/Module/AMQP.php
```diff
@@ -305,6 +305,76 @@ public function seeMessageInQueueContainsText($queue, $text)
...
+     * <?php
+     * $I->pushToQueue('queue.emails', 'Hello, davert');
+     * $I->seeNumberOfMessageInQueue('queue.emails',1);
+     * ?>
+     * ```
+     *
+     * @param string $queue
+     * @param int $expected
+     */
+    public function seeNumberOfMessageInQueue($queue, $expected)
```
Naktibalda: I don't like being a grammar nazi, but this method should be named seeNumberOfMessage**s**InQueue.

Everything else is good.

kardagan: @Naktibalda  done

Naktibalda:
Wercker build failed because of an issue which was fixed in 2.5 branch few weeks ago.
If you care about it, fetch 2.5 branch and rebase on it, to make build pass.
If you don't, I will merge this pull request tomorrow anyway.

PR #5285, opened 2018-11-27
add addQueue method for dynamic confid of queue liste

kardagan:
Actually we have to declare queues in codeception.yml but all apps must allready avec list declare in other place. with this method we can parse the app conf in the _initialize of the module suite and call define it without duplicate conf


src/Codeception/Module/AMQP.php
```diff
@@ -90,6 +91,18 @@ public function _initialize()
         }
     }
 
+    /**
+     * Add a queue to purge list
+     *
+     * @param string $queue
+     */
+    public function addQueue($queue)
```
DavertMik: `addQueue` sounds too general and doesn't represent the effect taken. 
Maybe it should be `scheduleQueueCleanup` or something like that

kardagan: ok :) I didn't have good idea for the name.

kardagan: done !

PR #5292, opened 2018-12-06
Save InnerBrowser's page source as .fail.json if content-type is application/json

freiondrej:


freiondrej:
@Naktibalda thanks. I added support for xml, though I cannot think of a better notation without the ternary here :/

PR #5294, opened 2018-12-13
Add Environment Name To Descriptor Unique Signatures

Tenzian:
Fixes codeception/phpunit-wrapper#4 

When running tests using multiple environments in serial, i.e.

    codecept run --html --env=one --env=two --env=three 

Test descriptor function `getTestSignatureUnique()` does _not_ return unique signatures across environments. Features using the unique signature (e.g. the phpunit-wrapper html report) can then conflate the tests.

In the case of the html report this causes a test that fails for an environment to appear as failed in the report for subsequent environments, even if it passes in those other environments. 

Fixed by adding the test case's environment name to the unique signature.


src/Codeception/Test/Descriptor.php
```diff
@@ -32,15 +32,22 @@ public static function getTestSignature(\PHPUnit\Framework\SelfDescribing $testC
      */
     public static function getTestSignatureUnique(\PHPUnit\Framework\SelfDescribing $testCase)
     {
+        $env     = null;
```
Naktibalda: Empty string would be a better choice for default value,
You are only using it for string concatenation.

src/Codeception/Test/Descriptor.php
```diff
@@ -32,15 +32,22 @@ public static function getTestSignature(\PHPUnit\Framework\SelfDescribing $testC
...
         $example = null;
 
+        if (method_exists($testCase, 'getScenario')
+            && $env = $testCase->getScenario()->current('env')
```
Naktibalda: If `$env` evaluates to some falsy value, it skips concatenation below, but is used in return line.
It would be better to assign temporary value to a different variable, e.g. `$currentEnv`.

Issue #5193, opened 2018-09-24
 [ConnectionException] Can't connect to Webdriver at http://127.0.0.1:9515/wd/hub. Please make sure that Selenium Server or PhantomJS is running. 


chenxiumiao:
#### What are you trying to achieve?
open chrome and click something When I install chromedriver
#### What do you get instead?

```bash
UserCest: Try to test
Signature: UserCest:tryToTest
Test: tests/acceptance/UserCest.php:tryToTest
Scenario --
  WebDriver::debugWebDriverLogs method has been called when webDriver is not set
  WebDriver::_saveScreenshot method has been called when webDriver is not set
  WebDriver::_savePageSource method has been called when webDriver is not set
  Screenshot and page source were saved into '/Users/cx/Downloads/code/fangPHP/data/web/CV/tests/_output/' dir
 ERROR 


1) UserCest: Try to test
 Test  tests/acceptance/UserCest.php:tryToTest
                                                                                                                                                    
  [ConnectionException] Can't connect to Webdriver at http://127.0.0.1:9515/wd/hub. Please make sure that Selenium Server or PhantomJS is running.  
                                                                                                                                                    
#1  phar:///usr/local/bin/codecept/src/Codeception/Module/WebDriver.php:1449
#2  phar:///usr/local/bin/codecept/src/Codeception/Module/WebDriver.php:426
#3  Codeception\Subscriber\Module->before
#4  phar:///usr/local/bin/codecept/vendor/symfony/event-dispatcher/EventDispatcher.php:212
#5  phar:///usr/local/bin/codecept/vendor/symfony/event-dispatcher/EventDispatcher.php:44
#6  phar:///usr/local/bin/codecept/codecept:37
#7  /usr/local/bin/codecept:7

> Provide console output if related. Use `-vvv` mode for more details.
```
> Provide test source code if related

```php
	public function tryToTest ( AcceptanceTester $I )
	{
		
		$I->amOnPage( '_enter.php?m=user&a=login' );
		$I->seeInTitle('用户登入');
		$I->fillField( 'email' , 'sipc@115.com');
		$I->fillField( 'password' , 'sipc115');
		$I->click('登入');
		$I->wait( 5);
		$I->seeInTitle( '我的简历');
	}
```
### Details

* Codeception version:2.4.6
* PHP Version:7.2.9
* Operating System:MAC
* Installation type: Phar 
* List of installed packages (`composer show`) erusev/parsedown 
* Suite configuration:
```yml
[...]

PR #5315, opened 2018-12-28
WebDriver: print curl error to debug output if webdriver failed to connect

Naktibalda:
It will help to debug issues like #5193

Example of output:

```
- WebDriverTest: Am on subdomain
  Curl error: Curl error thrown for http POST to /session with params: {"desiredCapabilities":{"browserName":"chrome"}}

  Failed to connect to 127.0.0.1 port 4444: Connection refused
  WebDriver::debugWebDriverLogs method has been called when webDriver is not set
  WebDriver::_saveScreenshot method has been called when webDriver is not set
  WebDriver::_savePageSource method has been called when webDriver is not set
  Screenshot and page source were saved into '/c/Projects/Codeception/tests/log/' dir
E WebDriverTest: Am on subdomain
```

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -1446,6 +1446,7 @@ class WebDriver extends CodeceptionModule implements
             $this->setBaseElement();
             $this->initialWindowSize();
         } catch (WebDriverCurlException $e) {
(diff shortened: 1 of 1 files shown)


DavertMik:
Thanks, I think it would be useful!

PR #5327, opened 2019-01-07
Update for dotenv v3.0

josephzidell:
`vlucas/phpdotenv` v 3.0 now has a different format for loading `.env` files

### DIFF
--- a/src/Codeception/Lib/ParamsLoader.php
+++ b/src/Codeception/Lib/ParamsLoader.php
@@ -115,7 +115,7 @@ class ParamsLoader
     protected function loadDotEnvFile()
     {
         if (class_exists('Dotenv\Dotenv')) {
-            $dotEnv = new \Dotenv\Dotenv(codecept_root_dir(), $this->paramStorage);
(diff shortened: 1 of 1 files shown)


Naktibalda:
Please update version constraint in composer.json to get tests passings.

Is it necessary to keep compatibility with phpdotenv v2?

Naktibalda:
Superseded by #5366

Naktibalda:
The best solution is to update wp-browser to use dotenv v3.

Naktibalda:
How did you manage to update Codeception to 2.5.3, composer should refuse to do that if some dependencies are incompatible?

defunctl:
Probably because wp-browser includes Codeception as a dependency, so you don't even require it in your composer.json: https://github.com/lucatume/wp-browser/blob/master/composer.json#L27

I've got a temporary work around by installing Codeception 2.5.2 until wpbrowser gets updated: https://github.com/lucatume/wp-browser/issues/210#issuecomment-460747403

Naktibalda:
Ah, it is `require-dev` dependency, so it does not prevent other libraries from requiring different versions.
I agree that it would be good to keep support for dotenv v2.

Issue #5328, opened 2019-01-08
Phalcon upgrade for 3.4.x 


chilimatic:
#### What are you trying to achieve?
This is a ticket to be referenced in the PR. Version bump fix for phalcon 3.4.x

# pretext
I upgraded my PHP version to 7.3 and for this I needed to upgrade my phalcon version to 3.4 because it's the first stable version to compile to 7.3.

# problem
in version 3.4.x they changed the Session interface MemorySession so the method signature contracts fail.

# reason for structure
To avoid collision I added a second module Phalcon_34x since I cannot guarantee it does not have collisions with <3.3.x versions.

# problem that gets fixed
The interface contract is a fatal, so I get a series of of errors

it's not a Codeception bug it's a phalcon internal upgrade which did break. I want to unitest

chilimatic:
Invalid issue - you @sergeyklay just informed me that you need phalcon 3.4.x compiled with zephir 10.x. in order to work. 

phalcon v.4 will break these interfaces anyway so solving an intermediate edge-case has not immanent value (this explanation is just if someone has a similar problem)

PR #5329, opened 2019-01-08
#5328 adds phalcon version 3.4.x module

chilimatic:
#5328 

I did only fix the interfaces and left the other code, besides the module name intact.

I don't mind better solutions but since phalcon is working on 4.x as their current release target and I did not want to impose any structure. 

This is a 5 minutes patch, so if there is a method like an internal switch for different versions based on a parameter I am more than willing to adapt this PR.


src/Codeception/Module/Phalcon_34x.php
```diff
@@ -0,0 +1,674 @@
...
+ *             part: services
+ *             bootstrap: 'app/config/bootstrap.php'
+ *             cleanup: true
+ *             savepoints: true
+ *         - WebDriver:
+ *             url: http://your-url.com
+ *             browser: phantomjs
+ * ```
+ */
+class Phalcon_34x extends Framework implements ActiveRecord, PartedModule
```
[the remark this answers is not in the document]
chilimatic: fixed in https://github.com/Codeception/Codeception/pull/5329/commits/653757eefb932ab3a2da342356d4b0a7ceffb37d

src/Codeception/Lib/Connector/Phalcon_34x/MemorySession.php
```diff
@@ -0,0 +1,309 @@
...
+        return true;
+    }
+
+    /**
+     * @inheritdoc
+     *
+     * @param bool $deleteOldSession
+     * @return \Phalcon\Session\AdapterInterface
+     */
+    public function regenerateId(bool $deleteOldSession = NULL): \Phalcon\Session\AdapterInterface
```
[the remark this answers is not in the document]
chilimatic: fixed in https://github.com/Codeception/Codeception/pull/5329/commits/653757eefb932ab3a2da342356d4b0a7ceffb37d

src/Codeception/Lib/Connector/Phalcon_34x.php
```diff
@@ -0,0 +1,159 @@
...
+use ReflectionProperty;
+use Codeception\Util\Stub;
+use Phalcon\Mvc\Application;
+use Symfony\Component\BrowserKit\Cookie;
+use Symfony\Component\BrowserKit\Client;
+use Phalcon\Mvc\Micro as MicroApplication;
+use Symfony\Component\BrowserKit\Response;
+use Codeception\Lib\Connector\Shared\PhpSuperGlobalsConverter;
+
+class Phalcon_34x extends Client
```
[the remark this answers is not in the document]
chilimatic: fixed in https://github.com/Codeception/Codeception/pull/5329/commits/653757eefb932ab3a2da342356d4b0a7ceffb37d

src/Codeception/Module/Phalcon34x.php
```diff
@@ -0,0 +1,674 @@
...
+use Codeception\Lib\Framework;
+use Phalcon\Mvc\RouterInterface;
+use Phalcon\Mvc\Model as PhalconModel;
+use Phalcon\Mvc\Router\RouteInterface;
+use Codeception\Util\ReflectionHelper;
+use Codeception\Exception\ModuleException;
+use Codeception\Lib\Interfaces\ActiveRecord;
+use Codeception\Lib\Interfaces\PartedModule;
+use Codeception\Exception\ModuleConfigException;
+use Codeception\Lib\Connector\Phalcon as PhalconConnector;
```
Naktibalda: It looks like your module is using the old connector here.
If module code is exactly the same, it would be better to update Phalcon module to select connector dynamically, based on some version check or existance of some class, method or property that was introduced or removed in 3.4.0.

src/Codeception/Lib/Connector/Phalcon34x/MemorySession.php
```diff
@@ -0,0 +1,309 @@
+<?php
+namespace Codeception\Lib\Connector\Phalcon;
```
Naktibalda: This line should be `namespace Codeception\Lib\Connector\Phalcon34x;`

src/Codeception/Module/Phalcon34x.php
chilimatic: fair enough.

src/Codeception/Lib/Connector/Phalcon34x/MemorySession.php
chilimatic: thx :)

src/Codeception/Module/Phalcon34x.php
Naktibalda: If the only difference is in MemorySession class, than please create it dynamically and keep using the same connector too,

chilimatic: @Naktibalda I simplified everything to just replace the session based on the internal phalcon version + I extended the session class so only the necessary changes are reflected.

chilimatic: didn't even read your comment but came to the same conclusion. :)

src/Codeception/Module/Phalcon.php
```diff
@@ -671,4 +670,15 @@ private function getApplicationDomainRegex()
 
         return '/^' . str_replace('.', '\.', $domain) . '$/';
     }
+
+    /**
+     * @return string
+     */
+    private function getMemorySessionForVersion() {
+        if (strpos(\Phalcon\Version::get(), '3.4') !== false) {
+            return PhalconConnector\MemorySession34x::class;
```
Naktibalda: This class will work with Phalcon 4.x too, right?
It would be better to use [version_compare](http://php.net/manual/en/function.version-compare.php)
`version_compare(\Phalcon\Version::get(), '3.4.0') >= 0)`

chilimatic: It's unclear if it does work with 4.x I'll compile and test it to check. Good point though.

Naktibalda: It is more likely to work than the old class.

chilimatic: yes :D probability says it is still if I don't need heuristics I won't use them :) 10 mins :) and we know it for sure I already changed it. I just verify it

chilimatic: @Naktibalda currently there is not even an Session adapter interface in branch 4.0.x so let @sergeyklay decide how to proceed :). Maybe it's not consolidated atm.

src/Codeception/Lib/Connector/Phalcon/MemorySession34x.php
```diff
@@ -0,0 +1,317 @@
...
+
+        $this->options = $options;
+    }
+
+    /**
+     * @inheritdoc
+     *
+     * @return array
+     */
+    public function getOptions(): array
```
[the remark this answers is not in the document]
chilimatic: @sergeyklay but the branch 3.4.x 

```
	/**
	 * Get internal options
	 */
	public function getOptions() -> array;
```
compiles to
```
public function getOptions(): array
```
that's the whole reason why I wrote this patch my local codeception was not working anymore :)

please correct me if I am wrong.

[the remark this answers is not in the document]
sergeyklay: Let me explain. Phalcon 3.4.x as well as current Codeception _must_ works on PHP 5.x. This fact confirmed by tests in the corresponding repositories. Such a function definition isn't compatible with PHP 5 syntax:
```
public function getOptions(): array
```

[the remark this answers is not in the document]
sergeyklay: And yes, this is not PHP:
```zep
public function getOptions() -> array;
```
This is Zephir.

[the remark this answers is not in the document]
chilimatic: @sergeyklay hmm well I know that but than it's incompatible with your implementation. so I need to build an extra check for the version > 7. and rename it. Because the only reason why i need this adapter is because phalcon breaks codeception in version 7.3 because phalcon changed the interfaces / the compiler now generates strict signature interface contracts for version 7+.

So I guess I need to check for the php >=7 ∩ Phalcon 3.4.x? with 7.2 and phalcon 3.3.x compile with zephir 10.x i didn't have the problem.

but php 7.3 + zephir 11.x + phalcon 3.4.x breaks it and 3.3.x or 4.0.x have different issues in this combination one time I need psr-ext and the other time I get a list of all the internal signature contracts that are wrong. Besides my gcc downgrades needed to compile it. (rust needs a newer gcc)

does this make sense to you? 

tl;dr; 
if I just do what you ask php 5 will work but 7.3 with phalcon 3.4.x compiled with zephir 11.x will break

that's it in a nutshell.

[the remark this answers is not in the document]
sergeyklay: Backward incompatible changes are in Zephir 0.11.x and Phalcon 4.x. So you may need to compile Phalcon 3.4.x using Zephir 0.10.x. Also take a look at https://stackoverflow.com/a/54066201/1661465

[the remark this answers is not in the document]
chilimatic: @sergeyklay fair enough, I use arch and I compile things per hand. So good for rpm and deb architectures. https://aur.archlinux.org/packages/php-phalcon/ currently defines it to 7.2 so also no luck there.

So I will close this PR as invalid, since I already fixed it in my fork and I don't have to care about compatibility. 

However it would be nice if you mention the zephir version locking with v3 of phalcon in the phalcon compilation readme guide. :) 

my mistake and sorry for wasting all of your time :)

chilimatic:
invalid due to zephir compiler to phalcon target version. 
using zephir 10.x + phalcon 3.4.x resolves the problem

PR #5335, opened 2019-01-11
[ZendExpressive] Set Cookie header in request

Naktibalda:
Because `Dflydev\FigCookies\Cookies::fromRequest($request);` reads cookies from header.

Cookies can be retrieved using `$request->getCookieParams()`, but this change helps if someone makes the same mistake as I did.


Naktibalda:
If other frameworks could benefit from this functionality, it could be moved to [InnerBrowser::clientRequest](https://github.com/Codeception/Codeception/blob/2.5.2/src/Codeception/Lib/InnerBrowser.php#L161) method.

Issue #5320, opened 2019-01-02
JsonType - Alternative filters after a string:regex are dropped


Tenzian:
#### What are you trying to achieve?

Provide alternative values for JSON Types by postfixing a type to `string:regex` i.e.

- `string:regex(<pattern>)|null` 

#### What do you get instead?

This used to be fine, but fails to work since #5273 was merged.

It looks like the new filtering in https://github.com/ellisgl/Codeception/blob/be75ceb95f463f653cb111a60b12411dc7d161ba/src/Codeception/Util/JsonType.php#L133-L148 is causing postfixed options to be dropped.

> Provide console output if related. Use `-vvv` mode for more details.

```bash
$ codecept run unit -vvv
Codeception PHP Testing Framework v2.5.2
Powered by PHPUnit 7.3.5 by Sebastian Bergmann and contributors.
Running with seed: 


Unit Tests (2) ------------------------------------------------------------------------------------------------------------------------------
Modules: CodeHelper, EmulateModuleHelper
---------------------------------------------------------------------------------------------------------------------------------------------
✔ JsonTypeTest: Regex filter with prefixed alternatives (0.00s)
✖ JsonTypeTest: Regex filter with postfixed alternatives (0.00s)
---------------------------------------------------------------------------------------------------------------------------------------------


Time: 90 ms, Memory: 10.00MB

There was 1 failure:

---------
1) JsonTypeTest: Regex filter with postfixed alternatives
 Test  tests/unit/codeception/util/JsonTypeTest.php:testRegexFilterWithPostfixedAlternatives
Failed asserting that '`test: NULL` is of type `string:regex(~^(\d\d@):\d\d$~)|null`' is true.
/Users/tenz/Repositories/Codeception/vendor/phpunit/phpunit/src/Framework/Constraint/Constraint.php:112
/Users/tenz/Repositories/Codeception/vendor/phpunit/phpunit/src/Framework/Constraint/Constraint.php:60
/Users/tenz/Repositories/Codeception/vendor/phpunit/phpunit/src/Framework/Assert.php:1857
/Users/tenz/Repositories/Codeception/vendor/phpunit/phpunit/src/Framework/Assert.php:1012
/Users/tenz/Repositories/codeception-demo/tests/unit/codeception/util/JsonTypeTest.php:24
/Users/tenz/Repositories/Codeception/vendor/phpunit/phpunit/src/Framework/TestCase.php:1150
[...]

PR #5339, opened 2019-01-12
[JsonType] Fixed issue #5320

Naktibalda:
I changed regex matcher to use the first character after `regex(` as a delimiter (back reference `\2`),
so it matches pattern from `regex(~` until nearest `~)`,

Thanks to @Tenzian for reporting and test code.

@ellisgl Do you see any potential issues that this change could cause?

### DIFF
--- a/src/Codeception/Util/JsonType.php
+++ b/src/Codeception/Util/JsonType.php
@@ -130,10 +130,10 @@ class JsonType
                 continue;
             }
 
-            $regexMatcher = '/:regex\(((.).*?\2)\)/';
+            $regexMatcher = '/:regex\((((\()|(\{)|(\[)|(<)|(.)).*?(?(3)\)|(?(4)\}|(?(5)\]|(?(6)>|\7)))))\)/';
             $regexes = [];
 
-            // Match the string ':regex(' and any characters until a regex delimiter (matches 99.999% use cases) followed by character ')'
+            // Match the string ':regex(' and any characters until a ending regex delimiter followed by character ')'
             // Place the 'any character' + delimiter matches in to an array.
             preg_match_all($regexMatcher, $type, $regexes);
 
--- a/tests/unit/Codeception/Util/JsonTypeTest.php
+++ b/tests/unit/Codeception/Util/JsonTypeTest.php
@@ -239,4 +239,14 @@ class JsonTypeTest extends \Codeception\Test\Unit
         // currently produces a false negative
         $this->assertTrue($jsonType->matches(['test'    => 'string:regex(~^(\d\d@):\d\d$~)|integer']));
     }
+
+    public function testRegexFilterWithSpecialDelimiters()
+    {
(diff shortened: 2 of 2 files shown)


ellisgl:
Actually, I wanted to do that, but was thinking that I couldn't do that for some reason. The only problem I see is that it would be searching for the nearest match for "\2)", which could prematurely cut off, which is why I had the "(?:\|(boolean|integer|double|float|string|array|object|resource|resource \(closed\)|null|unknown type)|$)" at the end.

ellisgl:
Tried this, it works. 
```
$jsonType = new JsonType(['str' => 'Something.(~xsdf~).something']);
$this->assertTrue($jsonType->matches(['str' => 'string:regex(~Something\.\(\~.*\~\)\.something~)']));
```

ellisgl:
I think I have a fix for bracket delimiters stuff. I'll do a pull request to this in a minute.

Naktibalda:
@ellisgl I have tried to come up with a reason to use negative look-behind assertion for backslash character in regex matcher, but I got some unexpected results.

Here is my test code:

```php
/**
 * backslash can be used to escape character, which is the same as closing delimiter, in the pattern
 */
public function testRegexFilterWithEscapedDelimiter()
{
    $jsonType = new JsonType(['test' => 'xy/z']);
    $this->assertTrue($jsonType->matches(['test'  => 'string:regex(/(xy\/)z/)']));
    $this->assertFalse($jsonType->matches(['test' => 'string:regex(/(xy\/)w/)']));
}
```

Second assertion unexpectedly fails,
```
Failed asserting that '`test: 'xy/z'` is of type `string:regex(/(xy\/)w/)`' is false.
```
A little bit of debugging shows that the pattern used in `matchFilter` method was `/(xy/)w/`.
Can you see a reason why pattern `/(xy\/)w/` matches text `xy/z`?

ellisgl:
If I read that correctly, you must escape the delimiter character you used if you are searching for that character in the string.

ellisgl:
Shouldn't the assertFalse be assertNotTrue? ```/(xy\/)w/``` shouldn't match ```xy/z``` unless somehow it's interpreting it as \w....???

Naktibalda:
> Shouldn't the assertFalse be assertNotTrue? 

All other regex tests use `assertFalse`.

> Try \\/ instead of \/

It makes no difference, PHP string interpolation removes that extra backslash.
but changing `\/` to `/` makes preg_match fail ` preg_match(): Unknown modifier ')'`
so one backslash does  the job.

ellisgl:
The test file has no assertFalse anywhere in it:
https://github.com/Codeception/Codeception/blob/4839a93782bcbbf03976ff3fa4043647ec3f21d9/tests/unit/Codeception/Util/JsonTypeTest.php

ellisgl:
No biggy, i had to double check when you said you had issues.

PR #5344, opened 2019-01-14
Fix XPath example

sdlins:
XPath does not work without `"`. 

Just test something like `$I->click('//form[1]/button[@type=submit]')` vs  `$I->click('//form[1]/button[@type="submit"]');`. The first will not find the button submit.

### DIFF
--- a/docs/modules/PhpBrowser.md
+++ b/docs/modules/PhpBrowser.md
@@ -288,7 +288,7 @@ $I->click('Submit');
 // CSS button
 $I->click('#form input[type=submit]');
 // XPath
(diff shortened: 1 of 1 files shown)


Naktibalda:
@sdlins thanks for fix,
nut you updated the wrong file, PhpBrowser.md file is autogenerated from dockblocks,
please update https://github.com/Codeception/Codeception/blob/2.5/src/Codeception/Lib/Interfaces/Web.php#L304 too.

sdlins:
https://github.com/Codeception/Codeception/pull/5371 

Explanations about the reason for two PR are there (sorry).

PR #5364, opened 2019-01-27
Update Ftp module to be compatible with phpseclib v2 (Issue #5363)

### LINKED ISSUE #5363: FTP module
Hi,

The FTP module require "phpseclib/phpseclib": "0.3.6" and phpseclib is now in 2.0.13.
In this version \Net_SFTP and \Crypt_RSA don't exist animore, remplaced by phpseclib\Crypt\RSA and phpseclib\Net\SFTP.

I see 3 way to fix it : 
- juste remplace classes but will not work anymore for those who use the old version.
- add version in config to know witch class to use (my prefered)
- create a new ftp module

what's your opinion ?

kardagan:
Closes #5363

### DIFF
--- a/src/Codeception/Module/FTP.php
+++ b/src/Codeception/Module/FTP.php
@@ -2,6 +2,7 @@
 namespace Codeception\Module;
 
 use Codeception\TestInterface;
+use Codeception\Exception\ModuleException;
 
 /**
  *
@@ -25,7 +26,7 @@ use Codeception\TestInterface;
  * For SFTP, add [phpseclib](http://phpseclib.sourceforge.net/) to require list.
  * ```
  * "require": {
- *  "phpseclib/phpseclib": "0.3.6"
+ *  "phpseclib/phpseclib": "^2.0.14"
  * }
  * ```
  *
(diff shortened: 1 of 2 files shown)


src/Codeception/Module/FTP.php
```diff
@@ -25,7 +26,7 @@
  * For SFTP, add [phpseclib](http://phpseclib.sourceforge.net/) to require list.
  * ```
  * "require": {
- *  "phpseclib/phpseclib": "0.3.6"
+ *  "phpseclib/phpseclib": ">=0.3.6"
```
Naktibalda: It would be better to give an instruction to install  phpseclib 2.*, this module can be incompatible with phpseclib 3.

Naktibalda: Support for 0.3.6 is for backwards compatibility, there is no reason to tell new users that they can use that old version.
It would be the best to change example to `"phpseclib/phpseclib": "^2.0.14"`

kardagan: done

src/Codeception/Module/FTP.php
```diff
@@ -836,15 +837,30 @@ private function modified($filename)
...
         }
 
         if (isset($this->config['key'])) {
             $keyFile = file_get_contents($this->config['key']);
-            $password = new \Crypt_RSA();
+
+            if (class_exists('Crypt_RSA')) {
+                $password = new \Crypt_RSA();
+            } elseif (class_exists('phpseclib\Net\SFTP')) {
+                $password = new \phpseclib\Crypt\RSA();
```
Naktibalda: I apologise for nitpicking, but checking one class and instantiating another doesn't look right to me.

kardagan: you're right, bad copy/past :p fixed

PR #5369, opened 2019-01-30
issue-5367 grabFromDatabase() returns string

### LINKED ISSUE #5367: Fix return type of grabFromDatabase() method to string | bool
grabFromDatabase() sometimes returns string. That will be more honest and will help with IDE notices.

SanzhiyevMergen:
Fixes #5367

### DIFF
--- a/src/Codeception/Module/Db.php
+++ b/src/Codeception/Module/Db.php
@@ -869,7 +869,7 @@ class Db extends CodeceptionModule implements DbInterface
      * @param string $column
      * @param array  $criteria
      *
-     * @return string|bool
+     * @return mixed
      */
     public function grabFromDatabase($table, $column, $criteria = [])
     {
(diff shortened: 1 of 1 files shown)


src/Codeception/Module/Db.php
```diff
@@ -869,7 +869,7 @@ public function grabColumnFromDatabase($table, $column, array $criteria = [])
      * @param string $column
      * @param array  $criteria
      *
-     * @return array
+     * @return string|bool
```
Naktibalda: This method can return other types too, so it would be the best to change it to `@return mixed`

SanzhiyevMergen: Ok, at least `null` can be returned. Agree.

SanzhiyevMergen: Done.

Naktibalda: integer too and probably float, because some database libraries return native types.

PR #5371, opened 2019-02-01
Update Web.php

sdlins:
https://github.com/Codeception/Codeception/pull/5344

Sorry, I had not time to make a clone, edit and push. So I've made the two PR via browser (not found a way to merge the files in only one PR via browser).


Naktibalda:
thanks

PR #5382, opened 2019-02-08
Phalcon module see number of records

Zaszczyk:
I have added new method seeNumberOfRecords in Phalcon module, to check number of records in the table.

### DIFF
--- a/src/Codeception/Module/Phalcon.php
+++ b/src/Codeception/Module/Phalcon.php
@@ -397,7 +397,13 @@ class Phalcon extends Framework implements ActiveRecord, PartedModule
     {
         $records = $this->findRecords($model, $attributes);
         if ($records->count() != $number) {
-            $this->fail("Couldn't find $number records of $model with " . json_encode($attributes) .". Found: ".$records->count()." records.");
+            $this->fail(sprintf(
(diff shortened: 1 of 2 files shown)


src/Codeception/Module/Phalcon.php
```diff
@@ -379,6 +379,29 @@ public function seeRecord($model, $attributes = [])
...
+     * @param string $model Model name
+     * @param int $number int number of records
+     * @param array  $attributes Model attributes
+     * @part orm
+     */
+    public function seeNumberOfRecords($model, $number, $attributes = [])
+    {
+        $records = $this->findRecords($model, $attributes);
+        if ($records->count() != $number) {
+            $this->fail("Couldn't find $number records of $model with " . json_encode($attributes) .". Found: ".$records->count()." records.");
```
sergeyklay: Could you please do it more neat, using `sprintf` or something like that. I prefer avoid too long lines

Zaszczyk: Sure, done.

Zaszczyk:
Could you merge it? :)

sergeyklay:
cc @Naktibalda

PR #5403, opened 2019-02-20
Added link to MailCatcher

ThomasLandauer:


DavertMik:
sorry, all other items are not linked, so this should not be linked too. Anyway, google works fine for such cases

Issue #5401, opened 2019-02-20
Error with chromedriver 2.46


JorisVanEijden:
#### What are you trying to achieve?
Use chromedriver to run tests.

#### What do you get instead?
```
[Facebook\WebDriver\Exception\UnrecognizedExceptionException] invalid argument: invalid 'expiry'
(Session info: chrome=72.0.3626.81)
(Driver info: chromedriver=2.46.628388 (4a34a70827ac54148e092aafb70504c4ea7ae926),platform=Linux 4.9.125-linuxkit x86_64)

#1  /var/www/html/vendor/facebook/webdriver/lib/Exception/WebDriverException.php:158
#2  /var/www/html/vendor/facebook/webdriver/lib/Remote/HttpCommandExecutor.php:326
#3  /var/www/html/vendor/facebook/webdriver/lib/Remote/RemoteWebDriver.php:547
#4  /var/www/html/vendor/facebook/webdriver/lib/Remote/RemoteExecuteMethod.php:40
#5  /var/www/html/vendor/facebook/webdriver/lib/WebDriverOptions.php:55
#6  Codeception\Module\WebDriver->loadSessionSnapshot
```

### Details

* Codeception version: 2.5.3
* PHP Version: 7.1
* Operating System: Any
* Installation type: Composer
* Suite configuration:

```yml
modules:
    enabled:
        -
          WebDriver:
            browser: chrome
            capabilities:
              javascriptEnabled: true
              resourceTimeout: 300000
            host: %WEBDRIVER_HOST%
            url: %WEBSITE_URL%
```

Chromedriver 2.45 works fine.
In 2.46 they added extra cookie validation (https://chromium-review.googlesource.com/c/chromium/src/+/1366925)

Naktibalda:
Isn't it a bug of Chromedriver then?
I've seen this issue few weeks ago and it looked like Chromedriver broke setCookie in legacy mode completely, but facebook/php-webdriver library does not support W3C mode yet.

OneEyedSpaceFish:
@JorisVanEijden 
Imo, if this is a bug with Chromedriver, then it's worth raising an issue / PR with them instead of adding a workaround solution in CC.

JorisVanEijden:
I don't think it's a bug in Chromedriver to state that null is an invalid value for those cookie fields.
From https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol I would also expect null to be an invalid value.
[...]

PR #5404, opened 2019-02-21
Ensure cookie expiry value is not null

JorisVanEijden:
Workaround for bug #5401 to ensure cookie expiry value is not null which breaks in chromedriver 2.46

Sorry about the unrelated whitespace change.

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -824,6 +824,14 @@ class WebDriver extends CodeceptionModule implements
                 $params['domain'] = $urlParts['host'];
             }
         }
+        // #5401 Supply defaults, otherwise chromedriver 2.46 complains.
+        $params = array_filter($params);
+        $params += [
+            'path' => '/',
+            'expiry' => 0,
+            'secure' => false,
+            'httpOnly' => false,
+        ];
         $this->webDriver->manage()->addCookie($params);
         $this->debugSection('Cookies', json_encode($this->webDriver->manage()->getCookies()));
     }
@@ -3097,7 +3105,7 @@ class WebDriver extends CodeceptionModule implements
         }
         
         foreach ($this->sessionSnapshots[$name] as $cookie) {
-            $this->webDriver->manage()->addCookie($cookie);
+            $this->setCookie($cookie->getName(), $cookie->getValue(), $cookie->toArray());
         }
         $this->debugSection('Snapshot', "Restored \"$name\" session snapshot");
         return true;


Naktibalda:
@JorisVanEijden When I ran Codeception web suite with chromedriver 2.46 I also got invalid parameter errors for other parameters, could you investigate that?

> snapshot tests failed with `invalid argument: invalid 'secure'`, cookie tests failed with `invalid argument: invalid 'path'`, and test `Save session snapshots exclude invalid cookie domains` failed with `invalid argument: invalid 'domain'` which can probably be expected.

https://codeception.slack.com/archives/C93HPU7NZ/p1549722886036600

JorisVanEijden:
The problem is that Codeception and Facebook\webdriver are ok with these values being NULL, but chromedriver 2.46 is not.
I'll update the PR with the same fix for the other params.

reinholdfuereder:
First of all thanks for your efforts! (I stumbled over this problem ~3 weeks ago, and due to still being a newbie just hoped that I am not the only one...)
Anyway, the "All checks have failed" section it appears as if the installation of codeception dependencies via Composer is already the problem (cf. codeception/phpunit-wrapper), which brings me to the main question: why is the base branch of this pull request 2.6 (and not 2.5)? Because the codeception/phpunit-wrapper dependency has changed (or was broken?) for 2.6 just 16 days ago...

JorisVanEijden:
I've changed the base to 2.5 but now 5 other commits have appeared.

reinholdfuereder:
@JorisVanEijden I must admit that I cannot really follow you, because it looks quite OK now (but I am certainly no git/github expert), but maybe that is because you already found the solution in the shape of your last "force-pushed" action?

Now it seems that "only" the "semaphoreci" check fails anymore, because there 2 webdriver test cases need to be fixed or in fact maybe just adapted to your changes? And maybe supplemented with new test cases for your changes? => I guess an expert like @Naktibalda can give better hints...

JorisVanEijden:
I'm still working on this. But I'm finding it very difficult, with all the different libraries and versions.

JorisVanEijden:
I've made a PR against facebook/php-webdriver that will strip null values from cookie fields.
https://github.com/facebook/php-webdriver/pull/627
That seem to me to be a better place. If that gets in then it doesn't matter what values Codeception adds to the cookies, nulls will never reach chromedriver.

Naktibalda:
facebook/php-webdriver seems unmaintained, if nothing happens there, I will merge this pull request before next release (probably this weekend).

Naktibalda:
I tested this patch and it didn't solve all compatibility issues.

Before patch:
```
1) WebDriverTest: Cookies with path
 Test  tests/web/WebDriverTest.php:testCookiesWithPath
                                                                                                                                                                                             
  [Facebook\WebDriver\Exception\UnrecognizedExceptionException] invalid argument: invalid 'secure'
  (Session info: headless chrome=72.0.3626.121)
  (Driver info: chromedriver=2.46.628402 (536cd7adbad73a3783fdc2cab92ab2ba7ec361e1),platform=Windows NT 10.0.17763 x86_64)
```

After patch:
```
4) WebDriverTest: Cookies with path
 Test  tests/web/WebDriverTest.php:testCookiesWithPath
                                                                                                                                                                                             
  [Facebook\WebDriver\Exception\UnableToSetCookieException] unable to set cookie
  (Session info: headless chrome=72.0.3626.121)
  (Driver info: chromedriver=2.46.628402 (536cd7adbad73a3783fdc2cab92ab2ba7ec361e1),platform=Windows NT 10.0.17763 x86_64)
```

I updated Chrome and Chromedriver to 73 (version numbers are matching now) and it failed in the same way:

```
  [Facebook\WebDriver\Exception\UnableToSetCookieException] unable to set cookie
  (Session info: headless chrome=73.0.3683.86)
  (Driver info: chromedriver=73.0.3683.68 (47787ec04b6e38e22703e856e101e840b65afe72),platform=Windows NT 10.0.17763 x86_64)
```

Naktibalda:
Oh wait, I tested it incorrectly :)

Naktibalda:
`setCookie` no longer fails, but `seeCookie` fails because chromedriver does not return any cookies.

Naktibalda:
0 isn't a good default value for `expiry`, changing it to `time() + 86400` fixed almost all tests, so I guess that 0 was considered to be timestamp 0 - `1970-01-01`.

reinholdfuereder:
> `setCookie` no longer fails, but `seeCookie` fails because chromedriver does not return any cookies.

@Naktibalda Should this be worrying? This sounded to me as if nothing depending on already set cookies is supposed to work anymore (like session/cookie snapshot)!? But actually it looks okay!? Sorry for my lack of understanding...

Naktibalda:
I fixed it by setting expiry to now + 24 hours instead of 0.

reinholdfuereder:
Sorry for being ambiguous or not clear enough: I meant "_Should the failing `seeCookie`, because chromedriver does not return any cookies, be worrying_"

PR #5406, opened 2019-02-22
[ZendExpressive] Application recreate config entries are mixed up

artmnv:
Application should not be recreated between requests with this config:
```
recreateApplicationBetweenTests: true
recreateApplicationBetweenRequests: false
```
But it does. I can't inject anything into container in test because of this issue.

### DIFF
--- a/src/Codeception/Exception/ElementNotFound.php
+++ b/src/Codeception/Exception/ElementNotFound.php
@@ -7,7 +7,9 @@ class ElementNotFound extends \PHPUnit\Framework\AssertionFailedError
 {
     public function __construct($selector, $message = null)
     {
-        $selector = Locator::humanReadableString($selector);
+        if (!is_string($selector) || $selector[0] !== "'") {
+            $selector = Locator::humanReadableString($selector);
+        }
         parent::__construct($message . " element with $selector was not found.");
     }
 }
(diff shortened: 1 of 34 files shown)


Naktibalda:
Thanks.

Naktibalda:
@artmnv Please don't worry about merge conflicts, I will merge your commit to 2.5 and 2.6 branches.

I set default values of both parameters to `false` to avoid breaking existing tests.
In 2.6 I would like to change one of them to `true`, which one would you prefer - reset between tests or reset between requests?

artmnv:
@Naktibalda thank you!

I prefer `tests: true, request: false` options. But I don't think we should rely on my preferences in this question. I can dig and set custom properties by myself.
I think we should set default options for easy start for newcomers. It can be execution speed or reliability. But I can't imagine why would anyone use multiple requests in one test.

Naktibalda:
> But I can't imagine why would anyone use multiple requests in one test.

Wizard style multi-step interaction.
e.g. Add product to card, make payment.

Also login is a common first step for testing authenticated endpoints.

Naktibalda:
I cherry picked this change to 2.5 branch: https://github.com/Codeception/Codeception/commit/f4836229fe51ab8fb4fc42226101fb5c11999d57

PR #5409, opened 2019-02-26
Add possibility to dynamically load contexts

hansdubois:
This PR will add the possibility to dynamically load context for Gherkin tests. When using the suggested method of adding contexts to your ```codeception.yml``` file. This soon will become a long list of context files and quite a pain to maintain. 

By adding the following line to configuration contexts will automatically be loaded:

```yaml
gherkin:
    contexts:
        path: tests/_support/Steps
        namespace_prefix: Steps
        default:
            - AcceptanceTester
```

Unfortunately this cannot be solved via a Customer loader, because the GherkinSnippet class directly instantiates the Gherkin loader instead of the specified Gherkin loader. Something that could be fixed but with a lot more work.

### DIFF
--- a/docs/07-BDD.md
+++ b/docs/07-BDD.md
@@ -525,6 +525,17 @@ gherkin:
                 - "Step\FastLogin"
 ```
 
+Contexts can be autoloaded as well:
+```yaml
+gherkin:
+    contexts:
+        path: tests/_support/Steps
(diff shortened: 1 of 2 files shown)


DavertMik:
Thanks, looks good to me!
I will merge it as it should not break anything

Issue #5325, opened 2019-01-07
Update Docker to use PHP 7.3


OneEyedSpaceFish:
#### What are you trying to achieve?
Run tests in latest stable 7.3 php version

#### What do you get instead?
PHP 7.2 :)

This issue is pretty self explanatory.
Hopefully the migration should be easy enough as it would only contain updating the `docker from` to 7.3-cli, and pushing the new image to DockerHub.

This would need some testing locally, but the list of breaking changes for 7.3 were slim, so it's likely that everything will work out of the box as is. Obviously this will need checking first.

In terms of composer, the yml file already contains a `>=5.6.0 <8.0` constraint, so 7.3 version should already be included in there.

Carsak:
Does this issue actual?

OneEyedSpaceFish:
> Does this issue actual?
No. I've been on holiday.
That being said the docker image still doesn't build tho due to: #5384
Closing the issue now.

PR #5412, opened 2019-02-27
Docker image update [php-7.3, xdebug-2.7.0RC2]

OneEyedSpaceFish:
See (#5325)
Docker image updated:
 - updated `php-cli to 7.3`
 - updated `xdebug to 2.7.0RC2` (stable version rather than beta),
 - added `zip` functionality so composer could install from dist rather than source
 - added `classmap-authoritative` option to composer for extra performance
 - added `no-interaction` option to composer as without .lock files we'll pull latest version matched by .json definition, therefore build can fail any time if there's an interactive update [which is rare, but happens]

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,8 @@ RUN apt-get update && \
             git \
             zlib1g-dev \
             libssl-dev \
+            libzip-dev \
+            unzip \
         --no-install-recommends && \
         apt-get clean && \
         rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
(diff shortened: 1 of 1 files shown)


OneEyedSpaceFish:
Just in case; I tested this using two larger project of mine (one is low level api, one is front-end based).
Both seems to works fine.

Would you guys mind checking xdebug? (never used xdebug before, so I didn't check that).

Thanks. :)

OneEyedSpaceFish:
Shoot, I'll accidentally pushed a commit into my fork while the PR was still standing,
I'll revert https://github.com/Codeception/Codeception/pull/5412/commits/b499f2e0bedd3f739d2a07c4f6848e65c4b5eb52 soon-ish.

Naktibalda:
Please use feature branches in the future.

OneEyedSpaceFish:
Done, I'll use a feature branch for that other separate commit that relates to a different issue.

OneEyedSpaceFish:
Closing in favor of https://github.com/Codeception/Codeception/pull/5462

PR #5415, opened 2019-02-27
[FTP] Make sure _directory() will return its directory

bbatsche:
The `_directory()` method in `FTP` was missing its return statement, meaning all paths were being expanded from root, rather than the client's current working directory.


Naktibalda:
Thanks,
It is interesting that nobody noticed it in 5 years.
It is probably a breaking change, so I will merge it to 2.6

Naktibalda:
better later than never

PR #5422, opened 2019-03-02
2.6 Cleanup

Naktibalda:
* Deleted all HHVM specific code
* Remove workarounds used to make tests pass on PHP 5.4
* Dropped support for Guzzle 5 
* Removed deprecated setExpectedException method 
* Removed many methods that were deprecated long time ago


DavertMik:
I think everything is fine except 

```

---------
1) FriendWithStepsCept: Call friend with steps to ask expert work
 Test  tests/web/FriendWithStepsCept.php
                                                      
  [Error] Class '\WebGuy\RootWatcherSteps' not found  
```

In web tests

tests/web/_bootstrap.php
```diff
@@ -1,4 +1,4 @@
 <?php
 // Here you can initialize variables that will for your tests
 
-\Codeception\Util\Autoload::registerSuffix('Steps', __DIR__.DIRECTORY_SEPARATOR.'_steps');
+//\Codeception\Util\Autoload::registerSuffix('Steps', __DIR__.DIRECTORY_SEPARATOR.'_steps');
```
DavertMik: looks like this broke web tests

Naktibalda: That method was marked as deprecated.

Naktibalda: It would be better to rename `RootWatcherSteps` class to `\WebGuy\Steps\RootWatcher`, right?
In a similar manner as `FunctionalHelper` was renamed to `Helper\Functional` few versions ago.

DavertMik: Ok, I will try to take this

PR #5423, opened 2019-03-03
[WIP][Doctrine2] Change haveInRepository to invoke entities without constructor

### LINKED ISSUE #5420: Doctrine2 haveInRepository and Entities with constructors
I am trying to inject Entities into the repository. The entities have a constructor which accepts a single argument (but the issue can be reproduces with any number of arguments greater than or equal to 1).

#### What do you get instead?
I get the following error

```bash
 [ArgumentCountError] Too few arguments to function App\Entity\Foo::__construct(), 0 passed and exactly 1 expected 
```

```php
// FooTest.php

$I->haveInRepository(Foo::class, [
    'bar' => 'baz',
]);
```
```php
// src/Entity/Foo.php

namespace App\Entity;

class Foo
{
    private $bar;
 
    public function __construct(string $bar)
    {
        $this->bar = $bar;
    }

}

```
### Details

* Codeception version: 2.5.4
* PHP Version: 7.2.12
* Operating System: macOS Mojave 10.14.3
* Installation type: Composer
* Suite configuration:

```yml
// functional.suite.yml

actor: FunctionalTester
modules:
    enabled:
        - Asserts
        - REST
        - Doctrine2
        - \App\Tests\Helper\Functional
    config:
        Symfony:
            app_path: 'src'
            environment: 'test'
        REST:
            depends: Symfony
            url: '/v1'
        Doctrine2:
            depends: Symfony
```

Looking at the code it would seem that this line causes the exception https://github.com/Codeception/Codeception/blob/2.5/src/Codeception/Module/Doctrine2.php#L369

There are two solutions that I can think of. The first one would be to provide an optional third parameter to all doctrine2 module methods, which would be the optional constructor arguments that get passed into `newInstance()`, or even better `newInstanceArgs()` since this method wrapps a variable number of arguments into a single array.
The second solution is to use `newInstanceWithoutConstructor()`, which is what I think consumers really want in most cases, since they are setting the object properties directly via reflection. This approach has concerns regarding backward compatibility with existing tests though.

krukru:
This way you can persist entities which have a constructor that accepts
parameters. The down side is that you cannot relay on any side effects
from the constructor, but you shouldn't be doing that anyway.
Always set all expected fields using the second parameter `$data` of haveInRepository.

This is WIP, should this get accepted the docs also need to be updated to reflect this change.

There is a risk that tests which relied on constructor side-effect will no longer work with this change.
Since you have the possibility to set entity parameters via reflection and the primary concern of this method is storing entities for testing purposes, I think the constructor should have never been invoked in the first place.

Closes #5420.

### DIFF
--- a/src/Codeception/Module/Doctrine2.php
+++ b/src/Codeception/Module/Doctrine2.php
@@ -314,7 +314,7 @@ EOF;
     public function haveInRepository($entity, array $data)
     {
         $reflectedEntity = new \ReflectionClass($entity);
-        $entityObject = $reflectedEntity->newInstance();
+        $entityObject = $reflectedEntity->newInstanceWithoutConstructor();
         foreach ($reflectedEntity->getProperties() as $property) {
             /** @var $property \ReflectionProperty */
             if (!isset($data[$property->name])) {


krukru:
Agreed, it is impossible to change this now in this way. `persistEntity` is the way to go.
For reference: https://github.com/Codeception/Codeception/issues/5420#issuecomment-517902420

PR #5435, opened 2019-03-14
Adding Ability To Truncate Table Through Db Module

Fenikkusu:
I've had a few instances where I needed to truncate a table because the ORM is generating a lot of records. I cannot use the clean up because the DB has over 600 tables and this causes severe delays in processing. This gives the ability to call `truncate` easily. 

Note: I've mostly worked with MySQL and Sqlite. I'm not sure if the truncate command is different on the others (or if it even exists).

### DIFF
--- a/src/Codeception/Lib/Driver/Db.php
+++ b/src/Codeception/Lib/Driver/Db.php
@@ -378,4 +378,15 @@ class Db
     {
         return $this->options;
     }
+
+    /**
+     * Executes Truncate Against DB
+     * @param $tableName
+     *
+     * @throws \Codeception\Exception\ModuleException
+     */
+    public function truncateTable($tableName)
+    {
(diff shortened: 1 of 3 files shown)


Naktibalda:
Even in mysql TRUNCATE `cannot be performed for InnoDB tables with parent-child foreign key relationships.`, it would probably be better and compliant with all RDBMs to use `DELETE FROM table`;

https://dev.mysql.com/doc/refman/5.7/en/truncate-table.html

connorshea:
You could also just disable the foreign key constraints before truncating? (https://tableplus.io/blog/2018/08/mysql-how-to-temporarily-disable-foreign-key-constraints.html)

Naktibalda:
If someone wants to try to implement this feature again, please make a new pull request to https://github.com/Codeception/module-db

PR #5440, opened 2019-03-19
Using button formaction attr in proceedSubmitForm method

Dezzpil:
Using button formaction attr in proceedSubmitForm method (AcceptanceTesterActions::submitForm)

### DIFF
--- a/src/Codeception/Lib/InnerBrowser.php
+++ b/src/Codeception/Lib/InnerBrowser.php
@@ -822,6 +822,7 @@ class InnerBrowser extends Module implements Web, PageSourceSaver, ElementLocato
      */
     protected function proceedSubmitForm(Crawler $frmCrawl, array $params, $button = null)
     {
+        $url = null;
         $form = $this->getFormFor($frmCrawl);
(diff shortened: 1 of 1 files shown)


src/Codeception/Lib/InnerBrowser.php
```diff
@@ -834,10 +835,14 @@ protected function proceedSubmitForm(Crawler $frmCrawl, array $params, $button =
...
                 $requestParams[$button] = $btnCrawl->attr('value');
+                $formaction = $btnCrawl->attr('formaction');
+                if ($formaction) {
+                    $url = $formaction;
+                }
             }
         }
 
-        $url = $this->getFormUrl($frmCrawl);
+        $url = $url ? $url : $this->getFormUrl($frmCrawl);
```
Naktibalda: This expression could be shortened to `$url = $url ?: $this->getFormUrl($frmCrawl);`,
But I think that using `if` is a better style:
```php
if (!$url) {
    $url = $this->getFormUrl($frmCrawl);
}
```

PR #5442, opened 2019-03-20
Update FTP.php (key param is not needed for sftp type)

dimkanio:
for symphony sftp doesnt work
  [PHPUnit\Framework\Exception] file_get_contents(): Filename cannot be empty  

so key is not needed in this case

### DIFF
--- a/src/Codeception/Module/FTP.php
+++ b/src/Codeception/Module/FTP.php
@@ -117,7 +117,7 @@ class FTP extends Filesystem
         'timeout'  => 90,
         'user'     => 'anonymous',
         'password' => '',
-        //'key'      => '',
+        'key'      => '',
         'tmp'      => 'tests/_data',
         'passive'  => false,
         'cleanup'  => true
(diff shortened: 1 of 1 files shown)


src/Codeception/Module/FTP.php
```diff
@@ -82,7 +82,19 @@
  *              key: ''
  *              tmp: 'tests/_data/ftp'
  *              cleanup: false
+ * *  *  * #### Example (SFTP) for symphony
```
Naktibalda: what is this symphony thing?

dimkanio: framework symfony

Naktibalda: how is FTP module related to symfony framework?

dimkanio: i use codeception with this framework

DavertMik: I agree, symfony or symphony are not related here :)

Naktibalda:
The code change makes sense, it is just the description that didn't.

The problem this pull request fixes is that when the key field isn't set in suite configuration file,
it has the default empty string value and FTP module considers it as set and tries to load the file.
This issue can be fixed either by changing default value to `null` or by changing the code to consider empty strig as not set.

I will fix this issue in [module-ftp](https://github.com/Codeception/module-ftp)

Naktibalda:
I raised new pull request to module-ftp: https://github.com/Codeception/module-ftp/pull/1

PR #5444, opened 2019-03-20
Removed Facebook module

Naktibalda:
It was unmaintained for quite a few years and officially deprecated in January 2018.


DavertMik:
Ok!

PR #5445, opened 2019-03-20
Remove ZF1 module

Naktibalda:
Zend Framework 1 is dead since 2016

https://framework.zend.com/blog/2016-06-28-zf1-eol.html


DavertMik:
Ok!
I think we will name this version 3.0 then
To clarify the true breaking change

Naktibalda:
Not cool enough! This is maintenance release mostly.
If removal of these 2 modules make you want to bump a major version number, I would rather keep them for another year.

DavertMik:
I think the main reason for 3.0 is PHPUnit 8, let's be safe in update.

PR #5447, opened 2019-03-22
Add patch to reopen closed entity manager

krukru:
This patch is credited to @Amunak and can be found here https://github.com/Codeception/Codeception/issues/5107#issuecomment-430557559
Also relevant to the issue is this discussion https://github.com/symfony/symfony/issues/5339

This solves the issue of Entity Manager being closed (for example when a rollback is thrown in the application under test). The issue has been heavily discussed and the patch provided by @Amunak solved the issue for me.

Important note! 
For this patch to work, I also had to install the following `composer require --dev symfony/proxy-manager-bridge`, which I would presume requires Flex support.

### DIFF
--- a/src/Codeception/Module/Doctrine2.php
+++ b/src/Codeception/Module/Doctrine2.php
@@ -114,7 +114,7 @@ EOF;
 
             if($this->em->getConnection()->isTransactionActive()) {
                 try {
-                    $this->rollbackTransaction();
+                    $this->rollbackTransactions();
                     $this->debugSection('Database', 'Transaction cancelled; all changes reverted.');
                 } catch (\PDOException $e) {
                 }
@@ -135,7 +135,7 @@ EOF;
         }
         if ($this->config['cleanup'] && $this->em->getConnection()->isTransactionActive()) {
             try {
-                $this->rollbackTransaction();
+                $this->rollbackTransactions();
                 $this->debugSection('Database', 'Transaction cancelled; all changes reverted.');
             } catch (\PDOException $e) {
             }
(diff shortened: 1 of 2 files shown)


Amunak:
Beware that as @Dukecz [pointed out](https://github.com/Codeception/Codeception/issues/5107#issuecomment-431041330) in the linked thread this solution is still not enough in some cases where you have nested transactions and the inner ones fail.

You may wish to add their patch as well.

krukru:
@Amunak thanks for pointing that out! Most of Doctrine2 module already has this pattern applied, everywhere except in function `onReconfigure`. I added the same logic to this function as well. 
Hopefully this will not break anything.

DavertMik:
Sorry @krukru and @Amunak I still can't decide what to do with this patch. 

> For this patch to work, I also had to install the following composer require --dev symfony/proxy-manager-bridge, which I would presume requires Flex support.

This looks like a pretty strict requirement. Is there any way to make it without an additional package?

krukru:
If I understood why was this dependency important, maybe I could help out. Unfortunately I do not.
I also tried finding the original post where someone suggested adding this dependency, but cannot find it. 
Is there any other way of getting this feature to work?

felsspat:
Please do something with this.. If any tests fail, all following tests fail too, this is terrible :(

vincentwongso:
Are we able to merge this fix for the next release? Tested the patch seems to be working fine.

Naktibalda:
@krukru Could you rebase this branch on 3.1 branch?
`git pull --rebase 3.1`

It conflicts with later changes to Doctrine2 module.

krukru:
Hey sorry for the delayed response. I think I had deleted my fork quite some time ago and am unable to rebase the changes. I will open a new PR.
Just to be sure, you are telling me to merge it into branch 3.1?

Naktibalda:
Modules were separated from Codeception core, please make new pull requests to https://github.com/Codeception/module-symfony and https://github.com/Codeception/module-doctrine2

PR #5450, opened 2019-03-23
Console

DavertMik:
Replaced QuestionHelper from Symfony to HoaConsole for Hoa Console (which is better).
Added `pause()` to Actor for interactive shell


DavertMik:
Demo of interactive pause

codeception_pause

PR #5452, opened 2019-03-26
Add error output for non-fatal errors and non-loaded classes

nimestel:
As a result of these not-critical errors we can get message "Process finished with exit code 255" without any explanation. Error output makes debug easier. 
Sometimes errors can be duplicated, but most often it saves hours of life.

### DIFF
--- a/src/Codeception/Subscriber/ErrorHandler.php
+++ b/src/Codeception/Subscriber/ErrorHandler.php
@@ -107,8 +107,8 @@ class ErrorHandler implements EventSubscriberInterface
         }
         // not fatal
         if ($error['type'] > 1) {
-			echo "\n\n\nNot fatal error.\n";
-			echo sprintf("%s \nin %s:%d\n", $error['message'], $error['file'], $error['line']);
+            echo "\n\n\nNot fatal error.\n";
+            echo sprintf("%s \nin %s:%d\n", $error['message'], $error['file'], $error['line']);
             return;
         }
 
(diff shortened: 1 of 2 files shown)


Naktibalda:
> As a result of these not-critical errors we can take get message "Process finished with exit code 255" without any explanation.

If those errors end test execution, then the output of `non-fatal error` is misleading,
it would probably be better to remove that if statement completely and get fatal error output.

Could you provide an example of "non-fatal" error which ends test execution and example of output before and after your change?

nimestel:
@Naktibalda Your approach is much better :)

> example of "non-fatal" error which ends test execution

```php

public static function selectorFieldCell($field) {}

$this->selectorFieldCell($something)
```
Cannot make static method non static in class  

> example of output before and after your change

before
https://s.mail.ru/74So/j2Yd7kgsr

after 
https://s.mail.ru/Fxeu/BidAT9YeV

src/Codeception/Util/Autoload.php
```diff
@@ -123,6 +123,7 @@ public static function load($class)
             }
         }
 
+        echo "\nClass " . $class . "not loaded.\n";
```
Naktibalda: Please remove this `echo` statement.

src/Codeception/Subscriber/ErrorHandler.php
```diff
@@ -107,6 +107,8 @@ public function shutdownHandler()
         }
         // not fatal
         if ($error['type'] > 1) {
+            echo "\n\n\nNot fatal error.\n";
```
Naktibalda: Please remove this `if` block completely, because the error is actually fatal.

Naktibalda:
Please upgrade to 3.0.0 and check if your error is handled correctly.

PR #5455, opened 2019-03-28
[REST] Short API responses in debug mode

sebastianneubert:
- add a new config `shortDebugResponse` to limit large API responses at CLI in REST module

### DIFF
--- a/src/Codeception/Module.php
+++ b/src/Codeception/Module.php
@@ -295,7 +295,8 @@ abstract class Module
      * @param $chars
      * @return string
      */
-    protected function shortenMessage($message, $chars = 150) {
+    protected function shortenMessage($message, $chars = 150)
+    {
         return substr($message, 0, $chars) . "...";
     }
 
--- a/src/Codeception/Module.php
+++ b/src/Codeception/Module.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace Codeception;
 
 use Codeception\Exception\ModuleException;
@@ -17,7 +18,7 @@ use Codeception\Util\Shared\Asserts;
 abstract class Module
 {
     use Asserts;
-    
+
     /**
      * @var ModuleContainer
      */
@@ -176,7 +177,7 @@ abstract class Module
      */
     public function _getName()
     {
-        $moduleName = '\\'.get_class($this);
+        $moduleName = '\\' . get_class($this);
 
         if (strpos($moduleName, ModuleContainer::MODULE_NAMESPACE) === 0) {
             return substr($moduleName, strlen(ModuleContainer::MODULE_NAMESPACE));
@@ -297,7 +298,7 @@ abstract class Module
      */
     protected function shortenMessage($message, $chars = 150)
     {
-        return substr($message, 0, $chars) . "...";
+        return mb_substr($message, 0, $chars);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Codeception/Module.php
```diff
@@ -288,6 +288,17 @@ protected function debugSection($title, $message)
...
     }
 
+    /**
+     * Short text message to an amount of chars
+     *
+     * @param $message
+     * @param $chars
+     * @return string
+     */
+    protected function shortenMessage($message, $chars = 150) {
```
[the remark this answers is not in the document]
sebastianneubert: done

src/Codeception/Module.php
```diff
@@ -288,6 +288,18 @@ protected function debugSection($title, $message)
...
+    /**
+     * Short text message to an amount of chars
+     *
+     * @param $message
+     * @param $chars
+     * @return string
+     */
+    protected function shortenMessage($message, $chars = 150)
+    {
+        return substr($message, 0, $chars) . "...";
```
Naktibalda: 1. You should only add  `...` if string was longer than 150.
2. You should use mb_substr with `utf-8` charset instead of substr, because substr breaks utf-8 characters and Codeception doesn't work with anything else than utf-8. All `substr` and `strlen` calls were replaced with mb equivalents in the course of multiple pull requests to fix various issues.

A good similar example is at https://github.com/Codeception/Codeception/blob/2.5.5/src/Codeception/Step.php#L126-L127

src/Codeception/Module/REST.php
```diff
@@ -608,7 +615,20 @@ protected function execute($method, $url, $parameters = [], $files = [])
         if ($this->isBinaryData($printedResponse)) {
             $printedResponse = $this->binaryToDebugString($printedResponse);
         }
-        $this->debugSection("Response", $printedResponse);
+
+        $short = $this->_getConfig('shortDebugResponse');
+
+        if ($short !== null && $short !== false) {
+            if (!is_int($short) && $short <= 0) {
```
Naktibalda: I think that you wanted to use `||` instead of `&&` here.

src/Codeception/Module/REST.php
```diff
@@ -608,7 +615,20 @@ protected function execute($method, $url, $parameters = [], $files = [])
...
+                $short = $this->DEFAULT_SHORTEN_VALUE;
```
Naktibalda: It would be better to perform validation and normalization of `shortDebugResponse` parameter in `validateConfig` method.
You have to add such method to REST class and call `parent::validateConfig()` before doing your own logic.
https://github.com/Codeception/Codeception/blob/2.5.5/src/Codeception/Module.php#L148

@DavertMik is validateConfig the right place for this code?

DavertMik: I think yes

sebastianneubert: Sure, I try to solve all comments above next week and push a new MR.

sebastianneubert:
I switch back to "only int params" after getting in trouble of casting bool vars from config into integer. Thanks for reviewing the code.

sebastianneubert:
> Thanks for raising a pull request.
> Could you make a few changes?

I resolved all changes.

DavertMik:
Thanks, looks good.
But this config option is not documented. Could you add this parameter to the configuration section in REST?
https://github.com/Codeception/Codeception/blob/2.6/src/Codeception/Module/REST.php#L30

DavertMik:
Thanks, looks good. Merging!

Issue #5457, opened 2019-03-29
Gherkin Scenarios not loading when specified in group file


mozillalives:
#### What are you trying to achieve?

Load a specific scenario from a group file. For example, let's say we have the group file `tests/_groups/group1` and in that file the following.

```
tests/acceptance/refund.feature:Jeff returns a faulty microwave
````

#### What do you get instead?

No tests are found

### Details

* Codeception version: 2.5
* PHP Version: 7.1
* Operating System: MacOS
* Installation type: Phar || Composer
* List of installed packages (`composer show`)

```
behat/gherkin                      v4.6.0  Gherkin DSL parser for PHP 5.3
codeception/codeception            2.5.5   BDD-style testing framework
codeception/phpunit-wrapper        7.7.1   PHPUnit classes used by Codeception
codeception/stub                   2.1.0   Flexible Stub wrapper for PHPUnit's Mock Builder
doctrine/instantiator              1.2.0   A small, lightweight utility to instantiate objects in PHP without invoking their constructors
facebook/webdriver                 1.6.0   A PHP client for Selenium WebDriver
guzzlehttp/guzzle                  6.3.3   Guzzle is a PHP HTTP client library
guzzlehttp/promises                v1.3.1  Guzzle promises library
guzzlehttp/psr7                    1.5.2   PSR-7 message implementation that also provides common utility methods
myclabs/deep-copy                  1.8.1   Create deep copies (clones) of your objects
phar-io/manifest                   1.0.3   Component for reading phar.io manifest information from a PHP Archive (PHAR)
phar-io/version                    2.0.1   Library for handling version information and constraints
phpdocumentor/reflection-common    1.0.1   Common reflection classes used by phpdocumentor to reflect the code structure
phpdocumentor/reflection-docblock  4.3.0   With this component, a library can provide support for annotations via DocBlocks or otherwise retrieve information that is embedded in a DocBlock.
phpdocumentor/type-resolver        0.4.0
phpspec/prophecy                   1.8.0   Highly opinionated mocking framework for PHP 5.3+
phpunit/php-code-coverage          6.1.4   Library that provides collection, processing, and rendering functionality for PHP code coverage information.
phpunit/php-file-iterator          2.0.2   FilterIterator implementation that filters files based on a list of suffixes.
phpunit/php-text-template          1.2.1   Simple template engine.
phpunit/php-timer                  2.1.1   Utility class for timing
[...]

PR #5458, opened 2019-03-29
Proposed solution for #5457 gherkin scenarios not loaded from group file

mozillalives:
A possible solution for the issue described in #5457. I have a couple of open questions though.

* I am not familiar with the Codeception framework. Is `Codeception\Lib\GroupManager` the best place to address this issue or is there a better place?
* I would like to write some tests to verify this issue. Can someone point me to a similar test to help me get started?
* My use of `strcasecmp` looks like it will fail for multibyte strings. Are there suggestions on better alternatives? One I saw (that I linked to in a comment in the code) suggests using `mb_strtolower` instead.

Any help, tips or advice would be appreciated. Thanks.

### DIFF
--- a/src/Codeception/Lib/GroupManager.php
+++ b/src/Codeception/Lib/GroupManager.php
@@ -5,6 +5,7 @@ use Codeception\Configuration;
 use Codeception\Test\Interfaces\Reported;
 use Codeception\Test\Descriptor;
 use Codeception\TestInterface;
+use Codeception\Test\Gherkin;
 use Symfony\Component\Finder\Finder;
 use Symfony\Component\Finder\SplFileInfo;
 
@@ -118,10 +119,8 @@ class GroupManager
                 if (strpos($filename . ':' . $test->getName(false), $testPattern) === 0) {
                     $groups[] = $group;
                 }
-                // TODO: consider mb_strtolower per https://stackoverflow.com/a/5473569
-                if (method_exists($test, 'getMetadata')
-                    && strcasecmp($filename . ':' . $test->getMetadata()->getFeature(), $testPattern) === 0
-                ) {
+                if ($test instanceof Gherkin
+                    && mb_strtolower($filename . ':' . $test->getMetadata()->getFeature()) === mb_strtolower($testPattern)) {
                     $groups[] = $group;
                 }
                 if ($test instanceof \PHPUnit\Framework\TestSuite\DataProvider) {
--- a/src/Codeception/Lib/GroupManager.php
+++ b/src/Codeception/Lib/GroupManager.php
@@ -119,7 +119,10 @@ class GroupManager
                     $groups[] = $group;
                 }
                 // TODO: consider mb_strtolower per https://stackoverflow.com/a/5473569
-                if (strcasecmp($filename . ':' . $test->getMetadata()->getFeature(), $testPattern) === 0) {
+                if (
+                    method_exists($test, 'getMetadata')
+                    && strcasecmp($filename . ':' . $test->getMetadata()->getFeature(), $testPattern) === 0
+                ) {
                     $groups[] = $group;
                 }
                 if ($test instanceof \PHPUnit\Framework\TestSuite\DataProvider) {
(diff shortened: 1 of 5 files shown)


mozillalives:
okey-dokey - looking at the output I can see at least a couple of tests that are failing because of my change. I'll address those and push up a commit.

src/Codeception/Lib/GroupManager.php
```diff
@@ -118,6 +118,13 @@ public function groupsForTest(\PHPUnit\Framework\Test $test)
                 if (strpos($filename . ':' . $test->getName(false), $testPattern) === 0) {
                     $groups[] = $group;
                 }
+                // TODO: consider mb_strtolower per https://stackoverflow.com/a/5473569
+                if (
```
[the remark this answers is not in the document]
mozillalives: lol, ok. I will update that in another commit over the weekend

src/Codeception/Lib/GroupManager.php
```diff
@@ -118,6 +118,12 @@ public function groupsForTest(\PHPUnit\Framework\Test $test)
                 if (strpos($filename . ':' . $test->getName(false), $testPattern) === 0) {
                     $groups[] = $group;
                 }
+                // TODO: consider mb_strtolower per https://stackoverflow.com/a/5473569
+                if (method_exists($test, 'getMetadata')
+                    && strcasecmp($filename . ':' . $test->getMetadata()->getFeature(), $testPattern) === 0
```
edno: This needs to be UFT-8 compliant as put in your TODO
`strcasecmp($filename . ':' . $test->getMetadata()->getFeature(), $testPattern) === 0`to be replaced by `mb_strtolower($filename . ':' . $test->getMetadata()->getFeature()) === mb_strtolower($testPattern)`

src/Codeception/Lib/GroupManager.php
```diff
@@ -118,6 +118,12 @@ public function groupsForTest(\PHPUnit\Framework\Test $test)
                 if (strpos($filename . ':' . $test->getName(false), $testPattern) === 0) {
                     $groups[] = $group;
                 }
+                // TODO: consider mb_strtolower per https://stackoverflow.com/a/5473569
```
edno: Is there a case where `metadata()` method does not exist ?

src/Codeception/Lib/GroupManager.php
edno: Since current solution is focused on Gherkin feature, before using `$test->getMetadata()->getFeature()`, you should verify that the current Test object is a [Gherkin Test](https://github.com/Codeception/Codeception/blob/cc739e8a049309842e88d8f5389f47826549ca7f/src/Codeception/Test/Gherkin.php)

src/Codeception/Lib/GroupManager.php
mozillalives: In some tests it did not so I added the check to make sure it was there before I called it

src/Codeception/Lib/GroupManager.php
mozillalives: I can do that. Are there any other types of tests that might benefit from this comparison though?

mozillalives:
@edno I still have the bigger question of - what tests should I write to verify this? I'm thinking I need to verify GroupManager so I've been digging into that. I would definitely appreciate any help you could give though.

mozillalives:
@edno actually, I think I figured it out. I've added two tests that fail before my changes to GroupManager but pass after my changes. Do you see any issues with them?

edno:
I've not encountered this issue myself, but tests look good.
Until you don't break existing tests that should be 👍

mozillalives:
Awesome! If this is all good, can we get this merged? My team is waiting for this change to run our automated tests.

edno:
Not sure when PR will be merged. 
There're few of them in waiting, but I think that @Naktibalda and @DavertMik are quite busy with the next major releases.
In the meanwhile, you can use a workaround so your team will be using your codeception codebase instead of the official one by adapting your `composer.json` (not tested):
```json
{
  "repositories" : [
    {
      "type": "vcs",
      "url": "https://github.com/mozillalives/Codeception/tree/fix-gherkin-scenario-run"
    }
  ],
  "require-dev": {
    "codeception/codecetion":"2.5"
  }
}
```

DavertMik:
Thank you! Looks good to me

PR #5461, opened 2019-04-04
Update phpseclib Constant

bbatsche:
When switching to phpseclib version 2, all global constants were changed to class constants. This PR reflects that change as well.

Additionally, this PR updates the `$filepath` property in `FTP::_writeToFile()` to be the remote server's file path. This matches how `FTP::_openFile()` behaves.


DavertMik:
Thanks!

PR #5462, opened 2019-04-08
Updating docker PHP version to 7.3

### LINKED ISSUE #5325: Update Docker to use PHP 7.3

OneEyedSpaceFish:
Closes #5325

Docker image updated:

- updated php-cli to 7.3
- updated xdebug to 2.7.0RC2 (stable version rather than beta),
- added zip functionality so composer could install from dist rather than source
- added classmap-authoritative option to composer for extra performance
- added no-interaction option to composer as without .lock files we'll pull latest version matched by .json definition, therefore build can fail any time if there's an interactive update [which is rare, but happens]

As requested, this is the same change as https://github.com/Codeception/Codeception/pull/5412 , but this is branched from upstream 2.6.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM php:7.2-cli
+FROM php:7.3-cli
 
 MAINTAINER Tobias Munk tobias@diemeisterei.de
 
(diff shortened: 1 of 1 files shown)


DavertMik:
Thanks!

PR #5463, opened 2019-04-08
Adding timestamp information in recordings

### LINKED ISSUE #5239: Feature request: timestamp information in recordings
Our ops team are using the HTML reports and the generated recording to identify any outages that occurred in our system. Since this error is generated automatically and might send an alarm to operations if there are enough consecutive failures in a row. Our issue is that those reports are not generating any timestamps, so they can't easily debug issues (logs, etc.).

I made a custom reporter for them that outputs this information for them in the output, but ideally, this information would still be available in the reports.

* Codeception version: 2.5.0
* PHP Version: latest
* Operating System: debian (latest docker)
* Installation type: docker-compose

#### Suggestion
I had made a change to display this information instead of `Scroll up and down to see full page` step description, with a configurable boolean value to enable microsecond precision if it's needed.

Right now this is how it looks (this also contains my -just been merged - Recorder improvements)
screenshot_2018-10-23 recorder result

I just wanted to raise this as a potential improvement. The change is minimal, and in my opinion this feature would be really useful for other potential users as well.

Could someone from the CC team give me a heads-up if you'd like this merged?
If so, I'll create a new PR with my already done changes.

Thanks :)

OneEyedSpaceFish:
Added configurable timestamp reporting for each recorded step in Recorder extension.

Closes #5239

FYI: @DavertMik 
Just as a reminder it seems that the Recorder gif had not been updated in the website to the provided GIF.
(As mentioned in https://github.com/Codeception/Codeception/pull/5221)
Would you like me to generate a new one?

### DIFF
--- a/ext/Recorder.php
+++ b/ext/Recorder.php
@@ -39,6 +39,7 @@ use Codeception\Util\Template;
  * * `failure_color` (default: danger) - bootstrap values to be used for color representation for failed tests
  * * `error_color` (default: dark) - bootstrap values to be used for color representation for scenarios where there's an issue occurred while generating a recording
  * * `delete_orphaned` (default: false) - delete recording folders created via previous runs
+ * * `include_microseconds` (default: false) - enable microsecond precision for recorded step time details
  *
  * #### Examples:
  *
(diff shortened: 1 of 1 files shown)


DavertMik:
> Would you like me to generate a new one?

Yes, please! That would be cool!

Issue #5464, opened 2019-04-10
Code coverage still calls C3 URL even when "remote" for coverage is set to "false"


pgl:
#### What are you trying to achieve?

Run all tests, but exclude code coverage for remote (acceptance) tests by setting `remote: false` in `codeception.yml` under the coverage section.

#### What do you get instead?

> Provide console output if related. Use `-vvv` mode for more details.

```bash
[PETER-GA:peter]:~/src/pgl-symfony-demo $ vendor/bin/codecept run --coverage --coverage-xml --coverage-html
Codeception PHP Testing Framework v2.5.5
Powered by PHPUnit 7.5.8 by Sebastian Bergmann and contributors.
Running with seed:


App\Tests.acceptance Tests (2) -----------------------------------------------------------------------------------------------------------------------------------------------

In ErrorHandler.php line 83:

  file_get_contents(http://127.0.0.1:8000/c3/report/clear): failed to open stream: HTTP request failed! HTTP/1.0 500 Internal Server Error


run [-o|--override OVERRIDE] [-e|--ext EXT] [--report] [--html [HTML]] [--xml [XML]] [--phpunit-xml [PHPUNIT-XML]] [--tap [TAP]] [--json [JSON]] [--colors] [--no-colors] [--silent] [--steps] [-d|--debug] [--coverage [COVERAGE]] [--coverage-html [COVERAGE-HTML]] [--coverage-xml [COVERAGE-XML]] [--coverage-text [COVERAGE-TEXT]] [--coverage-crap4j [COVERAGE-CRAP4J]] [--coverage-phpunit [COVERAGE-PHPUNIT]] [--no-exit] [-g|--group GROUP] [-s|--skip SKIP] [-x|--skip-group SKIP-GROUP] [--env ENV] [-f|--fail-fast] [--no-rebuild] [--seed SEED] [--] [<suite> [<test>]]

PHP Fatal error:  Uncaught RuntimeException: Command Did Not Finish Properly in /mnt/c/Users/Peter/src/pgl-symfony-demo/vendor/codeception/codeception/src/Codeception/Subscriber/ErrorHandler.php:101
Stack trace:
#0 [internal function]: Codeception\Subscriber\ErrorHandler->shutdownHandler()
#1 {main}
  thrown in /mnt/c/Users/Peter/src/pgl-symfony-demo/vendor/codeception/codeception/src/Codeception/Subscriber/ErrorHandler.php on line 101
PHP Stack trace:
[...]

PR #5465, opened 2019-04-10
Fix for issue #5464 - moved remote c3 calls for code coverage to RemoteServer

pgl:
Fix for issue #5464


Naktibalda:
Not a bug.

pgl:
The Semaphore test failed on the test `WebDriverTest:testChangingCapabilities`. I've run this test locally a few times and tried to figure out what might be wrong, but I really can't figure out what might be the issue.

Any help would be appreciated.

pgl:
Oh... how comes it was closed?

Naktibalda:
You raised #5489 which is still open.

pgl:
OK, thanks. :)

PR #5467, opened 2019-04-10
Fixing Module.php constuctor and getConfig phpdoc parameter definitions

### LINKED ISSUE #5466: Faulty `null` phpdoc parameter signatures in `Module.php`
Run phpstan on my project with lvl7 without any warnings and errors.

#### What do you get instead?
Two types of issues discovered so far both located in `Module.php`
Both issues are related to a faulty phpdoc signature given for parameters.
- `Parameter #1 $key of method Codeception\Module::_getConfig() expects null, string given.`
- `Parameter #2 $config of method Codeception\Module::__construct() expects null, array given.` 

The parameters are optional with a default `null` value.
The parameter definition states only null.
This is not really the case tho, as in both occasions (array in first example, mixed in the second one) null is only the default value.

### Details
* Codeception version: 2.5.5
* PHP Version: 7.2.11
* Operating System: Docker
* Installation type: Docker pull

OneEyedSpaceFish:
Fixes #5466

### DIFF
--- a/src/Codeception/Module.php
+++ b/src/Codeception/Module.php
@@ -345,8 +345,8 @@ abstract class Module
     /**
      * Get config values or specific config item.
      *
-     * @param mixed|null $key
-     * @return array|mixed|null
+     * @param mixed $key
+     * @return mixed the config item's value or null if it doesn't exist
      */
     public function _getConfig($key = null)
     {
--- a/src/Codeception/Module.php
+++ b/src/Codeception/Module.php
@@ -66,7 +66,7 @@ abstract class Module
      * Requires module container (to provide access between modules of suite) and config.
      *
      * @param ModuleContainer $moduleContainer
-     * @param null $config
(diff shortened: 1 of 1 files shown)


src/Codeception/Module.php
```diff
@@ -345,7 +345,7 @@ protected function getModule($name)
     /**
      * Get config values or specific config item.
      *
-     * @param null $key
+     * @param mixed|null $key
```
Naktibalda: Is there any point to keep |null after mixed?
I think that this line and return line can be simplified to mixed type.
`@return array|mixed|null` -> `@return mixed`

OneEyedSpaceFish: 1/ maybe not
2/ agreed

OneEyedSpaceFish:
I am not sure why the Semaphore build failed, I assume it has nothing to do with the changes?

PR #5470, opened 2019-04-16
Avoid removing required fields in cookies.

JorisVanEijden:
A cookie with ['name' => 'test', 'value' => 0] would get it's value stripped, causing an "undefined index: value" notice in php-webdriver's Cookie.php


DavertMik:
Thanks!

PR #5471, opened 2019-04-16
Update WebDriver.md - typo in example

bchatard:
Missing `'` in an exemple which break the style


Naktibalda:
This file is generated automatically, please update the source file instead: https://github.com/Codeception/Codeception/blob/2.5/src/Codeception/Module/WebDriver.php#L2122

bchatard:
ok sorry, I will fix that

Issue #5473, opened 2019-04-17
REST module 'execute()' prefixes slash to url which already have leading slash


nicholascus:
#### What are you trying to achieve?
Consistency in URL construction when both PhpBrowser and REST module are used.

#### What do you get instead?

Requests sent with REST module get leading slash doubled. So server get log full of
192.168.176.9 - - [17/Apr/2019:01:05:58 +1000] "POST //

```yml
actor: FunctionalTester
modules:
    enabled:
        - REST:
            url: 'http://%HOST%:%PORT%/'
            depends: PhpBrowser
        - \Helper\Functional
        - PhpBrowser:
            url: 'http://%HOST%:%PORT%/'
        - Asserts
```

Naktibalda:
The patch was inconsistent with bug report and it looks like it didn't fix anything.
Example of configuration above uses absolute urls starting with `http://`,
but the code only prepends `/` if there is no `://` in configured url - `if (strpos($url, '://') === false) {`

Provided patch checks the same byte in a different way which makes no difference.

```php
$url = $this->config['url'] . $url;
strpos($url, '/') !== 0 && $this->config['url'][0] !== '/'
```

nmihaiv:
If the config url doesn't contain http://, then the request link will have a / in front of it.

Naktibalda:
Fixed by #5511

PR #5474, opened 2019-04-17
Test and  do not add leading slash when there is one already #5473

nicholascus:


DavertMik:
I waited a long for this PR!

src/Codeception/Module/REST.php
```diff
@@ -577,7 +577,7 @@ protected function execute($method, $url, $parameters = [], $files = [])
         // allow full url to be requested
         if (strpos($url, '://') === false) {
             $url = $this->config['url'] . $url;
-            if ($this->config['url'] && strpos($url, '://') === false && $this->config['url'][0] !== '/') {
+            if ($this->config['url'] && strpos($url, '://') === false && strpos($url, '/') !== 0 && $this->config['url'][0] !== '/') {
```
Naktibalda: Are you sure that this fix was correct?
Previous line overrides `$url` so the first byte of `$url` and `$this->config['url']` is the same.

nicholascus: I'll check and redo that if required.

nicholascus: https://github.com/Codeception/Codeception/pull/5511

PR #5480, opened 2019-04-23
2.5 Fixed loadSessionSnapshot with php-webdriver 1.1.3

### LINKED ISSUE #5456: Error using an array as object in the loading Session Snapshot
When codeception try to load a session snapshot previously stored in other test, the code fail because the system except that the property "$this->sessionSnapshots[$name]" is filled with instances of some class when is filled with associative arrays

#### What do you get instead?

FATAL ERROR. TESTS NOT FINISHED.

Call to a member function getName() on array 
in /var/www/html/vendor/codeception/codeception/src/Codeception/Module/WebDriver.php:3108

A few lines before we can see the following code (lines 3081  to 3089):

```php
foreach ($this->webDriver->manage()->getCookies() as $cookie) {
            if (in_array(trim($cookie['name']), [LocalServer::COVERAGE_COOKIE, LocalServer::COVERAGE_COOKIE_ERROR])) {
                continue;
            }

            if ($this->cookieDomainMatchesConfigUrl($cookie)) {
                $this->sessionSnapshots[$name][] = $cookie;
            }
        }
```

The property sessionSnapshots is filled with the variable $cookie, provided by $this->webDriver->manage()->getCookies() and the getCookies return an associative array.

```php
    /**
     * Get all the cookies for the current domain.
     *
     * @return array The array of cookies presented.
     */
    public function getCookies()
    {
        return $this->executor->execute(DriverCommand::GET_ALL_COOKIES);
    }
```
### Details

* Codeception version: 2.5.5
* PHP Version: 5.6.39
* Installation type: Composer

Naktibalda:
It is a minor issue which could affect a handful of Codeception 2.5 users
Closes #5456

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -3109,7 +3109,7 @@ class WebDriver extends CodeceptionModule implements
         }
         
         foreach ($this->sessionSnapshots[$name] as $cookie) {
-            $this->setCookie($cookie['name'], $cookie['value'], $cookie);
+            $this->setCookie($cookie['name'], $cookie['value'], (array)$cookie);
         }
         $this->debugSection('Snapshot', "Restored \"$name\" session snapshot");
         return true;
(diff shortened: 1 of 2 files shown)


DavertMik:
Unfortunately, this implementation breaks CI

Naktibalda:
Fixed

PR #5489, opened 2019-04-25
Fix for issue #5464 - moves calls to remote server from LocalServer.php to RemoteServer.php

pgl:
This basically just removes the calls to a remote server over HTTP from `LocalServer.php` to `RemoteServer.php`. Fix for #5464.


pgl:
NB: I've left the original PR open, since it's a valid fix for the 2.5 branch.

DavertMik:
LocalServer means a locally running webserver which is also communicated via HTTP requests.
As @Naktibalda described it previously, `remote: false` is about merging code coverage requests.
However, you can just disable codecoverage report for the acceptance suite if you don't need it.

Issue #2638, opened 2015-12-17
Provide a way to check if debug mode is active for modules


FractalizeR:
I'm writting some module helper and I have a heavy operation, which makes sense only if CodeCeption is run in debug mode.

`\Codeception\Util\Debug` class has its `$output` property not null only in debug mode, but there is no getter for it. And this check is indirect.

Could you please provide a better way to do that?

jefersonralmeida:
It makes sense since sometimes you inject an external library that can have a debug mode, and just makes sense to enable it when debug is enabled.

jefersonralmeida:
Working on a PR to implement it.

jefersonralmeida:
PR created: https://github.com/Codeception/Codeception/pull/5494

Naktibalda:
`\Codeception\Util\Debug::isEnabled()` method was implemented in 2.5.0 

https://github.com/Codeception/Codeception/commit/1d4bb38d1a5ac852ddb8a35f7aa88c18180176df#diff-259b7eb57d4f5a6c3e125568a2989589R55

PR #5494, opened 2019-05-03
Feature/2638 debug flag

jefersonralmeida:
Fix for issue [2638](https://github.com/Codeception/Codeception/issues/2638).


autoload.php
```diff
@@ -42,6 +42,16 @@ function codecept_debug($data)
     }
 }
 
+if (!function_exists('codecept_debug_mode')) {
```
Naktibalda: Wrong function name used in this condition.

jefersonralmeida: Ok. I'll create a new branch based on 3.0 and place another PR (fixing this function name issue).

jefersonralmeida: New PR created: https://github.com/Codeception/Codeception/pull/5502

PR #5502, opened 2019-05-09
Fix for issue 2638.

jefersonralmeida:
Fix for issue #2638


jefersonralmeida:
I really don't think that this `semaphoreci` build not passing is because of my change, since it's breaking on the acceptance tests for web interface, and I just changed the cli behavior.

DavertMik:
Sorry, I didn't get it. Why just `Codeception\Util\Debug::isEnabled` doesn't work for you?

Naktibalda:
`\Codeception\Util\Debug::isEnabled()` method was implemented in 2.5.0 

https://github.com/Codeception/Codeception/commit/1d4bb38d1a5ac852ddb8a35f7aa88c18180176df#diff-259b7eb57d4f5a6c3e125568a2989589R55

PR #5525, opened 2019-05-24
Update assertion method in seeInShellOutput

KartaviK:
Old assertion throws deprecation warning


Naktibalda:
dontSeeInShellOutput had the same issue.

PR #5526, opened 2019-05-24
Add —pause-on-fail option

mneuhaus:
before 3.0 i used to use a pauseExecution in the _failed hook combined with --fail-fast to have a way to jump in and debug when a Testcase fails. 
3.0 changed pauseExecution to pause which introduces the interactive shell, which is nice.
But combined with _failed that pause call is not as helpful anymore, because 3.0 does not output the failure information to the console anymore, so it's difficult to analyse what actually fails.
Sadly, i didn't find an easy way to have that behavior again in 3.0, so i looked around and came to a new option "--pause-on-fail" which fails as soon as a failure/error is encountered, prints the failure/error information to the console and then uses the pause trait to start that interactive shell.

### DIFF
--- a/src/Codeception/Codecept.php
+++ b/src/Codeception/Codecept.php
@@ -58,6 +58,7 @@ class Codecept
         'filter'          => null,
         'env'             => null,
         'fail-fast'       => false,
+        'pause-on-fail'   => false,
         'ansi'            => true,
         'verbosity'       => 1,
         'interactive'     => true,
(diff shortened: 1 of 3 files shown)


DavertMik:
That's nice, thank you! Just checked, you are right. This must behavior must be improved. 
But could you make it as an extension? This way it will be decoupled from internal core classes, and won't bring more options to the runner.

You can still execute it like:

```
codecept run --ext pauseOnFail
```

which should be fine.

DavertMik:
Sorry for the delay replying it. I really like the PR btw.

Naktibalda:
As @DavertMik wrote, please implement it as extension.

PR #5539, opened 2019-06-07
Ability to use a custom cookie domain for code coverage

### LINKED ISSUE #5536: Code coverage not being generated for subdomains (acceptance suite)

I'm trying to get code coverage for our entire suite of acceptance tests, which consists of tests doing their work across several subdomains (admin, api, ussd, mobi etc).

#### What do you get instead?

Code coverage is created only for the subdomain configured in `WebDriver` module's section, or for the one for which the `c3_url` is configured (if different from `WebDriver/url`).

Note that c3.php is included in every subdomain's code: each `index.php` of a subdomain is including a `common.php` which includes `c3.php`.

I've created a dummy test whitelisting only index files from 2 subdomains to show it in pictures.

Example 1:
- `c3_url: https://admin-test.mysite.com`
- Running an acceptance test that's visiting `admin-test` and `ussd-test` subdomains.
- Result: Code coverage generated only for requests hitting `https://admin-test.mysite.com`:

Screenshot_2019-06-06_at_14_54_51

Example 2:
- `c3_url: https://ussd-test.mysite.com`
- Running the same acceptance test from Example 1
- Result: Code coverage generated only for requests hitting `https://ussd-test.mysite.com`:

Screenshot_2019-06-06_at_14_53_33

...and so on, you get the pattern: code coverage doesn't get generated for other subdomains.

### Summary:
Code coverage is not triggered at all for those other subdomains. I can see that pages load fast on a subdomain not being examined for code coverage (`--steps`), while the pages being examined by code coverage are slow as expected.

I expected code coverage to work regardless of subdomains that are being tested. Okay, maybe tell Codeception that `c3_url` is "this" one so it can do it's work there without confusion, but when performing requests within the tests - why not do the same thing to all (sub)domains instead of just 1?

### Details

* Codeception version: v3.0.1
* PHP Version: 7.2.17
* Operating System: Ubuntu 18.04.1 LTS
* Installation type: Composer

maksimovic:
Fixes #5536 
(Depends on #5541 for builds to pass)

With this modification, I get the coverage triggered on both subdomains that my dummy cest visits:
Screenshot_2019-06-07_at_14_42_35

Now, I did want to add a test for this, but I got lost in all the existing tests and underlying logic for them. This would most likely need a custom vhost on travis + new server listening on it + either new suite config or modifying an existing one + cest which would see the cookie. I don't know if it's worth the hassle, but if a test is required for this then I can spend some more time on figuring out how to add this (and not to wreak havoc in the whole Codeception's suite of tests).


maksimovic:
As of PhpUnit 8.2, there is no more `\PHPUnit\Framework\TestSuite::isTestMethod()` and that's why tests are failing (branch `3.0` is failing as well,[ here's a copy of it being tested by Travis](https://travis-ci.org/maksimovic/Codeception/builds/542747583))

The method has been moved over to `\PHPUnit\Util\Test::isTestMethod()`, diff:
https://github.com/sebastianbergmann/phpunit/commit/bde8364d5b5bd3aaca367a3b77bbc16b5bdf049d

Naktibalda:
isTestMethod issue was reported at #5541

Naktibalda:
Code change looks good to me,
Please document this new parameter in https://github.com/Codeception/Codeception/blob/3.0/docs/11-Codecoverage.md#remote-codecoverage

maksimovic:
Shouldn't this be in~n~ (pardon the "improved" MacBook keyboard) https://github.com/Codeception/Codeception/blob/3.0/docs/11-Codecoverage.md#configuration since it's not exclusively for `remote=true` config?

Naktibalda:
I think that it should be in Remote Coverage Local Server section, next to documention of c3_url.

Naktibalda:
Please merge changes from Codeception 3.0 branch (or rebase) to get travis build passing on PHP 7.2

PR #5540, opened 2019-06-07
Update docs: replace old key "actor" with current "actor_suffix"

splinter89:
Config key has been renamed in e45a779


DavertMik:
Thanks!

PR #5552, opened 2019-06-15
Add check for console class before calling it

carnage:
Zf2.5+ is modular, can't always rely that the console package is installed.


Naktibalda:
Thanks

PR #5554, opened 2019-06-16
Retry: double wait interval after each execution

### LINKED ISSUE #5522: Retry interval comment is misleading about doubling the time
This comment line in `\Codeception\Lib\Actor\Shared\Retry` is misleading:
> Interval will be doubled on each unsuccessful execution.
https://github.com/Codeception/Codeception/blob/52dfbb5f31b74d042100a8836bbde792326ebb64/src/Codeception/Lib/Actor/Shared/Retry.php#L12

Actual implementation just waits given interval X times without changing the interval, see `\Codeception\Step\Retry::run()` method:
https://github.com/Codeception/Codeception/blob/52dfbb5f31b74d042100a8836bbde792326ebb64/src/Codeception/Step/Retry.php#L53
```
    public function run(ModuleContainer $container = null)
    {
        $retry = 0;
        $interval = $this->retryInterval;
        while (true) {
            try {
                return parent::run($container);
            } catch (\Exception $e) {
                $retry++;
                if ($retry > $this->retryNum) {
                    throw $e;
                }
                codecept_debug("Retrying #$retry in ${interval}ms");
                usleep($interval * 1000);
            }
        }
    }```

Naktibalda:
Fixes #5522


Naktibalda:
Appveyor is failing everywhere.

PR #5560, opened 2019-06-28
Add make commands to Laravel5 and Lumen modules

ibpavlov:
Add `make` and `makeMultiple` commands to [Laravel5] and [Lumen] modules. 
They will act like factory()->make()

### DIFF
--- a/src/Codeception/Module/Laravel5.php
+++ b/src/Codeception/Module/Laravel5.php
@@ -1167,6 +1167,63 @@ class Laravel5 extends Framework implements ActiveRecord, PartedModule
             $this->fail("Could not create model: \n\n" . get_class($e) . "\n\n" . $e->getMessage());
         }
     }
+    
+    /**
+     * Use Lumen's model factory to make a model instance.
+     * Can only be used with Lumen 5.1 and later.
+     *
+     * ``` php
+     * <?php
+     * $I->make('App\User');
+     * $I->make('App\User', ['name' => 'John Doe']);
+     * $I->make('App\User', [], 'admin');
+     * ?>
+     * ```
+     *
+     * @see https://lumen.laravel.com/docs/master/testing#model-factories
+     * @param string $model
+     * @param array $attributes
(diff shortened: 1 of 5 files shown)


src/Codeception/Module/Laravel5.php
```diff
@@ -1167,6 +1167,63 @@ public function haveMultiple($model, $times, $attributes = [], $name = 'default'
...
+     *
+     * ``` php
+     * <?php
+     * $I->make('App\User');
+     * $I->make('App\User', ['name' => 'John Doe']);
+     * $I->make('App\User', [], 'admin');
+     * ?>
+     * ```
+     *
+     * @see https://lumen.laravel.com/docs/master/testing#model-factories
```
Naktibalda: You left Lumen links in Laravel docblocks. Please fix them.
markdown files are regenerated by release process, so the changes you made there will be lost if they aren't the same as the changes in code files.

ibpavlov: Ok. Thank I'll fix them. 
For your previous comment: I've added that chamge to the DataFactory, but for Laravel it utilizes their factories.

src/Codeception/Module/Laravel5.php
```diff
@@ -1167,6 +1167,63 @@ public function haveMultiple($model, $times, $attributes = [], $name = 'default'
...
+     *
+     * ``` php
+     * <?php
+     * $I->make('App\User');
+     * $I->make('App\User', ['name' => 'John Doe']);
+     * $I->make('App\User', [], 'admin');
+     * ?>
+     * ```
+     *
+     * @see http://laravel.com/docs/5.1/testing#model-factories
```
DavertMik: I would ask you to write a bit more how data factories are defined. Link to laravel website is not enough as it doesn't say how to use it with codeception

Naktibalda:
@ibpavlov It would be great if you could make another pull request with more documentation about these methods.

PR #5562, opened 2019-07-03
Fix makeHtmlSnapshot step usage documentation

reinholdfuereder:


reinholdfuereder:
@Naktibalda That approval was more than quick :-)

AppVeyor build failure seems to be a general problem (that I dare to ignore)!?

Naktibalda:
Yes, AppVeyor was failing for a month.
I fixed it with #5563

reinholdfuereder:
Thanks a lot for both fixes!

PR #5564, opened 2019-07-04
Fix "conditional assertions" doc: enabled by default

reinholdfuereder:
Cf. https://github.com/Codeception/Codeception/blob/582ffe3c9747875062f8569e71694b04cfe6bc42/src/Codeception/Configuration.php#L114

I think current documentation is not correct or at least a bit misleading. Frankly I have doubts that my suggested change is ideal though, maybe it can be still improved...

Question concerning the documentation code comment (right afterwards) "_# or in codeception.yml insides suites section_":
* I naively tried and failed with the following approach in "codeception.yml":
```
...
extensions:
    ...
suites:
    step_decorators:
        - \Codeception\Step\Retry
```
* How is that supposed to work?

### DIFF
--- a/docs/03-AcceptanceTests.md
+++ b/docs/03-AcceptanceTests.md
@@ -273,7 +273,7 @@ $I->cantSeeInField('user[name]', 'Miles');
 
 Each failed assertion will be shown in the test results, but it won't stop the test.
 
-Conditional assertions are disabled in bootstrap setup. To enable them you should add corresponding step decorators to suite config:
+Conditional assertions are enabled by default suite config step decorator settings (unless overridden):
 
 > If you started project as `codecept init acceptance` they should be already enabled in config
 


reinholdfuereder:
Hm, in https://codeception.com/04-24-2019/codeception-3.0.html it says "_We also disabled conditional assertions for new setups (as people regularly misuse them), so if you want to use ``canSee`` methods you will need to enable them._"
* I have to actually confirm this message/statement wit respect to "misuse", but I think they are actually enabled by default.

reinholdfuereder:
Hm (number 2): or should actually the code be fixed (to match the current documentation and the Codeception v3 release blog)?

DavertMik:
Ok, I got your point.
This line in configuration you refer to exists for backwards compatibility (previous defaults). When a new project is created (bootstrap) step decorators are explicitly defined in the new config. And they are set to null.

Naktibalda:
If there is something to fix, please update https://github.com/Codeception/codeception.github.com/blob/master/docs/03-AcceptanceTests.md instead.

PR #5566, opened 2019-07-04
add support for ZF3

svycka:
should work when  https://github.com/Naktibalda/codeception-zf2-tests/pull/10 will be fixed


Naktibalda:
Lets see if it works, if it doesn't, I will rollback this merge.

Naktibalda:
It works with ZF3 and ZF 2.5.

PR #5583, opened 2019-07-15
Fix: mismatch between text and code example

richardbrinkman:
In the text Firefox is mentioned whereas in the code sample below Chrome is used.


Naktibalda:
I changed target branch to 3.0, because website is currently generated from that branch and 2.2 is no longer maintained.

PR #5588, opened 2019-07-17
Add runSymfonyConsoleCommand

wirwolf:


Naktibalda:
I am not sure if this method belongs to Symfony module which simulates accessing Symfony web applications over HTTP.

wirwolf:
Not over HTTP. I test console command by this function

src/Codeception/Module/Symfony.php
```diff
@@ -532,6 +532,32 @@ public function grabService($service)
         }
         return $container->get($service);
     }
+    
+     /**
+     * @param string  $command
```
DavertMik: Please add a valid description with a code example here in docblock

wirwolf: fixed

DavertMik:
Let's make it!

Ravetracer:
I just tried this feature and in the first approach it failed.
After some investigation I realized, that some use statements are missing from the Symfony.php module:

use Symfony\Bundle\FrameworkBundle\Console\Application;
use Symfony\Component\Console\Input\ArrayInput;
use Symfony\Component\Console\Output\BufferedOutput;

I just upgraded the version of Codeception in composer and updated then. 
BTW: I'm using Symfony 4.2.

wirwolf:
Yes, i found this bug today in my project.

PR #5593, opened 2019-07-18
Update extension example code

Nebulosar:
Second value in `$events` array did hold `SUITE_BEFORE`, while the function is `beforeTest` and calls for a `TestEvent`

previous code threw error:
```
Error: Argument 1 passed to MyCustomExtension::beforeTest() must  be an instance of 
Codeception\Event\TestEvent, instance of Codeception\Event\SuiteEvent given
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -55,7 +55,9 @@
         "league/factory-muffin": "For DataFactory module",
         "league/factory-muffin-faker": "For Faker support in DataFactory module",
         "symfony/phpunit-bridge": "For phpunit-bridge support",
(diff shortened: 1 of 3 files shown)


Nebulosar:
Sorry for that

Naktibalda:
It isn't as simple as changing base branch because after changing base branch 2 additional commits appear in the diff.

Naktibalda:
I cherry-picked it: https://github.com/Codeception/Codeception/commit/cdb2236bbfa9ab3ed821138f8e90f23c10c7fe08

PR #5598, opened 2019-07-20
[Doctrine2] add fixtures support [obsoleted by #5629]

### LINKED ISSUE #5594: [Doctrine2] Integrate fixtures
Doctrine has some mechanisms to load fixtures (specific fixture format, suitable for Doctrine -- classes that implement certain interface, all nicely bundled in official package `doctrine/data-fixtures`; same package is used as part official fixtures bundle for Symfony). I think it would be nice to have this functionality in `Doctrine2` module out of the box. The code is quite simple, but might be tedious to add to every project again and again:

```php
$loader = new \Doctrine\Common\DataFixtures\Loader();
$loader->addFixture($fixture);
	
$purger = new \Doctrine\Common\DataFixtures\Purger\ORMPurger($this->em);
$executor = new \Doctrine\Common\DataFixtures\Executor\ORMExecutor($this->em, $purger);
$executor->execute($loader->getFixtures());
```

I have prepared some sample implementation here: https://github.com/alexkunin/Codeception/commit/a839aa7b41b31e303252f58d055c60d5f9a18e53. It includes simple test, but lacks documentation at this point, and also might need a check if Doctrine fixtures package is actually installed.

What do you think? Worth adding to core Codeception repo?

What about method name, `haveFixtureLoaded`? I've seen `haveFixtures` somewhere, not sure which one is better. Also, maybe two methods, one for single fixture, another one for array? Or just detect argument type and use plural in method name?

alexkunin:
Closes #5594.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ matrix:
   - php: 7.2
     env: FRAMEWORK=Yii2 TEST_REPO="https://github.com/Codeception/yii2-tests"
 addons:
-  postgresql: "9.2"
+  postgresql: "9.6"
 
 branches:
   except:
@@ -65,6 +65,7 @@ services:
   - rabbitmq
   - postgresql
   - redis
+  - mysql
 
 before_install:
   - '$(php PruneTest.php)'
--- a/composer.json
+++ b/composer.json
@@ -45,7 +45,8 @@
         "vlucas/phpdotenv": "^3.0",
         "symfony/process": ">=2.7 <5.0",
         "doctrine/orm": "^2",
-        "doctrine/annotations": "^1"
+        "doctrine/annotations": "^1",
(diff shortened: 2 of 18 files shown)


alexkunin:
Accidentally added unintended code formatting changes. Fixed that by force-pushing version with corrected whitespace.

alexkunin:
PHP 5.6 fails, but only because it uses old PHPUnit 5.7.27 which has following quirk: if you do `expectExceptionMessageRegExp()` without doing `expectException()`, exception is not handled, instead it is reported as a failure within test. More recent versions of PHPUnit allow to skip `expectException()`.

Naktibalda:
Please rebase this pull request on 3.0 too.

alexkunin:
Commit log got messed up heavily, so, here is new PR: #5629.

PR #5601, opened 2019-07-22
ErrorHandler: add test for early exits

### LINKED ISSUE #5595: What is the purpose of Subscriber\ErrorHandler? Can we document it?
Hi, whenever using a simple `var_dump($var);exit;` in a test to do a trivial debugging, the output of a run is bloated by `ErrorHandler::shutdownHandler` output, that is registered as a shutdown function in the `SUITE_BEFORE` event.

```
$ vendor/bin/codecept run common/tests/unit/MyTest.php 
Codeception PHP Testing Framework v2.4.5
Powered by PHPUnit 7.5.6 by Sebastian Bergmann and contributors.

Common\tests\_support.unit Tests (1) --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
- MyTest: What
int(1)

PHP Fatal error:  Uncaught RuntimeException: Command Did Not Finish Properly in ./vendor/codeception/base/src/Codeception/Subscriber/ErrorHandler.php:101
Stack trace:
#0 [internal function]: Codeception\Subscriber\ErrorHandler->shutdownHandler()
#1 {main}
  thrown in ./vendor/codeception/base/src/Codeception/Subscriber/ErrorHandler.php on line 101

Fatal error: Uncaught RuntimeException: Command Did Not Finish Properly in ./vendor/codeception/base/src/Codeception/Subscriber/ErrorHandler.php:101
Stack trace:
#0 [internal function]: Codeception\Subscriber\ErrorHandler->shutdownHandler()
#1 {main}
  thrown in ./vendor/codeception/base/src/Codeception/Subscriber/ErrorHandler.php on line 101
```

What is the exact purpose of this?
Is all my code converts errors to exceptions, could I get rid of this handler?
Would you accept a PR to avoid subscribing this listener with a proper config in the `codeception.yml` file?

Slamdunk:
Closes #5595

Add test for #4604 /cc @Fenikkusu

### DIFF
--- a/src/Codeception/Subscriber/ErrorHandler.php
+++ b/src/Codeception/Subscriber/ErrorHandler.php
@@ -98,8 +98,10 @@ class ErrorHandler implements EventSubscriberInterface
         if (!$this->suiteFinished && (
             $error === null || !in_array($error['type'], [E_ERROR, E_COMPILE_ERROR, E_CORE_ERROR])
         )) {
-            throw new \RuntimeException('Command Did Not Finish Properly');
-        } elseif (!is_array($error)) {
+            echo "\n\n\nCOMMAND DID NOT FINISH PROPERLY.\n";
+            return;
+        }
+        if (!is_array($error)) {
             return;
         }
         if (error_reporting() === 0) {
--- a/src/Codeception/Subscriber/ErrorHandler.php
+++ b/src/Codeception/Subscriber/ErrorHandler.php
@@ -99,7 +99,7 @@ class ErrorHandler implements EventSubscriberInterface
             $error === null || !in_array($error['type'], [E_ERROR, E_COMPILE_ERROR, E_CORE_ERROR])
         )) {
             echo "\n\n\nCOMMAND DID NOT FINISH PROPERLY.\n";
-            return;
+            exit(255);
         }
         if (!is_array($error)) {
             return;
--- a/tests/cli/ErrorHandlerCest.php
+++ b/tests/cli/ErrorHandlerCest.php
@@ -1,0 +1,13 @@
+<?php
+class ErrorHandlerCest
+{
+    /**
+     * @param CliGuy $I
+     */
+    public function earlyExitWarnsTheUser(\CliGuy $I)
+    {
+        $I->executeFailCommand('run -c tests/data/first_test_exits');
+
+        $I->seeInShellOutput('COMMAND DID NOT FINISH PROPERLY');
+    }
+}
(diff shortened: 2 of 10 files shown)


src/Codeception/Subscriber/ErrorHandler.php
```diff
@@ -98,8 +98,10 @@ public function shutdownHandler()
         if (!$this->suiteFinished && (
             $error === null || !in_array($error['type'], [E_ERROR, E_COMPILE_ERROR, E_CORE_ERROR])
         )) {
-            throw new \RuntimeException('Command Did Not Finish Properly');
-        } elseif (!is_array($error)) {
+            echo "\n\n\nCOMMAND DID NOT FINISH PROPERLY.\n";
+            return;
```
Naktibalda: This would still look like successful run in CI system.
It would be better to use `die(1);` in order to set non-zero exist code.

Slamdunk: I was deceived by `executeFailCommand`: I thought it would check that the command fails, instead it only accept possible failing commands:

https://github.com/Codeception/Codeception/blob/feb566a9dc26993611602011ae3834d8e3c1dd7f/src/Codeception/Module/Cli.php#L52-L54

PR updated.

This should be valid also for the `FATAL ERROR. TESTS NOT FINISHED.`, shouldn't it?

src/Codeception/Subscriber/ErrorHandler.php
```diff
@@ -98,8 +98,10 @@ public function shutdownHandler()
         if (!$this->suiteFinished && (
             $error === null || !in_array($error['type'], [E_ERROR, E_COMPILE_ERROR, E_CORE_ERROR])
         )) {
-            throw new \RuntimeException('Command Did Not Finish Properly');
-        } elseif (!is_array($error)) {
+            echo "\n\n\nCOMMAND DID NOT FINISH PROPERLY.\n";
+            exit(255);
```
Slamdunk: 255 is the code used by uncaught exception, so I retained the same number.

tests/cli/ErrorHandlerCest.php
```diff
@@ -0,0 +1,13 @@
...
+class ErrorHandlerCest
+{
+    /**
+     * @param CliGuy $I
+     */
+    public function earlyExitWarnsTheUser(\CliGuy $I)
+    {
+        $I->executeFailCommand('run -c tests/data/first_test_exits');
+
+        $I->seeInShellOutput('COMMAND DID NOT FINISH PROPERLY');
```
Naktibalda: I think that it would be useful to test exit code using [seeResultCodeIs](https://codeception.com/docs/modules/Cli#seeResultCodeIs)

Slamdunk: Didn't know this, nice, implemented, thanks for the feedback

Issue #1233, opened 2014-07-22
Idea: Split codeception core and it's modules


vysinsky:
Hi,
I would like to use codeception for unit testing because of nice API. But I do not need extra things so I do not need all the codeship's dependecies and it takes long time to install eg. on Travis CI. 

Wouldn't it be better to split codeception core and its modules to different composer packages? Everyone can than install only modules which he needs. Each package could have own dependencies.

Naktibalda:
I started extracting modules and, unless someone finds a fatal flaw, I will start moving them to Codeception org and publishing on Packagist soon.
Here are the modules I extracted so far:
https://github.com/Naktibalda/module-rest
https://github.com/Naktibalda/module-zendexpressive
https://github.com/Naktibalda/module-apc

Naming convention:
modules use prefix `codeception/module-`, as above,
shared dependencies of module will use a `codeception/lib-` prefix, e.g. `codeception/lib-innerbrowser`,
shared dev depencies use a `codeception/util-` prefix, unless someone has a better idea,
examples: 
https://github.com/Codeception/util-robohelpers
https://github.com/Naktibalda/util-rest-app
https://github.com/Naktibalda/util-universalframework

extensions use `codeception/ext-` prefix, e.g. `codeception/ext-runfailed`, `codeception/ext-recorder`.

DavertMik:
@Naktibalda I see problems here 

### Dependency Graph

let's say to fix a bug in PhpBrowser we need to update innerbrowser. This will require PR into 2 repositories. Also - when PhpBrowser is updated and released, we need to update and release REST module as well. So this really looks like a huge dependency graph.

The same goes for tests. When I make a patch to lib/innerbrowser I need to be sure that this won't break PhpBrowser and all frameworks. So lib-innerbrowser should contain tests for those. 

If we could have phpbrowser/rest/frameworks in one repository that would simplify the maintanance of those modules and compatibility with each other. However, that makes a trickier release process. To release only one module on change we would need to make repo split and push this update to a module repository, then tag it. 

## Compatibility

Let's say I'm stuck at Codeception 3 but I want to use module from Codeception 4. For instance, I'm writing tests for Symfony 4 application and I can't upgrade to symfony components v5 but I want to use latest WebDriver. As WebDriver depends on php-webdriver only the upgrade should go smooth. I could use `codeception/base:2.0` + `codeception/module-webdriver`. 
[...]

PR #5602, opened 2019-07-22
[Async] add implementation of async support

alexkunin:
Want to see how hard tests are failing.


alexkunin:
@Naktibalda, @DavertMik, could you take a look please? Do you think this has potential? Specifically, `testMultipleAsyncMethods()` in `tests/unit/Codeception/Module/AsyncTest.php` which asynchronously starts server and client, does some full duplex comms, and checks results.

This approach could allow creating tests that require some (stub) server to check functionality, e.g. simple SMTP server etc. -- all from within PHP code, without any external binaries or services.

TBH there is not much that requires adding any code to Codeception core, but having async functionality out of the box would be great.

alexkunin:
Just to be clear:
- I'm totally not expecting this to be accepted -- this is far from the final result
- not sure re API and particular method names
- not sure about 'wonky' file-based IPC -- could use sockets instead

DavertMik:
Thanks, @alexkunin this looks really impressive.
And the code so far is clean and elegant, however, I don't see where it can be applied.
I was expecting to see some PHP Async stuff like maybe for ReactPHP but I saw just running functions in external processes. So I'm a bit confused.

Could you explain the use case. And maybe show an example of a test where this module is used?
Can it be used for testing Swoole or ReactPHP?

alexkunin:
My original use case is the following: an app sends emails as part of user registration process, and I needed to test that (acceptance level).

To make that possible, the following 'construct' was 'assembled':

1. Install `captbaritone/mailcatcher-codeception-module`.
2. Set up simple SMTP sink, specifically `schickling/mailcatcher` instance: `docker run -p 1080:1080 -p 1025:1025 schickling/mailcatcher`
- this thing also includes nice UI -- something that my solution (due to automated nature of testing) actively avoids
3. Configure test environment to use that SMTP sink: `MAILER_DSN=smtp://localhost:1025`
4. Actually write some tests.

While all of this is not a rocket science and is not very hard to reproduce, things get trickier if you want to setup CI pipeline, because test code depends on external service wrapped in a Docker container. Depending on particular CI solution, complexity of setup varies greatly.

So, my idea was: can I use some very simple pure PHP implementation of SMTP server? Can I start and stop it from inside tests (`_before`/`_after`)? Can I ditch Docker dependency?

To all these questions the answer is 'yes', and this `Async` module is kinda embodiment of that answer. But of course this is just a surface scratch, I have never expecting Codeception maintainers to just accept this PR -- it's merely an invite for discussion.

And here are some points and questions:

1. `Async` is quite overloaded term these days, by using it I've unintentionally set some expectations for you, probably for others too. Probably there is some better name.

2. ReactPHP can work here, as any other similar library or approach. In my experiments I was using plain sockets, though, to avoid complexity and possible dependency on additional PHP extensions.

3. While working on simple SMTP implementation, I had to test it too. And `Async` module proved to be useful there, as testing pure socket comms is not exactly an easy thing. Of course, one could stub socket IO related functions and get away with that -- it's a valid testing technique, for unit testing anyway. But integration tests can't be done like that, right? This is another use case for this module.

4. Pure PHP implementation of any server or service is a good candidate for use case: HTTP, XML-RPC, whatever. The main benefits (as I see it) -- simplification of setup: install some PHP packages, and you're good to go, no need to setup external services, containers, etc. Adding mail testing to my app was a real step up (considering CI part), and having it like x10 simpler would be much appreciated.

5. Maybe this is not should be a core module. I mean, I don't use anything in my code that makes it a requirement to be inside Codeception repo. But if (when) this module is robust and complete enough, having it as part of core repo would be a nice thing (even from a buzzword PoV). I mean one could say "Codeception supports variety of use cases, including testing of DBs and DB abstractions, many well-known frameworks, asynchronous comms, etc.".

Naktibalda:
> So, my idea was: can I use some very simple pure PHP implementation of SMTP server? Can I start and stop it from inside tests (`_before`/`_after`)? Can I ditch Docker dependency?

Codeception has `RunBefore` and `RunProcess` extensions for this purpose: https://codeception.com/extensions#RunBefore

> Maybe this is not should be a core module. I mean, I don't use anything in my code that makes it a requirement to be inside Codeception repo. 

@DavertMik wants to separate core modules from [core anyway](https://github.com/Codeception/Codeception/issues/1233) and it may finally happen in the next major release.
So it would be the best if you published it as a separate package on packagist.

alexkunin:
RunBefore and RunProcess requires alot of boilerplate, including process start, watch, some IPC etc. That's the purpose of this module -- to simplify all of that.

I agree, module might be a better format at this stage. So, I'm closing this, and will setup a separate project. Thank you for your feedback.

Issue #5612, opened 2019-07-30
Support MYSQL_ATTR_INIT_COMMAND


tadasauciunas:
#### What are you trying to achieve?

`MYSQL_ATTR_INIT_COMMAND` Should be supported when creating a PDO connection for initial queries like `SET NAMES utf8` which are commonly used

Naktibalda:
`SET NAMES utf8` was commonly used in old versions of PHP.
`charset=utf8` can be specified in DSN since PHP 5.3.6: https://www.php.net/manual/en/ref.pdo-mysql.connection.php

tadasauciunas:
`SET NAMES utf8` is still necessary in still cases, even in modern php versions. 
Passing the charset inside the dsn is possible, but having the freedom to pass other initializing queries via MYSQL_ATTR_INIT_COMMAND adds more flexibility.

Issue #5605, opened 2019-07-24
Create database if it does not exist


tadasauciunas:
#### What are you trying to achieve?

Empty database to be created before loading the schema sql dump if it does not exist already

#### What do you get instead?

If codeception is being run for the first time in the environment, an empty database has to be created manually before running the tests

This is extremely annoying in various CI tools which are sometimes updated and their existing databases are cleared, then somebody has to remember to re-create all the necessary test databases manually which are used by codeception, or else the tests will fail because of a non-existing DB.

PR #5614, opened 2019-07-30
Support MYSQL_ATTR_INIT_COMMAND

tadasauciunas:
This is a PR for fixing https://github.com/Codeception/Codeception/issues/5612

### DIFF
--- a/src/Codeception/Module/Db.php
+++ b/src/Codeception/Module/Db.php
@@ -54,6 +54,7 @@ use Codeception\Util\ActionSequence;
  * * ssl_verify_server_cert - disables certificate CN verification (MySQL specific, @see http://php.net/manual/de/ref.pdo-mysql.php)
  * * ssl_cipher - list of one or more permissible ciphers to use for SSL encryption (MySQL specific, @see http://php.net/manual/de/ref.pdo-mysql.php#pdo.constants.mysql-attr-cipher)
  * * databases - include more database configs and switch between them in tests.
+ * * init_command - command to execute when creating the PDO connection (MySQL specific, @see https://php.net/manual/de/ref.pdo-mysql.php#pdo.constants.mysql-attr-init-command)
  *
  * ## Example
  *
@@ -73,6 +74,7 @@ use Codeception\Util\ActionSequence;
  *              ssl_ca: '/path/to/ca-cert.pem'
  *              ssl_verify_server_cert: false
  *              ssl_cipher: 'AES256-SHA'
+ *              init_command: 'SET NAMES utf8;'
  *
  * ## Example with multi-dumps
  *     modules:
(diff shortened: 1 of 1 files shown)


src/Codeception/Module/Db.php
```diff
@@ -542,6 +544,10 @@ private function connect($databaseKey, $databaseConfig)
             $options[\PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT] = (boolean) $databaseConfig[ 'ssl_verify_server_cert' ];
         }
 
+        if (array_key_exists('init_command', $databaseConfig) && defined('\PDO::MYSQL_ATTR_INIT_COMMAND')) {
+            $options[\PDO::MYSQL_ATTR_INIT_COMMAND] = $databaseConfig['init_command'];
+        }
```
Naktibalda: Is there any advantage of using mysql-specific init command over calling [executeQuery](https://github.com/Codeception/Codeception/blob/3.0.3/src/Codeception/Lib/Driver/Db.php#L285) in [connect ](https://github.com/Codeception/Codeception/blob/3.0.3/src/Codeception/Lib/Driver/Db.php#L36)method?

tadasauciunas: No, I don't think so.
If we added executeQuery in the `__construct` of `Driver\Db` that would be much more flexible than depending on the mysql-specific `init_command`, and it would also give the possibility to solve https://github.com/Codeception/Codeception/issues/5605 as well by configuring queries for creating/using the database if it doesn't exist right after creating the PDO object. 
Adding executeQuery in the `connect` method like you offered wouldn't really work without extending the `Module\Db` class and overriding how `connectToDatabases/reconnectDatabases` works, unless I am missing something.

Naktibalda: I looked at this code again and I think that the best place to execute init_command is at the bottom of this method, either before or after this line:
```php
$this->debugSection('Db', 'Connected to ' . $databaseKey . ' ' . $this->drivers[$databaseKey]->getDb());`
```

tadasauciunas: I agree with the location, but do you mean to execute `executeQuery` instead of `init_command`? Because `init_command` is the mysql specific command which is executed during the creation of PDO connection, and the PDO connection would have already been created a few lines before.

Naktibalda: mysql-specific connection attribute exists for legacy reasons - because mysql C lib and mysql command line client has it. It made sense to have it as a connection attribute, because mysql client and old PHP mysql extension could reconnect automatically if connection was lost. Is PDO even able to reconnect if connection is lost? I think that it throws exception instead, so there is no reason to use connection attribute.

tadasauciunas: I'm okay with not using mysql-specific attributes for PDO connection.
Do you think a change like this would be accepted ? It would solve both of my issues.
Params array which is passed to `executeQuery` could be configured as well (shown as empty array in the following snippet)

```
if ($databaseConfig['waitlock']) {
    $this->_getDriver()->setWaitLock($databaseConfig['waitlock']);
}

// NEW CODE START
foreach ($databaseConfig['initial_queries'] as $initialQuery) {
    $this->drivers[$databaseKey]->executeQuery($initialQuery, []);
}
// NEW CODE END

$this->debugSection('Db', 'Connected to ' . $databaseKey . ' ' . $this->drivers[$databaseKey]->getDb());
```

Naktibalda: Yes, something like that could be accepted.

Please create a fresh branch from 3.0 to get builds passing
and implement a test for it in https://github.com/Codeception/Codeception/blob/3.0/tests/unit/Codeception/Module/Db/TestsForDb.php or one of database specific test files in that directory (I think that you will have to use `_reconfigure` to set your configuration options). Creating a separate test file is fine too, I think that it would look cleaner that way.

DavertMik: I agree with @Naktibalda here :+1:

Naktibalda:
Replaced by #5660

PR #5621, opened 2019-08-02
Relaxing docker php image tag version

OneEyedSpaceFish:
Now that the buster based PHP images are the default, and now that the existing default tags have been switched to from stretch, It makes sense to relax the docker tag to major version only.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM php:7.3.7-cli-buster
+FROM php:7.3-cli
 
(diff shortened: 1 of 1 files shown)


Naktibalda:
Thanks, it looks good to me, because it reverts the previous change,

OneEyedSpaceFish:
> Thanks, it looks good to me, because it reverts the previous change,

No worries. The previous change had been made, as the php team had not updated the base tags from stretch to buster straight away. Now that has been done, so no need for this anymore.

PR #5625, opened 2019-08-04
[Doctrine2] merge functionality of persistEntity into haveInRepository

### LINKED ISSUE #5624: [Doctrine2] suggestion to merge persistEntity and haveInRepository
These two methods do basically the same thing, but with a bit different approach, and one of them does not follow naming convention. So, why have both? There are cases when people were confused by this diversity.

I suggest to:

- add instance reuse functionality to `haveInRepository` (just a parameter type check -- use passed value as instance if it is an object, otherwise treat passed value as class name and call constructor manually)
- make `persistEntity` an alias to `haveInRepository`
- deprecate `persistEntity` (remove in 4.0?)

I have prepared a PR that makes this possible (see below). It does not include any deprecation notices, though.

alexkunin:
Closes #5624.

### DIFF
--- a/src/Codeception/Module/Doctrine2.php
+++ b/src/Codeception/Module/Doctrine2.php
@@ -12,8 +12,12 @@ use Codeception\Util\Stub;
 use Doctrine\Common\Collections\Criteria;
 use Doctrine\Common\Collections\Expr\Expression;
 use Doctrine\ORM\QueryBuilder;
+use InvalidArgumentException;
 use PlainEntity;
 use ReflectionException;
+use function gettype;
+use function is_object;
+use function is_scalar;
 
 /**
  * Access the database using [Doctrine2 ORM](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/).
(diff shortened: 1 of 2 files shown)


src/Codeception/Module/Doctrine2.php
```diff
@@ -236,29 +240,11 @@ public function flushToDatabase()
...
     {
-        if ($values) {
-            $rpa = new ReflectionPropertyAccessor();
-            $rpa->setProperties($obj, $values);
-            $this->populateEmbeddables($obj, $values);
-        }
-
-        $this->em->persist($obj);
-        $this->em->flush();
+        return $this->haveInRepository($obj, $values);
```
Naktibalda: Please use `Notification::deprecate` to emit deprecation message.
[Example](https://github.com/Codeception/Codeception/blob/3.0/src/Codeception/Util/Stub.php#L5-L12)

alexkunin: Done.

PR #5627, opened 2019-08-04
[Doctrine2] properly handle non-typical primary keys in haveInRepository

### LINKED ISSUE #5626: [Doctrine2] primary key name is hardcoded
`haveInRepository()` returns primary key value only if there is a getter with fixed name `getId`. While this behavior covers most cases, it can be better. PR linked below makes it possible to:

- work with primary key with names other than 'id'
- work with composite primary keys
- use Reflection API instead of getter method (not saying this is the best way, but at least it is in line with other methods which already ignore getters)

alexkunin:
Closes #5626.

### DIFF
--- a/src/Codeception/Module/Doctrine2.php
+++ b/src/Codeception/Module/Doctrine2.php
@@ -26,6 +26,8 @@ use ReflectionException;
 use function gettype;
 use function is_object;
 use function is_scalar;
+use function get_class;
+use function is_array;
 
 /**
  * Access the database using [Doctrine2 ORM](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/).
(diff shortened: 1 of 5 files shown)


Naktibalda:
It looks like this change will conflict with #5625 , won't it?

alexkunin:
Yes, it will, and I will fix the conflict if you are going to accept both.

TBH I did not know how to properly do these two PRs, they affect same piece of code, but are conceptually different (i.e. must be separate PRs). So, one of them (this one) would be more of a "statement" or "proposal", not a real PR ready to merge.

So, if you think both PRs are worth it, accept one (probably there other one, as it is bigger), and I'll adjust the other.

alexkunin:
Rebased onto upstream 3.0, all tests green.

PR #5623, opened 2019-08-02
[Doctrine2] expose refresh and clear [obsoleted by #5628]

alexkunin:
Closes #4526.

PR #5628, opened 2019-08-04
[Doctrine2] expose clear() and refresh()

### LINKED ISSUE #4526: Suggestion for Doctrine2 Module: refresh()
It looks like, if an entity gets modified by the app, there's no cool way to refresh it in Codeception.

Sample pseudo code:
```php
$user = $I->grabEntityFromRepository(...);
$I->assertTrue($user->isActive());
$I->submitForm(...); // some form which changes the user to inactive
// This fails, since `$user` is still the same as above:
$I->assertFalse($user->isActive());
```

Right now, I can't think of any other way than to reload the entity from the database, which means repeating the first line:
```php
// ... 
$modifiedUser = $I->grabEntityFromRepository(...); // same as in line 1 above
$I->assertFalse($modifiedUser ->isActive());
```

But something like this would be nicer:
```php
// ... 
$I->refreshEntity($user); // would be cool :-)
$I->assertFalse($user ->isActive());
```

### Details

* Codeception version: 2.3.5
* PHP Version: 7.1

alexkunin:
Closes #5623, closes #4526.

### DIFF
--- a/src/Codeception/Module/Doctrine2.php
+++ b/src/Codeception/Module/Doctrine2.php
@@ -277,7 +277,6 @@ EOF;
         $this->em->clear();
     }
 
-
     /**
      * Performs $em->flush();
      */
@@ -286,6 +285,30 @@ EOF;
         $this->em->flush();
     }
 
+    /**
+     * Performs $em->refresh() on every passed entity. This is useful in acceptance tests where
+     * entity can become invalid due to external changes (relative to entity manager used in tests).
+     *
+     * @param object|object[] $entities
+     */
+    public function haveEntitiesRefreshed($entities)
+    {
+        if (!is_array($entities)) {
+            $entities = [$entities];
+        }
+
+        foreach ($entities as $entity) {
+            $this->em->refresh($entity);
+        }
+    }
+
+    /**
+     * Performs $em->clear();
+     */
+    public function haveEntityManagerCleared()
+    {
+        $this->em->clear();
+    }
 
     /**
      * This method is deprecated in favor of `haveInRepository()`. It's functionality is exactly the same.
--- a/src/Codeception/Module/Doctrine2.php
+++ b/src/Codeception/Module/Doctrine2.php
@@ -291,7 +291,7 @@ EOF;
      *
      * @param object|object[] $entities
      */
-    public function haveEntitiesRefreshed($entities)
+    public function refreshEntities($entities)
     {
         if (!is_array($entities)) {
             $entities = [$entities];
@@ -305,7 +305,7 @@ EOF;
     /**
      * Performs $em->clear();
      */
-    public function haveEntityManagerCleared()
+    public function clearEntityManager()
     {
         $this->em->clear();
     }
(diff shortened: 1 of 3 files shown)


src/Codeception/Module/Doctrine2.php
```diff
@@ -272,6 +271,30 @@ public function flushToDatabase()
         $this->em->flush();
     }
 
+    /**
+     * Performs $em->refresh() on every passed entity. This is useful in acceptance tests where
+     * entity can become invalid due to external changes (relative to entity manager used in tests).
+     *
+     * @param object|object[] $entities
+     */
+    public function haveEntitiesRefreshed($entities)
```
Naktibalda: It isn't mandatory to use `have` prefix, it fact there was [discussion](https://github.com/Codeception/Codeception/issues/4268) proposing to get rid of it completely and [documentation](https://codeception.com/docs/02-GettingStarted#The-Codeception-Syntax) says that `Actions start with a plain english verb, like “click” or “fill”.`
So I think that it would be better to name these methods `refreshEntities` and `clearEntities`.

alexkunin: I agree, these are awkward. So, changing them to `refreshEntities` and `clearEntityManager` (I think this one is more correct as it does not work on individual entities, it rather changes entity manager's internal state).

alexkunin:
Oops, I think I have mixed up branch names. Let me check this one and #5627.

alexkunin:
So, it was just invalid branch name -- commits are correct. Changed method names, let's see what tests are going show us.

alexkunin:
Rebased onto upstream, all tests are green.

alexkunin:
Rebased again.

alexkunin:
All tests are green.

src/Codeception/Module/Doctrine2.php
```diff
@@ -286,6 +285,30 @@ public function flushToDatabase()
         $this->em->flush();
     }
 
+    /**
+     * Performs $em->refresh() on every passed entity. This is useful in acceptance tests where
+     * entity can become invalid due to external changes (relative to entity manager used in tests).
+     *
+     * @param object|object[] $entities
```
DavertMik: A code example here would be nice

src/Codeception/Module/Doctrine2.php
```diff
@@ -286,6 +285,30 @@ public function flushToDatabase()
...
+            $entities = [$entities];
+        }
+
+        foreach ($entities as $entity) {
+            $this->em->refresh($entity);
+        }
+    }
+
+    /**
+     * Performs $em->clear();
```
DavertMik: Please add a code example here too :+1:

alexkunin:
I have added some trivial sample code to both methods, but TBH the problem is in the following: both methods exists to fight couple of very specific situations, and meaningful example would be quite lengthy. I thought that people who want to use these methods could look at tests if they require better understanding.

With that said, I totally agree that singe line doc entry is not a good thing, and adding even a trivial example might help someone.

src/Codeception/Module/Doctrine2.php
alexkunin: Done.

src/Codeception/Module/Doctrine2.php
alexkunin: Done.

PR #5629, opened 2019-08-04
[Doctrine2] add fixtures support

### LINKED ISSUE #5594: [Doctrine2] Integrate fixtures

alexkunin:
Closes #5594, closes #5598.

### DIFF
--- a/src/Codeception/Module/Doctrine2.php
+++ b/src/Codeception/Module/Doctrine2.php
@@ -9,10 +9,17 @@ use Codeception\Lib\Interfaces\DoctrineProvider;
 use Codeception\TestInterface;
 use Codeception\Util\ReflectionPropertyAccessor;
 use Codeception\Util\Stub;
+use Doctrine\Common\DataFixtures\Executor\ORMExecutor;
+use Doctrine\Common\DataFixtures\FixtureInterface;
+use Doctrine\Common\DataFixtures\Loader;
+use Doctrine\Common\DataFixtures\Purger\ORMPurger;
+use Codeception\Exception\ModuleException;
+use Codeception\Exception\ModuleRequireException;
 use Doctrine\Common\Collections\Criteria;
 use Doctrine\Common\Collections\Expr\Expression;
 use Doctrine\ORM\QueryBuilder;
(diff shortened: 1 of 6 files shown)


src/Codeception/Module/Doctrine2.php
```diff
@@ -411,6 +418,141 @@ public function haveInRepository($entity, array $data)
...
+     * ```
+     *
+     * Note: this method requires `doctrine/data-fixtures` package to be installed.
+     *
+     * @param string|string[]|object[] $fixtures
+     * @param bool $append
+     * @throws ModuleException
+     * @throws ModuleRequireException
+     */
+    public function haveFixtures($fixtures, $append = true)
```
Naktibalda: It would be better to rename this method to `loadFixtures`, right?

alexkunin: Agree. Fixed.

Issue #5632, opened 2019-08-06
[Doctrine2] haveInRepository does not support constructor parameters


burned42:
#### What are you trying to achieve?
Call `$I->haveInRepository()` of the "Doctrine2" module with an entity which expects constructor parameters.

#### What do you get instead?
`[ArgumentCountError] Too few arguments to function App\Entity\Thing::__construct(), 0 passed and at least 5 expected`

### Solution
In `Codeception\Util\ReflectionPropertyAccessor::setPropertiesForClass` check if the class has constructor parameters and if matching data is supplied, pass it as a parameter to the constructor. I'd be happy to provide you with a pull request for this change.

What do you think about this?

alexkunin:
Right now this can be resolved by using `persistEntity`:

```php
$I->persistEntity(new \App\Entity\Thing($arg1, $arg2, ...), ['field1' => 'value', ...]);
```

In the future (PR #5624 has been merged into master already) same functionality will be baked into `haveInRepository`.

Your proposed solution actually sounds nice, very smart-DI-like (e.g. like Symfony's auto-wiring works), but also adds edge cases (e.g. parameter name matches field name, etc.), and also introduces lots of complexity. Not saying this will never be implemented (after all I'm not a maintainer), but current codebase already supports this feature as described above.

burned42:
@alexkunin Thank you, I didn't know about `persistEntity()` :)

About the edge cases, you're actually right, that could be problematic and/or add complexity. For reference, this would be an implementation of what I thought about (without considering those edge cases): https://github.com/burned42/Codeception/tree/doctrine_haveInRepository_constructor_parameters

But since we can also pass an object to `persistEntity` and with your changes from the mentioned PR also to `haveInRepository()`, that's also fine for me, I'll just need to write my tests a bit different than expected :)

On the other side, I don't think that the changes from my linked branch would break any existing code because currently calling `haveInRepository` with a class that has a constructor with parameters just results in an exception, so I think it wouldn't hurt to merge even if there is a workaround (using `persistEntity`) and also even if there could be edge cases that would still not work (which wouldn't work with the current state either).

alexkunin:
What I like about your approach is that we can just pass data to `haveInRepository` without actually calling constructor in our test, i.e. very declarative way. That... feels good, for some reason.

[...]

PR #5633, opened 2019-08-06
[Doctrine2] Make haveInRepository support constructor parameters

burned42:
Closes https://github.com/Codeception/Codeception/issues/5632 (please also note the comments there).

### DIFF
--- a/src/Codeception/Module/Doctrine2.php
+++ b/src/Codeception/Module/Doctrine2.php
@@ -424,6 +424,7 @@ EOF;
      * Persists record into repository.
      * This method creates an entity, and sets its properties directly (via reflection).
      * Setters of entity won't be executed, but you can create almost any entity and save it to database.
+     * If the entity has a constructor, for optional parameters the default value will be used and for non-optional parameters the given fields (with a matching name) will be passed when calling the constructor before the properties get set directly (via reflection).
      *
      * Returns primary key of newly created entity. Primary key value is extracted using Reflection API.
      * If primary key is composite, array of values is returned.
--- a/src/Codeception/Module/Doctrine2.php
+++ b/src/Codeception/Module/Doctrine2.php
@@ -520,8 +520,15 @@ EOF;
     private function instantiateAndPopulateEntity($className, array $data, array &$instances)
     {
         $rpa = new ReflectionPropertyAccessor();
-        list($scalars,) = $this->splitScalarsAndRelations($className, $data);
-        $instance = $rpa->createWithProperties($className, $scalars);
+        list($scalars,$relations) = $this->splitScalarsAndRelations($className, $data);
+        // Pass relations that are already objects to the constructor, too
+        $properties = array_merge(
+            $scalars,
+            array_filter($relations, function ($relation) {
+                return is_object($relation);
+            })
+        );
+        $instance = $rpa->createWithProperties($className, $properties);
         $this->populateEntity($instance, $data, $instances);
         return $instance;
     }
--- a/src/Codeception/Module/Doctrine2.php
+++ b/src/Codeception/Module/Doctrine2.php
@@ -440,6 +440,12 @@ EOF;
      * $I->haveInRepository(new User($arg), array('name' => 'davert'));
      * ```
      *
+     * Alternatively, constructor arguments can be passed by name. Given User constructor signature is `__constructor($arg)`, the example above could be rewritten like this:
+     *
+     * ```php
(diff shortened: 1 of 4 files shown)


burned42:
@alexkunin That's a good idea, I'm throwing an exception now in this case: b246c801ce6e4fd9b416ffe097cc647d0babb026

alexkunin:
Do you think adding a test or two is possible? There is a suite for `Doctrine2`, adding an entity with constructor parameters and a test case for that should not be hard. Let me know if you need help with that.

burned42:
@alexkunin Actually I already added some [tests](https://github.com/Codeception/Codeception/pull/5633/files#diff-22979855156ba9918a2157667379ffbb). Did you just miss them, or do you think I should add more tests?

alexkunin:
Yep, missed them, was writing my comment when https://github.com/Codeception/Codeception/pull/5633/commits/b246c801ce6e4fd9b416ffe097cc647d0babb026 was not there yet (or at least was not displayed by UI). They look good to me, nothing to add. Maybe mention in method docblock that fields that match constructor parameters are passed to constructor instead of direct property assignment, along with 1-2 lines of sample code? And after that I'd summon @Naktibalda or @DavertMik for their review.

burned42:
@alexkunin sure, adding a comment would be nice, let me just make sure I get this correctly. Where would be the best place to add this comment? On first thought I would add it to the method `setPropertiesForClass` that I actually changed, but on the other side, the docblock of `Doctrine2::haveInRepository` is what the end users see (this is what you see in the documentation, right?), so would that maybe be a better place to add such a comment?

alexkunin:
`Doctrine2::haveInRepository` is the only valid place to add the comment if you want it to make it to the docs. I mean, comments on that utility class are not exactly useless, but not a part of user docs. If you want to preview your doc changes, see https://github.com/Codeception/Codeception/blob/3.0/RELEASE.md for general info on building, and then use `php robo.phar build:docs` (IIRC) to build only docs (they go to `docs/` directory).

burned42:
@alexkunin I tried to explain the behavior in f9dd006138c6d3fc229abe21e369724a4e1b3143

I'm not quite sure how a code example could look like as it could just be the same as the examples that already exist, given the mentioned `Entity\User` would have a constructor. If you have any suggestions, I'd be happy to add those :)

alexkunin:
The following would make sense to me:

> ```php
> $I->haveInRepository(new User($arg), array('name' => 'davert'));
> ```
> 
> Alternatively, constructor arguments can be passed by name. Given `User` constructor signature is `__constructor($arg)`, example above could be rewritten like this:
> 
> ```php
> $I->haveInRepository('User', array('arg' => $arg, 'name' => 'davert'));
> ```

alexkunin:
Looks good to me. I'd say maintainers (given they are willing to merge this PR) will ask to squish commits, but maybe wait until they say so. Other than that, I can't see anything to add here except our edge cases. But this is probably should be discussed in the issue first.

burned42:
@alexkunin Thank you very much :)

alexkunin:
You're welcome! BTW, all tests are green, which means this PR is ready for real review from maintainers.

Naktibalda:
@alexkunin Please resolve merge conflict in the docblock.

burned42:
@Naktibalda since I opened the PR I tried to resolve the merge conflicts myself. Apparently it was a bit more than just the docblock because @alexkunin recently introduced the nested entity creation which broke my implementation :D Anyway, I rebased onto 3.0, resolved the merge conflicts and adapted my code to work with the latest changes (see last two commits). CI seems to be happy with it :)

Basster:
This PR broke `haveInRepository`, so you cannot pass a pre populated object wich has constructor arguments.

```php
class A(B $b) {}

classs B() {}

$b = new B();
$a = new A($b);

$I->haveInRepository($a)
```
Results in `[InvalidArgumentException] Constructor parameter "b" missing`.

I'm going to try submiting a fix today, as I'm already figuring out, how to fix this.

burned42:
@Basster you are totally correct :man_facepalming: I must've missed that when I resolved the merge conflicts (when I started this PR, `haveInRepository` didn't even have the possibility to work with pre-populated objects). Constructor parameters currently get checked always, even if a pre-populated object gets passed and if one is missing the exception gets thrown, but of course if we already have a pre-populated object then we don't need to have those constructor parameters at all.

Did you already find out how to fix this? Otherwise I can also open a PR to fix this.

Basster:
Will submit the PR in a few seconds ^^

PR #5641, opened 2019-08-07
[Doctrine2] implement recursive entity creation

### LINKED ISSUE #5640: [Doctrine2] Recursive entity creation
There are cases when your data model has relations, and you need to test them. As a result, something like this can be seen in tests:

```php
$userId = $I->haveInRepository(User::class, [
    'username' => 'user1',
]);
$I->haveInRepository(Post::class, [
    'title' => 'Post Title 1',
    'author' => $I->grabEntityFromRepository(User::class, ['id' => $userId]),
]);
```

This is awkward, and gets even worse if you have two or more levels of relations.

I suggest to add support for the following format:

```php
$I->haveInRepository(Post::class, [
    'title' => 'Post Title 1',
    'author' => [
        'username' => 'user1',
    ],
]);
```

Or the inverse of the above:

```php
$I->haveInRepository(User::class, [
    'username' => 'user1',
    'posts' => [
        [
            'title' => 'Post Title 1',
        ],
    ],
]);
```

See PR below for implementation.

alexkunin:
Closes #5640.

### DIFF
--- a/src/Codeception/Module/Doctrine2.php
+++ b/src/Codeception/Module/Doctrine2.php
@@ -1,33 +1,36 @@
 <?php
+
 namespace Codeception\Module;
 
-use Codeception\Lib\Interfaces\DataMapper;
-use Codeception\Lib\Notification;
-use Codeception\Module as CodeceptionModule;
 use Codeception\Exception\ModuleConfigException;
+use Codeception\Exception\ModuleException;
+use Codeception\Exception\ModuleRequireException;
+use Codeception\Lib\Interfaces\DataMapper;
 use Codeception\Lib\Interfaces\DependsOnModule;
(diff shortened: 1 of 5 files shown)


DavertMik:
Thanks, @alexkunin for bringing Doctrine support to the new level!
This module didn't get enough love in previous years, so I'm very glad to have your patches!

alexkunin:
Thanks for your encouragement, @DavertMik! I'm actually using `Doctrine2` in my projects, so, all PRs I make are based on real use cases, so, I hope they are useful for other Codeception users too.

BTW, merged with upstream, let's see if tests will pass.

DavertMik:
Tests pass. Merging!

PR #5644, opened 2019-08-07
Add possibility to use absolute path for `settings.bootstrap` key in codeception.yml

### LINKED ISSUE #5611: It is not possible to load global bootstrap file by absolute path

Load custom global `bootstrap` file by providing absolute path in the `codeception.yaml`

#### What do you get instead?

When I use absolute path in `settings.bootstrap`,  it is treated as a name and is being resolved  to incorrect real path. Example:

```yaml
# codeception.yaml at /project/root

settings:
  bootstrap: /path/to/bootstrap.php
```

https://github.com/Codeception/Codeception/blob/feb566a9dc26993611602011ae3834d8e3c1dd7f/src/Codeception/Subscriber/Bootstrap.php#L29-L34

In this example, `$settings['path']` is `/project/root/tests/unit` and after concatenation we get:

```yaml
/project/root/tests/unit/path/to/bootstrap.php
```

which is obviously incorrect.

```bash
# Bootstrap file /project/root/tests/unit/path/to/bootstrap.php can't be loaded
```

### Details

* Codeception version: 3.0.2
* PHP Version: 7.2
* Operating System: any
* Installation type: Composer

* Suite configuration:

```yml
paths:
    tests: tests
    output: tests/_output
    data: tests/_data
    support: tests/_support
    envs: tests/_envs
actor_suffix: Tester
extensions:
    enabled:

settings:
    bootstrap: /path/to/bootstrap.php

coverage:
    enabled: true
    include:
        - src/*


```

----

# Why do we need it?

[We are trying](https://github.com/infection/infection/issues/45) to integrate Codeception as a new Test Framework in [Infection](https://github.com/infection/infection) and the absence of this trivial feature blocks us.

Infection needs to inject its code into the process of testing for the project by adding custom autoload/bootstap file. This is how mutation testing works in Infection.

Please note, that both PHPUnit and PHPSpec has a setting to use a bootstrap (autoload) file that is included *before* the test framework execution.

# Any alternatives?

I tried to create a custom extension to load my `bootstrap` file, but since we need to load *external* class, we must `require` it before using, and this is impossible with extension:

```yaml
# codeception.yaml at /project/root
...
extensions:
    enabled:
        - Infection\Codeception\Extension\CustomAutoloader: { path: '/x/y/z.php' }

```

With the config above, Codeception fails with 

```php
"Class `Infection\Codeception\Extension\CustomAutoloader` is not defined. Autoload it or include into '_bootstrap.php' file of 'tests' directory"
```

maks-rafalko:
* Fixes #5611
* fully backward compatible
* intentionally targeted to 3.0 branch as it is bugfix (I treat it like this) or a very small feature

As simple as that.

### DIFF
--- a/src/Codeception/Subscriber/Bootstrap.php
+++ b/src/Codeception/Subscriber/Bootstrap.php
@@ -26,7 +26,10 @@ class Bootstrap implements EventSubscriberInterface
             return;
         }
 
-        $bootstrap = $settings['path'] . $settings['bootstrap'];
+        $bootstrap = codecept_is_path_absolute($settings['bootstrap'])
+            ? $settings['bootstrap']
+            : $settings['path'] . $settings['bootstrap'];
+
         if (!is_file($bootstrap)) {
             throw new ConfigurationException("Bootstrap file $bootstrap can't be loaded");
         }


Naktibalda:
Thanks, but @DavertMik decided to do something different.

Replaced by #5645

PR #5645, opened 2019-08-07
WIP Unified bootstrap loading

### LINKED ISSUE #5611: It is not possible to load global bootstrap file by absolute path

DavertMik:
* Unified bootstrap loading (using the same method)
* Allow loading absolute bootstrap paths
* Deprecated configuring bootstrap for all suites implicitly
* Added `--bootstrap` option to `run` command"

See http://phptest.club/t/bootstrap-deprecations-in-3-0/2196

Replaces #5644 
Fixes #5611


src/Codeception/Util/PathResolver.php
```diff
@@ -124,4 +124,13 @@ private static function isWindows($dirSep = DIRECTORY_SEPARATOR)
     {
         return ($dirSep == '\\');
     }
+
+    public static function isPathAbsolute($path)
+    {
+        if (DIRECTORY_SEPARATOR === '/') {
+            return mb_substr($path, 0, 1) === DIRECTORY_SEPARATOR;
```
Naktibalda: I think that `substr($path, 0, 1)` would work fine  because there is no other utf-8 character which starts with the same byte as '/'. 
All other places where `mb` functions are called specify `utf-8` as charset parameter.

tests/data/snapshots/tests/_support/_generated/DataTesterActions.php
```diff
@@ -941,6 +941,37 @@ public function haveInDatabase($table, $data) {
     public function seeInDatabase($table, $criteria = null) {
         return $this->getScenario()->runStep(new \Codeception\Step\Assertion('seeInDatabase', func_get_args()));
     }
+    /**
```
Naktibalda: Please delete generated files and .gitignore rule to make sure that they don't get added again.

src/Codeception/Configuration.php
```diff
@@ -251,21 +253,37 @@ public static function config($configFile = null)
...
+
+        $bootstrap = codecept_is_path_absolute($bootstrap)
+            ? $bootstrap
+            : rtrim($path, DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR . $bootstrap;
+
+
+        if (!file_exists($bootstrap)) {
+            throw new ConfigurationException("Bootstrap file $bootstrap can't be loaded");
         }
+        include_once $bootstrap;
```
Naktibalda: Why is this method using `include_once`? `Subscriber/Bootstrap.php` below used `require_once` before.
But I think that it would be better to use `require` rather than `require_once`, because boostrap files are more likely to contain linear code than class/file definitions and after this change it will be possible for multiple test suites to have the same absolute bootstrap file path.

DavertMik: I think it's dangerous to switch this to `require` as this can produce fatal errors in current setups.
Bootstrap is usually used for autoloaders and I think that this should be enough to execute it once

maks-rafalko:
Wooohoooo, thank you guys!

Do you have an idea when you can release it so I can continue integration Codeception with Infection?

Naktibalda:
I will try to make a release this weekend.

PR #5648, opened 2019-08-10
[Doctrine2] implement configurable property access

### LINKED ISSUE #5647: [Doctrine2] Configurable property access
Currently `Doctrine2` uses Reflection API to set properties of entities directly, without using setters. This should not be a problem in most cases, but it might be a show stopper for someone who has some fancy logic in setter -- maybe some type conversion, or validation, etc. Also, arguable one of the most widespread ways of using Doctrine is as ORM module of Symfony, which almost always means heavy usage of getters and setters via `symfony/property-access` package.

The PR below addresses this issue by abstracting away entity creation and modification, and then providing two implementations -- one using Reflection API and another using `symfony/property-access`. It also introduces new module configuration option (defaults to Reflection API), and module API method to dynamically change property access strategy for current test.

So, this is somewhat an edge case, but if encountered, not being able to use setters might prevent someone to use `Doctrine2` module altogether.

alexkunin:
Closes #5647.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,8 @@
         "symfony/process": ">=2.7 <5.0",
         "doctrine/orm": "^2",
         "doctrine/annotations": "^1",
-        "doctrine/data-fixtures": "^1"
+        "doctrine/data-fixtures": "^1",
+        "symfony/property-access": "^4"
     },
     "suggest": {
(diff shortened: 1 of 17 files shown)


DavertMik:
This pull request is OK but because we didn't see any real use cases for it, nor a big interest in the community in it, we are closing it. If you want configurable property access - write here and we will reopen that.

PR #5649, opened 2019-08-10
Support Stub 3.0

### LINKED ISSUE #5508: PhpStorm shows "multiple definitions exist" for MockObject because of class_aliases in shim.php

Write unit tests in PhpStorm with type-hinted mocks via inline annotations.

#### What do you get instead?

"Multiple definitions exist for class MockObject".

image

This seems to be because of `shim.php`:

image

It seems that with the new version of Codeception, `shim.php` is maybe no longer needed? I fixed this locally by just commenting out every line in that file and the unit tests run fine.

### Details

* Codeception version: 3.0
* PHP Version: 7.3
* Operating System: Windows
* Installation type: Phar

Naktibalda:
Which doesn't use shim file and supports only PHPUnit 6.5+

Fixes #5508


DavertMik:
Great!

Issue #5620, opened 2019-08-02
Provide a second codeception.phar with Verify/Specify included


mikehaertl:
#### What are you trying to achieve?

Use Specify and Verify with the global `codeception.phar`.

#### What do you get instead?

Codeception/Specify and Codeception/Verify are not included and I can not add them because the app I want to test has a fixed set of composer packages that I can not modify just for testing.

### Details

There's already issue #1200 and #1658 where this was discussed. I understand that adding the above packages may lead to conflicts. 

Would it be an option to provide a second `.phar` archive that contains the above 2 packages (and maybe even more)?

This way a developer could choose which one is best for his purposes.

Naktibalda:
I think that this is a rare edge case and it isn't worth the effort.
If you need such phar file, build it yourself.

Instructions how to build your own phar file from https://github.com/Codeception/Codeception/issues/2124#issuecomment-142232225
```
git clone https://github.com/Codeception/Codeception/
cd Codeception
composer require guzzlehttp/guzzle 5.3.*
wget http://robo.li/robo.phar
php -d phar.readonly=0 robo.phar build:phar
```

Replace `composer require guzzlehttp/guzzle 5.3.*` with packages that you want to add to your phar.

Note. [build:phar](https://github.com/Codeception/Codeception/blob/3.0.3/RoboFile.php#L169) currently targets PHP 7.0.0, if you want to target a different version of PHP, comment that line out or specify a version at [this line](https://github.com/Codeception/Codeception/blob/3.0.3/RoboFile.php#L148).

mikehaertl:
@Naktibalda Thanks, actually that's what I wanted to try but couldn't find any documentation (and neither any code) how to build the phar. So I will try this now.

Maybe it's worth to add some notes to the docs, how to get your own phar?

alexkunin:
Well, there is https://github.com/Codeception/Codeception/blob/3.0/RELEASE.md -- it does not specifically say how to build only .phar, but it does that among other things, and gives some useful hints re how things are built and released.

mikehaertl:
Ok, that should be fine. Closing this then. Thanks.

Naktibalda:
Apparently there is low demand for custom phar files, but separating `buildPhar` from `buildPhar7` (for PHP 7.0), would make it easier to target different versions of PHP.

kk738:
What I found really wierd is this sentence in the docu:

[...]

PR #5650, opened 2019-08-10
Separate build:phar from build:phar7

Naktibalda:
buildPhar is a public method which can be used for building custom phar files
buildPhar7 is part of release process and builds phar file specifically for PHP 7.0.0

Example of usecase for custom phar file: #5620


DavertMik:
:+1:

PR #5555, opened 2019-06-17
hiding Symfony 4.3 deprecation warning

DavertMik:
Let's hide this annoying message as this has no deal to Codeception users

PR #5651, opened 2019-08-10
[Symfony] Fix symfony 4.3 deprecation removal

juntereiner:
Fix https://github.com/Codeception/Codeception/pull/5555

The original code was never called.


Naktibalda:
Thanks

PR #5652, opened 2019-08-12
Add assert equals canonicalizing

lukebee:
Today I wanted to test if two arrays have equal elements but without taking their order into account. After some research, I found that PHPUnit has this kind of assertion called
"assertEqualsCanonicalizing" which is not supported in Codeception.

I hope you will take a look. Best regards! :)


lukebee:
I don't have a clue why CI fails.

"Error downloading remote file: One or more errors occurred.
Inner Exception: The request was aborted: Could not create SSL/TLS secure channel.
Command exited with code 2"

Naktibalda:
I will add them myself.

Naktibalda:
Replaced by #5656

PR #5658, opened 2019-08-19
Add ability to connect to selenium hub from inside a proxy

### LINKED ISSUE #5619: Add ability to connect to selenium hub from inside a proxy
#### Use case
I am using codeception behind a corporate proxy. My selenium grid is outside of the corporate network so I have to use the proxy server to connect to it.
In the suite config yml file you can specify `http_proxy` and `http_proxy_port`, but these are the settings that get injected into the browser itself to connect to the test object url. So we can't take use of them for this use case.

#### Solution
My suggested solution is to add two config keys to the suite config yml: `webdriver_proxy` and `webdriver_proxy_port`. These are passed to the facebook RemoteWebDriver which already supports this kind of proxy settings.

LeoVie:
Fixes #5619


Naktibalda:
Please raise a new pull request to https://github.com/Codeception/module-webdriver

Naktibalda:
I will close this pull request next time when I look at it.

PR #5660, opened 2019-08-21
Support initial queries execution after creating connection

### LINKED ISSUE #5605: Create database if it does not exist

### LINKED ISSUE #5612: Support MYSQL_ATTR_INIT_COMMAND

tadasauciunas:
Fixes #5612
Fixes #5605 
Replaces #5614


tadasauciunas:
This includes changes discussed in https://github.com/Codeception/Codeception/pull/5614

DavertMik:
Seems good to me!

PR #5662, opened 2019-08-22
[Doctrine2] fixing issue where `haveInRepository` doesn't supported instantiated objects anymore

Basster:
In addition to #5633.
Fixing issue where `haveInRepository` doesn't supported instantiated objects anymore.

In general I'm not a friend of comments in code, but the one in `ReflectionPropertyAccessor.php` describes the behaviour and why we do the `is_a()` call.

### DIFF
--- a/src/Codeception/Util/ReflectionPropertyAccessor.php
+++ b/src/Codeception/Util/ReflectionPropertyAccessor.php
@@ -53,11 +53,17 @@ class ReflectionPropertyAccessor
             foreach ($constructor->getParameters() as $parameter) {
                 if ($parameter->isOptional()) {
                     $constructorParameters[] = $parameter->getDefaultValue();
-                } elseif (array_key_exists($parameter->getName(), $data)) {
+                } elseif (\array_key_exists($parameter->getName(), $data)) {
                     $constructorParameters[] = $data[$parameter->getName()];
-                } else {
+                } elseif (false === \is_a($obj, $class)) {
+                    /*
+                     * In case we need to create the entity via reflection, we need to have all constructor parameters
+                     * in the $data array, in case we only want to enhance an existing instance, the given $obj has
+                     * already been constructed with all of its parameters outside of this function,
+                     * so we can skip throwing this exception.
+                     */
                     throw new InvalidArgumentException(
-                        'Constructor parameter "'.$parameter->getName().'" missing'
+                      'Constructor parameter "'.$parameter->getName().'" missing'
                     );
                 }
             }
--- a/src/Codeception/Util/ReflectionPropertyAccessor.php
+++ b/src/Codeception/Util/ReflectionPropertyAccessor.php
@@ -53,7 +53,7 @@ class ReflectionPropertyAccessor
          * already been constructed with all of its parameters outside of this function,
          * so we can skip creating it via reflection.
          */
-        if ($this->isObjectOf($obj, $class)) {
+        if ($this->isNoObjectOf($obj, $class)) {
             $obj = $this->createObjectViaReflection($reflectedEntity, $data);
         }
 
@@ -106,7 +106,7 @@ class ReflectionPropertyAccessor
      *
      * @return bool
      */
-    private function isObjectOf($obj, $class)
+    private function isNoObjectOf($obj, $class)
     {
         return false === \is_a($obj, $class);
     }
--- a/src/Codeception/Util/ReflectionPropertyAccessor.php
+++ b/src/Codeception/Util/ReflectionPropertyAccessor.php
@@ -130,7 +130,7 @@ class ReflectionPropertyAccessor
                     $constructorParameters[] = $data[$parameter->getName()];
                 } else {
                     throw new InvalidArgumentException(
-                      'Constructor parameter "'.$parameter->getName().'" missing'
+                        'Constructor parameter "'.$parameter->getName().'" missing'
                     );
                 }
             }
--- a/src/Codeception/Util/ReflectionPropertyAccessor.php
+++ b/src/Codeception/Util/ReflectionPropertyAccessor.php
@@ -51,7 +51,7 @@ class ReflectionPropertyAccessor
          * In case we need to create the entity via reflection, we need to have all constructor parameters
          * in the $data array, in case we only want to enhance an existing instance, the given $obj has
          * already been constructed with all of its parameters outside of this function,
-         * so we can skip throwing this exception.
+         * so we can skip creating it via reflection.
          */
         if ($this->isObjectOf($obj, $class)) {
             $obj = $this->createObjectViaReflection($reflectedEntity, $data);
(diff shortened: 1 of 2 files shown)


burned42:
@Basster 
> In general I'm not a friend of comments in code, but the one in ReflectionPropertyAccessor.php describes the behaviour and why we do the is_a() call.

I'm with you on this one :) So since I already took a look at the code, the solution I had in mind was a bit more like this:
```php
// instead of the constructor block and this line
$obj = $obj ?: $reflectedEntity->newInstance(...$constructorParameters);

// maybe something like this:
if (!$obj) { // or false === \is_a($obj, $class) like you proposed
    $constructorParameters = [];
    // ... rest of the "constructor block"

    $obj = $reflectedEntity->newInstance(...$constructorParameters);
}
```
So this way I think it's obvious that we only go into the "constructor parameter" path if we don't already have an object and all the constructor parameter juggling only happens if we need to have those to call `$reflectedEntity->newInstance(...$constructorParameters)` afterwards. What do you think about this?

Basster:
@burned42 

> So this way I think it's obvious that we only go into the "constructor parameter" path if we don't already have an object

Good catch. Will try this.

Basster:
As far as I can see the failing tests are not related to these changes...

DavertMik:
Looks fine to me. 
@alexkunin any comments on this change?

src/Codeception/Util/ReflectionPropertyAccessor.php
```diff
@@ -47,29 +47,23 @@ private function setPropertiesForClass($obj, $class, array $data)
...
-                    );
-                }
-            }
+        /*
+         * In case we need to create the entity via reflection, we need to have all constructor parameters
+         * in the $data array, in case we only want to enhance an existing instance, the given $obj has
+         * already been constructed with all of its parameters outside of this function,
+         * so we can skip creating it via reflection.
+         */
+        if ($this->isNoObjectOf($obj, $class)) {
```
alexkunin: I'm not exactly sure what is going on here, as `setPropertiesForClass` is a private method and is called with either `$obj = null`, or `$class = get_class($obj)`. So, `$obj` is always a `null` or instance of `$class`, guaranteed. Why do we use `is_a` here, and why `is_a` is wrapped into separate method? Not saying this is wrong, just can't see the reasoning behind this.

src/Codeception/Util/ReflectionPropertyAccessor.php
```diff
@@ -105,4 +99,43 @@ public function createWithProperties($class, array $data)
...
+    private function createObjectViaReflection(ReflectionClass $reflectedEntity, array $data)
+    {
+        $constructorParameters = [];
+        $constructor           = $reflectedEntity->getConstructor();
+        if (null !== $constructor) {
+            foreach ($constructor->getParameters() as $parameter) {
+                if ($parameter->isOptional()) {
+                    $constructorParameters[] = $parameter->getDefaultValue();
+                } elseif (\array_key_exists($parameter->getName(), $data)) {
+                    $constructorParameters[] = $data[$parameter->getName()];
```
alexkunin: With this order of `if` branches, does it mean we can't pass a value to optional constructor parameter? It looks like switching `if` and `elseif` would allow that.

alexkunin:
Please, see my comments above.

src/Codeception/Util/ReflectionPropertyAccessor.php
Basster: For me its the responsibility of `setPropertiesForClass` method itself to check whether we deal with an instantiated object of exactly this class and not the responsibility of the caller. So whenever anyone (in the future?) will call `setPropertiesForClass` with an object not instance of `$class`, it will create an instance of the given class instead.

The wrapper is just my understanding of clean code, so it reads better than any if clause with a `false` in it. (less cognitive load).

src/Codeception/Util/ReflectionPropertyAccessor.php
Basster: I've checked it (and will push another test in a minute). Here, we just collect constructor parameters and instantiate the obj with the default value(s). Since the instance continues to run through `setPropertiesForClass` any given value for an optional constructor parameter would be set afterwards.

src/Codeception/Util/ReflectionPropertyAccessor.php
alexkunin: As the one who wrote the method (and whole class), I assure you that this private method expects either `null` or object instance of specified class, by design -- this is a helper method, nothing more. Feel free to change name to a clearer one, or add a comment, but please do not change semantics without any reason and any comment.

Creating instance of given class instead (and discarding passed instance) is something I'd call "huge surprise". Throwing exception instead is much more clear behavior. Or, even better, leave original design alone.

alexkunin: Re wrapper, you're wrapping function which checks if object is an instance of given class with a function which... does literally the same, without adding anything new. This *is* more cognitive load, because future reader will look into that function to see basically nothing.

src/Codeception/Util/ReflectionPropertyAccessor.php
alexkunin: > Since the instance continues to run through setPropertiesForClass any given value for an optional constructor parameter would be set afterwards.

Is my understanding correct that this assumes the constructor does nothing with arguments and just sets private property value? In this case we could just drop whole thing, use `ReflectionClass::newInstanceWithoutConstructor` and then set all properties directly. I think whole point of that code (added by @burned42, IIRC) was to allow full use of constructor. Maybe it would be better to direct this question to him, or just fix order of branches.

alexkunin:
Replied to all comments. That said, I'm just a contributor, not affiliated with Codeception in any meaningful way. If @Naktibalda and/or @DavertMik think your PR can be accepted as is, I'm fine with that.

src/Codeception/Util/ReflectionPropertyAccessor.php
burned42: The reason why I decided to put the 'default parameter' branch first was that I thought this way existing tests, involving classes with only optional parameters, would still work the same way as before (before, no parameters where passed to the constructor). So I was basically just trying to avoid breaking existing tests. If you don't think that's a problem then I guess switching the order of branches would be a good way to go :)

DavertMik:
@alexkunin 

> I'm just a contributor, not affiliated with Codeception

That should be fixed :smile:

DavertMik:
@alexkunin I added you to our dev chat. Hope to see you there!

DavertMik:
Sorry, @Basster we can't accept this PR because it is overcomplicated. We can't maintain and support code we don't understand. Personally, I can't.  

@alexkunin suggests you start with a failing test to show the issue. Maybe it could be fixed in a simpler manner.

PR #5664, opened 2019-08-23
[Doctrine2] Let haveInRepository handle composite keys of related entities properly.

### LINKED ISSUE #5663: [Doctrine2] var_export does not handle circular references after extractPrimaryKey under certain circumstances

Hands down, I have a pretty complex Entity structure:

```php

/** @Entity **/
class A {}

/** @Entity **/
class B {}

/** @Entity **/
class C{}

/** @Entity **/
class D {
    /**
     * @ORM\Id
     * @ORM\ManyToOne(targetEntity="A", inversedBy="as")
     */
    private $a;

    /**
     * @ORM\Id
     * @ORM\ManyToOne(targetEntity="B", inversedBy="bs")
     */
    private $b;

    /**
     * @ORM\Id
     * @ORM\ManyToOne(targetEntity="C", inversedBy="cs")
     */
    private $c;
}
```
All `ManyToOne` relations are bidirectional!

Now I try to  (pretty simplified):

```php
$a = new A();
$b = new B();
$c = new C();
$d = new D($a, $b, $c);

$I->haveInRepository($d, ['a' => $a, 'b' => $b, 'c' => $c]);
``` 
`haveInRepository` is working correctly, *but* the debug line right before the `return` utilizes a `var_export` to display the primary keys and fails with `var_export does not handle circular references`!

### Details

* Codeception version: 3.1.0
* PHP Version: 7.3
* Operating System: doesn't matter
* Installation type: Composer
* List of installed packages: doesn't matter
* Suite configuration: doesn't matter

Basster:
Fixes #5663.

Maybe this is a BC break?!

Updated haveInRepository to return a hashmap of FQCN => $pk when composite ids are entities, as well.

### DIFF
--- a/src/Codeception/Module/Doctrine2.php
+++ b/src/Codeception/Module/Doctrine2.php
@@ -637,7 +637,11 @@ EOF;
         if ($metadata->isIdentifierComposite) {
             $pk = [];
             foreach ($metadata->identifier as $field) {
-                $pk[] = $rpa->getProperty($instance, $field);
+                $id = $rpa->getProperty($instance, $field);
+                if (is_object($id)) {
+                    $id = [get_class($id) => $this->extractPrimaryKey($id)];
+                }
+                $pk[] = $id;
             }
         } else {
             $pk = $rpa->getProperty($instance, $metadata->identifier[0]);
--- a/src/Codeception/Module/Doctrine2.php
+++ b/src/Codeception/Module/Doctrine2.php
@@ -427,7 +427,8 @@ EOF;
      * If the entity has a constructor, for optional parameters the default value will be used and for non-optional parameters the given fields (with a matching name) will be passed when calling the constructor before the properties get set directly (via reflection).
      *
      * Returns primary key of newly created entity. Primary key value is extracted using Reflection API.
-     * If primary key is composite, array of values is returned.
+     * If primary key is composite, array of values is returned. If primary key
+     * is composite of related entites, array of [classname, pk] arrays is returned.
      *
      * ```php
      * $I->haveInRepository('Entity\User', array('name' => 'davert'));
@@ -639,7 +640,7 @@ EOF;
             foreach ($metadata->identifier as $field) {
                 $id = $rpa->getProperty($instance, $field);
                 if (is_object($id)) {
-                    $id = [get_class($id) => $this->extractPrimaryKey($id)];
+                    $id = [get_class($id), $this->extractPrimaryKey($id)];
                 }
                 $pk[] = $id;
             }
--- a/tests/data/doctrine2_entities/CompositePrimaryKeyEntityWithManyToOneKeys.php
+++ b/tests/data/doctrine2_entities/CompositePrimaryKeyEntityWithManyToOneKeys.php
@@ -1,0 +1,37 @@
+<?php
+
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * Class EntityWithCompositeIdOfRelatedEntities
+ *
+ * @ORM\Entity
+ */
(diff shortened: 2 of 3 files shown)


Naktibalda:
@alexkunin Please review this pull request too.

Naktibalda:
@alexkunin This is your last chance to review this change, I will merge it tomorrow.

alexkunin:
Looks good to me, except not sure why return `['class_name' => $key_value]` instead of just `$key_value`. That's not a showstopper, just can't see the reason behind that.

Basster:
> not sure why return `['class_name' => $key_value]` instead of just `$key_value`.

From my point of view, when I say `haveInRepository(MyEntity::class, [...])` its pretty clear what kind of entity the PK belongs to. In case of composite keys of related entities, just values without a context would confuse more than help. Maybe I want to do more stuff with the related entities, so I can go for e.g. `grabFromRepository('class_name', $key_value)` and do more stuff with this entity...

alexkunin:
> its pretty clear what kind of entity the PK belongs to

True, but variable key is pretty cumbersome to handle (IMO even `['class', $value]` works better than `['class' => $value]`). I mean, if you know class name, how are you going to get value? Can you make a full sample that includes your snippet `grabFromRepository('class_name', $key_value)`, please?

> just values without a context

Context is pretty clear -- Doctrine entity; primary key values are in the same order as in Doctrine meta data. I.e. "least surprise" principle is not broken here.

Basster:
> True, but variable key is pretty cumbersome to handle (IMO even `['class', $value]` works better than `['class' => $value]`).

Yes, you are right. As I think about it, the variable key is in fact pretty cumbersome. I'd go with the `['class', $value]` approach as it differs from scalar values (to make clear its a related object, not a scalar) but keep the info to the related entity class.

alexkunin:
After giving it another thought, I think `['class', $value]` is not much better -- how one can get value if class name is known? By iterating over all elements of array, which is still cumbersome (though less then key-value format).

Perfectly usable would be using class name as key of the outer array, e.g. `['class1' => 123, 'class2' => 456]`, but there are two components of composite key that are of the same type, we'll get key clash.

So, there are two promising formats:

- `['field_name1' => 123, 'field_name2' => 456]`,
- `[123, 456]` (i.e. my original format)
- or both, as an array with mixed keys, but this is not very useful for testing, I think (due to verbosity, which leads to complex assert statements)

alexkunin:
Forgot to mention, my original idea was to return entity which was created/modified by `haveInRepository`. I have postponed that because this is backward incompatible change, but I think that's pretty good solution which is convenient, and also resolves all the problems mentioned above. Maybe maintainers (@Naktibalda, @DavertMik) have something to say about this? Maybe schedule it for 4.0? Or come up with a new method name and (eventually) deprecate `haveInRepository`?

Basster:
I understand your position. 
And I also think that the initial return value (array of objects) was fine and the most correct one, besides the fact that the `var_export` at the debug statement, has no limited depth and stopped working in scenarios with circular references, so I was looking for a more "plain" solution.

Maybe we should also drop this workaround and find a better way for the debug statement?

DavertMik:
>  Maybe schedule it for 4.0? Or come up with a new method name and (eventually) deprecate haveInRepository?

Maybe. For 4.0 we will decouple core from modules so you will be able to break things unrelated to the core.

DavertMik:
I can't get deep into this but seems fine to me if we have 2 approvals

DavertMik:
This will be merged on next release

DavertMik:
Sorry, I have to take my words back. I can't accept PR which I don't understand. And I'm not sure the case of using composite entity keys is widely used.

PR #5667, opened 2019-08-26
Do not fail steps for retry and tryTo step decorators

### LINKED ISSUE #5666: Retry And TryTo step decorators fail Steps
I want to check failed step in my custom extension with $stepEvent->getStep()->hasFailed() method.

#### What do you get instead?
I've got a wrong failed step (it's always before the real one). The reason is step decorators: Retry and TryTo. 
Retry should be failed only if all attempts left. TryTo shouldn't fails at all. 

```bash
I assert equals "tryTo","dontTryDoIt","TryTo fail"
  Failed to perform: TryTo fail
  Failed asserting that two strings are equal., skipping...
  Step: assert equals "tryTo","dontTryDoIt","TryTo fail"
  Is Failed: 1
I assert false true,"The real one problem"
  Step: assert false true,"The real one problem"
  Is Failed: 1

 Test  tests/acceptance/TempCest:checkTryTo
 Step  Assert equals "tryTo","dontTryDoIt","TryTo fail"
 Fail  The real one problem
```

Extension:
```php
<?php

use \Codeception\Events;

class FailInfoExtension extends \Codeception\Extension 
{
	public static $events = [
		Events::STEP_AFTER => 'stepAfter',
	];

	public function stepAfter(\Codeception\Event\StepEvent $e) {
		codecept_debug('Step: ' . $e->getStep());
		codecept_debug('Is Failed: ' . $e->getStep()->hasFailed());
	}
}
```
Test:
```php
class TempCest 
{
	public function checkTryTo(AcceptanceTester $I) {
		$I->amOnPage('/');
		$I->tryToAssertEquals('tryTo', 'dontTryDoIt', 'TryTo fail');
		$I->assertFalse(true, 'The real one problem');
	}
}
```

### Details

* Codeception version: 3.1.0
* PHP Version: 7.3.7
* Installation type: Composer
* Suite configuration:

```yml
step_decorators:
  - \Codeception\Step\Retry
  - \Codeception\Step\TryTo
```

Mitrichius:
fixes #5666

### DIFF
--- a/src/Codeception/Step.php
+++ b/src/Codeception/Step.php
@@ -34,6 +34,7 @@ abstract class Step
     protected $metaStep = null;
 
     protected $failed = false;
+    protected $isTry = false;
 
     public function __construct($action, array $arguments = [])
     {
@@ -264,6 +265,9 @@ abstract class Step
         try {
             $res = call_user_func_array([$activeModule, $this->action], $this->arguments);
         } catch (\Exception $e) {
+            if ($this->isTry) {
+                throw $e;
+            }
             $this->failed = true;
             if ($this->getMetaStep()) {
                 $this->getMetaStep()->setFailed(true);
--- a/src/Codeception/Step/Retry.php
+++ b/src/Codeception/Step/Retry.php
@@ -43,10 +43,11 @@ EOF;
         $interval = $this->retryInterval;
         while (true) {
             try {
+                $this->isTry = $retry < $this->retryNum;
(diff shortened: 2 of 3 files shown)


DavertMik:
Thanks!

DavertMik:
Thanks. The solution is elegant. A really nice addition!

DavertMik:
I will try to make tests for this

Issue #5532, opened 2019-06-02
 deprecated since Symfony 4.3


ronisaha:
#### What are you trying to achieve?

While I'm trying to run unit/functional/acceptance test i'm getting DEPRECATION message
 
#### What do you get instead?

> Provide console output if related. Use `-vvv` mode for more details.

```bash
19x DEPRECATION: Calling the "Symfony\Component\EventDispatcher\EventDispatcherInterface::dispatch()" method with the event name as first argument is deprecated since Symfony 4.3, pass it second and provide the event object first instead. /Users/Projects/bits/budget/testtest/vendor/symfony/event-dispatcher/EventDispatcher.php:58
DEPRECATION: The "Symfony\Component\HttpKernel\Client" class is deprecated since Symfony 4.3, use "Symfony\Component\HttpKernel\HttpKernelBrowser" instead. /Users/Projects/bits/budget/testtest/vendor/symfony/http-kernel/Client.php:14
4x DEPRECATION: The "Symfony\Component\BrowserKit\Response::getStatus()" method is deprecated since Symfony 4.3, use getStatusCode() instead. /Users/Projects/bits/budget/testtest/vendor/symfony/browser-kit/Response.php:100

```
> Provide test source code if related

```php
// paste test
```
### Details

* Codeception version: 3.0.1
* PHP Version: 7.3
* Operating System: macOS Mojave
* Installation type: Phar || Composer
* List of installed packages (`composer show`)

```bash
behat/gherkin                       v4.6.0     Gherkin DSL parser for PHP 5.3
codeception/codeception             3.0.1      BDD-style testing framework
codeception/phpunit-wrapper         8.0.4      PHPUnit classes used by Codeception
codeception/stub                    2.1.0      Flexible Stub wrapper for PHPUnit's Mock Builder
doctrine/annotations                v1.6.1     Docblock Annotations Parser
doctrine/cache                      v1.8.0     Caching library offering an object-oriented API for many cache backends
doctrine/collections                v1.6.1     PHP Doctrine Collections library that adds additional functionality on top of PHP arrays.
doctrine/common                     v2.10.0    PHP Doctrine Common project is a library that provides additional functionality that other Doctrine projects depend on such as b...
doctrine/dbal                       v2.9.2     Powerful PHP database abstraction layer (DBAL) with many features for database schema introspection and management.
doctrine/doctrine-bundle            1.11.1     Symfony DoctrineBundle
[...]

PR #5670, opened 2019-09-06
Use Symfony5 class names in connectors

Naktibalda:
#5532

Preparation for Symfony 5, also removes deprecation message when Symfony 4.3 components are used.

HttpKernelBrowser alias must be set in module file because Symfony/HttpKernel package isn't loaded yet by the time shim.php runs because it isn't dependency of Codeception.


DavertMik:
I thought that would be worse but I think we are fine with this fix. If class aliasing a few files fix this it is fine!

PR #5671, opened 2019-09-06
Minor updates to docs

DavertMik:
Well, I restructured and updated some documentation parts. Maybe we need full documentation refresh but I'm not the person who can do that.


docs/01-Introduction.md
```diff
@@ -16,56 +16,53 @@ To test the behavior of your application as a whole, you should write functional
 Codeception supports all three testing types.
 Out of the box you have tools for writing unit, functional, and acceptance tests in a unified framework.
 
-| | Codeception Unit Tests | Codeception Functional Tests | Codeception Acceptance Tests
+| | Unit Tests | Functional Tests | Acceptance Tests
```
omentes: May be we should add Codeception Api tests?

DavertMik: they can be functional or acceptance by their nature, so they won't fit into the table (

PR #5672, opened 2019-09-06
Tests for retry

DavertMik:
Tests for #5667


DavertMik:
I accidentally commited #5671 
ok, I will wait to merge that

tests/cli/RetryCest.php
```diff
@@ -31,8 +31,8 @@ public function failForTime(CliGuy $I)
     public function failForInterval(CliGuy $I)
     {
         $I->executeFailCommand('run --debug -g fail2');
-        $I->seeInShellOutput('Retrying #1 in 100ms');
         $I->seeInShellOutput('Retrying #2 in 200ms');
+        $I->seeInShellOutput('Retrying #1 in 100ms');
         $I->seeInShellOutput('Retrying #3 in 400ms');
```
Naktibalda: Have you changed the order of these assertions by accident?
Please change it back.

tests/unit/Codeception/Step/TryToTest.php
```diff
@@ -0,0 +1,37 @@
...
+    protected $shouldFail = true;
+
+    public function testTryToShouldReturnSuccess()
+    {
+        $moduleContainer = Stub::make(ModuleContainer::class, [
+            'moduleForAction' => $this
+        ]);
+        $try = new \Codeception\Step\TryTo('_executeFailedCode', []);
+        $val = $try->run($moduleContainer);
+        $this->assertEquals(false, $val);
```
Naktibalda: `assertFalse` would be better here.

tests/unit/Codeception/Step/RetryTest.php
```diff
@@ -0,0 +1,31 @@
...
+    protected $shouldFail = true;
+
+    public function testRetryStepShouldNotFailStep()
+    {
+        $moduleContainer = Stub::make(ModuleContainer::class, [
+            'moduleForAction' => $this
+        ]);
+        $retry = new \Codeception\Step\Retry('_executeFailedCode', [], 2, 0);
+        $retry->run($moduleContainer);
+
```
Naktibalda: I must admit that I can't understand retry tests, including older timing based tests, at all.
It would be good to have a detailed explanation in comments what actually happens here.

tests/cli/RetryCest.php
DavertMik: Ouch, I thought I didn't...

tests/unit/Codeception/Step/TryToTest.php
DavertMik: True

tests/unit/Codeception/Step/RetryTest.php
DavertMik: Ok

DavertMik:
@Naktibalda I updated tests. Please review again

PR #5674, opened 2019-09-08
Fixes the issue with paths in `groups` section when `codeception.yaml` is not in the root of the project

maks-rafalko:
* Fixes the issue with paths in `groups` section when `codeception.yaml` is not in the root of the project
* Add an ability to use absolute path for `groups` key in `codeception.yaml`

Example:

Imagine, custom `codeception.yaml` is located inside `tests/_data` folder.

We run codeception as

```bash
codecept -g x -c tests/_data/codeception.yaml
```

And we have the following group:

```yaml
groups:
    x:
        ../../tests/unit/SomeTest.php
```

Such path is resolved to `/path/to/project/tests/_data/../../tests/unit/SomeTest.php`
and `GroupManager.php` can not match this path with the real path:

```
/path/to/project/tests/unit/SomeTest.php
```

https://github.com/Codeception/Codeception/blob/014faba8ac4d0985a2076f194655000a7226d7e5/src/Codeception/Lib/GroupManager.php#L116-L118

Applying `realpath()` solves the issue, because after that paths are the same and the code above successfully matches them.

Also, I've added an ability to use absolute paths:

```yaml
groups:
    x:
        /path/to/project/tests/unit/SomeTest.php
```

We need it to integrate Codeception with [Infection](https://github.com/infection/infection).

---

Related to infection/infection#45


maks-rafalko:
To be honest, have no idea how to fix the tests, because when I try to `var_dump()` anything in `tests/unit/Codeception/Lib/GroupManagerTest.php`, it's not being displayed to the console :|

I would appreciate any help / advice here.

Naktibalda:
@maks-rafalko try using `codecept_debug` instead of `var_dump` in tests and run codeception with `-vv` or `-vvv` flag to see it.

I think that these tests could be failing because they use made-up file names, so `realfile` returns `false` - https://github.com/Codeception/Codeception/blob/3.1/tests/unit/Codeception/Lib/GroupManagerTest.php#L15-L49

tests/data/UserTest.php
```diff
@@ -0,0 +1,6 @@
+<?php
+
+final class PostTest
```
Naktibalda: Incorrect class name here.

tests/unit/Codeception/Lib/GroupManagerTest.php
```diff
@@ -14,15 +14,22 @@ class GroupManagerTest extends \Codeception\Test\Unit
...
         $test3 = $this->makeTestCase('UserTest.php', 'testNot');
         $this->assertContains('important', $this->manager->groupsForTest($test1));
         $this->assertContains('important', $this->manager->groupsForTest($test2));
         $this->assertNotContains('important', $this->manager->groupsForTest($test3));
     }
 
+    public function testRealPathForFileWithMethodName()
+    {
+        $this->manager = new GroupManager(['important' => ['tests/data/PostTest.php:testName']]);
+        $test = $this->makeTestCase('tests/data/PostTest.php', 'testName');
```
Naktibalda: There is a lot of files in tests/data already, it would be a little bit cleaner to move these files to subdirectory, e.g. `tests/data/GroupManagerTest`

maks-rafalko: I named it `group_manager_test` to be consistent with the other folder names on the same level

maks-rafalko:
@Naktibalda I have resolved your comments and tests are green

DavertMik:
Thank you, @maks-rafalko ! 
Let's have it merged. I hope we will make a release very soon

maks-rafalko:
Thank you both

PR #5680, opened 2019-09-11
[Doctrine2] Bugfix: calling haveInRepository with preconstructed entity requires providing constructor parameters

burned42:
As pointed out [here](https://github.com/Codeception/Codeception/pull/5633#issuecomment-523851706) my PR seems to have broken `haveInRepository` in that it doesn't work with preconstructed entities anymore if their constructor has non-optional parameters and you don't provide those properties when calling `haveInRepository`. :man_facepalming: 

As I already offered a pretty simple solution in a commend as a suggestion [here](https://github.com/Codeception/Codeception/pull/5662#issuecomment-523906151) and said PR was rejected as too complex, I'm now offering my suggested (hopefully simple enough) solution along with a test as a PR :)

### DIFF
--- a/src/Codeception/Util/ReflectionPropertyAccessor.php
+++ b/src/Codeception/Util/ReflectionPropertyAccessor.php
@@ -47,23 +47,26 @@ class ReflectionPropertyAccessor
     {
         $reflectedEntity = new ReflectionClass($class);
 
-        $constructorParameters = [];
-        $constructor = $reflectedEntity->getConstructor();
-        if (null !== $constructor) {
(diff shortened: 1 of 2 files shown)


DavertMik:
Semaphoreci fails for a non related reason, it's fine

@alexkunin please review

burned42:
@DavertMik I guess that failing CI might be because I based that branch on `3.0` instead of `3.1`, fixed that now.

burned42:
After I rebased the branch on `3.1` all CI tests have passed now :)

DavertMik:
Thanks!

PR #5683, opened 2019-09-12
Update Asserts.md

ruudboon:
Added notice about deprecated expectException()

### DIFF
--- a/docs/modules/Asserts.md
+++ b/docs/modules/Asserts.md
@@ -396,7 +396,9 @@ Checks that condition is positive.
 
 
 ### expectException
- 
+
+Note: Deprecated. Use expectThrowable()
+
 Handles and checks exception called inside callback function.
 Either exception class name or exception instance should be provided.
 


docs/modules/Asserts.md
```diff
@@ -396,7 +396,9 @@ Checks that condition is positive.
 
 
 ### expectException
- 
+
+Note: Deprecated. Use expectThrowable()
+
```
Naktibalda: This file is autogenerated from code, so your change will be lost.
The right place to make the change is https://github.com/Codeception/Codeception/blob/3.1/src/Codeception/Module/Asserts.php#L107

This method had `@deprecated` tag when `expectThrowable` was introduced, but it disappeared later: https://github.com/Codeception/Codeception/pull/5213/files#diff-b3d3622a11b4688b45ee6999ce435bc0

It looks like I tried to remove this method completely, but had to put it back for some reason and `@deprecated` line was lost in the process
https://github.com/Codeception/Codeception/commit/7511fd2ceabe071cb65c0f38ae459c2d392ebefa#diff-b3d3622a11b4688b45ee6999ce435bc0

DavertMik:
Ok, taken. I will do the fix by myself

ruudboon:
Thnx. Only seeing this now

Naktibalda:
Sebastian Bergmann changes his mind about exception testing so often.
expectException is not currently deprecated,
but it will be deprecated in PHPUnit 9 again: https://github.com/sebastianbergmann/phpunit/issues/3776

Here is some evidence that Sebastian updated some PHPUnit code to use expectException at the beginning of this year: https://github.com/sebastianbergmann/phpunit/commit/f56cb265dac29f717a3fda2c0ef5114341b50d6e

I only replaced really deprecated `setExpectedException` method with `expectException` in March, so I guess that I removed a previous `@deprecated` message that you reintroduced now.

@ruudboon Could you point me to a source which says that `expectException` is currently deprecated?

PR #5709, opened 2019-10-03
[Webdriver] fix cookies debug

### LINKED ISSUE #5696: Problems with cookies in console debug mode
I want to see cookies content when console debug mode is enabled. 

#### What do you get instead?
Content of the cookies is missing.
Also when I load session snapshot with many cookies, they are shown up in debug many times (see console output below)

```bash
  [Cookies] [{}]
  [Cookies] [{},{}]
  [Cookies] [{},{},{}]
  [Cookies] [{},{},{},{}]
  [Cookies] [{},{},{},{},{}]
  [Cookies] [{},{},{},{},{},{}]
```

```php
protected function runMultiple()
{
	return [
		['run' => 0],
		['run' => 1],
	];
}

/**
 * @dataProvider runMultiple
 */
public function cookieTest(AcceptanceTester $I, \Codeception\Example $example){
	$I->amOnUrl('https://ya.ru');
	if ($I->loadSessionSnapshot('snapshot')) {
		return;
	}
	$I->saveSessionSnapshot('snapshot');
}
```
### Details

* Codeception version: 3.1.1
* PHP Version: 7.3.7
* Installation type: Composer

Mitrichius:
fixes #5696

Current result:
```
I load session snapshot "snapshot"
  [Cookies] [{"name":"yp","value":"1571926489.ygu.1","path":"/","domain":"ya.ru","expiry":1569420892,"secure":true,"httpOnly":false},{"name":"yandex_gid","value":"177","path":"/","domain":"ya.ru","expiry":1569420892,"secure":true,"httpOnly":false},{"name":"zm","value":"m-white_yaru.css%3As3home-static_FSsLNuVhmDl3rxvbm-qvz0ZjVdk%3Al","path":"/","domain":"ya.ru","expiry":1569420892,"secure":true,"httpOnly":false},{"name":"mda","value":"0","path":"/","domain":"ya.ru","expiry":1569420892,"secure":true,"httpOnly":false},{"name":"yandexuid","value":"9868385691569334489","path":"/","domain":"ya.ru","expiry":1569420892,"secure":true,"httpOnly":false},{"name":"i","value":"QusP9QfwDnr5s5+Zvdp1lHzsPFy3jQFyJ81z2fdnVx6kVtKwrFroy450dp8iXl1LDSQAfiei1hNw9tjhFrv0i3HIJLY=","path":"/","domain":"ya.ru","expiry":1569420892,"secure":true,"httpOnly":false}]
  [Snapshot] Restored "snapshot" session snapshot
```

### DIFF
--- a/src/Codeception/Module/WebDriver.php
+++ b/src/Codeception/Module/WebDriver.php
@@ -795,6 +795,16 @@ class WebDriver extends CodeceptionModule implements
         $this->webDriver->manage()->window()->setSize(new WebDriverDimension($width, $height));
     }
 
+    private function debugCookies()
+    {
+        $result = [];
+        $cookies = $this->webDriver->manage()->getCookies();
+        foreach ($cookies as $cookie) {
+            $result[] = is_array($cookie) ? $cookie : $cookie->toArray();
+        }
+        $this->debugSection('Cookies', json_encode($result));
+    }
+
     public function seeCookie($cookie, array $params = [])
     {
         $cookies = $this->filterCookies($this->webDriver->manage()->getCookies(), $params);
@@ -804,7 +814,7 @@ class WebDriver extends CodeceptionModule implements
             },
             $cookies
         );
-        $this->debugSection('Cookies', json_encode($this->webDriver->manage()->getCookies()));
+        $this->debugCookies();
         $this->assertContains($cookie, $cookies);
     }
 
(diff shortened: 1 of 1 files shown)


src/Codeception/Module/WebDriver.php
```diff
@@ -3117,7 +3129,8 @@ public function loadSessionSnapshot($name)
         }
         
         foreach ($this->sessionSnapshots[$name] as $cookie) {
-            $this->setCookie($cookie['name'], $cookie['value'], (array)$cookie);
+            $showDebug = $cookie == end($this->sessionSnapshots[$name]);
+            $this->setCookie($cookie['name'], $cookie['value'], (array)$cookie, $showDebug);
```
Naktibalda: I think that instead of using this messy logic to detect the last cookie, it would be better to set `$showDebug` parameter to `false` in the loop and call `$this->debugCookies()` after loop.

Mitrichius: Yeah, fixed it

PR #5710, opened 2019-10-04
seeInDatabase with many rows

walterwhites:


walterwhites:
I would like check many rows with seeInDatabase:

instead of
```
$I->seeInDatabase('users', ['name' => 'Davert', 'email' => 'davert@mail.com']);
$I->seeInDatabase('users', ['name' => 'Pierre', 'email' => 'pierre@mail.com']);
$I->seeInDatabase('users', ['name' => 'Tom', 'email' => 'tom@mail.com']);
$I->seeInDatabase('users', ['name' => 'April', 'email' => 'april@mail.com']);
```

use
```
$I->seeInDatabase('users', [
    ['name' => 'Davert', 'email' => 'davert@mail.com'],
    ['name' => 'Pierre', 'email' => 'pierre@mail.com'],
    ['name' => 'Tom', 'email' => 'tom@mail.com'],
    ['name' => 'April', 'email' => 'april@mail.com']
]);
```

Naktibalda:
I don't think that this feature is necessary.

DavertMik:
This is nice to have but I see that current implementation is not complete.
It just goes row by row, and if one row fails stops the assertion entirely. I expect that such method will go through all rows and show in a failure message which rows were absent. So it should aggregate failures.

Also I would like to see correspondiong implementation for `dontSeeInDatabase`.

Naktibalda:
> This is nice to have

Is it?
It doesn't actually add any new functionality, it just reduces a size of the test file a little bit when there is a lot of repetition. As currently implemented, it is identical to series of `seeInDatabase` calls. If modified according to Davert's request, it would be identical to series of `canSeeInDatabase` calls.

If this is a valuable feature, then it would be even more valuable to have a multi-row `haveInDatabase` method.
Rather than making `seeInDatabase` do two different things: 1) check one row and 2) check multiple rows, it would be better to introduce new `seeRowsInDatabase` or `seeMultipleRowsInDatabase` method.

The next release will be the final release of Codeception 3, so it isn't a good time to introduce new features now, and we can't just change target branch to 4.0, because in Codeception 4, Db module has its own repository, 

Please raise a new pull request to https://github.com/Codeception/module-db

PR #5711, opened 2019-10-05
Fix for file extension of .fail files

jasny:
The `Content-Type` header [may contain a charset](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type#Syntax), which caused XML and JSON .fail files to have a .html extension.

    Content-Type: application/json; charset=utf-8

Also changed that if a REST API outputs text/plain content, the .fail file gets the .txt extension.


Naktibalda:
thanks

PR #5718, opened 2019-10-13
add --no-artifacts flag for run command

### LINKED ISSUE #5646: Add flag to suppress Artifacts reporting in 3.0
Hi, commit https://github.com/Codeception/Codeception/commit/cca5463cdefd12120783bca94081496f02f226e9#diff-cddc2fb9d072ba8acaf613bd6b2d14f6R368 introduced `Artifacts` printing when tests fail.

I have to say that even though this is a useful new features to be active by default, a lot of time they may be noisy and may bring no value while TDDing, so it would be nice to have a flag to disable that particular output from `Subscriber\Console::printReports` method.

Mitrichius:
Fixed #5646

### DIFF
--- a/src/Codeception/Command/Run.php
+++ b/src/Codeception/Command/Run.php
@@ -215,7 +215,7 @@ class Run extends Command
                 InputOption::VALUE_REQUIRED,
                 'Define random seed for shuffle setting'
             ),
-
+            new InputOption('no-artifacts', '', InputOption::VALUE_NONE, 'Don\'t report about artifacts'),
         ]);
 
         parent::configure();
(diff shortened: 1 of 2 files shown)


Naktibalda:
Thanks, it looks good to me, but this is a new feature, so I will merge it to 4.0 branch later.

4.0.0 will be released soon and it will be functionally identical to 3.1, but modules will be extracted to separate repositories, so they must be added to composer.json separately.

https://github.com/Codeception/Codeception/issues/1233#issuecomment-531565778

AndrewFeeney:
Thanks for this @Mitrichius! Now that 4.0.0 is out (and indeed 4.1.0), would it be possible to merge this @Naktibalda? Thanks for all your effort.

kowalk:
@Naktibalda when do you planning to merge this cool feature? :)

Naktibalda:
Released as 4.1.1

PR #5725, opened 2019-10-16
Use path resolver for bootstrap file detection

laoneo:
Somehow I got the following error:
_Call to undefined function Codeception\codecept_is_path_absolute() in /project/tests/vendor/codeception/codeception/src/Codeception/Configuration.php:278_

This patch uses the PathResolver.


laoneo:
Mot sure why travis failed, but it looks like that the error is not related to the patch.

Issue #5728, opened 2019-10-21
Dockerfile is correct?


bscheshirwork:
#### What are you trying to achieve?
I trying to compile my own version of Codeception. But I can't passin install composer step without copy `tests` folder (after 47 line):
https://github.com/Codeception/Codeception/blob/9c69ec708f15140be20341d342c330d68fb04197/Dockerfile#L47-L48
```
# Add test-code
COPY ./tests /repo/tests
```
in mainline all is work fine?

Naktibalda:
What error do you get?

bscheshirwork:
```
  [RuntimeException]                                                                                    
  Could not scan for classes inside "tests/cli/_steps" which does not appear to be a file nor a folder  
 ```

```sh
...

monolog/monolog suggests installing graylog2/gelf-php (Allow sending log messages to a GrayLog2 server)
monolog/monolog suggests installing sentry/sentry (Allow sending log messages to a Sentry server)
monolog/monolog suggests installing doctrine/couchdb (Allow sending log messages to a CouchDB server)
monolog/monolog suggests installing ruflin/elastica (Allow sending log messages to an Elastic Search server)
monolog/monolog suggests installing php-amqplib/php-amqplib (Allow sending log messages to an AMQP server using php-amqplib)
monolog/monolog suggests installing ext-amqp (Allow sending log messages to an AMQP server (1.0+ required))
monolog/monolog suggests installing ext-mongo (Allow sending log messages to a MongoDB server)
monolog/monolog suggests installing mongodb/mongodb (Allow sending log messages to a MongoDB server via PHP Driver)
monolog/monolog suggests installing aws/aws-sdk-php (Allow sending log messages to AWS services like DynamoDB)
monolog/monolog suggests installing rollbar/rollbar (Allow sending log messages to Rollbar)
monolog/monolog suggests installing php-console/php-console (Allow sending log messages to Google Chrome)
Writing lock file
Generating optimized autoload files

                                                                                                        
  [RuntimeException]                                                                                    
  Could not scan for classes inside "tests/cli/_steps" which does not appear to be a file nor a folder  
                                                                                                        

[...]

PR #5733, opened 2019-10-30
[Docker] Copying source code before composer install step

OneEyedSpaceFish:
Fixes: #5728


Naktibalda:
I made this change and also added a few modules and necessary PHP extensions in #5917

PR #5738, opened 2019-11-06
Add option to recurse includes when a single suite is selected (branch 3.1)

vertexvaar:
Same as #5737 but for the 3.1 branch (if still supported)


Naktibalda:
Branch 3.1 is unsupported.

PR #5750, opened 2019-11-16
Added upgrade script init template

DavertMik:
Here is a simple demonstration of upgrade script. It scans for all suites, gets all modules, and updates composer.json. Then runs composer update. Also I want to collect some info about our users and ask for donations. I think that's fair 🙃

codecept_upgrade

Current blocker for this: https://github.com/Codeception/module-db/blob/master/composer.json#L20
Modules depend on `@dev` version of Codeception. So `composer update` will fail. This should be fixed in modules.


Naktibalda:
I thought that `*@dev` allows any version to be installed, doesn't it.
I think that the actual problem for your script could be that we don't have a stable release of 4.0 yet.

AppVeyor jobs will be fixed by 4.0 cleanup PR, but I want to undo deletion of package/bin before it is merges.

src/Codeception/Template/Upgrade4.php
```diff
@@ -0,0 +1,137 @@
...
+        'ZF2' => 'codeception/module-zf2',
+        'Sequence' => 'codeception/module-sequence',
+        'SOAP' => 'codeception/module-soap',
+        'Redis' => 'codeception/module-redis',
+        'Queue' => 'codeception/module-queue',
+        'MongoDb' => 'codeception/module-mongodb',
+        'Memcache' => 'codeception/module-memcache',
+        'FTP' => 'codeception/module-ftp',
+        'DataFactory' => 'codeception/module-datafactory',
+        'Apc' => 'codeception/module-apc',
```
Naktibalda: It hurts my eyes, could you reorder modules in alphabetical order?

src/Codeception/Template/Upgrade4.php
```diff
@@ -0,0 +1,137 @@
...
+        $this->say('');
+        $result = $this->ask('<question>Did you fill in the survey?</question>', true);
+        if ($result) {
+            $this->say('Thank you! ');
+        } else {
+            $this->say('Anyway...');
+        }
+        if (!file_exists('composer.json')) {
+            $this->sayWarning('Please use composer installation of Codeception');
+            throw new \Exception('composer.json not found, can\'t run upgrade');
```
Naktibalda: Output of exception probably looks ugly, sayWarning followed by return would be better.

Naktibalda:
I'm not a fan of survey prompt.
Majority of users have multiple projects to update and you will ask them to fill survey every time.

Almost all users will run upgrade script because they read about it in announcement or documentation page.
I think that these articles are perfect and sufficient place for survey and open collective links.

Naktibalda:
I rebased this branch on 4.0

DavertMik:
> I think that these articles are perfect and sufficient place for survey and open collective links.

I thought about it as well. But unfortunately, no. People do ignore them. This is the only way to make communication. I know that's not the best option, but what to do when we need to collect real usage statistics. 

Anyway, the survey is absolutely optional. Everyone can skip it.

src/Codeception/Template/Upgrade4.php
Naktibalda: It looks good actually.

Naktibalda:
Actually, I didn't find any suites.

Naktibalda:
Upgrade script worked for a simple standard test structure of https://github.com/Naktibalda/codeception-zend-expressive-tests
```
+        "codeception/module-zendexpressive": "^1.0.0",
+        "codeception/module-rest": "^1.0.0"
```

Naktibalda:
I tested `codecept init api` and it failed with `Module REST could not be found and loaded`

Full output:
```
tttt$ ../codecept init api
Let's prepare Codeception for REST API testing

? Where tests will be stored? (tests)
? Start url for tests (http://localhost/api)
> Created test directories inside at tests
> Api helper has been created in tests/_support/Helper

In ModuleContainer.php line 80:

  Module REST could not be found and loaded


init [--path PATH] [--namespace [NAMESPACE]] [--] <template>
```

`init acceptance` failed with `Module WebDriver could not be found and loaded`
`init unit` didn't fail.

So api and acceptance templates must be updated to install relevant modules automatically.
Another thing that needs updating is Symfony recipe for Codeception.

Naktibalda:
When you are going to make announcement about this feature, please make a note that it isn't necessary to run it if the user is using phar file for running acceptance tests, because majority of modules is included in the phar file.

DavertMik:
@Naktibalda  I can remove the survey part if you are not ok with it. It's more about community building and business related stuff, not the code. But if you are not comfortable. I can move that out.

> So api and acceptance templates must be updated to install relevant modules automatically.

Yes, I didn't think about it. Thanks!


> Please get upgrade script working for yii-app-advanced.

Nope. If you upgrade advanced one you can run upgrade script for sub-suites. What I can improve is adding an option for config path. Which should be fine.

Naktibalda:
> @Naktibalda I can remove the survey part if you are not ok with it. It's more about community building and business related stuff, not the code. But if you are not comfortable. I can move that out.

No, it is fine. Actually I think that it isn't intrusive enough.
It would be better to make a 5 second pause right after printing the survey link.

> > Please get upgrade script working for yii-app-advanced.
> 
> Nope. If you upgrade advanced one you can run upgrade script for sub-suites. What I can improve is adding an option for config path. Which should be fine.

That's fine, but please document it. And please print a big red error if `Configuration::suites()` returned 0 suites.

DavertMik:
@Naktibalda not sure why and how but I spent 3 hours today making the changes :)
PLease, check. I updated all init scripts and error messages

src/Codeception/Template/Acceptance.php
```diff
@@ -99,6 +96,8 @@ public function setup()
         $this->gitIgnore($supportDir . DIRECTORY_SEPARATOR . '_generated');
         $this->sayInfo("Created test directories inside at $dir");
 
+        $this->addModulesToComposer(['WebDriver']);
```
Naktibalda: A special case for PHAR file is necessary.
When the phar file is used, composer.json file is not necessary and WebDriver and other modules are included to phar file.
Maybe something like this:
```php
if (!class_exists(\Codeception\Module\WebDriver:class)) {
  $this->addModulesToComposer(['WebDriver']);
}
```

Also when any modules is installed using composer, it effectively turns phar installation to composer installation.

composer.json
```diff
@@ -35,7 +35,12 @@
         "codeception/specify": "~0.3",
         "squizlabs/php_codesniffer": "~2.0",
         "vlucas/phpdotenv": "^3.0",
-        "symfony/process": ">=2.7 <5.0"
+        "symfony/process": ">=2.7 <5.0",
+        "codeception/module-filesystem": "^1.0",
+        "codeception/module-cli": "^1.0",
+        "codeception/module-asserts": "^1.1",
+        "codeception/module-phpbrowser": "^1.0",
+        "codeception/util-universalframework": "@dev"
```
Naktibalda: Please remove `composer.json` changes.

We have a circular dependency problem here.
In order to be able to test branches of those modules we must either remove module dependencies from this repo or add very loose dependencies as I already did in those modules:
https://github.com/Codeception/module-asserts/commit/7f0e038316d84cefc4098ea570225bf9debc0304.

Naktibalda: I raised pull request for that: #5772

DavertMik: Done!

src/Codeception/Template/Acceptance.php
DavertMik: Done

src/Codeception/Template/Upgrade4.php
```diff
@@ -0,0 +1,66 @@
...
+        if ($result) {
+            $this->say('Thank you! ');
+        } else {
+            $this->say('Anyway...');
+        }
+        $config = Configuration::config();
+        $modules = [];
+        $suites = Configuration::suites();
+        if (empty($suites)) {
+            throw new \Exception("No suites found in codeception.yml. Use upgrade script on config with suites");
```
Naktibalda: This message is confusing.
Suites are almost never configured in codeception.yml, unless it is single suite project where suite config is merged to codeception.yml.

`codecept init upgrade4 -c backend/` works, but it must be documented,

src/Codeception/Template/Upgrade4.php
```diff
@@ -0,0 +1,66 @@
...
+        }
+        foreach (Configuration::suites() as $suite) {
+            $suiteConfig = Configuration::suiteSettings($suite, $config);
+            $modules = array_merge($modules, Configuration::modules($suiteConfig));
+        }
+
+        $this->addModulesToComposer($modules);
+
+        $this->saySuccess("Done upgrading!");
```
Naktibalda: I think that this script should say something else when it had nothing to do,
`Done upgrading!` looks confusing when script did nothing to upgrade.
Printing something like 'Everything is fine already', 'I have nothing to do here' seems more appropriate.

```
$ codecept init upgrade4 -c backend/
Welcome to Codeception v4 Upgrade wizard!

Codeception is maintained since 2011, is free & open-source.
To make it better we need your feedback on it!

Please take a minute and fill in a brief survey:
http://bit.ly/codecept-survey

? Did you fill in the survey? (y/n) n
Anyway...
 Done upgrading!

Please consider donating to Codeception on regular basis:

https://opencollective.com/codeception

It's ok to pay for reliable software.
Talk to your manager & support further development of Codeception!
```

DavertMik:
Done!

DavertMik:
So are we ready to get this merged?

eXorus:
Good job @DavertMik  and @Naktibalda 
I'm so excited with Codeception 4, core, modules...

src/Codeception/Template/Upgrade4.php
```diff
@@ -0,0 +1,74 @@
...
+        } else {
+            $this->say('Anyway...');
+        }
+        $config = Configuration::config();
+        $modules = [];
+        $suites = Configuration::suites();
+        if (empty($suites)) {
+            $this->sayError("No suites found in current config.");
+            $this->sayWarning('If you use sub-configs with `include` option, run this script on subconfigs:');
+            $this->sayWarning('Example: php /vendor/bin/codecept init upgrade4 -c backend/');
```
Naktibalda: Please change `Example: php /vendor/bin/codecept`
to `Example: php vendor/bin/codecept`

DavertMik: Ouch, right!

DavertMik: Btw there is a way you could add suggestions so I can commit them automatically 
 
Selection_664

DavertMik:
```suggestion
            $this->sayWarning('Example: php vendor/bin/codecept init upgrade4 -c backend/');
```

DavertMik:
@Naktibalda done

PR #5765, opened 2019-11-23
WIP: Codeception 5

Naktibalda:


Naktibalda:
I'm changing my mind that it would be better to implement Symfony 5 support in Codeception 4.0 instead of releasing it as 5.0
Pros: Only one major version to maintain.
Cons: I don't get a clean sheet for PHPUnit integration just yet, but I can consider doing that when PHPUnit 9 comes out in February.

DavertMik:
I think it's not a good idea, as we will throw away everyone who wanted to use decoupled modules but old Symfony.

So I propose to stick to the current plan. 
Release 4.0 and 5.0
Maybe this will be the only 4.x version ever released but at least this will allow to install modules separately and receive new updates

Naktibalda:
@DavertMik This pull request was superseded by #5766

PR #5771, opened 2019-11-27
allow to override default Yii2 Logger class via DI container

albertborsos:
I have an enhanced Logger class which implements my `LoggerInterface`. All of my tests are failing because the default Logger class of Codeception is not implements this interface. I made a custom Logger class which extends the Codeception's Logger class and implements my LoggerInterface, but the Logger class is hardcoded.

With this PR it is possible to override  in the test application via the configuration of the DI container, with the following way:

```php
return [
     'components' => [
          ...
     ],
     'container' => [
          'definitions' => [
               \Codeception\Lib\Connector\Yii2\Logger::class => \tests\codeception\support\Logger::class,
          ],
     ],
];
```

### DIFF
--- a/src/Codeception/Lib/Connector/Yii2.php
+++ b/src/Codeception/Lib/Connector/Yii2.php
@@ -272,7 +272,7 @@ class Yii2 extends Client
         $config = $this->mockMailer($config);
         /** @var \yii\web\Application $app */
         Yii::$app = Yii::createObject($config);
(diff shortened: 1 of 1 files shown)


albertborsos:
similar ticket: #4943

Naktibalda:
There will be no more releases of Codeception 3.1 and Codeception 4.0 is modular, please make pull request to https://github.com/Codeception/module-yii2/ instead.

PR #5773, opened 2019-12-02
Support phpdotenv v4

sunspikes:
Support `vlucas/phpdotenv` `v4` along with the other versions and symfony dotenv.

### DIFF
--- a/src/Codeception/Lib/ParamsLoader.php
+++ b/src/Codeception/Lib/ParamsLoader.php
@@ -115,7 +115,14 @@ class ParamsLoader
     protected function loadDotEnvFile()
     {
         if (class_exists('Dotenv\Dotenv')) {
-            if (method_exists('Dotenv\Dotenv', 'create')) {
+            if (class_exists('Dotenv\Repository\RepositoryBuilder')) {
+                //dotenv v4
+                $repository = \Dotenv\Repository\RepositoryBuilder::create()
+                    ->withReaders([new \Dotenv\Repository\Adapter\ServerConstAdapter()])
+                    ->immutable()
+                    ->make();
+                $dotEnv = \Dotenv\Dotenv::create($repository, codecept_root_dir(), $this->paramStorage);
+            } elseif (method_exists('Dotenv\Dotenv', 'create')) {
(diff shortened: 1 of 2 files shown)


src/Codeception/Lib/ParamsLoader.php
```diff
@@ -115,7 +115,14 @@ protected function loadXmlFile()
...
         if (class_exists('Dotenv\Dotenv')) {
-            if (method_exists('Dotenv\Dotenv', 'create')) {
+            if (class_exists('Dotenv\Repository\RepositoryBuilder')) {
+                //dotenv v4
+                $repository = \Dotenv\Repository\RepositoryBuilder::create()
+                    ->withReaders([new \Dotenv\Repository\Adapter\ServerConstAdapter()])
+                    ->immutable()
+                    ->make();
+                $dotEnv = \Dotenv\Dotenv::create($repository, codecept_root_dir(), $this->paramStorage);
+            } elseif (method_exists('Dotenv\Dotenv', 'create')) {
```
Naktibalda: Backwards compatibility code with v3 and v2 was left intact below, but lowest supported version was raised to 4.0.
Usually we prefer to support as many versions as possible, because you can never know if some app is tied to specific version of a library.
Please change version constraint to `"^2.0 | ^3.0 | ^4.0"` or `">=2.0 <5.0"` or, if you have a good reason, remove initialization code of unsupported versions.

Naktibalda: Thanks

sunspikes: Yes, it makes sense include all the supported versions in version constraint 👍

GrahamCampbell:
This behaviour is different to 3.x now?

Naktibalda:
What behaviour?

GrahamCampbell:
in v3, you use all the adapters, but in v4, only the server const adapter.

GrahamCampbell:
Though, actually, you are using all the writers, in v4, but only one of the readers!

Naktibalda:
You are welcome to make a pull request.

PR #5781, opened 2019-12-07
New Feature: Add ability to stash/unstash commands in interactive mode.

pohnean:
It's currently very cumbersome to retrieve successful commands performed in Interactive Mode. You often have to copy and paste commands line by line, and the commands do not have a trailing semi-colon.

This PR improves the interactive shell by adding 4 new keyboard shortcuts to allow the user to quickly stash and unstash the last run command.

1. Pressing F5 stashes the last run command.
2. Pressing F8 displays all the stashed commands with trailing semi-colons, so that it can be quickly pasted back into your IDE.
3. Pressing F10 clears the stashed commands.
4. Pressing F6 toggles 'auto-stashing', which will automatically stash a command if it successfully executes.

The following screenshot illustrates how it works:
image

I have not written any test cases for this because I'm not sure about to go about writing test cases. Would appreciate if maintainers or other contributors help add it in.

### DIFF
--- a/src/Codeception/Lib/Actor/Shared/Pause.php
+++ b/src/Codeception/Lib/Actor/Shared/Pause.php
@@ -15,15 +15,91 @@ trait Pause
             throw new \Exception('Hoa Console is not installed. Please add `hoa/console` to composer.json');
         }
 
+        $autoStash = false;
+        $stashedCommands = [];
+
         $I = $this;
+        $output = new ConsoleOutput();
         $readline = new \Hoa\Console\Readline\Readline();
 
         $readline->setAutocompleter(
             new \Hoa\Console\Readline\Autocompleter\Word(get_class_methods($this))
         );
-        $output = new ConsoleOutput();
+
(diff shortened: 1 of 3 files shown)


src/Codeception/Lib/Actor/Shared/Pause.php
```diff
@@ -43,7 +119,11 @@ public function pause()
                     if (!is_object($result)) {
                         codecept_debug($result);
                     }
-                    codecept_debug('>> Result saved to $result variable, you can use it in next commands');
+                    codecept_debug('>> Result saved to $result variable, you can usZZe it in next commands');
```
Naktibalda: Please remove `ZZ` from `usZZe`.

DavertMik:
```suggestion
                    codecept_debug('>> Result saved to $result variable, you can use it in next commands');
```

pohnean:
Okay, I've studied the code from CodeceptJS and added a similar feature that will save stashed commands automatically to `_output/stashed-commands`
Also removed the `ZZ` from `usZZe`.

DavertMik:
Awesome, thank you very much @pohnean

Issue #5820, opened 2020-01-16
Wrong step in in failures when error occurs. 


ruudboon:
#### What are you trying to achieve?
Display list of errors and failures

#### What do you get instead?
List of errors and failures failures containing wrong step information

Codeception 4.0.2
Related to https://github.com/Codeception/Codeception/pull/4654

ruudboon:
<img width="1074" alt="Screenshot 2020-01-16 at 11 21 56" src="https://user-images.githubusercontent.com/7444246/72516743-7931c500-3852-11ea-84af-bc27abca0b41.png">

Naktibalda:
@ruudboon What's the state of this issue?

ruudboon:
@Naktibalda I think this is still occurring. Not sure when I have time to write tests for this.

PR #4654, opened 2017-11-29
Wrong failed step

eXorus:
Hello,

Context: In one cest file with three tests (1 success and 2 failed)

Actual result: 
```
There were 2 failures:

  ---------
  1) PartialFailedCest: Test case two
   Test  tests/scenario/PartialFailedCest.php:testCaseTwo
   Step  See file found "testcasethree.txt"                  <==== bug here
   Fail  File "testcasetwo.txt" not found at ""

  Scenario Steps:

   2. $I->seeFileFound("testcasetwo.txt") at tests/scenario/PartialFailedCest.php:14
   1. $I->amInPath(".") at tests/scenario/PartialFailedCest.php:13


  ---------
  2) PartialFailedCest: Test case three
   Test  tests/scenario/PartialFailedCest.php:testCaseThree
   Step  See file found "testcasethree.txt"
   Fail  File "testcasethree.txt" not found at ""

  Scenario Steps:

   2. $I->seeFileFound("testcasethree.txt") at tests/scenario/PartialFailedCest.php:20
   1. $I->amInPath(".") at tests/scenario/PartialFailedCest.php:19


  FAILURES!
  Tests: 3, Assertions: 3, Failures: 2.
```

Expected result: 
```
There were 2 failures:

  ---------
  1) PartialFailedCest: Test case two
   Test  tests/scenario/PartialFailedCest.php:testCaseTwo
   Step  See file found "testcasetwo.txt"
   Fail  File "testcasetwo.txt" not found at ""

  Scenario Steps:

   2. $I->seeFileFound("testcasetwo.txt") at tests/scenario/PartialFailedCest.php:14
   1. $I->amInPath(".") at tests/scenario/PartialFailedCest.php:13


  ---------
  2) PartialFailedCest: Test case three
   Test  tests/scenario/PartialFailedCest.php:testCaseThree
   Step  See file found "testcasethree.txt"
   Fail  File "testcasethree.txt" not found at ""

  Scenario Steps:

   2. $I->seeFileFound("testcasethree.txt") at tests/scenario/PartialFailedCest.php:20
   1. $I->amInPath(".") at tests/scenario/PartialFailedCest.php:19


  FAILURES!
  Tests: 3, Assertions: 3, Failures: 2.
```

Fix this previous PR #4607

PR #5821, opened 2020-01-16
Fixed issue #5820

ruudboon:
When an error occurs a failed step is always added. 
This needs to be shifted even if we're not using it.
Issue: #5820


Naktibalda:
Please create a test to demonstrate your issue,
Use https://github.com/Codeception/Codeception/pull/4654 as an example.

PR #5846, opened 2020-02-13
fix recorder filename with special chars

gimler:
When you have special chars in your scenario titles the recorder can't create the directory.

example `Scenario: Simple Login/Logout`

### DIFF
--- a/ext/Recorder.php
+++ b/ext/Recorder.php
@@ -623,7 +623,7 @@ EOF;
      */
     private function getTestName($e)
     {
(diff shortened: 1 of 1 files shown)


gimler:
travis fails are third party dependency problems not related to my change

Naktibalda:
> travis fails are third party dependency problems not related to my change

Codeception 3.1 is no longer maintained, I can only merge this change to 4.0 branch.

Naktibalda:
Released as 4.1.1

Issue #5495, opened 2019-05-07
Feature request: prevent redirect to vendor directory


DanielRuf:
#### What are you trying to achieve?
As a developer / user I want to create a `tools` directory on the same level as `composer.json` to use different versions of Codeception.

#### What do you get instead?

https://github.com/Codeception/Codeception/blob/60702493e37019ca37009b229ed055b30d62b466/autoload.php#L4

```bash
==== Redirecting to Composer-installed version in vendor/codeception ====
```

### What do I expect
A CLI parameter to prevent this redirect.

### Details

* Codeception version: 3.0.0
* PHP Version: 7.2
* Operating System: macOS
* Installation type: Phar
* List of installed packages (`composer show`)
* Suite configuration:

Naktibalda:
How about not installing Codeception to vendor dir if you don't want it loaded?

DanielRuf:
> How about not installing Codeception to vendor dir if you don't want it loaded?

This is not possible in our case as we have a TYPO3 project and it installs `codeception/codeception` for its internal testsuites.

DanielRuf:
And this might be the same case in other projects too and affects others too.

Naktibalda:
This redirect is here for a reason - it prevents odd errors when different versions of Codeception or dependencies are installed in vendor dir and external locations and some classes are loaded from one place and some - from another.

But you are welcome to implement this parameter in your fork and test it in your project.
If it works, please make a pull request.

DanielRuf:
Well the phar is bundled and separated from `vendor` in general so there should be no conflicts in general afaik ;-)

I am aware of the fact that it makes sense for globally composer-installed + locally composer-installed versions (but not for the phar bundles).

DanielRuf:
> But you are welcome to implement this parameter in your fork and test it in your project.
> If it works, please make a pull request.

Will do so.

Naktibalda:
> Well the phar is bundled and separated from `vendor` in general so there should be no conflicts in general afaik ;-)

Conflicts start happenning when Codeception includes `vendor/autoload/autoload.php` which is necessary for functional and unit tests.

Here is a couple of examples: https://github.com/Codeception/Codeception/issues/3999
https://github.com/Codeception/Codeception/issues/3999

DanielRuf:
> Conflicts start happenning when Codeception includes `vendor/autoload/autoload.php` which is necessary for functional and unit tests.

[...]

PR #5857, opened 2020-02-26
Add input definition for --no-redirect in build command

liamjtoohey:
Description
-----------

When hosting composer dependencies outside of the standard vendor directory (similar to https://github.com/Codeception/Codeception/issues/5495) I have been getting the following error when running the build command:

```
dev/dependencies/vendor/bin/codecept build -c dev/tests/codeception

==== Redirecting to Composer-installed version in vendor/codeception. You can skip this using --no-redirect ====
Building Actor classes for suites: acceptance, api
 -> AcceptanceTesterActions.php generated successfully. 110 methods added
\AcceptanceTester includes modules: WebDriver, Db, Percy, \Helper\Acceptance

In ModuleContainer.php line 110:

  Module REST is not installed.
  Use Composer to install corresponding package:

  composer require codeception/module-rest --dev
```

This is of course because I do not have `codeception/module-rest` installed in my `vendor` directory in project root - it is instead installed in `dev/dependencies/vendor`.

The console log does mention to add `--no-redirect` to this command, which will in turn pass the initial vendor redirection step here https://github.com/Codeception/Codeception/blob/4.0/autoload.php#L4. This however fails for the `build` command as there is no input definition, resulting in:

```
dev/dependencies/vendor/bin/codecept build -c dev/tests/codeception --no-redirect


  The "--no-redirect" option does not exist.
```

This PR simply adds a definition for the `--no-redirect` argument in the build command to prevent these errors. With this applied, I can now build and subsequently run tests from my `dev/dependencies` directory:

```
dev/dependencies/vendor/bin/codecept build -c dev/tests/codeception --no-redirect
Building Actor classes for suites: acceptance, api
 -> AcceptanceTesterActions.php generated successfully. 110 methods added
\AcceptanceTester includes modules: WebDriver, Db, Percy, \Helper\Acceptance
 -> ApiTesterActions.php generated successfully. 57 methods added
\ApiTester includes modules: REST, PhpBrowser, \Helper\Api
dev/dependencies/vendor/bin/codecept run api -c dev/tests/codeception/ --no-redirect --env local -v
Codeception PHP Testing Framework v4.1.1
Powered by PHPUnit 8.5.2 by Sebastian Bergmann and contributors.
Running with seed:


Api (local) Tests (14) -----------------------------------------------------------------------------------------------
Modules: REST, PhpBrowser, \Helper\Api
----------------------------------------------------------------------------------------------------------------------
```

### DIFF
--- a/autoload.php
+++ b/autoload.php
@@ -27,6 +27,8 @@ if (( !isset($argv) || (isset($argv) && !in_array('--no-redirect', $argv)) ) && 
     require_once __DIR__ . '/../../autoload.php';
 }
 unset($autoloadFile);
+$argv = array_diff($argv, ['--no-redirect']);
+$_SERVER['argv'] = array_diff($_SERVER['argv'], ['--no-redirect']);
 
 // @codingStandardsIgnoreStart
 
--- a/autoload.php
+++ b/autoload.php
@@ -27,8 +27,12 @@ if (( !isset($argv) || (isset($argv) && !in_array('--no-redirect', $argv)) ) && 
     require_once __DIR__ . '/../../autoload.php';
 }
 unset($autoloadFile);
-$argv = array_diff($argv, ['--no-redirect']);
-$_SERVER['argv'] = array_diff($_SERVER['argv'], ['--no-redirect']);
+if (isset($argv)) {
+    $argv = array_diff($argv, ['--no-redirect']);
+}
+if (isset($_SERVER['argv'])) {
+    $_SERVER['argv'] = array_diff($_SERVER['argv'], ['--no-redirect']);
+}
 
 // @codingStandardsIgnoreStart
 
--- a/src/Codeception/Command/Build.php
+++ b/src/Codeception/Command/Build.php
@@ -28,6 +28,19 @@ class Build extends Command
      */
     protected $output;
 
+    /**
+     * Sets Run arguments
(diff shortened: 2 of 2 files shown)


Naktibalda:
We have 2 options:

1. Add `no-redirect` to definion of all commands (all at once or one at a time).
2. Remove '--no-redirect' from $argv in [autoload.php](https://github.com/Codeception/Codeception/blob/4.0/autoload.php) because it is only used by autoload.php and commands don't need to know anything about it.

I prefer the second option.

liamjtoohey:
@Naktibalda Thanks for the response! I'm not sure I quite understand the approach for the second option however - would this not just remove the `--no-redirect` feature entirely if we remove the $argv code in https://github.com/Codeception/Codeception/blob/4.0/autoload.php? I'd imagine we should still keep `--no-redirect` as a configurable argument - for my use case we very much need this option as we run our tests suites from a separate composer.json from the project root.

With this considered, I'm more than happy to update this PR with option 1 in mind.

Naktibalda:
I mean something like this:
```php
$argv = array_diff($argv, ['--no-redirect']);
```

liamjtoohey:
@Naktibalda Just pushed a commit with the solution above. I did also have to remove this argument from the global server variable as it seemed this was being used when configuring commands.

Naktibalda:
Released in 4.1.2

PR #5862, opened 2020-02-28
Added return type hint to the generated actions above PHP 7.0

pezia:
The idea is that if the module specifies a return type hint, we generate actions that also have this hint. This way there are proper code completion and type checks in IDEs.

E.g.:
```php
<?php declare(strict_types=1);

namespace Foo\Bar;

use Codeception\Module;

class CustomModule extends Module {
    public function haveCustomModuleAction(): \DateTimeInterface {
       return new \DateTimeImmutable();
    }
}
```

```yaml
#foo.suite.yml
class_name: FooTester
modules:
  enabled:
    - \Foo\Bar\CustomModule
```

```php
<?php declare(strict_types=1);

class FooCest {
    public function foo(FooTester $I) {
        $I->haveCustomModuleAction()->format(DATE_ATOM);
    }
}

```

### DIFF
--- a/src/Codeception/Lib/Generator/Actions.php
+++ b/src/Codeception/Lib/Generator/Actions.php
@@ -38,7 +38,7 @@ EOF;
      {{doc}}
      * @see \{{module}}::{{method}}()
      */
-    public function {{action}}({{params}}) {
+    public function {{action}}({{params}}){{return_type}} {
         return \$this->getScenario()->runStep(new \Codeception\Step\{{step}}('{{method}}', func_get_args()));
     }
 EOF;
@@ -113,6 +113,7 @@ EOF;
         $methodTemplate = (new Template($this->methodTemplate))
             ->place('module', $module)
             ->place('method', $refMethod->name)
+            ->place('return_type', $this->createReturnTypeHint($refMethod))
             ->place('params', $params);
 
         if (0 === strpos($refMethod->name, 'see')) {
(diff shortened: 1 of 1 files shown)


pezia:
The idea is that if the module specifies a return type hint, we generate actions that also have this hint. This way there is proper code completion ant type checks present in IDEs also.

src/Codeception/Lib/Generator/Actions.php
```diff
@@ -205,4 +206,26 @@ public function getNumMethods()
...
+
+        if ($returnType === null) {
+            return '';
+        }
+
+        return strtr(
+            ': {nullable}{trailingBackslash}{type}',
+            [
+                '{nullable}' => $returnType->allowsNull() ? '?' : '',
+                '{trailingBackslash}' => $returnType->isBuiltin() ? '' : '\\',
```
Naktibalda: This backslash is leading, not trailing.

Also `strtr` syntax looks like overkill, simple string concatenation or sprinf(': %s%s%s', expressions) would be much more concise.

pezia: True :)

Naktibalda:
Released in 4.1.2

src/Codeception/Lib/Generator/Actions.php
```diff
@@ -38,7 +38,7 @@ trait {{name}}Actions
      {{doc}}
      * @see \{{module}}::{{method}}()
      */
-    public function {{action}}({{params}}) {
+    public function {{action}}({{params}}){{return_type}} {
```
stereomon: This leads to a `Fatal error: A void function must not return a value` when your return type is void.

Jamesking56: See my new PR that fixes this https://github.com/Codeception/Codeception/pull/5878

src/Codeception/Lib/Generator/Actions.php
```diff
@@ -205,4 +206,24 @@ public function getNumMethods()
...
+
+        if ($returnType === null) {
+            return '';
+        }
+
+        return sprintf(
+            ': %s%s%s',
+            $returnType->allowsNull() ? '?' : '',
+            $returnType->isBuiltin() ? '' : '\\',
+            (string)$returnType
```
Slamdunk: On PHP 7.4:
```
Deprecated: Function ReflectionType::__toString() is deprecated in ./vendor/codeception/codeception/src/Codeception/Lib/Generator/Actions.php on line 226
```
See https://www.php.net/manual/en/reflectiontype.tostring.php

Slamdunk: Already reported in https://github.com/Codeception/Codeception/issues/5875

Issue #5768, opened 2019-11-25
Inconsistent code coverage in non-executed classes


CJDennis:
#### What are you trying to achieve?
Consistent code coverage. 

#### What do you get instead?
Inconsistent code coverage. Lines 10, 11, & 13 are correctly marked as not executable in `CJDennis\ElseTest\IfTest`, while they are incorrectly marked as executable and not covered in `CJDennis\ElseTest\ElseTest`. Running the same coverage report under PHPUnit gives consistent results (neither file has those lines marked as executable).

> Provide console output if related. Use `-vvv` mode for more details.

```bash
codecept run unit --coverage-xml coverage.xml
```

### tests/_output/coverage.xml
```xml
<?xml version="1.0" encoding="UTF-8"?>
<coverage generated="1574649066">
  <project timestamp="1574649066">
    <package name="CJDennis\ElseTest">
      <file name="D:\server\Apache24\htdocs\libraries\cjdennis\else-test\src\CJDennis\ElseTest\ElseTest.php">
        <class name="CJDennis\ElseTest\ElseTest" namespace="CJDennis\ElseTest">
          <metrics complexity="2" methods="1" coveredmethods="0" conditionals="0" coveredconditionals="0" statements="7" coveredstatements="0" elements="8" coveredelements="0"/>
        </class>
        <line num="7" type="method" name="__construct" visibility="public" complexity="2" crap="6" count="0"/>
        <line num="8" type="stmt" count="0"/>
        <line num="9" type="stmt" count="0"/>
        <line num="10" type="stmt" count="0"/>
        <line num="11" type="stmt" count="0"/>
        <line num="12" type="stmt" count="0"/>
        <line num="13" type="stmt" count="0"/>
        <line num="14" type="stmt" count="0"/>
        <metrics loc="15" ncloc="15" classes="1" methods="1" coveredmethods="0" conditionals="0" coveredconditionals="0" statements="7" coveredstatements="0" elements="8" coveredelements="0"/>
      </file>
      <file name="D:\server\Apache24\htdocs\libraries\cjdennis\else-test\src\CJDennis\ElseTest\IfTest.php">
        <class name="CJDennis\ElseTest\IfTest" namespace="CJDennis\ElseTest">
          <metrics complexity="2" methods="1" coveredmethods="0" conditionals="0" coveredconditionals="0" statements="4" coveredstatements="3" elements="5" coveredelements="3"/>
        </class>
[...]

PR #5863, opened 2020-03-02
Bugfix/show uncovered

dekker-m:
Related to issue https://github.com/Codeception/Codeception/issues/5768

* Added test cases to test the behavior of `show_uncovered`
* Fixed the newly broken test cases by setting `addUncoveredFilesFromWhitelist` to `false` on CodeCoverage instances

Other minor fixes:
* Set /codecept binary to line feed line endings. On windows the docker container does not run, because git checks out the files with crlf line endings.
* Minor changes to the docker files to get it up and running


tests/data/claypit/tests/math2/MathCest.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+use \Math\CalcHelper as Calc;
+
+class MathCest
```
Naktibalda: Builds are failing with

> Fatal error: Cannot redeclare class MathCest in /home/travis/build/Codeception/Codeception/tests/data/claypit/tests/math2/MathCest.php on line 6

Rename it to Math2Cest.

dekker-m: I just updated the branch. Is there any change this change is being merged into the 3.x branch as well? I think a lot of us still use version 3. Without this fix it would be impossible to switch to pcov and not lose code coverage metrics.

Naktibalda:
> Is there any change this is being merged into the 3.x branch as well? I think a lot of us still use version 3. Without this fix it would be impossible to switch to pcov and not lose code coverage metrics.

No chance. What's stopping you from upgrading to Codeception 4?

dekker-m:
> > Is there any change this is being merged into the 3.x branch as well? I think a lot of us still use version 3. Without this fix it would be impossible to switch to pcov and not lose code coverage metrics.
> 
> No chance. What's stopping you from upgrading to Codeception 4?

We have a number of custom modules in every project. One of those modules extends from the Framework class under Codeception 3. As soon as we upgrade to version 4 the class did not exist anymore. I couldn't find any migrations guide or changelog that explained what had changed, so instead of upgrading we kept running version 3.

Naktibalda:
You should have asked, it is in `codeception/lib-innerbrowser` package now.

https://github.com/Codeception/lib-innerbrowser/blob/master/src/Codeception/Lib/Framework.php

dekker-m:
I see. I wasn't aware Codeception 4 had split the modules in their own packages. It would have been nice this was mentioned on the release page of Codeception 4. I just read there is an upgrade script, but it still would have been nice if the release page contained notes about what this upgrade script does (probably add the required packages to composer).

Anyway, do you have an idea when this bugfix will be released into a new release?

src/Codeception/Coverage/SuiteSubscriber.php
```diff
@@ -62,6 +62,7 @@ protected function applySettings($settings)
             }
         }
         $this->coverage->setProcessUncoveredFilesFromWhitelist($this->settings['show_uncovered']);
+        $this->coverage->setAddUncoveredFilesFromWhitelist(false);
```
Naktibalda: It looks like this line doesn't matter or it isn't covered by test.
The change in Subscriber/Printer class is enough to make your tests pass.

.gitattributes
```diff
@@ -3,17 +3,18 @@
...
+*.js        eol=lf
+*.sql       eol=lf
+*.php       eol=lf diff=php
+*.yml       eol=lf
+*.xml       eol=lf
+*.dist      eol=lf
+*.conf      eol=lf
+*.json      eol=lf
+*.lock      eol=lf
+/codecept   eol=lf
```
Naktibalda: Please revert changes made to .gitattributes and Dockerfile.
If you want to improve them, please raise a new pull request.

tests/coverage/LocalShowUncoveredCept.php
```diff
@@ -0,0 +1,14 @@
...
+$I->executeCommand('run math2_show_uncovered Math2ShowUncoveredCest:testAddition --coverage', null, '-d pcov.directory=' . getcwd());
+$I->seeInShellOutput('OK (1 test');
+$I->seeInShellOutput('Classes: 50.00%');
+$I->seeInShellOutput('Methods: 50.00%');
+
+$I->wantTo('run local code coverage for cest and test without show uncovered');
+$I->executeCommand('run math2 Math2Cest:testAddition --coverage', null, '-d pcov.directory=' . getcwd());
+$I->seeInShellOutput('OK (1 test');
+$I->seeInShellOutput('Classes: 100.00%');
+$I->seeInShellOutput('Methods: 100.00%');
```
Naktibalda: Does anyone actually want coverage results without show_uncovered?
It looks like cheating.

pvsaintpe:
@krakjoe 

I have a similar problem when running tests in parallel and merge code coverage via php-cov-coverage, ex.:

Screenshot 2020-03-10 at 17 10 04
Screenshot 2020-03-10 at 17 10 14
Screenshot 2020-03-10 at 17 10 18
Screenshot 2020-03-10 at 17 10 43

I use the following versions of packages:
- Laravel 5.8.36
- phpunit 8.5.2
- php-code-coverage 7.0.10
- codeception 3.1.2

Softwares:
- PHP 7.3.15
- PCOV 1.0.6 (Without XDebug, because I removed it manually).

pvsaintpe:
@sebastianbergmann, what can you say about this? Have you encountered a similar problem?

I want to understand if this is a [php-code-coverage|https://github.com/sebastianbergmann/php-code-coverage] or [PCOV|https://github.com/krakjoe/pcov] problem?

pvsaintpe:
In theory, the problem is on the side of the one who generate report the code coverage, since if the code inside the method is covered (that is, we entered to the method and successfully exited it), then the brackets should be green and be included in the "Lines" section.

Now, the brackets aren't counted as lines of code, and as a result, if our method throws an Exception, the opening bracket included in report, and the closing bracket is not, and this is a normal situation, so the bracket must be considered as executable code, anyhow.

dekker-m:
The documentation quoted by @Naktibalda  explains perfectly what the expected behavior should be depending on the values of addUncoveredFilesFromWhitelist and processUnconveredFilesFromWhitelist. The pcov extension does not generate the red lines from @pvsaintpe his example, the php-code-coverage library adds those lines because the method `addUncoveredFilesFromWhitelist` is executed. However, this is the default setting and the documented behavior. So in short, the php-code-coverage and pcov library both seem to return the expected results.

@Naktibalda In your latest comment you suggest `addUncoveredFilesFromWhitelist` should be set to `true`. Could you explain why? Because I think disabling it solve all problems:

`show_uncovered: false` will not show any code coverage if processUncoveredFilesFromWhitelist and addUncoveredFilesFromWhitelist are both set to false.
`show_uncovered: true` will show all uncovered files, because processUncoveredFilesFromWhitelist is being set to true.

Could you explain why you think `addUncoveredFilesFromWhitelist` should be set to true?

There still is another case which is indeed not covered by the test cases. If I remember correctly it happens when codeception runs multiple test suites in a single process. Within a single process the coverage can only be collected once for every file. When the second suite starts collecting code coverage `processUncoveredFilesFromWhitelist` does not work, now  `addUncoveredFilesFromWhitelist` will start adding non executable lines all over the place. During a merge of the code coverage results from both suites the non executable lines show up in the reports. (Correctly me if I'm wrong, but codeception uses a separate `new CodeCoverage()` object for every suite, right?)

I'll try to update the pull request with an additional test case that show this behavior.

Naktibalda:
> Could you explain why you think addUncoveredFilesFromWhitelist should be set to true?

My understanding is that if you have 100 files of code and only one of them is tested, then with `addUncoveredFilesFromWhitelist = true`  code coverage result will be `1/100`
and with `addUncoveredFilesFromWhitelist = false`  code coverage result will be `1/1`, is that right?

> Correctly me if I'm wrong, but codeception uses a separate new CodeCoverage() object for every suite, right?

It's a good question, it looks like the same instance is used in all suites, but there is something odd happening in SuiteSubscriber::applyFilter method which temporarily assigned a new instance with mocked CodeCoverage Driver before reassigning the original instance:
https://github.com/Codeception/Codeception/blob/4.0/src/Codeception/Coverage/SuiteSubscriber.php#L81-L91

I will respond to other points later.

dekker-m:
> My understanding is that if you have 100 files of code and only one of them is tested, then with `addUncoveredFilesFromWhitelist = true` code coverage result will be `1/100`
> and with `addUncoveredFilesFromWhitelist = false` code coverage result will be `1/1`, is that right?

That's right, but `processUncoveredFilesFromWhitelist = true` will achieve the same effect if `addUncoveredFilesFromWhitelist = false`. `processUncoveredFilesFromWhitelist = true` even does a better job, because xdebug/pcov will determine the lines which are actually executable.

Using `addUncoveredFilesFromWhitelist = true` and `processUncoveredFilesFromWhitelist = true` makes no sense. Both methods try to achieve the same goals, but with different means and slightly different end results. In some weird cases when both options are true, the end result will contain the non executed lines, which are not executable at all.

I'll try to add some additional test cases later. I think there is some weird behavior when multiple suites are executed in the same process.

Naktibalda:
Now we have competing pull request #5874 which adds `add_uncovered` setting.
I have no objections to that - adding parameter is definitely better than reversing behaviour.

If it solves your problem, I will merge #5874 and I will merge your tests from this pull request too.

dekker-m:
I see @tifeit did some extensive research on the issue as well. For me, both pull requests would solve the issue and I think the issues of others should be resolved for other as well (if they use the right configuration).

Which pull request you choose to merge is actually up to you. I think arguments can be made in favor of both pull requests, it really depends on the philosophy of the Codeception team.

When I created this pull request I assumed `show_uncovered` would control whether uncovered files should show up in code coverage results. If people know about the two underlying parameters for php-code-coverage should not matter, the end result should reflect what the name of the configuration parameter suggests:
* Uncovered files should show up in the coverage results
* Uncovered files should not show up in the coverage results

The pull request I created makes sure the configuration parameter actually produces the results as described above.

The pull request @tifeit created allows users to configure the setting of `addUncoveredFilesFromWhitelist`. It allows users to tweak the underlying parameters of the PHP Code Coverage library. @tifeit Could you explain why you made the decision to add this parameter? 
I could not think about a reason why this parameter should be configurable, because `processUncoveredFilesFromWhitelist` and `addUncoveredFilesFromWhitelist` seem to have the same intention: let uncovered files show up in the results. I'm really interested if you came up with a reason why those parameters should be configurable.

I think it is up to the Codeception team to decide which solution has their preference.

PS: @sebastianbergmann Could you explain why the php-code-coverage library actually has two parameter, which seem to have kind of the same behavior? Over the past couple of days I've seen multiple issues created in different testing framework libraries, all related to both settings. From what I can see v1 of php-code-coverage already had the `addUncoveredFilesFromWhitelist` option (but it was named processUncoveredFilesFromWhitelist), Version 2 already seemed to have both option, while `processUncoveredFilesFromWhitelist` was renamed to `addUncoveredFilesFromWhitelist` and a new `processUncoveredFilesFromWhitelist` was added. After all the discussions I'm really interested in the views of the author of the php-code-coverage library.

SamMousa:
I think this is superseded by #5922 which proposes a way to expose all configuration of php-code-coverage.

dekker-m:
I see the other pull request is closed. After the changes mentioned, can this one still be merged? If so, I will make the required changes.

Naktibalda:
Another pull request was closed because @SamMousa couldn't be bothered to support multiple versions of PHPUnit.
`setAddUncoveredFilesFromWhitelist` method no longer exists in PhpCodeCoverage 9, so it is necessary to check if the method exists and call appropriated methods in order to support PhpCodeCoverage 9 and older versions.
Like here https://github.com/Codeception/Codeception/pull/5956/files#diff-5f152311bc517f54f5310a8eab767c52R65-R75

I would prefer if you took over #5922 and made it compatible with both interfaces.

Naktibalda:
I expect #5922 to be updated and merged to Codeception 5.0

PR #5867, opened 2020-03-05
Add Mezzio module

Slamdunk:
See https://github.com/Codeception/module-mezzio/pull/1


Naktibalda:
I'm not sure if this change is necessary, the main reason for `$packages` array to exist is migration from Codeception v3 to v4. 
Nobody is using Mezzio module with Codeception 3.
As a matter of precedent, Phalcon4 module is not there.

Issue #5868, opened 2020-03-05
Error on instalation


gabriel-lima96:
#### What are you trying to achieve?
Install Codeception in the root of the project and having the composer vendor folder in another path. 
#### What do you get instead?
Instaled via composer, then *path-to-vendor*/bin/codecept bootstrap, and it gives an error:
```bash
Generating autoload files
PHP Warning:  require(vendor/autoload.php): failed to open stream: No such file or directory in /home/gabriel/projects/portal-de-denuncias/source/shared/lib/vendor/codeception/codeception/src/Codeception/InitTemplate.php on line 334
PHP Fatal error:  require(): Failed opening required 'vendor/autoload.php' (include_path='.:/usr/share/php') in /home/gabriel/projects/portal-de-denuncias/source/shared/lib/vendor/codeception/codeception/src/Codeception/InitTemplate.php on line 334
```
I'm pretty sure it is because i put my composer.json in the root folder with
```json
{
    "config": {
        "vendor-dir": "source/shared/lib/vendor"
    },
   ...
}
```
And in the InitTemplate.php, line 334 it try a require as if the composer was used with the default vendor path folder.
### Details

* Codeception version:^4.1
* PHP Version:7.2
* Operating System:Ubuntu 19.10
* Installation type: Composer
* List of installed packages: only codeception and his dependencies

Naktibalda:
Stop doing that, install Codeception in the same place as other dependencies.

gabriel-lima96:
It is. But i want all my configs (.yml, .conf etc) and the tests folder on the root project.

gabriel-lima96:
If composer have an option to keep my vendor folder in another path, and Codeception do stuff using composer, it should take care of this option as well. 
It can be easily "fixed", since Codeception already parse composer.json. It just have to check if `$composer->config->vendor_dir` is set and send it as a parameter to updateComposerClassMap so it can use to find the path to the vendor folder.

Naktibalda:
Please make a pull request.

Naktibalda:
Sorry, I misunderstood this issue as a yet another variation of two vendor directories.

Naktibalda:
Fixed by #5871

PR #5871, opened 2020-03-06
Check the composer option config.vendor_dir when updating composer

gabriel-lima96:
Details on issue #5868.

Check if the composer.json have a [vendor_dir](https://getcomposer.org/doc/06-config.md#vendor-dir) and then use this path on the method updateComposerClassMap of the class InitTemplate, used on the `codecept bootstrap`.


Naktibalda:
Released in 4.1.2

Issue #5817, opened 2020-01-15
CodeCoverage slows down between test suites


tweber68:
#### Acceleration of CodeCoverage

I suggest to adapt Configuration.php function mergeConfigs line 674:

currently:
```
return array_merge_recursive($a2, $a1);
```

to:
```
return array_merge(array_unique(array_merge_recursive($a2, $a1), SORT_REGULAR));
```

#### What is the problem?

I had the problem that lots of calls of `mergeConfigs()` made the internally handled `$this->config` variable explode. 
```php
$a = ['a','b','c'];
$b = ['a','b','c'];

var_export(array_merge_recursive($a,$b));
```

output:
```
array (
  0 => 'a',
  1 => 'b',
  2 => 'c',
  3 => 'a',
  4 => 'b',
  5 => 'c',
)
```

I think this is not intended. In my project `$this->config` of Codeception/Lib/ModuleContainer should contain 5 entries, but it grew up to 100000s of entries by this logic.

### Details

* Codeception version: 4.0.2
* PHP Version: 7.2
* Operating System: Ubuntu 16.04
* Installation type: Composer
* List of installed packages (`composer show`):
```
behat/gherkin                      v4.6.0             Gherkin DSL parser for PHP 5.3
codeception/codeception            4.0.2              BDD-style testing framework
codeception/lib-asserts            1.10.1             Assertion methods used by Codeception core and Asserts module
codeception/module-asserts         1.1.1              Codeception module containing various assertions
codeception/module-cli             1.0.0              Codeception module for testing basic shell commands and shell output
codeception/module-filesystem      1.0.2              Codeception module for testing local filesystem
codeception/module-mongodb         1.0.0              MongoDB module for Codeception
codeception/phpunit-wrapper        8.1.1              PHPUnit classes used by Codeception
codeception/stub                   3.6.0              Flexible Stub wrapper for PHPUnit's Mock Builder
composer/ca-bundle                 1.2.6              Lets you find a path to the system CA bundle, and includes a fallback to the Mozilla CA bundle.
composer/xdebug-handler            1.4.0              Restarts a process without Xdebug.
dnode/dnode                        v0.1.3             DNode RPC protocol for PHP 5.3
doctrine/instantiator              1.3.0              A small, lightweight utility to instantiate objects in PHP without invoking their constructors
domnikl/statsd                     2.10.0             a PHP client for statsd
evenement/evenement                v1.0.0             Événement is a very simple event dispatching library for PHP 5.3
[...]

PR #5873, opened 2020-03-10
Prevent merged config array ballooning in memory

AndrewFeeney:
Closes https://github.com/Codeception/Codeception/issues/5817

This change prevents an issue where the config array gets duplicated every time the mergeConfigs array is called causing the array to balloon in size in memory and the test suite to slow down as it progresses.

Original issue raised by and fix suggested by @tweber68

### DIFF
--- a/src/Codeception/Configuration.php
+++ b/src/Codeception/Configuration.php
@@ -670,7 +670,7 @@ class Configuration
 
         // for sequential arrays
         if (isset($a1[0], $a2[0])) {
-            return array_merge_recursive($a2, $a1);
+            return array_merge(array_unique(array_merge_recursive($a2, $a1), SORT_REGULAR));
         }
 
         // for associative arrays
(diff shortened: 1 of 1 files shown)


src/Codeception/Configuration.php
```diff
@@ -670,7 +670,7 @@ public static function mergeConfigs($a1, $a2)
 
         // for sequential arrays
         if (isset($a1[0], $a2[0])) {
-            return array_merge_recursive($a2, $a1);
+            return array_merge(array_unique(array_merge_recursive($a2, $a1), SORT_REGULAR));
```
Naktibalda: Why are you using `array_merge` here?
I've never seen it used with a single argument.

tweber68:
@n8whnp I use it in order to make the indexes correct. For example:
```
0 => 'first',
3 => 'second',
4 => 'third',
```
array_merge will transform it to:
```
0 => 'first',
1 => 'second',
2 => 'third',
```

Naktibalda:
array_values would be more to the point.

tweber68:
You're right

AndrewFeeney:
@Naktibalda @tweber68 I made the suggested change to `array_values()` instead of `array_merge()`

PR #5878, opened 2020-03-16
Ensure that the return keyword is not used when method returns void type

### LINKED ISSUE #5877: A void function must not return a value error since 4.1.2
Caused by #5862 

Return types on generated actions for methods that return `void` should not use `return` since you cannot return a `void`:

```php
public function voidMethod(): void {
    $this->getScenario()->runStep(new \Codeception\Step\Action('voidMethod', func_get_args()));
}
```

#### What do you get instead?

```php
public function voidMethod(): void {
    return $this->getScenario()->runStep(new \Codeception\Step\Action('voidMethod', func_get_args()));
}
```

This causes PHP errors:

```
Fatal error:  A void function must not return a value
```
### Details

* Codeception version: 4.1.2
* PHP Version: 7.3.15
* Operating System: Ubuntu 18.04.3
* Installation type: Composer

Jamesking56:
Fixes #5877 

When Codeception generates a method action for a method which returns a `void` type, the method still uses the `return` keyword causing a PHP error:

```
Fatal error:  A void function must not return a value
```

This PR makes the `return` keyword variable, so that when we're dealing with a method which has a return type of `void`, we don't use the `return` keyword at all.

### DIFF
--- a/src/Codeception/Lib/Generator/Actions.php
+++ b/src/Codeception/Lib/Generator/Actions.php
@@ -39,7 +39,7 @@ EOF;
      * @see \{{module}}::{{method}}()
      */
     public function {{action}}({{params}}){{return_type}} {
-        return \$this->getScenario()->runStep(new \Codeception\Step\{{step}}('{{method}}', func_get_args()));
+        {{return}}\$this->getScenario()->runStep(new \Codeception\Step\{{step}}('{{method}}', func_get_args()));
     }
 EOF;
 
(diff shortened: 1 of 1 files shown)


src/Codeception/Lib/Generator/Actions.php
```diff
@@ -113,7 +113,8 @@ protected function addMethod(\ReflectionMethod $refMethod)
         $methodTemplate = (new Template($this->methodTemplate))
             ->place('module', $module)
             ->place('method', $refMethod->name)
-            ->place('return_type', $this->createReturnTypeHint($refMethod))
+            ->place('return_type', $returnType = $this->createReturnTypeHint($refMethod))
```
Naktibalda: The patch looks good, however I really don't like this inline assignment, 
please extract this assignment to separate statement above `$methodTemplate = (new Template($this->methodTemplate))` line.

Jamesking56: No problem will do :+1:

Jamesking56: Updated 2b8d8aa

Naktibalda:
This PR received a lot of attention.

Please review my PR too: #5876

Naktibalda:
Released as 4.1.3, new phar file will be published soon.

PR #5880, opened 2020-03-18
Fix bug with void type not being picked up correctly

### LINKED ISSUE #5877: A void function must not return a value error since 4.1.2

Jamesking56:
Fixes #5877 

My previous PR (#5878) had a slight problem in that I assumed that `$returnType` just had the return type on its own in it (e.g. `void`). However it does not, it also includes the `: ` before it.


Jamesking56:
Tested this with my project (editing vendor directly) and it works!

Jamesking56:
@Naktibalda tests added and passing

Naktibalda:
Thank you

Naktibalda:
Released as 4.1.4

PR #5882, opened 2020-03-19
Fixed condition for void return type on generated action.

KacerCZ:
Fixes condition used in #5878 - colon is part of return type string.


KacerCZ:
Duplicate of #5880.

Issue #3441, opened 2016-08-14
Multiple environment configurations not working


maxfysho:
#### What are you trying to achieve?

I am trying to run a collection of tests using two environment configurations as outlined here (http://codeception.com/docs/07-AdvancedUsage)

Running the following command "codecept run acceptance --env dev,mobile -g screenshot -vvv"
#### What do you get instead?

Only the last environment configuration in list was used

> Provide console output if related. Use `-vvv` mode for more details.
#### Environment file dev.yml

``` yml
# `dev` environment config goes here
modules:
    enabled:
         - WebDriver:
             url: https://test.fysho.com
```
#### Environment file mobile.yml

``` yml
# `mobile` environment config goes here
modules:
    enabled:
         - WebDriver:
             window_size: 375x667
```

``` bash
odeception PHP Testing Framework v2.2.3
Powered by PHPUnit 5.4.7 by Sebastian Bergmann and contributors.
[Groups] screenshot 

Acceptance (dev, mobile) Tests (3) --------------------------------------------------------------------
Modules: WebDriver, \Helper\Acceptance
-------------------------------------------------------------------------------------------------------
landing: Navigate to Default Landing Page
Signature: landing:Navigate to Default Landing Page
Test: tests/acceptance/landing.feature:Navigate to Default Landing Page
Scenario --
 As a visitor
 I want to be able to findout a little about fysho
 So that I can decide whether or not to sign up
 Given i am a vistor 
   I am on page "/user/logout"
  [GET] http://local.fysho.com/user/logout
 When i navigate to fysho 
   I am on page "/"
  [GET] http://local.fysho.com/
 Then i see the landing page 
   I see in title "Fysho - Find Fishing Reports"
   I see "FIND FISHING REPORTS IN","h1"
   I see "SIGN UP",".nav"
   I see "LOGIN",".nav"
 And i take a screenshot called "landing" 
   I make screenshot "landing"
  Screenshot saved to /Users/Max/Documents/Development/PHP/Fysho/app/tests/_output/debug/landing.png
 PASSED 

login: Can navigate to login directly
Signature: login:Can navigate to login directly
Test: tests/acceptance/login.feature:Can navigate to login directly
Scenario --
 As a visitor
 I want to login to my account
 So that I can begin to use Fysho
 Given i am a vistor 
   I am on page "/user/logout"
  [GET] http://local.fysho.com/user/logout
 When i navigate to "login" page 
   I am on page "/login"
[...]

PR #5948, opened 2020-08-04
Remove complicated merge logic for environment configurations.

Sasti:
* The logic was flawed because it prevented environments to modify the config subsequently if they where not complete.
* Only the last configured environment won and controlled the resulting configuration set. This should be fixed now.

### DIFF
--- a/src/Codeception/Codecept.php
+++ b/src/Codeception/Codecept.php
@@ -153,40 +153,47 @@ class Codecept
         );
 
         $config = $config ?: Configuration::config();
-
-        $settings = Configuration::suiteSettings($suite, $config);
+        $config = Configuration::suiteSettings($suite, $config);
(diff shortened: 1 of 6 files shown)


Sasti:
This pull request fixes issue #3441. Changes include additional tests in RunEnvironmentCest which verify that the change actually implements the desired effect. 
The change itself removed some unnecessary code and instead used direct access to environments in the already loaded configurations.

DavertMik:
Thank you! The logic is so complicated that we still need to some time to check how this improves current behavior

DavertMik:
Let's have it! :tada:

simonberger:
The removal of `suiteEnvironments` broke our test package. Not the best to remove a function of a common classes in a patch release.

PR #5966, opened 2020-09-01
Support absolute output dir path on Windows

Naktibalda:
Use `codecept_is_path_absolute` function, which supports windows `C:\` style paths instead of using unix-specific condition.
Fixes https://github.com/infection/infection/issues/1296

I haven't had time to test it yet.


sidz:
Will it be possible to backport this fix to the 3.x?

Naktibalda:
@sidz no, it won't be backported, upgrade to Codeception 4.

PR #5975, opened 2020-09-15
Add compatibility with vlucas/phpdotenv v5

johanzandstra:


johanzandstra:
@Naktibalda can we have merged this into a new release? We need this as Laravel 8 is using Dotenv 5

src/Codeception/Lib/ParamsLoader.php
```diff
@@ -78,53 +80,29 @@ protected function loadYamlFile()
...
-                }
-            }
-
-            return $a;
-        };
+        return parse_ini_file($this->paramsFile);
+    }
 
-        return $paramsToArray(simplexml_load_file($this->paramsFile));
+    protected function loadPhpFile()
```
Naktibalda, line 88: Was it really necessary to reorder methods?
It would be much easier to merge this PR if the only affected method was `loadDotEnvFile`.

johanzandstra: Sorry for that, that was my auto formatter I will update the PR

johanzandstra: @Naktibalda Please let me know if you want me to change more!

PR #5978, opened 2020-09-18
Fix issue when using includes in combination with override config

bramstroker:
I am using include configs in my project to run multiple applications/modules in a single test run. https://codeception.com/docs/08-Customization#One-Runner-for-Multiple-Applications

In my root `codeception.yml`:
```yml
include:
    - src/OtherModule/*
paths:
    tests: tests
...
```

In my `src/OtherModule` I have another `codeception.yml` and seperate tests.
This is working perfectly and tests of the other module are also executed when I run `vendor/bin codecept run`.

However when using these feature the overriden configs (using `-o` parameter) are removed from the test run.
This was caused by line 365-368 which reset the complete config object to that of the main `codeception.yml`, but don't apply the `override` config again.
PHPStorm also uses this method to set a custom reporter `-o "reporters: report: PhpStorm_Codeception_ReportPrinter"`, but because this was removed the complete UI reporting in PHPStorm was broken for me.

I also refactored to ONLY override the config when the test should be run by an include config. This seems less fragile to me.

### DIFF
--- a/src/Codeception/Command/Run.php
+++ b/src/Codeception/Command/Run.php
@@ -337,19 +337,22 @@ class Run extends Command
                     // Find if the suite begins with an include path
                     if (strpos($suite, $include) === 0) {
                         // Use include config
-                        $config = Configuration::config($projectDir.$include);
+                        $includeConfig = Configuration::config($projectDir.$include);
 
-                        if (!isset($config['paths']['tests'])) {
+                        if (!isset($includeConfig['paths']['tests'])) {
                             throw new \RuntimeException(
                                 sprintf("Included '%s' has no tests path configured", $include)
                             );
                         }
 
-                        $testsPath = $include . DIRECTORY_SEPARATOR.  $config['paths']['tests'];
+                        $testsPath = $include . DIRECTORY_SEPARATOR.  $includeConfig['paths']['tests'];
 
                         try {
                             list(, $suite, $test) = $this->matchTestFromFilename($suite, $testsPath);
-                            $isIncludeTest = true;
+                            $config = $includeConfig;
+                            if (count($this->options['override'])) {
+                                $config = $this->overrideConfig($this->options['override']);
+                            }
                         } catch (\InvalidArgumentException $e) {
                             // Incorrect include match, continue trying to find one
                             continue;
@@ -360,11 +363,6 @@ class Run extends Command
                             list(, $suite, $test) = $result;
                         }
                     }
-                }
-
-                // Restore main config
-                if (!$isIncludeTest) {
-                    $config = Configuration::config($projectDir);
                 }
             } elseif (!empty($suite)) {
                 $result = $this->matchSingleTest($suite, $config);


bramstroker:
Any update on this PR?

DavertMik (approved):

DavertMik:
Looks very good to me. If no objections we can merge this and release a new version

DavertMik:
@bramstroker is it possible to set up a test for it? 
probably that will require to create a new test preset in https://github.com/Codeception/Codeception/tree/4.1/tests/data
like this one https://github.com/Codeception/Codeception/tree/4.1/tests/data/included

and create a test in "cli" suite to run tests using `-o` parameter to prove that parameters substitute correctly

bramstroker:
@DavertMik thanks. yes I can have a look into this somewhere next week.

Naktibalda:
@bramstroker I am waiting for test.

bramstroker:
I'm having a look into it now.

bramstroker:
I want to provide a custom phpunit reporter within the test suite so I can use this reporter in my test.
This way I can check if the reporter can be overridden using the `-o` parameter and can check in the shell output if the custom reporter was executed.
However I am not sure what the right location is for this custom reporter so PSR-4 autoloading would work correctly.
I also see a custom report printer in `tests/data/claypit/tests/_data/MyReportPrinter.php`, and this claypit location is in composer autoload definitions. Not sure what claypit is however and if I need to put my report printer in the same location.
Could you provide some insights?

Naktibalda:
tests/data/claypit directory is copied to tests/data/sandbox before running each test in Cli test suite and tests are executed against sandbox directory because many tests generate code.
https://github.com/Codeception/Codeception/blob/4.1/tests/support/CliHelper.php#L17

Yes, you can put your custom printer to tests/data/claypit/tests/_data directory.

calvinalkan:
Is anybody still working on this?

What's missing to merge this?

This is a major pain point in a multi-application setup.

calvinalkan:
I have found that no matter what settings you put into the "root apps" configuration file, it has zero effects on included apps when you run it as a multi-app setup.

Neither options passed as command-line arguments nor options in the root codeception.yml file.

This causes the described PHPStrom UI bug here. But it's not exclusive to reporters. 

I find this a bit counterintuitive:

I would expect it to work like this:

- Anything that you pass via command-line or in the root config has priority over the hardcoded default config in `Configuration.php`.
- The config of included apps has priority over the root config.

But how it currently works is:

- Root config and command-line arguments are ignored 
- included app config is merged with hardcoded default config.

DavertMik:
@calvinalkan 

> What's missing to merge this?

A test. Could you add one?

> I would expect it to work like this:

> Anything that you pass via command-line or in the root config has priority over the hardcoded default config in Configuration.php.
> The config of included apps has priority over the root config.

I agree, please improve that behavior and we will merge it

calvinalkan:
> @calvinalkan
> 
> > What's missing to merge this?
> 
> A test. Could you add one?
> 
> > I would expect it to work like this:
> 
> > Anything that you pass via command-line or in the root config has priority over the hardcoded default config in Configuration.php.
> > The config of included apps has priority over the root config.
> 
> I agree, please improve that behavior and we will merge it

I can try, the static `Configuration` class is a bit tricky to navigate.

calvinalkan:
@DavertMik I fixed this including a test in #6452.

Naktibalda:
Replaced by #6452 , released as 4.2.0

PR #5980, opened 2020-09-22
RunFailed: fix path to file in the docblock

Naktibalda:
Correct path was used at the bottom of docblock, but this line at the top is misleading.

This docblock is used to generate https://codeception.com/extensions#RunFailed page.


DavertMik (approved):

DavertMik:
Thanks!

PR #5982, opened 2020-09-22
More practical ext logger

varp:
- `.log` extension is added to suite log files
-  create logger with a full test name in the `beforeTest`

### DIFF
--- a/ext/Logger.php
+++ b/ext/Logger.php
@@ -86,13 +86,13 @@ class Logger extends Extension
 
     public function beforeSuite(SuiteEvent $e)
     {
-        $suite = str_replace('\\', '_', $e->getSuite()->getName());
+        $suite = str_replace('\\', '_', $e->getSuite()->getName()) . '.log';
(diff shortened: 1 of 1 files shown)


DavertMik (approved):

ext/Logger.php
```diff
@@ -86,13 +86,13 @@ public static function getLogger()
 
     public function beforeSuite(SuiteEvent $e)
     {
-        $suite = str_replace('\\', '_', $e->getSuite()->getName());
+        $suite = str_replace('\\', '_', $e->getSuite()->getName()) . '.log';
```
Naktibalda, line 89: Now the name of this variable is really wrong.
Could you rename it to something like $suiteLogFile?

There is no information about suite log files in documentation of this extension.
Could you decribe them in the docblock above this class?

varp: @Naktibalda  I've made what you asked for. Please, comment if this ok or not =)

PR #5988, opened 2020-09-29
Update Test.php

ThomasLandauer:
Added `use Codeception\Test\Unit` for nicer syntax.

Would be nice if you could do the same for `@var \{{actorClass}}` - I don't know how :-(

### DIFF
--- a/src/Codeception/Lib/Generator/Test.php
+++ b/src/Codeception/Lib/Generator/Test.php
@@ -12,7 +12,8 @@ class Test
 
     protected $template = <<<EOF
 <?php {{namespace}}
-class {{name}}Test extends \Codeception\Test\Unit
+use Codeception\Test\Unit;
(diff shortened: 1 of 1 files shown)


RazvanFarte:
It looks nicer
But does the Unit class contain any function name that might interfere with the current scope?

ThomasLandauer:
Don't know how/why this could be a problem...
But take a look yourself: https://github.com/Codeception/Codeception/blob/4.1/src/Codeception/Test/Unit.php

DavertMik:
Sorry, to my preference this doesn't look better
Also it breaks our tests. 
So I'd rather not change this part

ThomasLandauer:
OK, it's not that important. But isn't it a best practice to have all your `use`s on top? See https://www.php-fig.org/psr/psr-12/#41-extends-and-implements

PR #5990, opened 2020-09-30
Update Web.php

ThomasLandauer:
Integrating the findings of https://github.com/Codeception/module-webdriver/issues/24

Do you know what the reason for this delay could be?


DavertMik:
Yes, cookie is not set immediately, this is how browser works. I always reload page after cookie was set

Issue #5989, opened 2020-09-30
Can't run Coverage


roomm:
#### What are you trying to achieve?
Trying to run this
`php vendor/bin/codecept run --coverage`

#### What do you get instead?
```bash
Codeception PHP Testing Framework v4.1.7
Powered by PHPUnit 9.3.10 by Sebastian Bergmann and contributors.
Running with seed: 


  [Connecting To Db] {"config":{"populate":false,"cleanup":false,"reconnect":false,"waitlock":0,"dump":null,"populator":null,"dsn":"mysql:host=*****;port=*****;dbname=****","user":"*****","password":"****"},"options":[]}
  [Db] Connected to default foodradar_testing

App\Tests.acceptance Tests (13) -----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Modules: WebDriver, Db, Symfony, Doctrine2, \App\Tests\Helper\Acceptance, Asserts
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

In ErrorHandler.php line 83:
                                                                                                                                        
  [PHPUnit\Framework\Exception (2)]                                                                                                     
  file_get_contents(http://localhost//c3/report/clear): failed to open stream: HTTP request failed! HTTP/1.1 500 Internal Server Error  
                                                                                                                                        

Exception trace:
  at /var/www/vendor/codeception/codeception/src/Codeception/Subscriber/ErrorHandler.php:83
 Codeception\Subscriber\ErrorHandler->errorHandler() at n/a:n/a
 file_get_contents() at /var/www/vendor/codeception/codeception/src/Codeception/Coverage/Subscriber/LocalServer.php:167
 Codeception\Coverage\Subscriber\LocalServer->c3Request() at /var/www/vendor/codeception/codeception/src/Codeception/Coverage/Subscriber/LocalServer.php:72
 Codeception\Coverage\Subscriber\LocalServer->beforeSuite() at /var/www/vendor/symfony/event-dispatcher/EventDispatcher.php:264
 Symfony\Component\EventDispatcher\EventDispatcher->doDispatch() at /var/www/vendor/symfony/event-dispatcher/EventDispatcher.php:239
 Symfony\Component\EventDispatcher\EventDispatcher->callListeners() at /var/www/vendor/symfony/event-dispatcher/EventDispatcher.php:73
[...]

PR #5991, opened 2020-09-30
Switch off c3Request call if 'remote' set to false

dereuromark:
Might have something to do with https://github.com/Codeception/Codeception/issues/5989

But for us running it locally never works and fails with
Screenshot 2020-09-14 at 16 32 32

This solves it.


DavertMik:
Thank you! That's a good fix!

PR #5994, opened 2020-10-06
Added Cobertura XML coverage support

zachkknowbe4:
Now that PHPUnit has support for Cobertura XML, I've added support so that Codeception can make use of it.

### DIFF
--- a/src/Codeception/Coverage/Subscriber/Printer.php
+++ b/src/Codeception/Coverage/Subscriber/Printer.php
@@ -6,6 +6,7 @@ use Codeception\Coverage\Filter;
 use Codeception\Coverage\PhpCodeCoverageFactory;
 use Codeception\Event\PrintResultEvent;
 use Codeception\Events;
+use Codeception\Exception\ConfigurationException;
 use Codeception\Subscriber\Shared\StaticEvents;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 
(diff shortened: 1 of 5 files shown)


src/Codeception/Coverage/Subscriber/Printer.php
```diff
@@ -146,6 +150,12 @@ protected function printCrap4j()
         $writer->process(self::$coverage, $this->absolutePath($this->options['coverage-crap4j']));
     }
 
+    protected function printCobertura()
+    {
+        $writer = new \SebastianBergmann\CodeCoverage\Report\Cobertura;
```
Naktibalda, line 155: Please check if this class is defined before using it and throw exception that php-code-coverage >=9.2.0 is required.

tests/coverage/RemoteWithCoberturaCept.php
```diff
@@ -0,0 +1,8 @@
+<?php
+$I = new CoverGuy($scenario);
+$I->wantTo('try generate remote cobertura xml report');
```
Naktibalda, line 3: This test fails with PHPUnit 8 and older so it must be skipped if `\SebastianBergmann\CodeCoverage\Report\Cobertura` class does not exist.

Please check if this code works:
```php
if (!class_exists(\SebastianBergmann\CodeCoverage\Report\Cobertura::class)) {
  $scenario->skip('Cobertura report requires php-code-coverage 9.2');
}
````

Naktibalda (changes requested):

Naktibalda:
Also please document your new parameter in the docblock of Run command: https://github.com/Codeception/Codeception/blob/4.1/src/Codeception/Command/Run.php#L74-L77

It is used to generate this reference page: https://codeception.com/docs/reference/Commands#Run

Naktibalda (approved):
Looks good now.

zachkknowbe4:
Awesome thank you.

PR #5998, opened 2020-10-14
Update Api.php

ThomasLandauer:
See https://github.com/Codeception/module-rest/pull/28


Naktibalda:
Thanks, but it would be great if you could try to use meaningful commit messages.

Issue #5556, opened 2019-06-19
$I->seeElement required html tag? 


cosinus90:
I have AJAX endpoint that returns html markup of the items loaded on the page scroll. This isn't a jquery and isn't a full HTML document which I mean starts from HTML tag. My endpoint returns items that later will be included in the document like

`$('#id').append(html);`

I guess it's a common situation. But when I try to test this result for an element - it fails:

```
$I->sendAjaxRequest('POST', '/channel/search?sort_column=false&sort_dir=false', [...]);
$I->seeElement('.offset');
```

First line of AJAX response is 

````
 <tr class="offset" data-offset="10" ></tr>
````

Looks like the seeElement function (and others) require the HTML document starting from HTML tag? Is there any possibility to test the response in my case?

**Comment by Naktibalda** [comment]
You could use [seeInSource](https://codeception.com/docs/modules/PhpBrowser#seeInSource) to match a bit of HTML.
```php
$I->seeInSource('<tr class="offset"');
```

ThomasLandauer:
So what are the exact requirements for `seeElement()` to work? And does the same apply to `$I->see('...', [XXX])` as well?

Naktibalda:
How does it fail?

ThomasLandauer:
```php
$I->sendAjaxPostRequest(...);
$I->seeElement(['css' => 'form']);
```
fails with:
> Step  See element {"css":"form"}
> Fail  Element located either by name, CSS or XPath element with '"css":"form"' was not found.

ThomasLandauer:
Just did some testing with `see()`:
* When I just pass the string (`see('foo');`), it works.
* When I also pass a CSS selector (`see('foo', ['css' => '.class']);`), it does never work - no matter if the Ajax response has an `<html><body>` skeleton or not.

Naktibalda:
I looked into this, and I think that it is a bug that any assertion sees response of Ajax request.

As far as I remember, the intention was that Ajax methods make requests that cause some side effects in backend, e.g. update database record, but the browser stays on the same page where it was before making Ajax request and the test can `click` or `submitForm` on that page.

@DavertMik Can you confirm that?

DavertMik:
`sendAjaxRequest` was one of the first methods introduced to Codeception to test my project I worked on in 2012. The fun fact it didn't change much since that time.

Originally, it was dealing with HTML responses because our site used to display popups via HTML.

[...]

PR #6004, opened 2020-10-21
Update Web.php

ThomasLandauer:
Mentioning `<html>` tag requirement. See https://github.com/Codeception/Codeception/issues/5556
Question: Is `<html>` enough, or are there more requirements?


Naktibalda:
This file was moved to https://github.com/Codeception/lib-web/blob/main/src/Lib/Interfaces/Web.php

Issue #5987, opened 2020-09-29
PHPUnit 9.3 assertEquals error


Pi-George:
Argument 3 passed to PHPUnit\Framework\Assert::assertEquals() must be of the type string, null given, called in /srv/www/vendor/codeception/lib-asserts/src/Codeception/Util/Shared/InheritedAsserts.php on line 243

Pi-George:
So it's easy enough to fix by simply providing a message instead of null, however the default value is null so something is wrong for sure here https://imgur.com/RXngQk1

Naktibalda:
So the problem is that the default value is null, it should be an empty string.

Naktibalda:
It looks like the problem is in this generator:
https://github.com/Codeception/Codeception/blob/4.1.7/src/Codeception/Lib/Generator/Actions.php#L158
it uses null as default value for all optional arguments.

The same issue was fixed in Actor generator 5 years ago:
https://github.com/Codeception/Codeception/commit/60160d88679b296f92b229b9829480aa8c1fab0f

PR #6016, opened 2020-10-27
BUG FIX: Use fully qualified name for class constant defaults

lastzero:
A recent fix for #5987 added support for action parameter defaults, primarily used for generating actor traits from what I see.
This PR makes sure `ReflectionHelper::getDefaultValue()` returns fully qualified names for class constant defaults referencing "self".

In addition, it adds unit tests and makes sure empty defaults have the same type as the parameter they belong to if null is not allowed (even though that may not cause issues just yet).

*Note: This was developed & tested using PHP 7.4.3 while >=5.6.0 is supported according to composer.json (true?). Minor changes might be needed for full compatibility. Travis CI should report potential issues.*

**Our test suite broke completely after upgrading Codeception from  v4.1.8 to v4.1.9 - likely to be the case for other users as well, so I'd consider this an important bug fix that should be released as soon as possible. Thank you very much! :)**

### DIFF
--- a/src/Codeception/Util/ReflectionHelper.php
+++ b/src/Codeception/Util/ReflectionHelper.php
@@ -125,7 +125,7 @@ class ReflectionHelper
         }
 
         // Default to 'null' for PHP versions < 7.1.
-        if(PHP_VERSION_ID < 70100) {
+        if (PHP_VERSION_ID < 70100) {
             return 'null';
         }
 
--- a/src/Codeception/Util/ReflectionHelper.php
+++ b/src/Codeception/Util/ReflectionHelper.php
@@ -23,7 +23,7 @@ class ReflectionHelper
     public static function readPrivateProperty($object, $property, $class = null)
     {
         if (is_null($class)) {
-            $class = get_class($object);
+            $class = $object;
         }
 
         $property = new ReflectionProperty($class, $property);
@@ -45,7 +45,7 @@ class ReflectionHelper
     public static function invokePrivateMethod($object, $method, $args = [], $class = null)
     {
         if (is_null($class)) {
-            $class = get_class($object);
+            $class = $object;
         }
 
         $method = new ReflectionMethod($class, $method);
--- a/src/Codeception/Util/ReflectionHelper.php
+++ b/src/Codeception/Util/ReflectionHelper.php
@@ -126,15 +126,29 @@ class ReflectionHelper
 
         $type = $param->getType();
 
+        // Return 'null' as default if explicitly allowed or there is no specific type hint.
         if (!$type || $type->allowsNull() || !$type->isBuiltin()) {
             return 'null';
         }
 
-        $result = null;
-
-        settype($result, $type->getName());
-
-        return var_export($result, true);
+        // Default value should match the parameter type if 'null' is _not_ allowed.
+        switch ($type->getName()) {
+            case 'string':
+                return "''";
+            case 'array':
+                return '[]';
+            case 'boolean':
+                return 'false';
+            case 'int':
+            case 'integer':
+            case 'float':
+            case 'double':
+            case 'number':
+            case 'numeric':
+                return '0';
+            default:
+                return 'null';
+        }
     }
 
     /**
--- a/src/Codeception/Util/ReflectionHelper.php
+++ b/src/Codeception/Util/ReflectionHelper.php
@@ -1,9 +1,10 @@
 <?php
 namespace Codeception\Util;
 
-use ReflectionClass;
 use ReflectionException;
 use ReflectionParameter;
+use ReflectionProperty;
+use ReflectionMethod;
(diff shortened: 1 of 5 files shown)


src/Codeception/Util/ReflectionHelper.php
```diff
@@ -121,7 +124,17 @@ public static function getDefaultValue(\ReflectionParameter $param)
...
+
+        if (!$type || $type->allowsNull() || !$type->isBuiltin()) {
+            return 'null';
+        }
+
+        $result = null;
+
+        settype($result, $type->getName());
+
+        return var_export($result, true);
```
Naktibalda, line 137: I still down know why does the `phpEncodeValue` method `json_encode` string values.
If there is a good reason for that, it would be better to call `phpEncodeValue` here instead of var_export.

lastzero: I was wondering as well why this is needed? There might be better ways to deal with binary / unicode strings. Don't want to break existing functionality, or introduce security issues, though.

lastzero: Oh, maybe I got this wrong: You say there might be a good reason to call `phpEncodeValue()` instead **because** it uses `json_encode()` for strings?

For this particular use case, the only difference is that `json_encode()`  uses double instead of single quotes (single quotes may be a tiny bit faster in PHP, but probably doesn't matter for empty strings): `""` vs `''`

Generating PHP source using JS encoding generally seems like a bad idea to me.

lastzero:
Similar to recent pull requests of other contributors, some test environments on Travis CI failed with:

**No composer.json present in the current directory... Package hirak/prestissimo at version has a PHP requirement incompatible with your PHP version, PHP extensions and Composer version**

Is there anything I can do about it or just ignore?

Screenshot 2020-10-27 at 19 28 53

Screenshot 2020-10-27 at 19 27 56

src/Codeception/Util/ReflectionHelper.php
lastzero: Replaced `var_export()` with a more explicit switch statement. For callbacks and objects, `getDefaultValue()` continues to return "null" as before.

lastzero:
With regard to `phpEncodeValue()` and `json_encode()`, I'd consider this a separate issue. I'm working with PHP since the 90s and honestly can't tell you what might happen, especially with unicode string defaults. First step would be to add tests for this.

My impression is that it was used to perform some kind of escaping, especially for `phpEncodeArray()`.

mwi-gofore:
This apparently relates to errors like `Error: Class '_generated\My\PSR\FooBar' not found` when used as default value for a function argument in a helper class?
`public function doSomething($argWithStaticDefault = FooBar::DEFAULT_VALUE) {
    ...
}
`

lastzero:
> This apparently relates to errors like `Error: Class '_generated\My\PSR\FooBar' not found` when used as default value for a function argument in a helper class?
> `public function doSomething($argWithStaticDefault = FooBar::DEFAULT_VALUE) { ... } `

Exactly 👍

mwi-gofore:
Oh, you probably want to remove the dependency on hirak/prestissimo, as Travis uses composer 2 now by default

lastzero:
> Oh, you probably want to remove the dependency on hirak/prestissimo, as Travis uses composer 2 now by default

Was trying to focus on the actual bug and not to fix unrelated issues. If tests need to be green and that's the only way, I'll do it. How to proceed @Naktibalda?

mwi-gofore:
> Was trying to focus on the actual bug and not to fix unrelated issues. If tests need to be green and that's the only way, I'll do it. How to proceed @Naktibalda?

Looks like they may have changed the auto update of composer to have version 1.x again now.. Can you trigger a new test run?

Naktibalda (approved):
Thank you, this change looks good to me.

mwi-gofore:
Whoa. Nice backlog of builds in Travis.. Hope that #6018 fixes the build issue (might need to wait some more hours for results though)

mwi-gofore:
> Whoa. Nice backlog of builds in Travis.. Hope that #6018 fixes the build issue (might need to wait some more hours for results though)

Looks like it worked. Can this PR be merged or should it be rebased first?

lastzero:
@mwi-gofore better safe than sorry 🧐

lastzero:
Going to need slight changes for PHP 5.6 and 7.0 compatibility:
- `$type->__toString()` is deprecated in favor of `$type->getName()`, which is not available in PHP <= 7.0 though
- PHP 5.6 reports a parse error, should be easy to fix

lastzero:
Hope that's it! Tests green locally using PHP 7.4, 7.0 and 5.6...

lastzero:
@Naktibalda What is missing for this to be merged?

PR #6018, opened 2020-10-29
Remove global dependency hirak/prestissimo

mwi-gofore:
Composer 2.0 should take care of everything now, mostly testing if this minor change fixes the test runs


mwi-gofore:
> Composer 2.0 should take care of everything now, mostly testing if this minor change fixes the test runs

This will slightly slow down the builds that run with composer 1.x, but it's going to be deprecated "soon" (whatever that means) and then finally stop working, so  it should not be a big issue. It could be possible to check the version, and only optionally try to install hirak/prestissimo, but that check would then become useless later

Issue #6024, opened 2020-11-02
Bug: Class constant as default function value in Helper translated to wrong namespace in `...TesterActions.php`


ThomasLandauer:
I'm not sure if this will automatically be fixed through https://github.com/Codeception/Codeception/issues/5727

When I have this in my `Helper/Acceptance.php`:
```php
use App\Entity\Foo;
public function whatever(string $input = Foo::BAR) { }
```
... the generated method in `AcceptanceTesterAction.php` looks like this:
```php
public function whatever($whatever = App\Entity\Foo::BAR) { }
```
...which leads to this error message:
> [Error] Class 'App\Tests\_generated\App\Entity\Foo' not found

... cause `App\Entity\Foo` is resolved *relatively* to the `AcceptanceTesterAction`s namespace `App\Tests\_generated`.

### Details

* Codeception version: 4.1.9

javabudd:
confirmed, after updating to 4.1.9 this is happening to my code as well

Naktibalda:
Fixed by #6016 , released as 4.1.10

mwi-gofore:
Actually, the fix only works for cases where the constant is defined in the same class. If the constant comes from another class, it still fails :(

PR #6027, opened 2020-11-03
Prepend backslash on class constant default values

mwi-gofore:
Need to prepend a '\\' on imported class constants also, not just constants from the same class (which was already fixed by #6016).
Fixes the issue from #6024

### DIFF
--- a/src/Codeception/Util/ReflectionHelper.php
+++ b/src/Codeception/Util/ReflectionHelper.php
@@ -115,6 +115,8 @@ class ReflectionHelper
                     list($class, $const) = explode('::', $constName);
                     if (in_array($class, ['self', 'static'])) {
                         $constName = '\\' . $param->getDeclaringClass()->getName() . '::' . $const;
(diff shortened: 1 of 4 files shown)


mwi-gofore:
I'm not sure if this possibly could break something, but seems to fix our builds at least.

Generated code
```PHP
public function createModel($name, $policy = \My\Namespace\Security::DEFAULT_POLICY) {
    return $this->getScenario()->runStep(new \Codeception\Step\Action('createModel', func_get_args()));
}
```
whereas before it was

```PHP
public function createModel($name, $policy = My\Namespace\Security::DEFAULT_POLICY) {
    return $this->getScenario()->runStep(new \Codeception\Step\Action('createModel', func_get_args()));
}
```
which would fail with the mentioned
```
[Error] Class '_generated\My\Namespace\Security' not found
```

Naktibalda:
Released as 4.1.11, because I did all the work to release 4.1.10 in the morning, except actually tagging it.

PR #6028, opened 2020-11-03
Prevent reformating of method names in CLI output

ThomasLandauer:
Currently the output is
> ✔ FooCest: Try to create a new user (0.13s)

This PR changes it to:

> ✔ FooCest: TryToCreateANewUser (0.13s)

Reason: Leave the method name as-is, to allow copy-pasting it.

TODO: The very first letter is still being capitalized - but before searching where this happens, I'm waiting for some overall feedback on this :-)

### DIFF
--- a/src/Codeception/Test/Cest.php
+++ b/src/Codeception/Test/Cest.php
@@ -65,11 +65,7 @@ class Cest extends Test implements
 
     public function getSpecFromMethod()
     {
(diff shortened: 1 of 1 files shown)


DavertMik:
Nope, I think test names should be readable. 
In `debug` mode you have the test names to copy.

PR #6031, opened 2020-11-10
Fix PHP types being treated as a classes

cs278:
Some of our tests have started failing with latest versions of Codeception because the DI code is trying to find an `int` class.﻿

### DIFF
--- a/src/Codeception/Util/ReflectionHelper.php
+++ b/src/Codeception/Util/ReflectionHelper.php
@@ -85,7 +85,7 @@ class ReflectionHelper
         }
 
         $type = $parameter->getType();
-        if ($type === null) {
+        if ($type === null || $type->isBuiltin()) {
             return null;
         }
         $typeString = $type->getName();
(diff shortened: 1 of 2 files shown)


src/Codeception/Util/ReflectionHelper.php
```diff
@@ -85,7 +85,7 @@ public static function getClassFromParameter(ReflectionParameter $parameter)
         }
 
         $type = $parameter->getType();
-        if ($type === null) {
+        if ($type === null || $type->isBuiltin()) {
```
Naktibalda, line 88: Cool, there is a method for types, I thought that it may be necessary to hardcode a list.

Naktibalda:
Regarding your tests, have you got a good reason to use literal parameters in public methods of test classes?

One example I've seen so far had incorrect visibility of trait method, it wasn't supposed to be public.

cs278:
One case is a public method which probably should be private but the other cases are private methods.

Naktibalda:
Private methods are not treated as tests, so they don't get dependencies injected, it is only that public method that's causing errors.

Naktibalda:
Released as 4.1.12

cs278:
Thanks for the merge. I've just verified this does affect private methods, if you do something like the code below:

```php
class mycest
{
    /**
     * @before createEntity
     */
    public function someTest($I)
    {
        // Blah blah
    }

    private function createEntity($I, int $value = 0)
    {
        // Create the entity
    }
}
```

Issue #5740, opened 2019-11-06
Explain how to exit the interactive shell (`pause()`)


ThomasLandauer:
When starting the interactive shell, it should also be explain how to exit it:

> Type in commands to try them, ENTER to continue, TAB to auto-complete

It looks like typing "exit" exits the interactive console and continues with the test, right? So maybe:

> Type in commands to try them, ENTER to continue, TAB to auto-complete, 'exit' to leave the console

I noticed that pressing Ctrl+C quits the Codeception run completely, but leaves the browser window (in acceptance test) open. Is there a command to gracefully stop the test and close the browser?

https://github.com/Codeception/Codeception/blob/9c69ec708f15140be20341d342c330d68fb04197/src/Codeception/Lib/Actor/Shared/Pause.php#L26

Naktibalda:
There is no such explanation anymore since hoa/console was replaced by PsySH in #6385

PR #6034, opened 2020-11-17
Update Pause.php

ThomasLandauer:
Adding 'exit' information. Closes https://github.com/Codeception/Codeception/issues/5740

### DIFF
--- a/src/Codeception/Lib/Actor/Shared/Pause.php
+++ b/src/Codeception/Lib/Actor/Shared/Pause.php
@@ -101,7 +101,6 @@ trait Pause
         $output->writeln("  - <info>F6</info> to toggle auto-stashing of successful commands");
         $output->writeln("  - <info>F8</info> to view stashed commands");
         $output->writeln("  - <info>F10</info> to clear stashed commands");
-        $output->writeln("  - <info>exit</info> to leave the Interactive Shell");
 
         $result = '';
 
--- a/src/Codeception/Lib/Actor/Shared/Pause.php
+++ b/src/Codeception/Lib/Actor/Shared/Pause.php
@@ -94,7 +94,7 @@ trait Pause
         $readline->addMapping($tput->get('key_f10'), $clearStashFn);
 
         $output->writeln("  <comment>Execution PAUSED, starting Interactive Shell...</comment>");
-        $output->writeln("  Type in PHP commands to try them, or press:");
+        $output->writeln("  Type in commands to try them, or press:");
         $output->writeln("  - <info>ENTER</info> (on blank line) to leave the Interactive Shell");
         $output->writeln("  - <info>TAB</info> to auto-complete");
         $output->writeln("  - <info>F5</info> to stash a command");
--- a/src/Codeception/Lib/Actor/Shared/Pause.php
+++ b/src/Codeception/Lib/Actor/Shared/Pause.php
@@ -93,7 +93,7 @@ trait Pause
         $readline->addMapping($tput->get('key_f8'), $viewStashedFn);
         $readline->addMapping($tput->get('key_f10'), $clearStashFn);
 
-        $output->writeln("  <comment>Execution PAUSED, starting interactive shell...</comment>");
+        $output->writeln("  <comment>Execution PAUSED, starting Interactive Shell...</comment>");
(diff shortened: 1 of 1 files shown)


src/Codeception/Lib/Actor/Shared/Pause.php
```diff
@@ -93,14 +93,15 @@ public function pause()
...
-        $output->writeln("  <comment>Execution PAUSED, starting interactive shell...</comment>");
+        $output->writeln("  <comment>Execution PAUSED, starting Interactive Shell...</comment>");
         $output->writeln("  Type in commands to try them:");
         $output->writeln("  - <info>ENTER</info> to continue");
         $output->writeln("  - <info>TAB</info> to auto-complete");
         $output->writeln("  - <info>F5</info> to stash a command");
         $output->writeln("  - <info>F6</info> to toggle auto-stashing of successful commands");
         $output->writeln("  - <info>F8</info> to view stashed commands");
         $output->writeln("  - <info>F10</info> to clear stashed commands");
+        $output->writeln("  - <info>exit</info> to leave the Interactive Shell");
```
Naktibalda, line 104: Based on the code below, entering `exit` has the same effect as pressing ENTER on empty line.

I think that the line which must be fixed is `Type in commands to try them`, because the only command that can be typed is `exit`, all others are keyboard keys that can be pressed.

Naktibalda (changes requested):

Naktibalda:
It would be great if you could start putting short description in commit message field, using default value provided by Github is not a good practice.

src/Codeception/Lib/Actor/Shared/Pause.php
ThomasLandauer: Yeah, I only noticed that afterwards ;-) I updated it now.
2 questions:
* Then why is `exit` in the code at all? Is there a use-case where ENTER may not work? If not, I'll delete the `exit` line again.
* Just noticed that Backspace is not working, so there's no way to fix a typo?! Is this a limitation of `hoa/console`?

src/Codeception/Lib/Actor/Shared/Pause.php
```diff
@@ -94,8 +94,8 @@ public function pause()
         $readline->addMapping($tput->get('key_f10'), $clearStashFn);
 
         $output->writeln("  <comment>Execution PAUSED, starting Interactive Shell...</comment>");
-        $output->writeln("  Type in commands to try them:");
-        $output->writeln("  - <info>ENTER</info> to continue");
+        $output->writeln("  Type in PHP commands to try them, or press:");
```
Naktibalda, line 97: `PHP Commands` is not accurate, what you actually have to enter is the name of actor method and the list of parameters, e.g. `amOnPage('/info')`,  because command is executed by this code:
```php
 $value = eval("return \$I->$command;");
```

Clearly command injection is possible with semicolon, but it doesn't look like intended use case for me.

ThomasLandauer:
Sorry, I never understood the difference between those two input fields GitHub is giving me; so I didn't know what I'm expected to put into which. I'll try to improve :-)

src/Codeception/Lib/Actor/Shared/Pause.php
ThomasLandauer: I thought the trailing `;` is mandatory... Now removed "PHP" again :-)

src/Codeception/Lib/Actor/Shared/Pause.php
Naktibalda: Please delete the `exit` line.

I can't answer question about hoa/console, because I am not familiar with it.

There is some ambiguity in the current text that it doesn't tell to press ENTER after typing command.

ThomasLandauer: > Please delete the exit line.

Done :-)

> There is some ambiguity in the current text that it doesn't tell to press ENTER after typing command.

I guess that's clear to anybody :-)

ThomasLandauer: Just found a way to correct errors: `UP` key!

Naktibalda:
hoa/console was replaced by PsySH in #6385

Issue #6047, opened 2020-12-14
Repository test suite does not run


ctrl-f5:
#### What are you trying to achieve?

I want to make a PR. to do this I want to clone the repo and run the tests.

#### What do you get instead?

a failing test suite:

as per documentation in https://github.com/Codeception/Codeception/blob/4.1/tests/README.md:

    git clone git@github.com:Codeception/Codeception.git 
    cd Codeception/tests 
    docker-compose build
    docker-compose run --rm codecept run cli

Does not run without failures:

    ERRORS!
    Tests: 208, Assertions: 659, Errors: 11, Failures: 22, Skipped: 3.

As a side note:

> https://github.com/Codeception/Codeception/blob/4.1/CONTRIBUTING.md

states

> Breaking features and major improvements should be sent into master. 

But there is no master branch in the repo.

Full output:

```
Time: 00:28.123, Memory: 14.00 MB

There were 11 errors:

---------
1) GeneratePageObjectCest: Generate global page object in different path
 Test  tests/cli/GeneratePageObjectCest.php:generateGlobalPageObjectInDifferentPath
                                                                                                                                          
  [InjectionException] Failed to inject dependencies in instance of 'GeneratePageObjectCest'. Class CliGuy\GeneratorSteps does not exist  
                                                                                                                                          
#1  /repo/app.php:46
#2  /repo/app.php:47
#3  /repo/codecept:7

---------
2) SnapshotCest: Generate suite snapshot
 Test  tests/cli/SnapshotCest.php:generateSuiteSnapshot
                                                                                                                                
  [InjectionException] Failed to inject dependencies in instance of 'SnapshotCest'. Class CliGuy\GeneratorSteps does not exist  
                                                                                                                                
#1  /repo/app.php:46
#2  /repo/app.php:47
#3  /repo/codecept:7

---------
3) GeneratePageObjectCest: Generate global page object
 Test  tests/cli/GeneratePageObjectCest.php:generateGlobalPageObject
                                                                                                                                          
  [InjectionException] Failed to inject dependencies in instance of 'GeneratePageObjectCest'. Class CliGuy\GeneratorSteps does not exist  
                                                                                                                                          
#1  /repo/app.php:46
#2  /repo/app.php:47
#3  /repo/codecept:7

---------
4) GeneratePageObjectCest: Generate suite page object
 Test  tests/cli/GeneratePageObjectCest.php:generateSuitePageObject
                                                                                                                                          
[...]

PR #6050, opened 2020-12-14
udpated docker setup, fixed some tests

ctrl-f5:
As explained in #6047 I was having issues with the docker-compose setup.

Either I was missing something, or the setup is just outdated and doesn't work anymore.

The tests also used the root Dockerfile, which installed everything with `--no-dev` which means autoloading did not include certain test files, resulting in a lot of `Class not found` errors.

Also found some tests that I think could never succeed, so i'm really struggling to see how these pass the travis builds and you guys' local test runs... (`WarningTest` throws an exception, which is an error, not a warning, type hint test checked for an `int` on a `string` function)  
Please enlighten me if I am missing something.

In any case, I got it working locally with this setup:

- `.dockerignore` only for main `Dockerfile`
- `Dockerfile-dev` used for the tests docker-compose build, this now includes the entire repo, when doing a composer install your vendor is also updated locally so you can actually use it for development.
- added the packages that were added in the `Dockerfile` also to the `composer.json` dev requirements, not sure why dockerfile installs `--no-dev` though
- fixed the tests that didn't run for me


samdark (approved):

composer.json
```diff
@@ -36,11 +36,20 @@
         "squizlabs/php_codesniffer": "~2.0",
         "vlucas/phpdotenv": "^2.0 | ^3.0 | ^4.0 | ^5.0",
         "symfony/process": ">=2.7 <6.0",
+        "codeception/module-apc": "*@dev",
```
samdark, line 39: Do we need all these requirements for tests?

composer.json
```diff
@@ -36,11 +36,20 @@
...
         "codeception/module-filesystem": "*@dev",
+        "codeception/module-ftp": "*@dev",
+        "codeception/module-memcache": "*@dev",
+        "codeception/module-mongodb": "*@dev",
         "codeception/module-phpbrowser": "*@dev",
+        "codeception/module-redis": "*@dev",
+        "codeception/module-rest": "*@dev",
+        "codeception/module-sequence": "*@dev",
+        "codeception/module-soap": "*@dev",
+        "codeception/module-webdriver": "*@dev",
```
Naktibalda, line 52: If this is for Docker only, please add `composer require` commands to Dockerfile.

Naktibalda (changes requested):

composer.json
ctrl-f5: Well there is no "docker only" either it's needed to run the tests and it should be in `require-dev` or it should not be anywhere. requiring packages only in the dockerfile makes no sense imho?

composer.json
ctrl-f5: They were required by the original docker image: https://github.com/Codeception/Codeception/blob/4.1/Dockerfile#L59

If they are actually used, I dont know, but if not, they shouldn't be in that image either? or does that image serve a special purpose?

samdark: No idea.

samdark:
@ctrl-f5 we've switched to GitHub actions. Would you please merge master into this PR branch? Thanks.

composer.json
Naktibalda: Dockerfile is used to build images that are published at https://hub.docker.com/r/codeception/codeception
Some people may be using it to run their tests, so the docker image provides the same modules as phar file.

Modules that are currently listed in `require-dev` section are the only modules that are required to run Codeception test suite. The fewer, the better.

Dockerfile-dev
```diff
@@ -0,0 +1,64 @@
+FROM php:7.4-cli
+
+LABEL maintainer="Tobias Munk <tobias@diemeisterei.de>"
+
+ENV CODECEPT_USER codecept
```
Naktibalda, line 5: Wouldn't it be better to move this file to tests/ directory?

Dockerfile-dev
```diff
@@ -0,0 +1,64 @@
...
+RUN pecl install \
+        apcu \
+        memcached \
+        mongodb \
+        xdebug-2.9.5 && \
+    docker-php-ext-enable \
+        apcu.so \
+        memcached.so \
+        mongodb.so \
+        xdebug
```
Naktibalda, line 41: If the goal is to run tests, then the only PECL extension you need is xdebug.

Dockerfile-dev
```diff
@@ -0,0 +1,64 @@
...
+        rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+# Install php extensions
+RUN docker-php-ext-install \
+    bcmath \
+    mysqli \
+    pdo pdo_mysql pdo_pgsql \
+    soap \
+    sockets \
+    zip
```
Naktibalda, line 29: The only database extension used by tests is sqlite.
Github Actions enable these extensions - curl, mbstring, openssl, pdo, pdo_sqlite.
Other extensions aren't necessery in Dockerfile which runs tests.

tests/cli/BuildCest.php
```diff
@@ -51,10 +51,10 @@ public function usesTypehintsWherePossible(CliGuy $I, Scenario $scenario)
         $I->seeInThisFile('use _generated\CliGuyActions');
         $I->seeFileFound('CliGuyActions.php', 'tests/support/_generated');
         $I->openFile(codecept_root_dir('tests/support/CliHelper.php'));
-        $I->seeInThisFile('public function grabFromOutput($regex): int');
```
Naktibalda, line 54: This test wasn't failing because typehint was set by str_replace above, not by real code.

tests/README.md
```diff
@@ -170,17 +170,10 @@ Using `docker-compose` for test configurations
...
     docker-compose build
-
-Start
-
-    docker-compose up -d
-
-By default the image has `codecept` as its entrypoint, to run the tests simply supply the `run` command
-
-    docker-compose run --rm codecept help
+    docker-compose run --rm --entrypoint composer codecept install
```
Naktibalda, line 176: codecept install ?

Naktibalda (changes requested):

composer.json
ctrl-f5: alright, that makes sense, already had a hunch the dockerfile had some other purpose than just local development. If I had some spare time during the hollidays I'll update the PR

PR #6051, opened 2020-12-15
detect the suite from a test path relative to the current working dir

ctrl-f5:
Detect the suite from a test file or path relative to the current working directory.

This fixes the issue that the suite cant be detected when `codeception.yml` is in any subdirectory, and subsequently also fixes the issue when it's inside the tests directory, for which there was a single use case fix in place.

### DIFF
--- a/src/Codeception/Command/Run.php
+++ b/src/Codeception/Command/Run.php
@@ -117,7 +117,6 @@ class Run extends Command
      */
     protected $output;
 
-
     /**
      * Sets Run arguments
      * @throws \Symfony\Component\Console\Exception\InvalidArgumentException
@@ -429,12 +428,6 @@ class Run extends Command
 
     protected function matchSingleTest($suite, $config)
     {
-        // Workaround when codeception.yml is inside tests directory and tests path is set to "."
-        // @see https://github.com/Codeception/Codeception/issues/4432
-        if (isset($config['paths']['tests']) && $config['paths']['tests'] === '.' && !preg_match('~^\.[/\\\]~', $suite)) {
-            $suite = './' . $suite;
-        }
-
         // running a single test when suite has a configured path
         if (isset($config['suites'])) {
             foreach ($config['suites'] as $s => $suiteConfig) {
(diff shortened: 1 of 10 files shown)


samdark:
@Naktibalda worth merging master into this one and checking test results with GitHub actions.

ctrl-f5:
@samdark updated with upstream, seems to run fine in the new actions setup!

samdark:
It doesn't break current tests. That's good. It would be great to add some tests verifying the change.

ctrl-f5:
It was basically already tested by `CodeceptionYmlInTestsDirCest` since that's one use case. I've also added a more convoluted test case with some more directories to traverse, based on how we use it in our projects.

src/Codeception/Command/Run.php
```diff
@@ -443,6 +436,9 @@ protected function matchSingleTest($suite, $config)
                 }
                 $testsPath = $config['paths']['tests'] . DIRECTORY_SEPARATOR . $suiteConfig['path'];
                 if ($suiteConfig['path'] === '.') {
+                    if ($config['paths']['tests'] === '.') {
+                        return ['', $s, $suite];
```
Naktibalda, line 440: I don't like this line at all. 

This code is inside `matchSingleTest` method and the third element in result array is supposed to be a path to specific test file.
This condition doesn't even check if `$suite` is a path.

ctrl-f5: The type of check you describe was not there in the past, nor in this function (line 444), nor in `matchTestFromFilename`. They both just do a `preg_match` on the input and return the result.

`$suite` is never validated as a path in any other code branch.

ctrl-f5: anyway, I've removed this code and put back the original hack at the start of the function

Naktibalda:
@ctrl-f5 Your change broke `codecept run tests` - it is fails with this error:
```
In Run.php line 545:

  Test file can't be matched
```
#6103

Could you fix it?

Naktibalda:
I implemented patch, which will make your change to apply only if relative path is inside tests directory #6105

PR #6059, opened 2020-12-21
Remove Symfony 2.7 Support

TavoNiievez:
Symfony 2.x has been without support of any kind for years.
However, Symfony 3.4 is still compatible with PHP 5.6 and has security support for another year.


TavoNiievez:
I'm closing this since intentions to support only Symfony 4.4 have already been announced.

PR #6062, opened 2020-12-24
ISSUE-6061: Run Module after.. and fail hooks in reverse order

### LINKED ISSUE #6061: Module `_after` methods should run in reverse order
`ModuleA` depends on `ModuleB`.
In `A::_after()`  I want to use `ModuleB` to do some cleanup. 

#### What do you get instead?
`ModuleB::_after()` has already run and so is not in a usable state. 

### Details

* Codeception version:  3.1.2
* PHP Version: 7.4
* Operating System: osx
* Installation type: Composer
* Suite configuration:

```yml
actor: ApiTester
modules:
    enabled:
        - REST:
              depends: PhpBrowser
              part: Json
        - Asserts
        - ModuleB:
            depends:
              - REST
         - ModuleA:
            depends:
              - ModuleB
```

WillemHoman:
Run Module `after...` hooks in reverse order so a module can use it's dependencies in `after...`

Fixes #6061

### DIFF
--- a/src/Codeception/Subscriber/Module.php
+++ b/src/Codeception/Subscriber/Module.php
@@ -25,7 +25,16 @@ class Module implements EventSubscriberInterface
         Events::SUITE_AFTER  => 'afterSuite'
     ];
 
+    /** @var \Codeception\Module[] */
     protected $modules = [];
+
+    /**
+     * @param \Codeception\Module[] $modules
+     */
+    public function __construct(array $modules = [])
+    {
+        $this->modules = $modules;
+    }
 
     public function beforeSuite(SuiteEvent $e)
     {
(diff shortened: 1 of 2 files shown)


Naktibalda:
This change looks like a good idea, but you never know what it may break.
I will merge it to 5.0 branch.

See https://github.com/Codeception/Codeception/discussions/6064

WillemHoman:
Thanks @Naktibalda 
Probably wise to make it 5.0. 

I was investigating this further and it also seems to be something worth anywhere that subscribes to `Codeception\Events` such as `Events::SUITE_AFTER` like `Extensions` although I suspect that would more work. 

This PR will at least provide one place where they run in the reverse order but it will make things inconsistent  between modules hooks and Extensions.

DavertMik (approved):

DavertMik:
I changed the base branch to 5.0 to ensure we merge it correctly

PR #6071, opened 2021-01-02
[WIP] New folder structure

DavertMik:
Please take a look on this PR and tell what you think of it as it introduces some new defaults to Codeception

The idea is to make all classes in Codeception to be PSR-compatible

1. all new installations have default namespace: `Tests`
2. all support files are stored in `TestSupport` directory under `Tests\TestSupport` namespace
3. all suite names are uppercases like `Acceptance`
4. all tests in suites contain suite name in a namespace like `Tests\Acceptance`
5. directory structure was changed:

```
tests/
  _output => yeah we got used to this folder so much, it doesn't need to belong to namespace
  Acceptance
  Functional
  TestSupport/
    Data/ => ok, let's  put all data things here, because Data can also contain classes
    Generated/ => we still need generated classes
    Helper/
  Unit/
```
7. Init templates updated correspondingly
8. These changes are not affecting any of the current installations

Additional changes:

* generate:test produces PHP 7.4 ready code with typed properties:

```php
<?php

namespace Tests\Unit;

use Tests\TestSupport\UnitTester;

class UserTest extends Codeception\Test\Unit
{
    protected UnitTester $tester;  

    //...
```

* bootstrap and init template does not create Helper files
* Upgrade4 init template was renamed to `Dependencies`, it is used to install all required modules for existing project:

```
./vendor/bin/codecept init dependencies
```
* Added new CLI tests v5Cest which test generators for new v5 project

PR #6072, opened 2021-01-02
Add line break between opening tag and namespace in generated files

Naktibalda:
Test and Cest generators generate namespace on the same line as opening tag.
e.g.
```php
<?php namespace foo;
```

This change changes it to
```php
<?php
namespace foo;
```

### DIFF
--- a/src/Codeception/Lib/Generator/Cest.php
+++ b/src/Codeception/Lib/Generator/Cest.php
@@ -11,8 +11,8 @@ class Cest
     use Namespaces;
 
     protected $template = <<<EOF
-<?php {{namespace}}
-
+<?php
+{{namespace}}
 class {{name}}Cest
 {
     public function _before({{actor}} \$I)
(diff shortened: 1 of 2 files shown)


TavoNiievez:
after I generate a Test, I always change:
```php
<?php namespace App\Tests;

class MyTest extends \Codeception\Test\Unit
```
by:
```php
<?php

namespace App\Tests;

use Codeception\Test\Unit;

final class MyTest extends Unit
```
could this be the default behavior?

Naktibalda:
I think that will be a part of #6071.

Naktibalda:
I misread the question last night, so it probably won't be a part of that pull request.

DavertMik:
@Naktibalda yes, let's make it part of #6071

DavertMik:
The original idea of this was to not introduce empty line if no namespace is used

Naktibalda:
@DavertMik this is a non-breaking change, so it can be merged to 4.1.
4.1 isn't dead yet.

PR #6073, opened 2021-01-02
[5.0] Update code standards in Commands

TavoNiievez:
- [x] Use strict types in `src/Codeception/Command`.
- [x] Add type hints to Commands.
- [x] Remove unnecessary qualifiers in `src/Codeception/Command`.
- [x] Add 'use function' statements for performance reasons.

**Added from feedback:**

- [x] Fix naming in `src/Codeception/Command`.
- [x] Rename `src/Codeception/Command/Shared` traits.

### DIFF
--- a/src/Codeception/Command/Run.php
+++ b/src/Codeception/Command/Run.php
@@ -9,10 +9,11 @@ use Codeception\Configuration;
 use Codeception\Exception\ConfigurationException;
 use Codeception\Exception\ParseException;
 use Exception;
-use PHPUnit\Runner\Version;
+use InvalidArgumentException;
+use PHPUnit\Runner\Version as PHPUnitVersion;
 use RuntimeException;
 use Symfony\Component\Console\Command\Command;
-use Symfony\Component\Console\Exception\InvalidArgumentException;
+use Symfony\Component\Console\Exception\InvalidArgumentException as SymfonyConsoleInvalidArgumentException;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
@@ -128,7 +129,7 @@ class Run extends Command
 
     /**
      * Sets Run arguments
-     * @throws InvalidArgumentException
+     * @throws SymfonyConsoleInvalidArgumentException
      */
     protected function configure()
     {
@@ -277,7 +278,7 @@ class Run extends Command
 
         if (!$this->options['silent']) {
             $this->output->writeln(
-                Codecept::versionString() . "\nPowered by " . Version::getVersionString()
+                Codecept::versionString() . "\nPowered by " . PHPUnitVersion::getVersionString()
             );
             $this->output->writeln(
                 "Running with seed: " . $this->options['seed'] . "\n"
@@ -366,7 +367,7 @@ class Run extends Command
                         try {
                             list(, $suite, $test) = $this->matchTestFromFilename($suite, $testsPath);
                             $isIncludeTest = true;
-                        } catch (\InvalidArgumentException $e) {
+                        } catch (InvalidArgumentException $e) {
                             // Incorrect include match, continue trying to find one
                             continue;
                         }
@@ -531,7 +532,7 @@ class Run extends Command
         $res = preg_match("~^$testsPath/(.*?)(?>/(.*))?$~", $filename, $matches);
 
         if (!$res) {
-            throw new \InvalidArgumentException("Test file can't be matched");
+            throw new InvalidArgumentException("Test file can't be matched");
         }
         if (!isset($matches[2])) {
             $matches[2] = null;
--- a/src/Codeception/Command/Run.php
+++ b/src/Codeception/Command/Run.php
@@ -215,7 +215,7 @@ class Run extends Command
                 InputOption::VALUE_OPTIONAL,
                 'Generate CodeCoverage PHPUnit report in path'
             ),
-            new InputOption('no-exit', '', InputOption::VALUE_NONE, 'Don\'t finish with exit code'),
+            new InputOption('no-exit', '', InputOption::VALUE_NONE, "Don't finish with exit code"),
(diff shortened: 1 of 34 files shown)


TavoNiievez:
@Naktibalda

Naktibalda (changes requested):

TavoNiievez:
@Naktibalda please review today's commits.

src/Codeception/Command/SelfUpdate.php
```diff
@@ -18,12 +23,17 @@
...
+    /**
+     * @var string
+     */
     const GITHUB_REPO = 'Codeception/Codeception';
-    const PHAR_URL = 'http://codeception.com/';
-    const PHAR_URL_PHP56 = 'http://codeception.com/php56/';
+    /**
+     * @var string
+     */
+    const PHAR_URL = 'https://codeception.com/';
```
Naktibalda, line 36: Please set value to `https://codeception.com/php73/`  to distingush it from PHP 7.2 build of Codeception 4.

src/Codeception/Command/Shared/Config.php
```diff
@@ -47,7 +59,7 @@ protected function overrideConfig($configOptions)
         return Configuration::append($updatedConfig);
     }
 
-    protected function enableExtensions($extensions)
+    protected function enableExtensions($extensions): ?array
```
Naktibalda, line 62: I think that all methods in this class return `array`.
If for some reason one of Configuration properties is not initialized, the things must be so bad that execution is stopped before these methods can possibly be executed.

src/Codeception/Command/Shared/FileSystem.php
```diff
@@ -43,7 +55,7 @@ protected function removeSuffix($classname, $suffix)
         return preg_replace("~$suffix$~", '', $classname);
     }
 
-    protected function createFile($filename, $contents, $force = false, $flags = null)
```
Naktibalda, line 46: Was there a reason why you couldn't add typehints for parameters?

Naktibalda (changes requested):

src/Codeception/Command/SelfUpdate.php
```diff
@@ -18,12 +23,17 @@
...
      */
     const NAME = 'Codeception';
+    /**
+     * @var string
+     */
     const GITHUB_REPO = 'Codeception/Codeception';
-    const PHAR_URL = 'http://codeception.com/';
-    const PHAR_URL_PHP56 = 'http://codeception.com/php56/';
+    /**
+     * @var string
```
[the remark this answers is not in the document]
TavoNiievez: See [Why Class Constants Should be Typed](https://tomasvotruba.com/blog/2020/06/22/why-class-constants-should-be-typed/).

src/Codeception/Command/Shared/Config.php
TavoNiievez: Done.
> I think that all methods in this **class** return `array`.

Now that you mention it...
I would like to add the suffix 'Trait' ... to the names of those traits.
So that it looks like: `ConfigTrait` `FileSystemTrait` and `StyleTrait`.
It is a convention to include these suffixes in Traits and Interfaces, but I don't see it very often in Codeception.
Do you agree with them being renamed?

src/Codeception/Command/Run.php
```diff
@@ -404,29 +430,30 @@ public function execute(InputInterface $input, OutputInterface $output)
...
-                exit(1);
-            }
+        if (!$input->getOption('no-exit') && !$this->codecept->getResult()->wasSuccessful()) {
+            exit(1);
         }
         return 0;
     }
 
+    /**
+     * @return mixed|void
```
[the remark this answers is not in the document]
TavoNiievez: However, I had to add return null; explicitly, because otherwise, CI jobs would fail:
`PHP Fatal error:  Uncaught TypeError: Codeception\Command\Run::matchSingleTest(): Return value must be of type ?array, none returned in /home/runner/work/Codeception/Codeception/src/Codeception/Command/Run.php:487`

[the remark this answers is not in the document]
TavoNiievez: However, I had to add `return null`; explicitly, because otherwise, CI jobs would fail:
`
PHP Fatal error:  Uncaught TypeError: Codeception\Command\Run::matchSingleTest(): Return value must be of type ?array, none returned in /home/runner/work/Codeception/Codeception/src/Codeception/Command/Run.php:487
`

src/Codeception/Command/SelfUpdate.php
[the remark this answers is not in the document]
Naktibalda: Thanks

src/Codeception/Command/Run.php
```diff
@@ -498,7 +525,7 @@ protected function currentNamespace()
         return $config['namespace'];
     }
 
-    protected function runSuites($suites, $skippedSuites = [])
+    protected function runSuites($suites, $skippedSuites = []): int
```
Naktibalda, line 528: Both parameters look like arrays to me.

src/Codeception/Command/Shared/FileSystem.php
```diff
@@ -37,13 +49,13 @@ protected function completeSuffix($filename, $suffix)
...
 
-    protected function removeSuffix($classname, $suffix)
+    protected function removeSuffix(string $classname, string $suffix): string
     {
         $classname = preg_replace('~\.php$~', '', $classname);
         return preg_replace("~$suffix$~", '', $classname);
     }
 
-    protected function createFile($filename, $contents, $force = false, $flags = null)
+    protected function createFile(string $filename, $contents, bool $force = false, int $flags = 0): bool
```
Naktibalda, line 58: `$contents` looks like string to me.

Naktibalda (changes requested):
I have no objections to ading Trait suffix to traits.

Naktibalda (approved):
Looks good to me.

PR #6079, opened 2021-01-10
suggestion on missing modules

c33s:
had the problem that i wasn't able to connect to a module, which i included in the config. took quite a while to debug. it was a simple diff in the case `LockPick ` vs `Lockpick`

so i added some code to add a  suggestion which module could be meant to the error message. 

```
[ModuleException] Codeception\Module: Module \C33s\Codeception\Module\LockPick couldn't be connected 
```
vs
```
[ModuleException] Codeception\Module: Module \C33s\Codeception\Module\LockPick couldn't be connected (did you mean '\C33s\Codeception\Module\Lockpick'?)
```

have not adapted any tests yet, wanted to wait for comments if you would merge this before.

### DIFF
--- a/src/Codeception/Lib/ModuleContainer.php
+++ b/src/Codeception/Lib/ModuleContainer.php
@@ -289,7 +289,7 @@ class ModuleContainer
     public function throwMissingModuleExceptionWithSuggestion($className, $moduleName)
     {
         $suggestedModuleNameInfo = $this->getModuleSuggestion($moduleName);
-        throw new ModuleException($className, "Module $moduleName couldn't be connected$suggestedModuleNameInfo");
+        throw new ModuleException($className, "Module $moduleName couldn't be connected" . $suggestedModuleNameInfo);
     }
 
     protected function getModuleSuggestion($missingModuleName)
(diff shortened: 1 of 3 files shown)


DavertMik:
This looks really nice to me! Thank you :+1: 
This UX fix will save my time as well  😃

Yes, please add some tests into ModuleContainerTest

c33s:
@DavertMik ready to merge. 
can you squash the commits via the github interface or do i have to do it (only used to gitlab, there it's possible to squash on merge)?

src/Codeception/Lib/ModuleContainer.php
```diff
@@ -275,12 +280,37 @@ public function hasModule($moduleName)
...
+            $this->throwMissingModuleExceptionWithSuggestion(__CLASS__, $moduleName);
         }
 
         return $this->modules[$moduleName];
     }
 
+    public function throwMissingModuleExceptionWithSuggestion($className, $moduleName)
+    {
+        $suggestedModuleNameInfo = $this->getModuleSuggestion($moduleName);
+        throw new ModuleException($className, "Module $moduleName couldn't be connected$suggestedModuleNameInfo");
```
Naktibalda, line 292: I really don't like that there is no space in `connected$suggestedModuleNameInfo`.
String concatenation would look better.

PR #6082, opened 2021-01-11
Removing `--debug` and `--silent`

ThomasLandauer:
https://codeception.com/docs/reference/Commands is generated from here, right?

Follow up of https://github.com/Codeception/Codeception/discussions/6081

* Since `-v/vv/vvv` can't be removed (Symfony!), I'd like to drop `--debug`. The clean way would be to raise a deprecation note, but I don't know how to do that ;-) So I'm just removing it from the docs for now.
* Since `--quiet` is the Symfony standard, I'd like to drop `--silent` too.


DavertMik:
let's keep debug )

PR #6085, opened 2021-01-12
Added nunomaduro/collision for better stack traces

DavertMik:
Tired of boring stack traces? 
Let's add some code in it!

Powered by nunomaduro/collision

image

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,8 @@
         "symfony/yaml": ">=4.4 <6.0",
         "symfony/css-selector": ">=4.4 <6.0",
         "behat/gherkin": "^4.6.2",
-        "codeception/stub": "^3.7"
+        "codeception/stub": "^3.7",
+        "nunomaduro/collision": "^5.1"
     },
     "require-dev": {
         "vlucas/phpdotenv": "^4.0 | ^5.0",
--- a/src/Codeception/Subscriber/Console.php
+++ b/src/Codeception/Subscriber/Console.php
@@ -22,6 +22,7 @@ use Codeception\Test\Descriptor;
 use Codeception\Test\Interfaces\ScenarioDriven;
 use Codeception\TestInterface;
 use Codeception\Util\Debug;
+use NunoMaduro\Collision\Writer;
 use PHPUnit\Framework\AssertionFailedError;
 use PHPUnit\Framework\ExceptionWrapper;
 use PHPUnit\Framework\ExpectationFailedException;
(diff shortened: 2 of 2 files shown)


Naktibalda:
Is this screenshot real?
Arrow points to wrong line and stack trace doesn't match.

Does it work with Cept, Cest and Gherkin formats?

DavertMik:
@Naktibalda screenshot is real

1. arrow point highlights a failed command, this one is a multiline one, so it is showing on the second line on the assertion. I think everything is correct here.
2. for Gherkin it is not shown. This is shown only in places where we had printed exceptions, so it is just prepended to exception trace.

Naktibalda:
> P.S. Why all the jobs are failing?

I think that because of your change. Most of failing tests test errors.

Naktibalda:
I rebased this branch too.

DavertMik:
Thanks!

Naktibalda:
@DavertMik I tested this change and found a bug.

It was failing with `preg_match(): No ending delimiter '/' found at ../../../vendor/nunomaduro/collision/src/Writer.php:208` because the last / in the first pattern was escaped.

I fixed it in this branch and this feature is working.
Though there is a smaller issue that it doesn't respect `colors: false` in codeception.yaml

image

Naktibalda:
It is a known issue of Collision library: https://github.com/nunomaduro/collision/issues/63

Naktibalda (approved):

Naktibalda (changes requested):
Maybe we should disable this feature when `$this->output->getFormatter()->isDecorated()` returns false?

`Failed asserting that true is false.` was displayed twice, there is no way to disable dispaying this message (description) in Collision library, so we have to remove it from Codeception code.

composer.json
```diff
@@ -28,7 +28,8 @@
         "symfony/yaml": ">=4.4 <6.0",
         "symfony/css-selector": ">=4.4 <6.0",
         "behat/gherkin": "^4.6.2",
-        "codeception/stub": "^3.7"
+        "codeception/stub": "^3.7",
+        "nunomaduro/collision": "^5.1"
```
TavoNiievez, line 32: Codeception 5 supports Symfony 4.4 as a minimum, but this collision release requires `symfony/console ^5.0` as a minimum.

TavoNiievez (changes requested):

composer.json
Naktibalda: This is a very good point. In order to support Symfony 4.4 we would have to use nunomaduro/collision v3, which is [not supported anymore](https://github.com/nunomaduro/collision/pull/173).
I think that it is more important to keep support for Symfony 4.4.

Naktibalda:
I raised pull request so make nunomaduro/collision support Symfony 4.4

https://github.com/nunomaduro/collision/pull/179

Naktibalda:
Using unmaintained old version of dependency for new minor feature is a bad idea, this feature is not important enough.

PR #6092, opened 2021-01-18
Update code standards in Coverage and Event

TavoNiievez:
- Use strict types in `src/Codeception/Coverage` and `src/Codeception/Event`.
- Add type hints to Coverage Subscribers and Events.
- Longer aliases for ambiguous class names.
- Fix naming in `src/Codeception/Coverage` and `src/Codeception/Event`.

### DIFF
--- a/src/Codeception/Coverage/Filter.php
+++ b/src/Codeception/Coverage/Filter.php
@@ -1,4 +1,7 @@
 <?php
+
+declare(strict_types=1);
+
 namespace Codeception\Coverage;
 
 use Codeception\Configuration;
--- a/src/Codeception/Coverage/Filter.php
+++ b/src/Codeception/Coverage/Filter.php
@@ -137,7 +137,7 @@ class Filter
         return $this;
     }
 
-    protected function matchWildcardPattern($pattern): Finder
+    protected function matchWildcardPattern(string $pattern): Finder
     {
         $finder = Finder::create();
         $fileOrDir = str_replace('\\', '/', $pattern);
--- a/src/Codeception/Coverage/Filter.php
+++ b/src/Codeception/Coverage/Filter.php
@@ -10,11 +10,19 @@ use Codeception\Exception\ModuleException;
 use SebastianBergmann\CodeCoverage\CodeCoverage;
 use Symfony\Component\Finder\Exception\DirectoryNotFoundException;
 use Symfony\Component\Finder\Finder;
+use function array_pop;
+use function count;
+use function explode;
+use function implode;
+use function is_array;
+use function method_exists;
+use function str_replace;
+use function strpos;
 
 class Filter
 {
     /**
-     * @var CodeCoverage
+     * @var CodeCoverage|null
      */
     protected $phpCodeCoverage = null;
 
@@ -24,7 +32,7 @@ class Filter
     protected static $c3;
 
     /**
-     * @var \SebastianBergmann\CodeCoverage\Filter
+     * @var \SebastianBergmann\CodeCoverage\Filter|null
      */
     protected $filter = null;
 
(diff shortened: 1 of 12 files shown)


src/Codeception/Coverage/Filter.php
```diff
@@ -131,29 +137,26 @@ public function blackList($config)
         return $this;
     }
 
-    protected function matchWildcardPattern($pattern)
+    protected function matchWildcardPattern($pattern): Finder
```
Naktibalda, line 140: `$pattern` is a string.

src/Codeception/Coverage/Subscriber/LocalServer.php
```diff
@@ -129,10 +179,10 @@ public function afterSuite(SuiteEvent $e)
     /**
      * Allows Translating Remote Paths To Local (IE: When Using Docker)
      *
-     * @param \SebastianBergmann\CodeCoverage\CodeCoverage $coverage
+     * @param CodeCoverage $coverage
      * @return $this
```
Naktibalda, line 183: `@return LocalServer` would be better, right?

src/Codeception/Coverage/SuiteSubscriber.php
```diff
@@ -26,31 +39,62 @@ abstract class SuiteSubscriber implements EventSubscriberInterface
...
+     * @var array
+     */
     protected $filters = [];
+    /**
+     * @var array
+     */
     protected $modules = [];
-
+    /**
+     * @var CodeCoverage
```
Naktibalda, line 55: In some other classes this variable woud be typed as `@var CodeCoverage|null`

src/Codeception/Coverage/SuiteSubscriber.php
```diff
@@ -26,31 +39,62 @@ abstract class SuiteSubscriber implements EventSubscriberInterface
...
+     */
+    protected function applySettings(array $settings): void
     {
         try {
             $this->coverage = PhpCodeCoverageFactory::build();
         } catch (CodeCoverageException $e) {
-            throw new \Exception(
-                'XDebug is required to collect CodeCoverage. Please install xdebug extension and enable it in php.ini'
+            throw new Exception(
+                'XDebug is required to collect CodeCoverage. Please install xdebug extension and enable it in php.ini',
```
Naktibalda, line 95: This message looks outdated, but lets deal with it later.

Naktibalda (changes requested):

src/Codeception/Coverage/Subscriber/LocalServer.php
TavoNiievez: **TL;DR**: Yes.

There are three options for this:
```php
return $this
return self
return LocalServer
```
The first is the auto-generated by phpStorm and it seems to be the worst of the 3 just for readability.
Now the key here is to just be consistent. If you're going to do it like this:
```php
     * @return LocalServer
     */
    protected function preProcessCoverage(CodeCoverage $coverage): self
```
it's okay, and if you're going to do it like this:
```php
     * @return self
     */
    protected function preProcessCoverage(CodeCoverage $coverage): LocalServer
```
it's ok as well. Only when choosing one, you should do it that way **everywhere**.

If we think ahead the best thing is that the name of the class is in the annotation exactly as you suggest. Since from php 8 `static` it is possible as a return data type. And it is better to switch between `self` and `static` than between class name and `static`.

src/Codeception/Coverage/SuiteSubscriber.php
TavoNiievez:
```php
    // src/Codeception/Coverage/Filter.php

    /**
     * @var CodeCoverage|null
     */
    protected $phpCodeCoverage = null;
```
However, the difference is that in `Filter` the variable is initialized to `null`.

I will change it anyway. Thanks

src/Codeception/Coverage/Subscriber/LocalServer.php
```diff
@@ -129,10 +179,10 @@ public function afterSuite(SuiteEvent $e)
     /**
      * Allows Translating Remote Paths To Local (IE: When Using Docker)
      *
-     * @param \SebastianBergmann\CodeCoverage\CodeCoverage $coverage
-     * @return $this
+     * @param CodeCoverage $coverage
+     * @return LocalServer
```
Naktibalda, line 183: Are `@param` and `@return` annotations in docblock even necessary when type hints provide the same information?

TavoNiievez: It depends. Tools like `robo` are able to infer the types and generate the documentation for parameters and return values ​​anyway? PHPStorm suggests adding them... but I don't think they are strictly necessary

Naktibalda: We don't generate documentation for this class.

Naktibalda (approved):

PR #6100, opened 2021-01-26
Update code standards in Steps

TavoNiievez:
- Use strict types in `src/Codeception/Step`.
- Add some type hints to Steps.

### DIFF
--- a/src/Codeception/Step/Comment.php
+++ b/src/Codeception/Step/Comment.php
@@ -1,37 +1,41 @@
 <?php
+
+declare(strict_types=1);
+
 namespace Codeception\Step;
 
 use Codeception\Lib\ModuleContainer;
 use Codeception\Step as CodeceptionStep;
+use function mb_strcut;
 
 class Comment extends CodeceptionStep
 {
-    public function __toString()
+    public function __toString(): string
     {
         return (string) $this->getAction();
     }
 
-    public function toString($maxLength)
+    public function toString($maxLength): string
     {
-        return mb_strcut($this->__toString(), 0, $maxLength, 'utf-8');
+        return mb_strcut((string) $this, 0, $maxLength, 'utf-8');
     }
 
-    public function getHtml($highlightColor = '#732E81')
+    public function getHtml($highlightColor = '#732E81'): string
     {
         return '<strong>' . $this->getAction() . '</strong>';
     }
 
-    public function getPhpCode($maxLength)
+    public function getPhpCode($maxLength): string
     {
         return '// ' . $this->getAction();
     }
 
-    public function run(ModuleContainer $container = null)
+    public function run(ModuleContainer $container = null): void
     {
         // don't do anything, let's rest
     }
 
-    public function getPrefix()
+    public function getPrefix(): string
     {
         return '';
     }
--- a/src/Codeception/Step/ConditionalAssertion.php
+++ b/src/Codeception/Step/ConditionalAssertion.php
@@ -24,23 +24,23 @@ class ConditionalAssertion extends Assertion implements GeneratedStep
         }
     }
 
-    public function getAction(): ?string
+    public function getAction(): string
     {
         $action = 'can' . ucfirst($this->action);
-        return preg_replace('/^canDont/', 'cant', $action);
+        return (string)preg_replace('/^canDont/', 'cant', $action);
     }
 
-    public function getHumanizedAction()
+    public function getHumanizedAction(): string
     {
         return $this->humanize($this->action . ' ' . $this->getHumanizedArguments());
     }
 
-    public static function getTemplate(Template $template)
+    public static function getTemplate(Template $template): ?Template
     {
         $action = $template->getVar('action');
 
         if ((0 !== strpos($action, 'see')) && (0 !== strpos($action, 'dontSee'))) {
-            return '';
+            return null;
         }
 
         $conditionalDoc = "* [!] Conditional Assertion: Test won't be stopped on fail\n     " . $template->getVar('doc');
@@ -57,7 +57,7 @@ class ConditionalAssertion extends Assertion implements GeneratedStep
             ->place('step', 'ConditionalAssertion');
     }
 
-    public function match($name): bool
+    public function match(string $name): bool
     {
         return 0 === strpos($name, 'see') || 0 === strpos($name, 'dontSee');
     }
--- a/src/Codeception/Step/ConditionalAssertion.php
+++ b/src/Codeception/Step/ConditionalAssertion.php
@@ -1,26 +1,33 @@
 <?php
+
+declare(strict_types=1);
+
 namespace Codeception\Step;
 
 use Codeception\Exception\ConditionalAssertionFailed;
 use Codeception\Lib\ModuleContainer;
 use Codeception\Util\Template;
+use PHPUnit\Framework\AssertionFailedError;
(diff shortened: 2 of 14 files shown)


src/Codeception/Step/ConditionalAssertion.php
```diff
@@ -1,26 +1,33 @@
...
         try {
             parent::run($container);
-        } catch (\PHPUnit\Framework\AssertionFailedError $e) {
+        } catch (AssertionFailedError $e) {
             throw new ConditionalAssertionFailed($e->getMessage(), $e->getCode(), $e);
         }
     }
 
-    public function getAction()
+    public function getAction(): ?string
```
Naktibalda, line 27: I think that probability of preg_replace failing is so low or even non-existent that the return type of this method should be `: string`

src/Codeception/Step/Comment.php
```diff
@@ -1,37 +1,41 @@
...
     }
 
-    public function getHtml($highlightColor = '#732E81')
+    public function getHtml($highlightColor = '#732E81'): string
     {
         return '<strong>' . $this->getAction() . '</strong>';
     }
 
-    public function getPhpCode($maxLength)
+    public function getPhpCode($maxLength): string
```
Naktibalda, line 28: $maxLength should be int, right?

src/Codeception/Step/Comment.php
```diff
@@ -1,37 +1,41 @@
...
 
-    public function toString($maxLength)
+    public function toString($maxLength): string
     {
-        return mb_strcut($this->__toString(), 0, $maxLength, 'utf-8');
+        return mb_strcut((string) $this, 0, $maxLength, 'utf-8');
     }
 
-    public function getHtml($highlightColor = '#732E81')
+    public function getHtml($highlightColor = '#732E81'): string
```
Naktibalda, line 23: It is odd that unused argument has default value.

src/Codeception/Step/ConditionalAssertion.php
```diff
@@ -1,26 +1,33 @@
...
     {
         $action = 'can' . ucfirst($this->action);
-        $action = preg_replace('/^canDont/', 'cant', $action);
-        return $action;
+        return preg_replace('/^canDont/', 'cant', $action);
     }
 
     public function getHumanizedAction()
```
Naktibalda, line 33: It returns string, right?

src/Codeception/Step/Executor.php
```diff
@@ -1,15 +1,21 @@
...
+use Closure;
 use Codeception\Step as CodeceptionStep;
 use Codeception\Lib\ModuleContainer;
 
 class Executor extends CodeceptionStep
 {
+    /**
+     * @var Closure|null
+     */
+    protected $callable;
```
Naktibalda, line 16: I think that this field can be typed as Closure, because it is assigned in constructor.

src/Codeception/Step/Meta.php
```diff
@@ -1,27 +1,35 @@
...
 
 class Meta extends CodeceptionStep
 {
-    public function run(ModuleContainer $container = null)
+    public function run(ModuleContainer $container = null): void
     {
     }
 
-    public function setTraceInfo($file, $line)
+    public function setTraceInfo($file, $line): void
```
Naktibalda, line 21: string, int

src/Codeception/Step/Meta.php
```diff
@@ -1,27 +1,35 @@
...
     {
         $this->file = $file;
         $this->line = $line;
     }
 
-    public function setPrefix($actor)
+    public function setPrefix($actor): void
```
Naktibalda, line 27: I think that $actor is a string here.

src/Codeception/Step/Meta.php
```diff
@@ -1,27 +1,35 @@
...
     {
         $this->prefix = $actor;
     }
 
-    public function getArgumentsAsString($maxLength = 200)
+    public function getArgumentsAsString($maxLength = 200): string
```
Naktibalda, line 32: int

src/Codeception/Step/ConditionalAssertion.php
```diff
@@ -50,7 +57,7 @@ public static function getTemplate(Template $template)
             ->place('step', 'ConditionalAssertion');
     }
```
Naktibalda, line 58: `getTemplate` above returns `Template` or empty string.

src/Codeception/Step/ConditionalAssertion.php
```diff
@@ -50,7 +57,7 @@ public static function getTemplate(Template $template)
...
 
-    public function match($name)
+    public function match($name): bool
```
Naktibalda, line 60: string

src/Codeception/Step/Meta.php
```diff
@@ -35,7 +43,7 @@ public function getArgumentsAsString($maxLength = 200)
         return $result;
     }
 
-    public function setFailed($failed)
+    public function setFailed($failed): void
```
Naktibalda, line 46: I have no time to check what $failed is

src/Codeception/Step/TryTo.php
```diff
@@ -1,18 +1,24 @@
...
+    public function run(ModuleContainer $container = null): bool
     {
         $this->isTry = true;
         try {
             parent::run($container);
-        } catch (\Exception $e) {
+        } catch (Exception $e) {
             codecept_debug("Failed to perform: {$e->getMessage()}, skipping...");
             return false;
         }
```
Naktibalda, line 24: `getTemplate` method below returns either `Template` or null.
Please check where `getTemplate` is used and make it more consistent.

Naktibalda (changes requested):

src/Codeception/Step/Comment.php
TavoNiievez: Yes, what happens is that to add this type-hint I should also change the abstract class as well, and I didn't decide to touch those files yet (`src/Codeception/*.php`). I'll do it here anyway.

TavoNiievez:
Some of the requested changes require changes in abstract classes or interfaces, as well as inferring the use of certain methods in the tests. Also, CI jobs randomly fail with exact commits that previously had correct builds, sometimes only on certain versions of PHP or sometimes on all. Given the time consuming waiting for tests and/or re-running them to make/undo changes in more classes than expected, I am converting this into a draft until I can find some time for this.

src/Codeception/Step/ConditionalAssertion.php
Naktibalda: It is not acceptable for this preg_replace to fail, so if you think that it can return null, please check if result is null and throw exception.

src/Codeception/Step/Comment.php
TavoNiievez: some tests assume that this default value exists

src/Codeception/Step/Meta.php
TavoNiievez: I will make this change in a future PR when I edit the files in the path: `src/Codeception/*.php`

src/Codeception/Step/Comment.php
TavoNiievez: I will make this change in a future PR when I edit the files in the path: `src/Codeception/*.php`

Naktibalda (approved):

PR #6106, opened 2021-02-02
Remove old version conditionals

TavoNiievez:
- Conditionals of PHP and PHPUnit versions not supported in Codeception 5 are removed.
- Codeception version constant is updated to `5.0.0`.


src/Codeception/Test/Loader/Unit.php
```diff
@@ -43,13 +43,7 @@ public function getTests()
...
     protected function createTestFromPhpUnitMethod(\ReflectionClass $class, \ReflectionMethod $method)
     {
-        if (method_exists(\PHPUnit\Framework\TestSuite::class, 'isTestMethod')) {
-            //PHPUnit <8.2
-            if (!\PHPUnit\Framework\TestSuite::isTestMethod($method)) {
-                return;
-            }
-            $test = \PHPUnit\Framework\TestSuite::createTest($class, $method->name);
-        } elseif (method_exists(\PHPUnit\Util\Test::class, 'isTestMethod')) {
+        if (method_exists(\PHPUnit\Util\Test::class, 'isTestMethod')) {
```
Naktibalda, line 46: This condition is not necessary, `\PHPUnit\Util\Test::isTestMethod` exists in all supported versions of PHPUnit.

src/Codeception/Test/Loader/Unit.php
```diff
@@ -43,13 +43,7 @@ public function getTests()
...
             //PHPUnit >=8.2
```
Naktibalda, line 47: Comment isn't necessary too.

tests/unit/Codeception/Util/ReflectionTestClass.php
```diff
@@ -1,8 +1,4 @@
 <?php
 namespace Codeception\Util;
 
-if (PHP_VERSION_ID < 70000) {
-    require_once 'ReflectionTestClass56.php';
-} else {
-    require_once 'ReflectionTestClass70.php';
-}
\ No newline at end of file
+require_once 'ReflectionTestClass70.php';
```
Naktibalda, line 4: Please rename ReflectionTestClass70.php to ReflectionTestClass.php

Naktibalda (changes requested):

src/Codeception/Test/Loader/Unit.php
TavoNiievez: Yes, I thought I would leave it like that in case it disappears in the future.
But if you don't see it necessary, I'll remove it, the less the better.

src/Codeception/Lib/Parser.php
```diff
@@ -119,9 +119,6 @@ public static function validate($file)
 
     public static function load($file)
     {
-        if (PHP_MAJOR_VERSION < 7) {
-            self::validate($file);
-        }
```
Naktibalda, line 124: I think that validate method can be deleted now, the only other place where it is used is a [Cept Loader](https://github.com/Codeception/Codeception/blob/c26038596f7f35cc684b380f6e3fe186331e7820/src/Codeception/Test/Loader/Cept.php#L18), but Cept Format [handles ParseError](https://github.com/Codeception/Codeception/blob/c26038596f7f35cc684b380f6e3fe186331e7820/src/Codeception/Test/Cept.php#L41-L44).

TavoNiievez: thanks

Naktibalda (approved):

PR #6107, opened 2021-02-02
Update code standards in Subscribers

TavoNiievez:
- [x] Use strict types in `src/Codeception/Subscriber`.
- [x] Add type hints to Subscribers.
- [x] Remove unnecessary qualifiers in `src/Codeception/Subscriber`.
- [x] Add 'use function' statements for performance reasons.
- [x] Fix naming in `src/Codeception/Subscriber`.
- [x] Rename `src/Codeception/Subscriber/Shared/StaticEvent` trait.

### DIFF
--- a/src/Codeception/Subscriber/BeforeAfterTest.php
+++ b/src/Codeception/Subscriber/BeforeAfterTest.php
@@ -59,7 +59,7 @@ class BeforeAfterTest implements EventSubscriberInterface
         $this->runHooks('afterClass');
     }
 
-    protected function runHooks($hookName): void
+    protected function runHooks(string $hookName): void
     {
         foreach ($this->hooks as $className => $hook) {
             foreach ($hook[$hookName] as $method) {
--- a/src/Codeception/Subscriber/BeforeAfterTest.php
+++ b/src/Codeception/Subscriber/BeforeAfterTest.php
@@ -22,7 +22,7 @@ class BeforeAfterTest implements EventSubscriberInterface
     /**
      * @var array<string, string|int[]|string[]>
      */
-    public static $events = [
+    protected static $events = [
         Events::SUITE_BEFORE => 'beforeClass',
         Events::SUITE_AFTER  => ['afterClass', 100]
     ];
--- a/src/Codeception/Subscriber/BeforeAfterTest.php
+++ b/src/Codeception/Subscriber/BeforeAfterTest.php
@@ -3,25 +3,43 @@ namespace Codeception\Subscriber;
 
 use Codeception\Event\SuiteEvent;
 use Codeception\Events;
+use PHPUnit\Framework\Test;
+use Rector\Core\Tests\Issues\Issue4191\DoNotDuplicateComment\Fixture\a;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use function call_user_func;
+use function get_class;
+use function is_callable;
+use function strstr;
 
 class BeforeAfterTest implements EventSubscriberInterface
 {
     use Shared\StaticEvents;
 
+    /**
+     * @var array<string, string|int[]|string[]>
+     */
     public static $events = [
         Events::SUITE_BEFORE => 'beforeClass',
         Events::SUITE_AFTER  => ['afterClass', 100]
     ];
 
+    /**
+     * @var array
+     */
     protected $hooks = [];
+    /**
+     * @var array
+     */
     protected $startedTests = [];
+    /**
+     * @var array
+     */
     protected $unsuccessfulTests = [];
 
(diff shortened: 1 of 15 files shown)


src/Codeception/Subscriber/BeforeAfterTest.php
```diff
@@ -1,45 +1,66 @@
...
+use function is_callable;
+use function strstr;
 
 class BeforeAfterTest implements EventSubscriberInterface
 {
-    use Shared\StaticEvents;
+    use Shared\StaticEventsTrait;
 
+    /**
+     * @var array<string, string|int[]|string[]>
```
Naktibalda, line 23: I don't think that this signature matches `['afterClass', 100]`.
Also declaring different type of the same property in different classes doesn't look right to me.

According to [docblock in symfony/event-dispatcher](https://github.com/symfony/event-dispatcher/blob/v5.2.3/EventSubscriberInterface.php#L39-L41), events can be defined in 3 formats:

```
     *  * ['eventName' => 'methodName']
     *  * ['eventName' => ['methodName', $priority]]
     *  * ['eventName' => [['methodName1', $priority], ['methodName2']]]
```

Must $events property be public? I didn't find any code accessing it directly - only getSubscribedEvents method in StaticEventsTrait.

src/Codeception/Subscriber/Console.php
```diff
@@ -564,7 +644,7 @@ public function detectWidth()
             } else {
```
Naktibalda, line 644: I think that return type can be set to : int after casting result of getenv('COLUMNS')

src/Codeception/Subscriber/Console.php
```diff
@@ -576,18 +656,14 @@ public function detectWidth()
         return $this->width;
     }
 
-    private function isWin()
+    private function isWin(): bool
     {
         return strtoupper(substr(PHP_OS, 0, 3)) === 'WIN';
```
Naktibalda, line 661: I had to use this logic in Run command :)

src/Codeception/Subscriber/GracefulTermination.php
```diff
@@ -1,22 +1,35 @@
...
 
     /**
      * @var SuiteEvent
      */
     protected $suiteEvent;
 
-    public function handleSuite(SuiteEvent $event)
+    public function handleSuite(SuiteEvent $event): void
     {
         if (PHP_MAJOR_VERSION === 7 && PHP_MINOR_VERSION === 0) {
```
Naktibalda, line 34: Wow, it is still here, another piece of code to remove.

Naktibalda (changes requested):

src/Codeception/Subscriber/GracefulTermination.php
TavoNiievez: It was removed in the other PR, you have not merged it yet

src/Codeception/Subscriber/Console.php
TavoNiievez: if there was a dedicated class for this it would be nice, better than repeating that from `strtoupper(substr(...` in various places.

src/Codeception/Subscriber/BeforeAfterTest.php
TavoNiievez: > I don't think that this signature matches ['afterClass', 100].

It was the only one that psalm was happy with, it also raised doubts for me, but the other option is to simply leave it as an array. Nested `array<>` syntax is not well interpreted by some static code analyzers.

> Must $events property be public?

I don't think so. I'll change it to `protected`.

TavoNiievez: In the other classes I left it as: `@var array<string, string>` but it should probably be the same everywhere. So I'll just leave it as you tell me.

src/Codeception/Subscriber/Console.php
Naktibalda: Worse, it is [different](https://github.com/Codeception/Codeception/commit/4d62088152e5459e977a941e0d6a881fe61dd5f9).

[PathResolver](https://github.com/Codeception/Codeception/blob/5.0/src/Codeception/Util/PathResolver.php#L130-L142) has `isWindows` method which only checks DIRECTORY_SEPARATOR and `isPathAbsolute` method which uses regexp pattern for Windows paths.

[Coverage\Subscriber\Printer](https://github.com/Codeception/Codeception/blob/5.0/src/Codeception/Coverage/Subscriber/Printer.php) has its own `absolutePath` method which checks if the first character is `/` or the second character is `:`.
I discovered these methods through your PRs.

Naktibalda (approved):

PR #6110, opened 2021-02-05
Update code standards in Extensions

TavoNiievez:
- Use strict types in `src/ext`.
- Add some type hints to Extensions.

### DIFF
--- a/ext/DotReporter.php
+++ b/ext/DotReporter.php
@@ -43,14 +43,24 @@ class DotReporter extends Extension
      * @var Console
      */
     protected $standardReporter;
-
+    /**
+     * @var array
+     */
     protected $errors = [];
+    /**
+     * @var array
+     */
     protected $failures = [];
-
+    /**
+     * @var int
+     */
     protected $width = 10;
+    /**
+     * @var int
+     */
     protected $currentPos = 0;
 
-    public function _initialize()
+    public function _initialize(): void
     {
         $this->options['silent'] = false; // turn on printing for this extension
         $this->_reconfigure(['settings' => ['silent' => true]]); // turn off printing for everything else
@@ -58,7 +68,11 @@ class DotReporter extends Extension
         $this->width = $this->standardReporter->detectWidth();
     }
 
-    // we are listening for events
+    /**
+     * We are listening for events
+     *
+     * @var array<string, string>
+     */
     public static $events = [
         Events::SUITE_BEFORE => 'beforeSuite',
         Events::TEST_SUCCESS => 'success',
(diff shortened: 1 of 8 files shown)


ext/DotReporter.php
```diff
@@ -68,42 +85,42 @@ public function _initialize()
...
 
-    public function skipped()
+    public function skipped(): void
     {
         $this->printChar('S');
     }
-    
-    protected function printChar($char)
+
+    protected function printChar($char): void
```
Naktibalda, line 113: parameter is a string.

ext/Logger.php
```diff
@@ -59,12 +69,15 @@ class Logger extends Extension
 
     protected $path;
```
Naktibalda, line 70: $path is string, $logHandler above is RotatingFileHandler

ext/Recorder.php
```diff
@@ -617,19 +630,18 @@ protected function isStepIgnored(StepEvent $e)
...
     {
-        return basename($e->getTest()->getMetadata()->getFilename()) . '_' . preg_replace('/[^A-Za-z0-9\-\_]/', '_', $e->getTest()->getMetadata()->getName());
+        return basename($event->getTest()->getMetadata()->getFilename()) . '_' . preg_replace('/[^A-Za-z0-9\-\_]/', '_', $event->getTest()->getMetadata()->getName());
     }
 
     /**
      * @param string $message
      */
-    protected function writeln($message)
+    protected function writeln($message): void
```
Naktibalda, line 644: Parameter is a string

Naktibalda (changes requested):

ext/Recorder.php
TavoNiievez: Abstract class `src/Codeception/Extension.php` needs to be modified, I will do it in a future PR.

TavoNiievez: I'm waiting for https://github.com/Codeception/Codeception/pull/6071 to be merged so as not to generate conflicts.

Naktibalda: Then take a break, we have 2 months.

TavoNiievez: At the moment I think I can move forward in `src/PHPUnit` but pending folders or files in `src/Codeception` are being occupied by **DavertMik**.

TavoNiievez: Of course, I will only have one PR of this type open at a time, only when you merge one will I continue with another.

Naktibalda: Then don't worry about conflicts, I will rebase other branches after merging this one.

Majority of code in src/PHPUnit was copied from PHPUnit. It would be the best to keep differences from original minimal and refresh them with changes implemented by PHPUnit.
Please update files in src/PHPUnit/Constraint/ directory, /src/PHPUnit/Util/Log/ directory, ConsolePrinter and FilterTest only.

TavoNiievez: I understood, I will, thanks

TavoNiievez: I also fixed this

src/Codeception/Extension.php
```diff
@@ -82,7 +82,7 @@ protected function write($message)
         }
     }
 
-    protected function writeln($message)
```
Naktibalda, line 85: Could you add types to all methods here?

src/Codeception/Extension.php
```diff
@@ -49,33 +69,32 @@ public static function getSubscribedEvents()
...
      */
-    public function _initialize()
+    public function _initialize(): void
     {
         $this->_reconfigure(); // hook for BC only.
     }
 
-    protected function write($message)
+    /**
+     * @param string|iterable $message
```
TavoNiievez, line 95:
```php
	# Symfony\Console\Output\OutputInterface

    /**
     * Writes a message to the output.
     *
     * @param string|iterable $messages The message as an iterable of strings or a single string
     * @param bool            $newline  Whether to add a newline
     * @param int             $options  A bitmask of options (one of the OUTPUT or VERBOSITY constants), 0 is considered the same as self::OUTPUT_NORMAL | self::VERBOSITY_NORMAL
     */
    public function write($messages, bool $newline = false, int $options = 0);
```

PR #6111, opened 2021-02-06
5.0 Fix name of IncompleteTestError exception

Naktibalda:
I overlooked this change in #6107 but noticed it while rebasing #6085


TavoNiievez:
@Naktibalda Did you do a search for other uses of `PHPUnit_` / `Framework_` ?

Naktibalda:
I searched for Framework_ before and for PHPUnit_ now and there were no matches.

TavoNiievez (approved):

PR #6112, opened 2021-02-06
Update code standards to some src/PHPUnit classes

TavoNiievez:
- Use strict types in `src/PHPUnit/ConsolePrinter`, `src/PHPUnit/Constraint`, `src/PHPUnit/FilterTest`, `src/PHPUnit/Util/Log`.
- Add some type hints to `src/PHPUnit`.

### DIFF
--- a/src/PHPUnit/Constraint/Crawler.php
+++ b/src/PHPUnit/Constraint/Crawler.php
@@ -6,6 +6,7 @@ namespace Codeception\PHPUnit\Constraint;
 
 use Codeception\Exception\ElementNotFound;
 use Codeception\Lib\Console\Message;
+use DOMElement;
 use Facebook\WebDriver\WebDriverBy;
 use PHPUnit\Framework\ExpectationFailedException;
 use SebastianBergmann\Comparator\ComparisonFailure;
@@ -14,9 +15,12 @@ use function strpos;
 
 class Crawler extends Page
 {
-    protected function matches($nodes) : bool
+    /**
+     * @param SymfonyDomCrawler $nodes
+     * @return bool
+     */
+    protected function matches($nodes): bool
     {
-        /** @var SymfonyDomCrawler $nodes **/
         if (!$nodes->count()) {
             return false;
         }
@@ -33,13 +37,12 @@ class Crawler extends Page
     }
 
     /**
-     * @param mixed $nodes
-     * @param string|array|WebDriverBy $selector
+     * @param SymfonyDomCrawler $nodes
+     * @param string $selector
      * @param ComparisonFailure|null $comparisonFailure
      */
-    protected function fail($nodes, $selector, ComparisonFailure $comparisonFailure = null):void
+    protected function fail($nodes, $selector, ComparisonFailure $comparisonFailure = null): void
     {
-        /** @var SymfonyDomCrawler $nodes **/
         if (!$nodes->count()) {
             throw new ElementNotFound($selector, 'Element located either by name, CSS or XPath');
         }
@@ -62,7 +65,11 @@ class Crawler extends Page
         );
     }
 
-    protected function failureDescription($other) : string
+    /**
+     * @param DOMElement[] $other
+     * @return string
+     */
+    protected function failureDescription($other): string
     {
         $description = '';
         foreach ($other as $o) {
@@ -71,11 +78,11 @@ class Crawler extends Page
         return $description;
     }
 
-    protected function nodesList(SymfonyDomCrawler $domCrawler, $contains = null): string
+    protected function nodesList(SymfonyDomCrawler $domCrawler, string $contains = null): string
     {
         $output = '';
         foreach ($domCrawler as $node) {
-            if ($contains && strpos($node->nodeValue, (string) $contains) === false) {
+            if ($contains && strpos($node->nodeValue, $contains) === false) {
                 continue;
             }
             $output .= "\n+ " . $node->C14N();
--- a/src/PHPUnit/Constraint/Crawler.php
+++ b/src/PHPUnit/Constraint/Crawler.php
@@ -1,16 +1,22 @@
 <?php
+
+declare(strict_types=1);
+
 namespace Codeception\PHPUnit\Constraint;
 
 use Codeception\Exception\ElementNotFound;
 use Codeception\Lib\Console\Message;
-use Symfony\Component\DomCrawler\Crawler as DomCrawler;
+use Facebook\WebDriver\WebDriverBy;
+use PHPUnit\Framework\ExpectationFailedException;
 use SebastianBergmann\Comparator\ComparisonFailure;
+use Symfony\Component\DomCrawler\Crawler as SymfonyDomCrawler;
+use function strpos;
 
 class Crawler extends Page
 {
     protected function matches($nodes) : bool
     {
-        /** @var $nodes DomCrawler  * */
+        /** @var SymfonyDomCrawler $nodes **/
         if (!$nodes->count()) {
             return false;
         }
@@ -26,14 +32,19 @@ class Crawler extends Page
         return false;
     }
 
+    /**
+     * @param mixed $nodes
+     * @param string|array|WebDriverBy $selector
+     * @param ComparisonFailure|null $comparisonFailure
+     */
     protected function fail($nodes, $selector, ComparisonFailure $comparisonFailure = null):void
     {
-        /** @var $nodes DomCrawler  * */
+        /** @var SymfonyDomCrawler $nodes **/
         if (!$nodes->count()) {
             throw new ElementNotFound($selector, 'Element located either by name, CSS or XPath');
         }
 
-        $output = "Failed asserting that any element by '$selector'";
+        $output = "Failed asserting that any element by '{$selector}'";
         $output .= $this->uriMessage('on page');
         $output .= " ";
 
@@ -45,7 +56,7 @@ class Crawler extends Page
         }
         $output .= "\ncontains text '{$this->string}'";
 
-        throw new \PHPUnit\Framework\ExpectationFailedException(
+        throw new ExpectationFailedException(
             $output,
             $comparisonFailure
         );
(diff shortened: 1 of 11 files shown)


src/PHPUnit/Constraint/Crawler.php
```diff
@@ -1,16 +1,22 @@
...
+use Symfony\Component\DomCrawler\Crawler as SymfonyDomCrawler;
+use function strpos;
 
 class Crawler extends Page
 {
-    protected function matches($nodes) : bool
+    protected function matches($nodes): bool
     {
-        /** @var $nodes DomCrawler  * */
+        /** @var SymfonyDomCrawler $nodes **/
```
Naktibalda, line 19: Can you set parameter type in method signature?
The same in other methods.

src/PHPUnit/Constraint/Crawler.php
```diff
@@ -45,26 +56,30 @@ protected function fail($nodes, $selector, ComparisonFailure $comparisonFailure
...
         foreach ($other as $o) {
-            $desc .= parent::failureDescription($o->textContent);
+            $description .= parent::failureDescription($o->textContent);
         }
-        return $desc;
+        return $description;
     }
 
-    protected function nodesList(DomCrawler $nodes, $contains = null)
+    protected function nodesList(SymfonyDomCrawler $domCrawler, $contains = null): string
```
Naktibalda, line 78: In the only place where `nodesList` method is used, it is called with second parameter `$this->string` and $contains is only used as string so parameter can have `string` type too.

src/PHPUnit/Constraint/WebDriver.php
```diff
@@ -63,14 +75,14 @@ protected function failureDescription($nodes) : string
         return $desc;
     }
 
-    protected function nodesList($nodes, $contains = null)
+    protected function nodesList($nodes, $contains = null): string
```
Naktibalda, line 78: $contains is string.

src/PHPUnit/Util/Log/JSON.php
```diff
@@ -30,17 +42,17 @@ class JSON extends \PHPUnit\Util\Printer implements \PHPUnit\Framework\TestListe
     /**
      * An error occurred.
      *
-     * @param \PHPUnit\Framework\Test $test
-     * @param \Throwable $e
+     * @param TestCase|string $test
```
Naktibalda, line 45: Where did the string come from?
Nobody has every raised such issue, I think that this change is mistake.

src/PHPUnit/Constraint/Page.php
```diff
@@ -22,36 +40,30 @@ public function __construct($string, $uri = '')
...
+    public function toString(): string
     {
         return sprintf(
             'contains "%s"',
             $this->string
         );
     }
 
-    protected function failureDescription($pageContent) : string
+    protected function failureDescription($pageContent): string
```
Naktibalda, line 66: parameter is a string.

src/PHPUnit/Constraint/WebDriver.php
```diff
@@ -1,25 +1,32 @@
...
 use SebastianBergmann\Comparator\ComparisonFailure;
+use function count;
+use function htmlspecialchars_decode;
+use function strpos;
 
 class WebDriver extends Page
 {
-
-    protected function matches($nodes) : bool
+    protected function matches($nodes): bool
```
Naktibalda, line 19: Constraint logic is complicated and requires testing, but it seems that $nodes in WebDriver constraints is an array of WebDriverElements

src/PHPUnit/Constraint/Page.php
```diff
@@ -22,36 +40,30 @@ public function __construct($string, $uri = '')
      *
      * @return bool
      */
-    protected function matches($other) : bool
+    protected function matches($other): bool
```
Naktibalda, line 43: $other is a string.

src/PHPUnit/Constraint/Crawler.php
```diff
@@ -45,26 +56,30 @@ protected function fail($nodes, $selector, ComparisonFailure $comparisonFailure
...
             $comparisonFailure
         );
     }
 
-    protected function failureDescription($other) : string
+    /**
+     * @param mixed $other
+     * @return string
+     */
+    protected function failureDescription($other): string
```
Naktibalda, line 69: There is very specific `$o->textContent` in this method, so I guess that $other is an array of DOMElement

Naktibalda (changes requested):

src/PHPUnit/Util/Log/JSON.php
TavoNiievez: In PHPUnit 8/9 the interfaces inside `src/Runner/Hook/*.php` type these parameters as strings, I found it too strange to completely overwrite that signature. Do I leave it as it was before?

Naktibalda: I think that this is related to TestListener deprecation in PHPUnit 8/9.
Codeception hasn't migrated from TestListener to TestHooks.
TestListener is no longer deprecated in PHPUnit 10, and TestHooks are deprecated now: https://github.com/sebastianbergmann/phpunit/issues/4595
There is no need to support them.

TavoNiievez: @Naktibalda OK, I wasn't aware of that. Thank you. I re-implemented the TestListener interface and made the corresponding changes.

src/PHPUnit/Constraint/Crawler.php
```diff
@@ -1,16 +1,26 @@
...
+use function strpos;
 
 class Crawler extends Page
 {
-    protected function matches($nodes) : bool
+    /**
+     * @param SymfonyDomCrawler $nodes
+     * @return bool
+     */
+    protected function matches($nodes): bool
```
Naktibalda, line 22: was there an issue that prevented setting type `SymfonyDomCrawler` in method signature?

src/PHPUnit/Constraint/Crawler.php
```diff
@@ -26,14 +36,18 @@ protected function matches($nodes) : bool
         return false;
     }
 
-    protected function fail($nodes, $selector, ComparisonFailure $comparisonFailure = null):void
+    /**
+     * @param SymfonyDomCrawler $nodes
+     * @param string|array|WebDriverBy $selector
```
Naktibalda, line 41: Based on usage below, $selector can only be string.
This constraint isn't used by WebDriver module.

src/PHPUnit/Util/Log/JSON.php
```diff
@@ -30,17 +46,17 @@ class JSON extends \PHPUnit\Util\Printer implements \PHPUnit\Framework\TestListe
     /**
      * An error occurred.
      *
-     * @param \PHPUnit\Framework\Test $test
-     * @param \Throwable $e
+     * @param TestCase|PHPUnitTest $test
```
Naktibalda, line 49: PHPUnit\Framework\TestCase implements PHPUnit\Framework\Test interface, so it isn't necessary to document TestCase separately.
I think that @param annotation can be removed.

Naktibalda (changes requested):
My comments about parameters in Page constraint were resolved without making any change or explanation, so I unresolved them.

src/PHPUnit/Util/Log/JSON.php
TavoNiievez: `writeCase` method below is typed to receive a TestCase. 
```php
protected function writeCase(string $status, float $time, array $trace = [], string $message = '', ?TestCase $test = null): void
```
This is more of a help to the IDE, if I remove the annotation it would be saying that any object that implements the `Test` interface can call `writeCase` and that is not true:

`Expected parameter of type '\PHPUnit\Framework\TestCase|void', '\PHPUnit\Framework\Test' provided.`

src/PHPUnit/Constraint/Crawler.php
TavoNiievez: the base class [`vendor/phpunit/phpunit/src/Framework/Constraint/Constraint.php`](https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/Constraint/Constraint.php) does not have these parameters typed.
I don't feel comfortable overwriting that signature even if only a DomCrawler is used (or a `string` in the case of the methods you mentioned above).

What I did was add/update the docblock with the type that you suggested.

src/PHPUnit/Constraint/Page.php
TavoNiievez:
```diff
-     * @param mixed $other Value or object to evaluate.
+     * @param string $other Value or object to evaluate.
```
the `PHPUnit/Framework/Constraint/Constraint` base class doesn't have that parameter typed, I know the only use it is given here is as a string, but that generates a lot of warnings in the IDE and static code analyzers.

src/PHPUnit/Constraint/Page.php
TavoNiievez:
```diff
+    /**
+     * @param string $pageContent
+     * @return string
+     */
    protected function failureDescription($pageContent): string
```

TavoNiievez:
@Naktibalda I'm sorry that I have resolved the conversations, I forgot that I made changes in the docblocks instead of the signature of the methods when seeing the warnings that it generated.

src/PHPUnit/Constraint/Page.php
TavoNiievez: Making this change also generates errors in the tests:

```shell
1) PhpBrowserTest: See text after js comparision operator
 Test  tests/unit/Codeception/Module/PhpBrowserTest.php:testSeeTextAfterJSComparisionOperator
                                                                                                                  
  [PHPUnit\Framework\Exception] Declaration of Codeception\PHPUnit\Constraint\Page::matches(string $other): bool should be compatible with PHPUnit\Framework\Constraint\Constraint::matches($other): bool at ../../../src/PHPUnit/Constraint/Page.php:42  
```

src/PHPUnit/Constraint/Crawler.php
TavoNiievez: Making this change also generates errors in the tests:

```shell
2) PhpBrowserTest: Submit form with queries
 Test  tests/unit/Codeception/Module/PhpBrowserTest.php:testSubmitFormWithQueries
                                                                                                                                                                                                                                                                                                 
  [PHPUnit\Framework\Exception] Declaration of Codeception\PHPUnit\Constraint\Crawler::matches(Symfony\Component\DomCrawler\Crawler $nodes): bool should be compatible with Codeception\PHPUnit\Constraint\Page::matches(string $other): bool at ../../../src/PHPUnit/Constraint/Crawler.php:18  
```

src/PHPUnit/Constraint/Page.php
TavoNiievez: Making the change you suggest also generates errors in the tests:

```shell
 1) PhpBrowserTest: See text after js comparision operator
 Test  tests/unit/Codeception/Module/PhpBrowserTest.php:testSeeTextAfterJSComparisionOperator
                                                                                                                                                                                                                                                                                          
  [PHPUnit\Framework\Exception] Declaration of Codeception\PHPUnit\Constraint\Page::failureDescription(string $pageContent): string should be compatible with PHPUnit\Framework\Constraint\Constraint::failureDescription($other): string at ../../../src/PHPUnit/Constraint/Page.php:65  
```

src/PHPUnit/Util/Log/JSON.php
TavoNiievez: I changed it just as you suggested with a co-authored commit.

Naktibalda (changes requested):
Please rebase on 5.0 branch

PR #6119, opened 2021-02-20
Don't suppress failed assertions

Naktibalda:
This issue was introduced by renaming some variables, $e is passed to hooks below.

It is impossible to test for this issue, because Codeception is self-testing and the test can't fail if all assertions are passing.
We could implement more robust test suite by moving some cli test to separate suite which is directly executed by PHPUnit, but it requires a lot of effort.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,6 +55,9 @@ jobs:
       - name: Composer install
         run: composer install --no-interaction --no-progress --optimize-autoloader --ansi
 
+      - name: Test that failing test really fails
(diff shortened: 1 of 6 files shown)


Naktibalda:
I fixed all other issues that were suppressed by original issue too.

Naktibalda:
I added step to Github Actions workflow to test that failing test is really failing, it passes in this branch.

Example of that test failing: https://github.com/Codeception/Codeception/runs/1945591998?check_suite_focus=true

TavoNiievez (approved):
A tricky issue.
For me the way you ensure it fails is fine. I'll keep it in mind. Thank you.

eXorus (approved):

PR #6120, opened 2021-02-23
Update code standards to Unit tests

TavoNiievez:


tests/unit/Codeception/Util/PathResolverTest.php
```diff
@@ -19,7 +21,7 @@ public function testGetRelativeDir($path, $projDir, $dirSep, $expectedOutput)
      *
      * @return array(array(strings))
```
Naktibalda, line 22: Based on another example this method returns `string[][]`

Naktibalda (changes requested):

PR #6124, opened 2021-02-26
Action file generator improvements

### LINKED ISSUE #6122: Call to undefined method ReflectionUnionType::getName()
Run acceptance test with Webdriver (selenium)

#### What do you get instead?
PHP Fatal error:  Uncaught Error: Call to undefined method ReflectionUnionType::getName() in ..... vendor\codeception\codeception\src\Codeception\Lib\Generator\Actions.php:228
 
### Details
* Codeception version: 4.1.18
* PHP Version: 8.0
* Operating System: Windows
* Installation type: Composer

Naktibalda:
* Don't crash on union type in PHP 8
* Generate return union types
* Generate parameter typehints
* Fix tests, because they were running assertions agaist source file instead of generated file

Closes #6122

### DIFF
--- a/src/Codeception/Lib/Generator/Actions.php
+++ b/src/Codeception/Lib/Generator/Actions.php
@@ -155,11 +155,18 @@ EOF;
     {
         $params = [];
         foreach ($refMethod->getParameters() as $param) {
+            $type = '';
+            if (PHP_VERSION_ID >= 70000) {
+                $reflectionType = $param->getType();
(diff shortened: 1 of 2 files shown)


TavoNiievez (approved):
I was able to reproduce the issue that this PR solves.
It only gave an error with the union return types, the union parameter types did work.
This solves the problem.
both tests and changes look good to me.

PR #6129, opened 2021-03-02
Update code standards to Lib and Templates

TavoNiievez:
- [x] Use strict types in `src/Codeception/Lib` and `src/Codeception/Template`.
- [x] Add type hints to Libs and Templates.
- [x] Remove unnecessary qualifiers in `src/Codeception/Lib` and `src/Codeception/Template`.
- [x] Fix naming in `src/Codeception/Lib` and `src/Codeception/Template`.

### DIFF
--- a/src/Codeception/Lib/Actor/Shared/Comment.php
+++ b/src/Codeception/Lib/Actor/Shared/Comment.php
@@ -29,7 +29,7 @@ trait Comment
     {
         $role = trim($role);
 
-        if (stripos('aeiou', (string) $role[0]) !== false) {
+        if (stripos('aeiou', $role[0]) !== false) {
             return $this->comment('As an ' . $role);
         }
 
--- a/src/Codeception/Lib/Actor/Shared/Comment.php
+++ b/src/Codeception/Lib/Actor/Shared/Comment.php
@@ -1,45 +1,47 @@
 <?php
+
+declare(strict_types=1);
+
 namespace Codeception\Lib\Actor\Shared;
+
+use Codeception\Scenario;
 
 trait Comment
 {
-    /**
-     * @return \Codeception\Scenario
-     */
-    abstract protected function getScenario();
+    abstract protected function getScenario(): Scenario;
 
-    public function expectTo($prediction)
+    public function expectTo(string $prediction)
     {
         return $this->comment('I expect to ' . $prediction);
     }
 
-    public function expect($prediction)
+    public function expect(string $prediction)
     {
         return $this->comment('I expect ' . $prediction);
     }
 
-    public function amGoingTo($argumentation)
+    public function amGoingTo(string $argumentation)
     {
         return $this->comment('I am going to ' . $argumentation);
     }
 
-    public function am($role)
+    public function am(string $role)
     {
         $role = trim($role);
 
-        if (stripos('aeiou', $role[0]) !== false) {
+        if (stripos('aeiou', (string) $role[0]) !== false) {
             return $this->comment('As an ' . $role);
         }
 
         return $this->comment('As a ' . $role);
     }
 
-    public function lookForwardTo($achieveValue)
+    public function lookForwardTo(string $achieveValue)
     {
         return $this->comment('So that I ' . $achieveValue);
     }
 
-    public function comment($description)
+    public function comment(string $description)
     {
         $this->getScenario()->comment($description);
         return $this;
--- a/src/Codeception/Lib/Console/DiffFactory.php
+++ b/src/Codeception/Lib/Console/DiffFactory.php
@@ -1,4 +1,7 @@
 <?php
+
+declare(strict_types=1);
+
 namespace Codeception\Lib\Console;
 
 use SebastianBergmann\Comparator\ComparisonFailure;
@@ -9,26 +12,17 @@ use SebastianBergmann\Diff\Differ;
  **/
 class DiffFactory
 {
-    /**
-     * @param ComparisonFailure $failure
-     * @return string|null
-     */
-    public function createDiff(ComparisonFailure $failure)
+    public function createDiff(ComparisonFailure $failure): ?string
     {
         $diff = $this->getDiff($failure->getExpectedAsString(), $failure->getActualAsString());
-        if (!$diff) {
+        if ($diff === '') {
             return null;
         }
 
         return $diff;
     }
 
-    /**
-     * @param string $expected
-     * @param string $actual
-     * @return string
-     */
-    private function getDiff($expected = '', $actual = '')
+    private function getDiff(string $expected = '', string $actual = ''): string
     {
         if (!$actual && !$expected) {
             return '';
--- a/src/Codeception/Lib/Console/DiffFactory.php
+++ b/src/Codeception/Lib/Console/DiffFactory.php
@@ -7,19 +7,11 @@ namespace Codeception\Lib\Console;
 use SebastianBergmann\Comparator\ComparisonFailure;
 use SebastianBergmann\Diff\Differ;
 
-/**
- * DiffFactory
- **/
 class DiffFactory
 {
-    public function createDiff(ComparisonFailure $failure): ?string
+    public function createDiff(ComparisonFailure $failure): string
     {
-        $diff = $this->getDiff($failure->getExpectedAsString(), $failure->getActualAsString());
-        if ($diff === '') {
-            return null;
-        }
-
-        return $diff;
+        return $this->getDiff($failure->getExpectedAsString(), $failure->getActualAsString());
     }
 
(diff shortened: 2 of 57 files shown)


src/Codeception/Lib/Actor/Shared/Comment.php
```diff
@@ -1,45 +1,47 @@
...
         return $this->comment('I am going to ' . $argumentation);
     }
 
-    public function am($role)
+    public function am(string $role)
     {
         $role = trim($role);
 
-        if (stripos('aeiou', $role[0]) !== false) {
+        if (stripos('aeiou', (string) $role[0]) !== false) {
```
Naktibalda, line 32: Is this type casting really necesssary?

src/Codeception/Lib/Console/Message.php
```diff
@@ -78,44 +95,41 @@ public function append($string)
...
     }
 
-    public function getLength($includeTags = false)
+    public function getLength(bool $includeTags = false): int
     {
         return mb_strwidth($includeTags ? $this->message : strip_tags($this->message), 'utf-8');
     }
 
-    public static function ucfirst($text)
+    public static function ucfirst(?string $text): string
```
Naktibalda, line 127: Questionmark in signature and all the casting looks suspicious.
What code passes nulls to this method? I think that it should be fixed.

src/Codeception/Lib/Console/DiffFactory.php
```diff
@@ -9,26 +12,17 @@
...
-    /**
-     * @param ComparisonFailure $failure
-     * @return string|null
-     */
-    public function createDiff(ComparisonFailure $failure)
+    public function createDiff(ComparisonFailure $failure): ?string
     {
         $diff = $this->getDiff($failure->getExpectedAsString(), $failure->getActualAsString());
-        if (!$diff) {
+        if ($diff === '') {
```
Naktibalda, line 18: `createDiff` method is called in a [single place](https://github.com/Codeception/Codeception/pull/6129/files#diff-d3024ad9f62f7f2a33ee201f9b31494397d8d18c0936c30f49275dd673ff4e7bL49) which could be updated to handle empty string response and simplify return type to `: string`.

src/Codeception/Lib/Console/MessageFactory.php
```diff
@@ -22,38 +22,26 @@ class MessageFactory
...
 
-    /**
-     * @param ComparisonFailure $failure
-     * @return string
-     */
-    public function prepareComparisonFailureMessage(ComparisonFailure $failure)
+    public function prepareComparisonFailureMessage(ComparisonFailure $failure): string
     {
         $diff = $this->diffFactory->createDiff($failure);
         if (!$diff) {
```
Naktibalda, line 40: Changing this condition to `$diff !== ''` would allow to simplify the signature of createDiff method.

src/Codeception/Lib/Generator/Actor.php
```diff
@@ -30,10 +51,19 @@ class {{actor}} extends \Codeception\Actor
 
 EOF;
 
+    /**
+     * @var string
+     */
     protected $inheritedMethodTemplate = ' * @method {{return}} {{method}}({{params}})';
 
     protected $settings;
```
Naktibalda, line 59: Based on usage and changes made in other generators, $settings must be an array, right?

src/Codeception/Lib/GroupManager.php
```diff
@@ -171,10 +173,9 @@ public function groupsForTest(\PHPUnit\Framework\Test $test)
                 }
                 if ($test instanceof \PHPUnit\Framework\TestSuite\DataProvider) {
                     $firstTest = $test->testAt(0);
-                    if ($firstTest != false && $firstTest instanceof TestInterface) {
-                        if (strpos($filename . ':' . $firstTest->getName(false), $testPattern) === 0) {
-                            $groups[] = $group;
-                        }
+                    $firstTestName = $firstTest->getName(false);
+                    if ($firstTest != false && $firstTest instanceof TestInterface && strpos($filename . ':' . $firstTestName, (string) $testPattern) === 0) {
```
Naktibalda, line 177: Potential bug here, `getName` method is called before checking that $firstTest is not false.

src/Codeception/Lib/ModuleContainer.php
```diff
@@ -286,17 +274,17 @@ public function getModule($moduleName)
...
-    public function throwMissingModuleExceptionWithSuggestion($className, $moduleName)
+    public function throwMissingModuleExceptionWithSuggestion(string $className, string $moduleName): void
     {
         $suggestedModuleNameInfo = $this->getModuleSuggestion($moduleName);
-        throw new ModuleException($className, "Module $moduleName couldn't be connected" . $suggestedModuleNameInfo);
+        throw new ModuleException($className, "Module {$moduleName} couldn't be connected" . $suggestedModuleNameInfo);
     }
 
-    protected function getModuleSuggestion($missingModuleName)
+    protected function getModuleSuggestion($missingModuleName): string
```
Naktibalda, line 283: Parameter must be string.

src/Codeception/Lib/Parser.php
```diff
@@ -25,13 +30,13 @@ public function __construct(Scenario $scenario, Metadata $metadata)
...
 
-    public function prepareToRun($code)
+    public function prepareToRun($code): void
     {
         $this->parseFeature($code);
         $this->parseScenarioOptions($code);
     }
 
-    public function parseFeature($code)
+    public function parseFeature($code): void
```
Naktibalda, line 39: It seems that you forgot to add types to method parameters in this file.
I see lots of strings here.

src/Codeception/Template/Bootstrap.php
```diff
@@ -146,24 +159,23 @@ public function createGlobalConfig()
...
             $namespace = rtrim($this->namespace, '\\');
-            $str = "namespace: $namespace\n" . $str;
+            $str = "namespace: {$namespace}\n" . $str;
         }
         $this->createFile('codeception.yml', $str);
     }
 
-
-    protected function createSuite($suite, $actor, $config)
+    protected function createSuite($suite, $actor, $config): void
```
Naktibalda, line 174: Please add types of method parameters to this file too.

Naktibalda (changes requested):

src/Codeception/Lib/Actor/Shared/Comment.php
TavoNiievez: It's not, it seems like I added this before typing the parameter and then forgot to remove it, sorry.

src/Codeception/Lib/Console/Message.php
TavoNiievez: The uses are here.. `src/Codeception/Test/Cept.php` : 

```php
    public function toString(): string
    {
        return $this->getSignature() . ': ' . Message::ucfirst($this->getFeature());
    }
```

... and here `src/Codeception/Test/Cest.php` :

```php
    public function toString(): string
    {
        return sprintf('%s: %s', ReflectionHelper::getClassShortName($this->getTestClass()), Message::ucfirst($this->getFeature()));
    }
```

Then, both the class `src/Codeception/Test/Feature/ScenarioLoader.php`:

```php
    public function getFeature(): ?string
    {
        return $this->getScenario()->getFeature();
    }
```

as the interface `src/Codeception/Test/Interfaces/ScenarioDriven.php`:

```php
    public function getFeature(): ?string;
```
    
have the return value nullable.

What should I do?

TavoNiievez: @Naktibalda

TavoNiievez: @Naktibalda also look at the failed build: https://github.com/Codeception/Codeception/actions/runs/617723034 .

Naktibalda: I think that it is Gherkin's fault that Feature can be null, and it would be the best to fix this problem at the source and add casting to string to https://github.com/Codeception/Codeception/blob/5.0/src/Codeception/Test/Gherkin.php#L63

TavoNiievez: @Naktibalda Please review my last commit.
Also note that I will finish adding typehints to the files in `src/Codeception/*php` in the next PR.

src/Codeception/Lib/Generator/Actor.php
```diff
@@ -30,10 +51,24 @@ class {{actor}} extends \Codeception\Actor
...
+     * @var array
+     */
     protected $modules;
+
+    /**
+     * @var array
+     */
     protected $actions;
 
     public function __construct($settings)
```
Naktibalda, line 74: $settings typehint is missing here.

Naktibalda (approved):

PR #6130, opened 2021-03-03
Update code standards to src/Codeception classes

TavoNiievez:
- [x] Use strict types in `src/Codeception/*.php`.
- [x] Add type hints to `src/Codeception/*.php` classes.
- [x] Remove unnecessary qualifiers in `src/Codeception/*.php`.


src/Codeception/Configuration.php
```diff
@@ -621,6 +633,11 @@ public static function append(array $config = [])
         return self::$config;
     }
 
+    /**
+     * @param array|bool $a1
+     * @param array|bool $a2
```
Naktibalda, line 638: Parameters can be `null` too, because `getConfFromFile` can return `null`.

src/Codeception/Example.php
```diff
@@ -98,8 +106,8 @@ public function count()
      * <b>Traversable</b>
      * @since 5.0.0
```
Naktibalda, line 107: These version numbers may be confusing now. They refer to PHP version, right?

src/Codeception/Module.php
```diff
@@ -268,58 +269,49 @@ public function _failed(TestInterface $test, $fail)
     /**
      * Print debug message to the screen.
      *
-     * @param $message
+     * @param array|string $message
```
Naktibalda, line 272: It can be anything, so the type is really `mixed`.

src/Codeception/Module.php
```diff
@@ -345,10 +335,10 @@ protected function getModule($name)
     /**
      * Get config values or specific config item.
      *
-     * @param mixed $key
-     * @return mixed the config item's value or null if it doesn't exist
+     * @param string|null $key
+     * @return array|bool|null the config item's value or null if it doesn't exist
```
Naktibalda, line 339: I think that you missed `string` and `int` in return type and it should be `mixed` too.

src/Codeception/Scenario.php
```diff
@@ -117,54 +127,46 @@ public function getHtml()
...
+            $text .= $step->getPrefix() . "{$step} \r\n";
         }
         $text = trim(str_replace(['"\'', '\'"'], ["'", "'"], $text));
-        $text = mb_strtoupper('I want to ' . $this->getFeature(), 'utf-8') . "\r\n\r\n" . $text . "\r\n\r\n";
-        return $text;
+        return mb_strtoupper('I want to ' . $this->getFeature(), 'utf-8') . "\r\n\r\n" . $text . "\r\n\r\n";
     }
 
-    public function comment($comment)
+    public function comment(?string $comment): void
```
Naktibalda, line 143: Gherkin is passing `null` again, it would be better to make it accept strings only and update `Codeception\Test\Gherkin::runStep` to pass an empty string instead.

src/Codeception/Step.php
```diff
@@ -1,50 +1,78 @@
...
+     */
     protected $failed = false;
+
+    /**
+     * @var bool
+     */
     protected $isTry = false;
 
-    public function __construct($action, array $arguments = [])
+    public function __construct(?string $action, array $arguments = [])
```
Naktibalda, line 69: If Comment stops accepting null, $action parameter can be a string too, right?

src/Codeception/SuiteManager.php
```diff
@@ -45,15 +51,30 @@ class SuiteManager
      */
```
Naktibalda, line 51: `$testLoader` field above is unused, it is a good time to delete it.

Naktibalda (changes requested):

src/Codeception/Configuration.php
```diff
@@ -14,6 +26,9 @@
 
 class Configuration
 {
+    /**
+     * @var int[]|string[]
+     */
     protected static $suites = [];
```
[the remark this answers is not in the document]
Naktibalda: Well, I know that it comes from the signature of `array_keys` but we can't possibly have any integer keys in this array.

src/Codeception/Configuration.php
```diff
@@ -404,10 +428,10 @@ protected static function loadEnvConfigs($path)
      *
      * @param string $contents Yaml config file contents
      * @param string $filename which is supposed to be loaded
-     * @return array
+     * @return array|null
```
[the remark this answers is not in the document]
TavoNiievez: @Naktibalda, can you please look at the failed build considering my last commit?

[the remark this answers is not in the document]
Naktibalda: https://github.com/Codeception/Codeception/blob/c26038596f7f35cc684b380f6e3fe186331e7820/src/Codeception/Configuration.php#L184 doesn't check if the file exists.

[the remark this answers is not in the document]
TavoNiievez: thanks

src/Codeception/Configuration.php
```diff
@@ -621,6 +633,11 @@ public static function append(array $config = [])
         return self::$config;
     }
 
+    /**
+     * @param array|bool $a1
+     * @param array|bool $a2
+     * @return array|bool
```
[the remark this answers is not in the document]
TavoNiievez: Sorry I'm not really sure how to deal with this, so it always returns an `array`

src/Codeception/Step.php
```diff
@@ -145,7 +173,11 @@ public function getArgumentsAsString($maxLength = self::DEFAULT_MAX_LENGTH)
         return implode(',', $arguments);
     }
 
-    protected function stringifyArgument($argument)
+    /**
+     * @param string|resource|array|object $argument
```
Naktibalda, line 177: This method can receive APP php data types, so `mixed` is the most appropriate.

src/Codeception/Step.php
```diff
@@ -169,13 +201,16 @@ protected function stringifyArgument($argument)
...
         }
         $arg_str = json_encode($argument, JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES);
-        $arg_str = str_replace('\"', '"', $arg_str);
-        return $arg_str;
+        return str_replace('\"', '"', $arg_str);
     }
 
-    protected function getClassName($argument)
+    /**
+     * @param Closure|MockObject $argument
```
Naktibalda, line 208: This type is incorrect, because this method can receive instance of any class, but only 2 special cases are mentioned.

Naktibalda (changes requested):

PR #6135, opened 2021-03-07
Merge changes from 4.1

Naktibalda:


TavoNiievez (approved):

Naktibalda:
I reviewed this PR myself and removed `PHP_VERSION_ID >= 70000` condition, because it won't be necessary in Codeception 5.

src/Codeception/Lib/Generator/Actions.php
```diff
@@ -255,10 +260,24 @@ private function createReturnTypeHint(ReflectionMethod $refMethod): string
...
         }
 
+        return ': ' . $this->stringifyType($returnType);
+    }
+
+    /**
+     * @param \ReflectionType $returnType
+     * @return string
+     */
+    private function stringifyType(\ReflectionType $returnType)
```
TavoNiievez, line 270: I think you can `: string` and remove that docblock.

TavoNiievez (changes requested):

src/Codeception/Command/DryRun.php
```diff
@@ -93,7 +93,7 @@ public function execute(InputInterface $input, OutputInterface $output): int
         $eventDispatcher->dispatch(new SuiteEvent($suiteManager->getSuite(), null, $settings), Events::SUITE_BEFORE);
 
         foreach ($tests as $test) {
-            if ($test instanceof PHPUnitDataProvider) {
+            if ($test instanceof DataProviderTestSuite) {
```
TavoNiievez, line 96: I saw the opinion of **DavertMik** about it, but in `5.0` I think it is better to remove it... According https://github.com/Codeception/Codeception/pull/6136#discussion_r589233283.

Naktibalda: I would rather remove it from Codeception 4 because it counts as a new feature now and it could be a breaking change.

src/Codeception/Lib/Generator/Actions.php
Naktibalda: Done

src/Codeception/Command/DryRun.php
Naktibalda: Dead DataProviderTestSuite conditions were removed from 4.1 by #6138 ,
This PR will remove them from 5.0

tests/cli/CleanCest.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+class CleanCest
```
TavoNiievez, line 3:
```suggestion
<?php

declare(strict_types=1);

final class CleanCest
```

tests/cli/BuildCest.php
```diff
@@ -48,9 +48,28 @@ public function usesTypehintsWherePossible(CliGuy $I, Scenario $scenario)
...
         $I->seeFileFound('CliGuyActions.php', 'tests/support/_generated');
-        $I->openFile(codecept_root_dir('tests/support/CliHelper.php'));
-        $I->seeInThisFile('public function grabFromOutput($regex): int');
-        $I->seeInThisFile('return $match[1]');
+        $I->seeInThisFile('public function grabFromOutput(string $regex): int');
+    }
+
+    public function generatedUnionReturnTypeOnPhp8(CliGuy $I, Scenario $scenario)
+    {
+        if (PHP_MAJOR_VERSION < 8) {
```
TavoNiievez, line 56: I'm wondering if Codeception supports [`@requires`](https://phpunit.readthedocs.io/en/9.5/incomplete-and-skipped-tests.html#skipping-tests-using-requires) annotation. If so, I think it would be good to use it, since it is cleaner than this conditional and in the future it can be replaced with its homonymous attribute in php 8

Naktibalda: I tried it, it has no effect.

tests/cli/CleanCest.php
TavoNiievez: I suggested including `final` as well since the other classes in this folder are declared like this. This was done because there is no implicit intention of inheritance here. It would be good to be congruent and decide whether to have final test classes or not. But the point is, they should be the same.

tests/cli/BuildCest.php
TavoNiievez: Oh 😞 it would be great to have that. Thanks.

tests/cli/CleanCest.php
Naktibalda: Done

TavoNiievez (approved):

PR #6136, opened 2021-03-07
4.1 Use correct name of DataProviderTestSuite in conditions

Naktibalda:
I accidently discovered that Codeception was using name of non-existant class in conditions.

PHPUnit_Framework_TestSuite_DataProvider class was renamed to PHPUnit\Framework\DataProviderTestSuite in PHPUnit 6.0.0, but Codeception code was updated to use PHPUnit\Framework\TestSuite\DataProvider in conditions.

2 things are obvious:
1. We have no tests covering these cases.
2. It hasn't caused any issues that are worth reporting.

Wouldn't it be better to delete the code in these blocks? It was dead for 5 years.

### DIFF
--- a/src/Codeception/Lib/GroupManager.php
+++ b/src/Codeception/Lib/GroupManager.php
@@ -149,7 +149,7 @@ class GroupManager
         if ($test instanceof \PHPUnit\Framework\TestCase) {
             $groups = array_merge($groups, \PHPUnit\Util\Test::getGroups(get_class($test), $test->getName(false)));
         }
-        if ($test instanceof \PHPUnit\Framework\TestSuite\DataProvider) {
+        if ($test instanceof \PHPUnit\Framework\DataProviderTestSuite) {
             $firstTest = $test->testAt(0);
             if ($firstTest != false && $firstTest instanceof TestInterface) {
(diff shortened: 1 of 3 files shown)


src/Codeception/Lib/GroupManager.php
```diff
@@ -169,7 +169,7 @@ public function groupsForTest(\PHPUnit\Framework\Test $test)
                     && mb_strtolower($filename . ':' . $test->getMetadata()->getFeature()) === mb_strtolower($testPattern)) {
                     $groups[] = $group;
                 }
-                if ($test instanceof \PHPUnit\Framework\TestSuite\DataProvider) {
+                if ($test instanceof \PHPUnit\Framework\DataProviderTestSuite) {
                     $firstTest = $test->testAt(0);
```
Naktibalda, line 173: @DavertMik This code is so dead. testAt method was removed in PHPUnit 9.3.0 - https://github.com/sebastianbergmann/phpunit/commit/7d948a75c6cc66059873386124291fd890f9ee2a#diff-ce07bbf9be7fe14e3102ab699c8a2d6c2a995bc23f8f93871fb4f3dde098dfaa

PR #6137, opened 2021-03-07
Standardize code style

TavoNiievez:
With this PR I do a general review of my previous PRs, and I make sure to fix any inadvertently omitted changes, in addition to taking the opportunity to give *(the much desired)* consistency to the source code:

* Removed unnecessary typing docblocks
* Standardize `#` as PREG delimiter in regex
* Add `{}` to interpolate variables inside strings
* Stop using `list()` in favor of `[]`
* Add strict comparisons
* Add `JSON_THROW_ON_ERROR` to json functions where missing
* Initialize properties that do not get initial value in constructor, but are typed
* Type constants
* Use pre-increment and pre-decrement instead of increment and decrement
* Remove invalid type annotations
* Fix naming
* `assertNull` instead of `assertSame(null, ...)`

This is the last PR from me before the `5.0` release.

### DIFF
--- a/src/Codeception/Application.php
+++ b/src/Codeception/Application.php
@@ -80,7 +80,6 @@ class Application extends BaseApplication
      * Validate and get the name of the command
      *
      * @param CustomCommandInterface|string $commandClass
-     * @return string
      * @throws ConfigurationException
      */
     protected function getCustomCommandName($commandClass): string
@@ -149,7 +148,7 @@ class Application extends BaseApplication
             $argv = $_SERVER['argv'];
 
             for ($i = 0; $i < count($argv); ++$i) {
-                if (preg_match('/^(?:-([^c-]*)?c|--config(?:=|$))(.*)$/', $argv[$i], $match)) {
+                if (preg_match('#^(?:-([^c-]*)?c|--config(?:=|$))(.*)$#', $argv[$i], $match)) {
                     if (!empty($match[2])) { //same index
                         $this->preloadConfiguration($match[2]);
                     } elseif (isset($argv[$i + 1])) { //next index
--- a/src/Codeception/Command/Run.php
+++ b/src/Codeception/Command/Run.php
@@ -262,8 +262,6 @@ class Run extends Command
     /**
      * Executes Run
      *
-     * @param InputInterface $input
-     * @param OutputInterface $output
      * @return int|null|void
      * @throws ConfigurationException|ParseException
      */
@@ -282,7 +280,7 @@ class Run extends Command
         $config = $this->getGlobalConfig();
 
         // update config from options
-        if (count($this->options['override'])) {
+        if (!empty($this->options['override'])) {
             $config = $this->overrideConfig($this->options['override']);
         }
         if ($this->options['ext']) {
@@ -385,7 +383,7 @@ class Run extends Command
                         $testsPath = $include . DIRECTORY_SEPARATOR.  $config['paths']['tests'];
 
                         try {
-                            list(, $suite, $test) = $this->matchTestFromFilename($suite, $testsPath);
+                            [, $suite, $test] = $this->matchTestFromFilename($suite, $testsPath);
                             $isIncludeTest = true;
                         } catch (InvalidArgumentException $e) {
                             // Incorrect include match, continue trying to find one
@@ -394,7 +392,7 @@ class Run extends Command
                     } else {
                         $result = $this->matchSingleTest($suite, $config);
                         if ($result) {
-                            list(, $suite, $test) = $result;
+                            [, $suite, $test] = $result;
                         }
                     }
                 }
@@ -406,7 +404,7 @@ class Run extends Command
             } elseif (!empty($suite)) {
                 $result = $this->matchSingleTest($suite, $config);
                 if ($result) {
-                    list(, $suite, $test) = $result;
+                    [, $suite, $test] = $result;
                 }
             }
         }
@@ -458,7 +456,7 @@ class Run extends Command
     {
         // Workaround when codeception.yml is inside tests directory and tests path is set to "."
         // @see https://github.com/Codeception/Codeception/issues/4432
-        if (isset($config['paths']['tests']) && $config['paths']['tests'] === '.' && !preg_match('~^\.[/\\\]~', $suite)) {
+        if (isset($config['paths']['tests']) && $config['paths']['tests'] === '.' && !preg_match('#^\.[/\\\]#', $suite)) {
             $suite = './' . $suite;
         }
 
@@ -472,7 +470,7 @@ class Run extends Command
                 if ($suiteConfig['path'] === '.') {
                     $testsPath = $config['paths']['tests'];
                 }
-                if (preg_match("~^$testsPath/(.*?)$~", $suite, $matches)) {
+                if (preg_match("~^{$testsPath}/(.*?)$~", $suite, $matches)) {
                     $matches[2] = $matches[1];
                     $matches[1] = $s;
                     return $matches;
@@ -492,7 +490,7 @@ class Run extends Command
             if (strpos($realTestDir, $cwd) === 0) {
                 $file = $suite;
                 if (strpos($file, ':') !== false) {
-                    list($file) = explode(':', $suite, -1);
+                    [$file] = explode(':', $suite, -1);
                 }
                 $realPath = $cwd . DIRECTORY_SEPARATOR . $file;
                 if (file_exists($realPath) && strpos($realPath, $realTestDir) === 0) {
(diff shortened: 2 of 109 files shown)


src/Codeception/Command/Shared/FileSystemTrait.php
```diff
@@ -51,8 +51,8 @@ protected function completeSuffix(string $filename, string $suffix): string
 
     protected function removeSuffix(string $classname, string $suffix): string
     {
-        $classname = preg_replace('~\.php$~', '', $classname);
-        return preg_replace("~$suffix$~", '', $classname);
+        $classname = preg_replace('#\.php$#', '', $classname);
+        return preg_replace("~{$suffix}$~", '', $classname);
```
Naktibalda, line 55: Is there a good reason to standardize on `#`? `/` is much more common.

Also you missed different delimiters in a few places - this one is the most glaring.

src/Codeception/Command/Run.php
```diff
@@ -588,15 +584,15 @@ protected function matchTestFromFilename($filename, $testsPath)
             //codecept run tests
             return ['', '', $filter];
         }
-        $res = preg_match("~^$testsPath/(.*?)(?>/(.*))?$~", $filename, $matches);
+        $res = preg_match("~^{$testsPath}/(.*?)(?>/(.*))?$~", $filename, $matches);
```
hason, line 587: `#` as PREG delimiter in regex

src/Codeception/Command/Shared/FileSystemTrait.php
TavoNiievez: > / is much more common.

Yes, however it is also too common to see slashes within the same expression but for another purpose.

> Is there a good reason to standardize on #?

The idea is to use a character that is easy to identify to increase readability.
It's a matter of choosing, but `/` would be my last pick.

> Also you missed different delimiters in a few places - this one is the most glaring.

Yes, sorry about that.
I try to do my best with these PRs, but there are a few things I still tend to skip. That is the goal of this PR itself and that is why I value reviews so much.

src/Codeception/Command/Run.php
TavoNiievez: @hason Oops, you're right :)

This time I also searched for all [PCRE functions](https://www.php.net/manual/en/ref.pcre.php) and tried to catch them all.

Thanks for reviewing.

src/Codeception/Command/Shared/FileSystemTrait.php
TavoNiievez: @Naktibalda I just did a commit to replace the ones that I may have missed

src/Codeception/Test/Loader/Gherkin.php
```diff
@@ -172,7 +172,7 @@ public function makePlaceholderPattern(string $pattern): string
             ); // or matching numbers with optional $ or € chars
 
             // params converting from :param to match 11 and "aaa" and "aaa\"aaa"
-            $pattern = preg_replace('~"?\\\:(\w+)"?~', $replacePattern, $pattern);
+            $pattern = preg_replace('#"?\\\:(\w+)"?#', $replacePattern, $pattern);
             $pattern = "/^{$pattern}$/u";
```
Naktibalda, line 176: Another pattern here.

src/Codeception/Command/Shared/FileSystemTrait.php
Naktibalda: The line above this comment hasn't been updated.

src/Codeception/SuiteManager.php
```diff
@@ -60,14 +60,15 @@ class SuiteManager
      * @var string
      */
     protected $path = '';
+
     protected $printer = null;
 
     protected $env = null;
```
Naktibalda, line 66: This class had suspiciously many untyped fields, so I checked them and it appeared that most of them are actually unused and can be removed.
Unused fields:
```
    public static $environment;
    public static $name;
    protected $testLoader;
    protected $tests = [];
    protected $debug = false;
    protected $printer = null;
```

`protected $path = '';` is assigned a value in constructor, but it is never used anywhere.

src/Codeception/SuiteManager.php
```diff
@@ -123,7 +124,7 @@ public function loadTests(string $path = null): void
     /**
      * @param TestInterface|DataProviderTestSuite $test
      */
-    protected function addToSuite($test): void
+    protected function addToSuite(\PHPUnit\Framework\Test $test): void
```
Naktibalda, line 127: This change looks like conflict between `@param` and type.
I think that `@param` can be removed.

tests/unit/Codeception/Util/StubTest.php
```diff
@@ -2,8 +2,8 @@
 
 declare(strict_types=1);
 
-use \Codeception\Stub;
-use \Codeception\Stub\Expected;
```
Naktibalda, line 6: This file should be deleted altogether.

Codeception\Util\Stub class was renamed to Codeception\Stub and extracted to separate package 3 years ago. I removed class alias Codeception\Util\Stub in 5.0.

The latest version of this test is at https://github.com/Codeception/Stub/blob/master/tests/StubTest.php#L20-L26

src/Codeception/Application.php
```diff
@@ -80,7 +80,6 @@ protected function readCustomCommandsFromConfig(): void
      * Validate and get the name of the command
      *
      * @param CustomCommandInterface|string $commandClass
-     * @return string
      * @throws ConfigurationException
      */
     protected function getCustomCommandName($commandClass): string
```
Naktibalda, line 85: Based on the code and [documentation](https://codeception.com/docs/08-Customization#Custom-Commands) $commandClass is a string.

Naktibalda (changes requested):

src/Codeception/Event/SuiteEvent.php
```diff
@@ -22,7 +22,7 @@ class SuiteEvent extends Event
     protected $result;
 
     /**
-     * @var array
+     * @var array|null
```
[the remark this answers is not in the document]
TavoNiievez: Fixed in [`dd8f09d`](https://github.com/Codeception/Codeception/pull/6137/commits/dd8f09d7a509c0e780bdb473abcfb74afea54054).

src/Codeception/Event/SuiteEvent.php
```diff
@@ -22,7 +22,7 @@ class SuiteEvent extends Event
...
      */
     protected $settings = [];
```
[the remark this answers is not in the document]
TavoNiievez: Yeah... Someone is also waiting for `getSuite` to return... a DateTime? See [this](https://github.com/Codeception/Codeception/blob/db9c5b4aa30f92f8e03dc24a5a22002cb2d1e86c/tests/unit/Codeception/Subscriber/ModuleTest.php#L44). And failed build [here](https://github.com/Codeception/Codeception/pull/6137/checks?check_run_id=2086207496).

src/Codeception/Subscriber/ExtensionLoader.php
```diff
@@ -97,7 +97,6 @@ public function stopSuiteExtensions(): void
     }
 
     /**
-     * @param array $config
      * @return array<class-string, EventSubscriberInterface>
```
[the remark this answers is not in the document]
TavoNiievez: See [`PHPStan`](https://phpstan.org/writing-php-code/phpdoc-types#class-string) and [`Psalm`](https://psalm.dev/docs/annotating_code/type_syntax/scalar_types/#class-string-interface-string) documentation about it

src/Codeception/Event/SuiteEvent.php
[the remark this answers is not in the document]
Naktibalda: It is a stupid test with fake mocks.
I disabled it in PHPUnit 10 branch, because Prophet is no longer a dependency of PHPUnit and it doesn't make sense to add it as dependency of Codeception for single test.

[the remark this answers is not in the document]
Naktibalda: https://github.com/Codeception/Codeception/blob/5.0-phpunit-10/tests/unit/Codeception/Subscriber/ModuleTest.php#L29

src/Codeception/SuiteManager.php
TavoNiievez: I couldn't find `protected $testLoader;` in this class.
I also typed `protected $env` as a string

src/Codeception/Event/SuiteEvent.php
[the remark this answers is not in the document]
TavoNiievez: Well, I added the `markTestIncomplete` line here for this to work.

PR #6138, opened 2021-03-09
4.1 Remove dead DataProviderTestSuite code

Naktibalda:
These conditions used wrong class name since release of PHPUnit 6.0, so this code was is effectively dead for 4 years.
I fixed conditions in #6136 but now I think that it was a wrong way to handle this issue.

The only place where Codeception creates DataProviderTestSuite is in [Test\Loader\Cest](https://github.com/Codeception/Codeception/blob/4.1.18/src/Codeception/Test/Loader/Cest.php#L79) but dry-run command does not receive it.

I think that DataProviderTest Suite is handled well by [Test\Loader\Unit](https://github.com/Codeception/Codeception/blob/4.1.18/src/Codeception/Test/Loader/Unit.php#L63-L68) and [Codeception\SuiteManager](https://github.com/Codeception/Codeception/blob/4.1.18/src/Codeception/SuiteManager.php#L114-L119)


TavoNiievez (approved):

Naktibalda:
I removed tests/data/snapshots/tests/_output/failed file from this commit.

PR #6141, opened 2021-03-11
WIP: PHPUnit 10 support

Naktibalda:


src/Codeception/Coverage/PhpCodeCoverageFactory.php
```diff
@@ -4,23 +4,33 @@
...
 {
     public static function build(): CodeCoverage
     {
-        if (method_exists(Driver::class, 'forLineCoverage')) {
-            //php-code-coverage 9+
+        $filter = new CodeCoverageFilter();
+        $coverageConfiguration = Configuration::config()['coverage'];
+        if (PHPUnit10::codeCoverageRunnerClassExists()) {
+            if (!PHPUnitCodeCoverage::isActive()) {
+                PHPUnitCodeCoverage::activate($filter, $coverageConfiguration['path_coverage'] ?? false);
```
Naktibalda, line 22: PHPUnit 10 support must be copied to c3 script too.

tests/data/claypit/c3.php
```diff
@@ -229,18 +243,37 @@ function __c3_factory($filename, $lock=false)
             } else {
                 $phpCoverage = unserialize(file_get_contents($filename));
             }
-            
-            return array($phpCoverage, $file);
-        } else {
-            if (method_exists(Driver::class, 'forLineCoverage')) {
-                //php-code-coverage 9+
+
+            return array(
```
TavoNiievez, line 247: please, let's follow the boy scout rule here as much as possible.
Stop using `array( )`, use `[ ]` instead.

tests/data/claypit/c3.php
```diff
@@ -229,18 +243,37 @@ function __c3_factory($filename, $lock=false)
...
+            );
+        }
+
+        $pathCoverage = false;
+        if (isset($coverageConfiguration['path_coverage'])) {
+            $pathCoverage = (bool)$coverageConfiguration['path_coverage'];
+        }
+
+        if (class_exists(CodeCoverageRunner::class)) {
+            //PHPUnit 10+
```
TavoNiievez, line 259: Is it possible to use the `Codeception\PHPUnit\Compatibility` class here?

tests/data/claypit/c3.php
```diff
@@ -368,11 +406,11 @@ function () use ($codeCoverage, $current_report) {
                 // read/write to the file at the same time as this request (leading to a corrupt file). flock() is a
                 // blocking call, so it waits until an exclusive lock can be acquired before continuing.
 
-                list($existingCodeCoverage, $file) = __c3_factory($current_report, true);
+                list($existingCodeCoverage, $file) = __c3_factory($currentReport, true);
```
TavoNiievez, line 409: Stop using `list( )` use `[ ]` instead.

tests/data/claypit/c3.php
Naktibalda: This file is copy of https://github.com/Codeception/c3/blob/main/c3.php
which must stay compatible with PHP 5.5 and Codeception 3.

Please review modernization PR at https://github.com/Codeception/c3/pull/72

TavoNiievez: well, he's asking if he should remove the compatibility logic as well, you know my opinion on that.
Personally, I never used C3 which is why I refrain from commenting.

Naktibalda: The file in claypit will be updated when we have an updated file in c3 main branch.

Naktibalda:
Replaced by https://github.com/Codeception/Codeception/pull/6332

PR #6142, opened 2021-03-11
Abstract PHPUnit compatibility logic

TavoNiievez:
Benefits of doing something like this:

- There is no need to comment the code.
- The compatibility logic is in one place, it is easy to know *where* it is used and *which* of them there are.
- With fewer literals out there and there, typos are less likely to occur.

I'm not sure if that namespace (`Codeception\PHPUnit\Compatibility`) was the best to place something like this, but that is subject to change.

### DIFF
--- a/src/Codeception/Coverage/PhpCodeCoverageFactory.php
+++ b/src/Codeception/Coverage/PhpCodeCoverageFactory.php
@@ -4,25 +4,23 @@ declare(strict_types=1);
 
 namespace Codeception\Coverage;
 
+use Codeception\PHPUnit\Compatibility\PHPUnit10;
 use PHPUnit\Runner\CodeCoverage as PHPUnitCodeCoverage;
 use SebastianBergmann\CodeCoverage\CodeCoverage;
 use SebastianBergmann\CodeCoverage\Driver\Driver;
 use SebastianBergmann\CodeCoverage\Filter as CodeCoverageFilter;
-use function method_exists;
 
 class PhpCodeCoverageFactory
 {
     public static function build(): CodeCoverage
     {
         $filter = new CodeCoverageFilter();
-        if (method_exists(PHPUnitCodeCoverage::class, 'activate')) {
-            // php-code-coverage 10
+        if (PHPUnit10::codeCoverageRunnerClassExists()) {
             if (!PHPUnitCodeCoverage::isActive()) {
                 PHPUnitCodeCoverage::activate($filter, false);
             }
             return PHPUnitCodeCoverage::instance();
         } else {
-            //php-code-coverage 9+
             $filter = new CodeCoverageFilter();
             $driver = Driver::forLineCoverage($filter);
 
--- a/src/PHPUnit/NonFinal/JUnit.php
+++ b/src/PHPUnit/NonFinal/JUnit.php
@@ -9,6 +9,7 @@
  */
 namespace Codeception\PHPUnit\NonFinal;
 
+use Codeception\PHPUnit\Compatibility\PHPUnit10;
 use DOMDocument;
 use DOMElement;
 use PHPUnit\Framework\AssertionFailedError;
@@ -324,11 +325,9 @@ class JUnit extends Printer implements TestListener
     {
         $numAssertions = 0;
 
-        if (\method_exists($test, 'numberOfAssertionsPerformed')) {
-            // PHPUnit 10
+        if (PHPUnit10::numberOfAssertionsPerformedMethodExists($test)) {
             $numAssertions = $test->numberOfAssertionsPerformed();
         } else {
-            // PHPUnit 9
             $numAssertions = $test->getNumAssertions();
         }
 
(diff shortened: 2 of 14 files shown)


TavoNiievez:
btw, I don't think it is necessary *at the moment* to comment `// PHPUnit 10` in the `else` since the minimum version is 9, it is understood that any other version is 10 or higher.

src/Codeception/Coverage/PhpCodeCoverageFactory.php
```diff
@@ -4,25 +4,23 @@
...
-use function method_exists;
 
 class PhpCodeCoverageFactory
 {
     public static function build(): CodeCoverage
     {
         $filter = new CodeCoverageFilter();
-        if (method_exists(PHPUnitCodeCoverage::class, 'activate')) {
-            // php-code-coverage 10
+        if (PHPUnit10::codeCoverageRunnerClassExists()) {
```
TavoNiievez, line 18: I changed this because apparently that class only exists in `PHPUnit 10`, first you were using the `isActive` method which could also be non-existent... So I think validating this makes more sense, am I correct?

Naktibalda:
1. What about this? https://github.com/Codeception/Codeception/blob/5.0-phpunit-10/src/Codeception/Test/Test.php#L20-L24

2. We need to think how this logic will be extended to support PHPUnit 11. As you noticed, `PHPUnit 9` comments mean that specific code only works with PHPUnit 9, but `PHPUnit 10` comment means that it works with PHPUnit 10 and is expected to work with later versions.
Next year we will have to add 3 new compability classes:
    1. PHPUnit 11 or later
    2. PHPUnit 9 and 10.
    3. PHPUnit 10 only (It will be the smallest class).

src/PHPUnit/NonFinal/JUnit.php
```diff
@@ -324,11 +325,9 @@ public function endTest(Test $test, float $time): void
     {
         $numAssertions = 0;
 
-        if (\method_exists($test, 'numberOfAssertionsPerformed')) {
-            // PHPUnit 10
+        if (PHPUnit10::numberOfAssertionsPerformedMethodExists($test)) {
             $numAssertions = $test->numberOfAssertionsPerformed();
         } else {
-            // PHPUnit 9
             $numAssertions = $test->getNumAssertions();
```
Naktibalda, line 331: `getNumAssertions` method does not exist in PHPUnit 9 only.
it is defined in [AssertionCounter trait](https://github.com/Codeception/Codeception/blob/5.0/src/Codeception/Test/Feature/AssertionCounter.php) too.

TavoNiievez:
> Next year we will have to add 3 new compability classes:
> 
>     PHPUnit 11 or later
>     PHPUnit 9 and 10.
>     PHPUnit 10 only (It will be the smallest class).

Hmmm, I don't see it that way. My expectation is that Codeception supports in a stable version only the [supported versions of PHPUnit](https://phpunit.de/supported-versions.html), which are usually two, one with active support and the other with maintenance support. So...
```diff
-PHPUnit 11 or later
+PHPUnit 11 only
-PHPUnit 9 and 10.
PHPUnit 10 only
```
So next year you just have to add PHPUnit11 and remove PHPUnit9.

Even in a scenario where we have to support 3 versions at the same time...

```php
if(PHPUnit11...) {
} else { // PHPUnit 9 and 10
}

if(PHPUnit9...) {
} else { // PHPUnit 10 or later
}

if(PHPUnit10...) {
} elseif(PHPUnit9...) {
} else { // PHPUnit 11
}
```

src/PHPUnit/NonFinal/JUnit.php
TavoNiievez: So the whole conditional isn't necessary?

Naktibalda:
> Hmmm, I don't see it that way. My expectation is that Codeception supports in a stable version only the supported versions of PHPUnit, which are usually two, one with active support and the other with maintenance support

Supporting 2 rolling versions of PHPUnit in one version of Codeception is unnecessary burden.
If you are going to drop support for 1 version of PHPUnit every year, then it is much better to do 1 to 1 mapping and support 2 versions of Codeception.

TavoNiievez:
> then it is much better to do 1 to 1 mapping and support 2 versions of Codeception.

@Naktibalda Ok I understand.

Well... I think that would be better.
Codeception 10 -> PHPUnit 10
Codeception 9 -> PHPUnit 9
If we are so tied into (and dependent on) PHPUnit's release cycle that would make more sense to me.
But I think you mentioned it once and @DavertMik preferred the current independent versioning strategy.

I thought of presenting this PR just because of what I saw so far in your other PR, It is up to you to consider whether this is best for future releases.

This PR is an idea (which can be discarded), so I'm open to any design changes you suggest.

src/PHPUnit/NonFinal/JUnit.php
Naktibalda: this is  `// PHPUnit 9 or Cest or Cept or Gherkin`

TavoNiievez: Done.

Naktibalda:
The decision to support only 2 versions of PHPUnit was a one off, applicable to this release only.
I even considered adding support for PHPUnit 8 if it isn't too much work: https://github.com/Codeception/Codeception/discussions/6064

We can defer decision what we are going to do later until the next year.

This abstraction looks like a good idea, I'm just afraid that it won't age well.

DavertMik (approved):
The abstraction looks good to me!
I'm not following PHPUnit API changes so can't say in detail.
But I like the approach and I like class namings

PR #6144, opened 2021-03-16
Deprecating `GenerateCept`

ThomasLandauer:
The Cept format has been deprecated some 4 years ago https://github.com/Codeception/Codeception/issues/3976 - so I think it's about time to start thinking about ways to actually remove it from the codebase :-)
I deleted the explanation in order to remove the commands from https://codeception.com/docs/reference/Commands.html


DavertMik:
We can remove generator it in 5.0

PR #6067, opened 2020-12-28
Codeception 5.0

Naktibalda:
- [x] Merge [the latest branch](https://github.com/Codeception/phpunit-wrapper/tree/10.0) of phpunit-wrapper back to Codeception/Codeception
- [x] Merge "Improved compatibility with PHPUnit 7+" changes made in [codeception/codeception](https://github.com/Codeception/Codeception/pull/5894) and [phpunit-wrapper](https://github.com/Codeception/phpunit-wrapper/pull/77) repos.
- [x] Remove some old PHPUnit compatibility hacks
- [x] Remove support for Symfony 3 (raise the lowest supported version to 4.4)
- [x] Remove support for old versions of php-dotenv 2 and 3 (possibly 4 too).
- [x] Remove support for codeception/stub v2
- [x] Upgrade to the latest major versions of the following dependencies:
  - [x] ~~codeception/specify 0.* to 1.*~~ Removed dependency instead because it wasn't used by tests
  - [x] ~~monolog/monolog 1.* to 2.* (used by ext/Logger.php)~~ Removed dependency instead because Logger has no tests.
  - [x] ~~squizlabs/php_codesniffer 2.* to 3.*~~ Removed because it wasn't part of CI and code style wasn't enforced. It can be readded latter.
- [x] Delete deprecated code
  - [x]  [Codeception\Util\Stub](https://github.com/Codeception/Codeception/blob/4.1/src/Codeception/Util/Stub.php) was deprecated in 2.3.8, 3 years ago
  - [x] [settings: bootstrap:](https://github.com/Codeception/Codeception/commit/55e91e14567c7476bc29e778eda976d336c24f58) setting was deprecated in 3.1.0 in August 2019
- [x] Review all phpunit-wrapper code in src/PHPUnit/.
- [x] Run tests of require-dev modules in Github Actions (this change will be backported to 4.1)
- [ ] Make [codeception/stub](https://github.com/Codeception/stub) compatible with PHPUnit 10.
- [ ] Update files in src/PHPUnit/NonFinal directory with the latest revision of their counterparts in PHPUnit (the only difference was that they have no `final` keyword, because some Codeception classes extend them).
- [ ] Make it work with PHPUnit 10 after final release.
- [ ] Check how much work it is to make with PHPUnit 9. Support it if it is simple (e.g. a few version or method_exists conditions).
- [ ] Repeat this process with PHPUnit 8.

[...]

PR #6145, opened 2021-03-16
Removing Cept

ThomasLandauer:


DavertMik:
thanks, it is a good idea to start removing it from mentioning anywhere
but probably for now, we should just keep this format inside the codebase for BC and to keep our internal tests running as well.

ThomasLandauer:
OK, so please merge this now, to remove it from docs.
I added the Cept removal to the v5 TODO list: https://github.com/Codeception/Codeception/pull/6067#issuecomment-800576390

gimler:
@ThomasLandauer why was the cept format removed?

ThomasLandauer:
@gimler See https://github.com/Codeception/Codeception/pull/4259 for the beginning of the story :-)

If you're asking why Cept is eventually going to be dropped from Codeception altogether, I can only tell you my personal opinion: In short: Cest is PHP, and Gherkin is (somewhat) natural language. Cept is something in between. And, AFAIK, while Gherkin is some sort of standard today, Cept has always been Codeception-only.

PR #6147, opened 2021-03-16
Explaining `--debug` as alias for `-vv`

ThomasLandauer:
See https://github.com/Codeception/Codeception/discussions/6081#discussioncomment-275698

### DIFF
--- a/src/Codeception/Command/Run.php
+++ b/src/Codeception/Command/Run.php
@@ -27,9 +27,9 @@ use Symfony\Component\Console\Output\OutputInterface;
  *
  * * `codecept run -v`:
  * * `codecept run --steps`: print step-by-step execution
- * * `codecept run -vv`:
- * * `codecept run --debug`: print steps and debug information
- * * `codecept run -vvv`: print internal debug information
+ * * `codecept run -vv`: print steps and debug information
+ * * `codecept run --debug`: alias for `-vv`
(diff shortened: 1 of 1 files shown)


src/Codeception/Command/Run.php
```diff
@@ -27,9 +27,9 @@
  *
  * * `codecept run -v`:
  * * `codecept run --steps`: print step-by-step execution
- * * `codecept run -vv`:
- * * `codecept run --debug`: print steps and debug information
- * * `codecept run -vvv`: print internal debug information
```
Naktibalda, line 32: What is `internal debug information`? I have no idea, so I doubt that anyone else knows what it is too.

As far as I know, the main (and only?) difference is that `-vvv` displays full stack trace in error output.

ThomasLandauer: I've taken the info from https://github.com/Codeception/Codeception/discussions/6081#discussioncomment-273324
Further down, I've called it "Codeception-internal debug" - is that better? The idea is to give users a hint that `vvv` is just for searching bugs inside Codeception (isn't it?)

Naktibalda: Probably it is correct.  
The difference is that in lower levels of verbosity PHPUnit and Codeception classes are removed from displayed stack trace.

ThomasLandauer: So merge it as-is? Or change something?

PR #6156, opened 2021-04-04
Explaining how to uncheck a pre-checked checkbox

ThomasLandauer:
Sounds like a tongue twister ... ;-)

I was expecting this text to be in https://github.com/Codeception/lib-innerbrowser/blob/master/src/Codeception/Lib/InnerBrowser.php#L941 - while in fact it's even in another module! So, again, a link at each function to get to its source code would really be nice - see https://github.com/Codeception/Codeception/discussions/6063#discussioncomment-244542
(More precisely: If the documentation isn't where the function is, then *two* links)

### DIFF
--- a/src/Codeception/Lib/Interfaces/Web.php
+++ b/src/Codeception/Lib/Interfaces/Web.php
@@ -153,13 +153,15 @@ interface Web
      * For example, given this sample "Sign Up" form:
      *
      * ``` html
-     * <form action="/sign_up">
+     * <form id="userForm">
      *     Login:
      *     <input type="text" name="user[login]" /><br/>
      *     Password:
      *     <input type="password" name="user[password]" /><br/>
      *     Do you agree to our terms?
      *     <input type="checkbox" name="user[agree]" /><br/>
+     *     Subscribe to our newsletter?
+     *     <input type="checkbox" name="user[newsletter]" value="1" checked="checked" /><br/>
      *     Select pricing plan:
      *     <select name="plan">
      *         <option value="1">Free</option>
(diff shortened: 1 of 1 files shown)


Naktibalda:
Isn't this just repeating that [uncheckOption](https://codeception.com/docs/modules/Symfony#uncheckOption) actually unchecks checkbox?

Is it a naming problem? Would it be better to rename `uncheckOption` to `uncheckCheckbox`?

ThomasLandauer:
Well, let's look at this from the very beginning: Codeception has two ways to submit a form (right?):
* The "human" way: `fillField`, `selectOption`, etc., then `click` the submit button
* The "machine" way: `submitForm` and pass all values you want

The problem is that this unchecking scenario is the only (?) one where you need to *mix* those two ways.

So the (to me) perfect solution would be: Change `submitForm` so that it *just* submits what you pass in the array. I.e. do not submit "2" for the "plan" field in the example at https://codeception.com/docs/modules/Symfony#submitForm
Cause when I want to use what's on the page, I can just `click`. And if I want to test "fake/spam/whatever" submissions, I don't care what's on the page (even worse: If I don't fully understand this concept, some pre-set values might compromise my tests, in that they submit stuff I didn't think about).

Well, but changing `submitForm` would be a BC break...

And, yes, renaming `uncheckOption` to `uncheckCheckbox` is a good idea in any case! "Option" is an ambiguous term - e.g. `selectOption` refers to `<select>` *and* radio: https://codeception.com/docs/modules/Symfony#selectOption But checkbox is checkbox is checkbox :-)

Issue #5881, opened 2020-03-18
[Facebook\WebDriver\Exception\UnableToSetCookieException] unable to set cookie


voku:
#### What are you trying to achieve?

Running ```codecept.phar run``` after upgrading from v3 ⇉ v4

#### What do you get instead?

```bash
[Facebook\WebDriver\Exception\UnableToSetCookieException] unable to set cookie
```

### Details

* Codeception version: 4.1.2
* PHP Version: 7.4
* Operating System: Ubuntu 18.04
* Installation type: Phar

Naktibalda:
Could you confirm that v3 works with the same version of Selenium and browser?

voku:
Yep, I only tried to update the phar file.

laoneo:
Same issue for me.

Naktibalda:
I'm trying to pin it down what exactly is causing it.

@OndraM so far it looks like facebook/webdriver is good, php-webdriver/webdriver is bad - version doesn't matter.

```
$ codecept run tests/web/WebDriverTest.php:CookiesWithP -vv
Codeception PHP Testing Framework v4.1.3
Powered by PHPUnit 9.0.1 by Sebastian Bergmann and contributors.
Running with seed:


Web Tests (1) -------------------------------------------------------------------------------------------------------------------------------------------------------------
Modules: WebDriver, WebHelper
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------
- WebDriverTest: Cookies with path
  [GET] http://localhost:8001/info
  [Cookies] [{"name":"cookie","value":"tasty","path":"/info","domain":"localhost","expiry":1584868313,"secure":false,"httpOnly":false}]
  [Cookies] [{"name":"cookie","value":"tasty","path":"/info","domain":"localhost","expiry":1584868313,"secure":false,"httpOnly":false}]
  [Cookies] [{"name":"cookie","value":"tasty","path":"/info","domain":"localhost","expiry":1584868313,"secure":false,"httpOnly":false}]
  [Cookies] []
  [Cookies] []
  [Cookies] []
✔ WebDriverTest: Cookies with path (0.65s)
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------


Time: 10.38 seconds, Memory: 14.00 MB

OK (1 test, 5 assertions)
$ composer update
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 1 install, 0 updates, 1 removal
[...]

PR #6210, opened 2021-05-19
Coverage: Don't attempt to set cookie domain when it is "localhost"

marcovtwout:
Fixes https://github.com/Codeception/Codeception/issues/5881

### DIFF
--- a/src/Codeception/Coverage/Subscriber/LocalServer.php
+++ b/src/Codeception/Coverage/Subscriber/LocalServer.php
@@ -201,6 +201,13 @@ class LocalServer extends SuiteSubscriber
             // we need to separate coverage cookies by host; we can't separate cookies by port.
             $cookieDomain = isset($c3Url['host']) ? $c3Url['host'] : 'localhost';
         }
+        
+        $cookieParams = [];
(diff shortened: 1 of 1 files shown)


src/Codeception/Coverage/Subscriber/LocalServer.php
```diff
@@ -201,6 +201,13 @@ protected function startCoverageCollection($testName)
...
         }
+        
+        $cookieParams = [];
+        if ($cookieDomain !== 'localhost') {
+            $cookieParams['domain'] = $cookieDomain;
+        }
+
+        $this->module->setCookie(self::COVERAGE_COOKIE, $value, $cookieParams);
 
         $this->module->setCookie(self::COVERAGE_COOKIE, $value, ['domain' => $cookieDomain]);
```
Naktibalda, line 212: You missed the old setCookie line here, please delete it.

Naktibalda (changes requested):

src/Codeception/Coverage/Subscriber/LocalServer.php
marcovtwout: Thanks - not sure how I missed that.

Issue #1485, opened 2014-10-24
acceptPopup doesn't work when using CodeCoverage


rafaelbdb:
My acceptance tests works perfectly when I don't use --coverage (or --coverage-html), but when I do the following error appears (and I have a "$I->acceptPopup();" in this point of the code).

Command:

> php codecept.phar run acceptance myTestCept.php --coverage-html -d

```
[UnexpectedAlertOpenException]
  unexpected alert open
    (Session info: chrome=38.0.2125.104)
    (Driver info: chromedriver=2.10.267521,platform=Windows NT 6.1 SP1 x86_64
  ) (WARNING: The server did not provide any stacktrace information)
  Command duration or timeout: 1 milliseconds: null
  Build info: version: '2.43.0', revision: '597b76b', time: '2014-09-09 20:52
  :14'
  System info: host: 'Rafael-PC', ip: '10.1.1.14', os.name: 'Windows 7', os.a
  rch: 'x86', os.version: '6.1', java.version: '1.7.0_71'
  Session ID: c609d18ac1e286650052fa88369ed083
  Driver info: org.openqa.selenium.chrome.ChromeDriver
  Capabilities [{platform=XP, acceptSslCerts=true, javascriptEnabled=true, br
  owserName=chrome, chrome={userDataDir=C:\Users\Cliente\AppData\Local\Temp\s
  coped_dir6936_23574}, rotatable=false, locationContextEnabled=true, version
  =38.0.2125.104, takesHeapSnapshot=true, cssSelectorsEnabled=true, databaseE
  nabled=false, handlesAlerts=true, browserConnectionEnabled=false, nativeEve
  nts=true, webStorageEnabled=true, applicationCacheEnabled=false, takesScree
  nshot=true}]
Exception trace:
 () at phar://E:/Auto_testing/Codecept/codecept.phar/vendor/facebook/webdriver/lib/WebDriverExceptions.php:94
 WebDriverException::throwException() at phar://E:/Auto_testing/Codecept/codecept.phar/vendor/facebook/webdriver/lib/remote/HttpCommandExecutor.php:253
 HttpCommandExecutor::curl() at phar://E:/Auto_testing/Codecept/codecept.phar/vendor/facebook/webdriver/lib/remote/HttpCommandExecutor.php:153
 HttpCommandExecutor::remoteExecute() at phar://E:/Auto_testing/Codecept/codecept.phar/vendor/facebook/webdriver/lib/remote/HttpCommandExecutor.php:117
[...]

PR #6211, opened 2021-05-19
Coverage: Don't attempt to read cookies while an alert is open

marcovtwout:
Fixes https://github.com/Codeception/Codeception/issues/1485

Please check and test this workaround thoroughly.


Naktibalda:
Wouldn't it be better to catch exception thrown by `grabCookie`?

marcovtwout:
@Naktibalda It's actually not `grabCookie()` that throws the exception. Calling `grabCookie()` (which internally calls `WebDriver->manage()->getCookies()`) unexpectedly dismisses the alert and execution continues. Then the next call in your acceptance tests such as `$I->acceptPopup()` will fail since the alert is no longer present.

PR #6220, opened 2021-05-29
Type the interfaces

TavoNiievez:


Naktibalda (approved):
Looks good to me

DavertMik:
This is a good initiative but I'd rather not update interfaces for modules. 
This will break compatibility with all our current modules and 3rd-party modules as well. 

It is a really good idea to update the internal interfaces of Codeception to new standards. But all interfaces that are reused to other modules must be kept the same.

TavoNiievez:
Hi @DavertMik 
 
> "This will break compatibility with all of our current modules."

It's true, but I also promised to update the vast majority of modules (both the most downloaded and those that have received at least one commit in the last year). So this will not really affect anyone who requires the new versions of the modules and Codeception 5 because they will be compatible from day one.

> "and 3rd-party modules as well."

The amount of time required to update the implementation of these interfaces is minimal since the interface itself will guide the change.

> "all interfaces that are reused to other modules must be kept the same."

Forever and ever?
Please keep in mind that a major version is a great opportunity to do this.

Also remember that this directly affects the implementation of the Fluent interface in Codeception 5 that you agreed with.

Please take some time to reconsider.

DavertMik (approved):

DavertMik:
Thanks, @TavoNiievez I agree on all points.
If all modules will be updated I'm ok with this.

PR #6222, opened 2021-06-03
Fix Maybe::value()

TavoNiievez:
`$val` can also be an `array` or an `ArrayAccess` implementation.

### DIFF
--- a/src/Codeception/Util/Maybe.php
+++ b/src/Codeception/Util/Maybe.php
@@ -29,7 +29,6 @@ use function range;
  * <?php
  * $user = new Maybe;
  * $user->posts->comments->count();
- * ?>
  * ```
  */
 class Maybe implements ArrayAccess, Iterator, JsonSerializable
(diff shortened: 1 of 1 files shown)


src/Codeception/Util/Maybe.php
```diff
@@ -39,7 +38,7 @@ class Maybe implements ArrayAccess, Iterator, JsonSerializable
      */
     protected $position = 0;
     /**
-     * @var null|object
+     * @var ArrayAccess|array|object|null
```
Naktibalda, line 41: How can it be object if constructor permits only array and null?

TavoNiievez: Well, in various places in the file `is_object` and `property_exists` are used with this property (`$val`), maybe that typing is wrong too (?).

Naktibalda: I think that it would be better to set it to `mixed` and remove all typehints.

There wree no restrictions on value types before.

PR #6242, opened 2021-08-07
Added more PHP types

TavoNiievez:
Final final final review of PHP types

### DIFF
--- a/src/Codeception/Command/Run.php
+++ b/src/Codeception/Command/Run.php
@@ -321,11 +321,7 @@ class Run extends Command
         $userOptions['interactive'] = !$input->hasParameterOption(['--no-interaction', '-n']);
         $userOptions['ansi'] = (!$input->hasParameterOption('--no-ansi') xor $input->hasParameterOption('ansi'));
 
-        if (!$this->options['seed']) {
-            $userOptions['seed'] = rand();
-        } else {
-            $userOptions['seed'] = (int) $this->options['seed'];
-        }
+        $userOptions['seed'] = $this->options['seed'] ? (int) $this->options['seed'] : rand();
         if ($this->options['no-colors'] || !$userOptions['ansi']) {
             $userOptions['colors'] = false;
         }
@@ -481,13 +477,13 @@ class Run extends Command
             // Run single test from working directory
             $realTestDir = realpath(Configuration::testsDir());
             $cwd = getcwd();
-            if (strpos($realTestDir, $cwd) === 0) {
+            if (strpos($realTestDir, (string) $cwd) === 0) {
                 $file = $suite;
                 if (strpos($file, ':') !== false) {
                     [$file] = explode(':', $suite, -1);
                 }
                 $realPath = $cwd . DIRECTORY_SEPARATOR . $file;
-                if (file_exists($realPath) && strpos($realPath, $realTestDir) === 0) {
+                if (file_exists($realPath) && strpos($realPath, (string) $realTestDir) === 0) {
                     //only match test if file is in tests directory
                     return $this->matchTestFromFilename(
                         $cwd . DIRECTORY_SEPARATOR . $suite,
@@ -554,7 +550,10 @@ class Run extends Command
         return $executed;
     }
 
-    protected function matchTestFromFilename($filename, $testsPath)
+    /**
+     * @return string[]
+     */
+    protected function matchTestFromFilename($filename, $testsPath): array
     {
         $filter = '';
         if (strpos($filename, ':') !== false) {
--- a/src/Codeception/Command/Run.php
+++ b/src/Codeception/Command/Run.php
@@ -475,15 +475,15 @@ class Run extends Command
             }
 
             // Run single test from working directory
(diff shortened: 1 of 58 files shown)


TavoNiievez:
I think Nitpick-CI's comments in this case can be considered false positives (?)

src/Codeception/Test/Cept.php
```diff
@@ -47,7 +47,7 @@ public function test(): void
         try {
             require $testFile;
         } catch (ParseError $error) {
-            throw new TestParseException($testFile, $error->getMessage(), $error->getLine());
```
Naktibalda, line 50: I think that it was mistake to set type of `$line` to `string`.

src/Codeception/Test/Cest.php
```diff
@@ -114,7 +117,7 @@ protected function executeHook($I, string $hook): void
         }
     }
 
-    protected function executeBeforeMethods($testMethod, $I): void
+    protected function executeBeforeMethods(?string $testMethod, $I): void
```
Naktibalda, line 120: Why is it nullable? It doesn't make sense.

src/Codeception/Test/Cest.php
```diff
@@ -123,7 +126,7 @@ protected function executeBeforeMethods($testMethod, $I): void
             }
         }
     }
-    protected function executeAfterMethods($testMethod, $I): void
+    protected function executeAfterMethods(?string $testMethod, $I): void
```
Naktibalda, line 129: same here

src/Codeception/Command/Run.php
```diff
@@ -481,13 +477,13 @@ protected function matchSingleTest($suite, $config): ?array
...
             $cwd = getcwd();
-            if (strpos($realTestDir, $cwd) === 0) {
+            if (strpos($realTestDir, (string) $cwd) === 0) {
                 $file = $suite;
                 if (strpos($file, ':') !== false) {
                     [$file] = explode(':', $suite, -1);
                 }
                 $realPath = $cwd . DIRECTORY_SEPARATOR . $file;
-                if (file_exists($realPath) && strpos($realPath, $realTestDir) === 0) {
+                if (file_exists($realPath) && strpos($realPath, (string) $realTestDir) === 0) {
```
Naktibalda, line 486: `$cwd` and `$realTestDir` are used as strings in other places too, I think that it would be better to do casting on assignment.

src/Codeception/Command/Run.php
```diff
@@ -554,7 +550,10 @@ protected function runSuites(array $suites, array $skippedSuites = []): int
         return $executed;
     }
 
-    protected function matchTestFromFilename($filename, $testsPath)
+    /**
+     * @return string[]
+     */
+    protected function matchTestFromFilename($filename, $testsPath): array
```
Naktibalda, line 556: no parameter type declarations here?

src/Codeception/Coverage/Subscriber/LocalServer.php
```diff
@@ -74,6 +74,9 @@ class LocalServer extends SuiteSubscriber
      */
     const COVERAGE_COOKIE_ERROR = 'CODECEPTION_CODECOVERAGE_ERROR';
 
+    /**
+     * @var string|null
+     */
     protected $suiteName;
```
Naktibalda, line 80: It is string in all places that matter.

src/Codeception/Lib/Console/Output.php
```diff
@@ -76,7 +76,7 @@ protected function clean(string $message): string
         return $message;
     }
 
-    public function debug($message): void
+    public function debug(string $message): void
```
Naktibalda, line 79: `debug` can print any type of data (see [Debug class](https://github.com/Codeception/Codeception/blob/5.0/src/Codeception/Util/Debug.php#L28))

src/Codeception/Util/Autoload.php
```diff
@@ -80,6 +80,9 @@ public static function addNamespace(string $prefix, string $baseDir, bool $prepe
         }
     }
 
+    /**
+     * @return string|false
+     */
     public static function load($class)
```
Naktibalda, line 86: `$class` must be `string`.

Naktibalda (changes requested):

src/Codeception/Test/Cest.php
TavoNiievez: `$testMethod` is nullable in PHPUnit 9.5 and lower https://github.com/sebastianbergmann/phpunit/blob/9550b9cb7461584ed4beeb74d119d01808000a8c/src/Util/Test.php#L328 .

do I prevent it from being nullable here?

Naktibalda: That means, that PHPUnit method can be used for something else. We don't use it for that purpose.

Cest methods must have names, so they are not nullable,.

TavoNiievez: thanks. Done.

Naktibalda (changes requested):

PR #6252, opened 2021-09-12
4.1 PHP 8.1 support

Naktibalda:
In PHP 8.1 htmlspecialchars uses different flags by default, so I set these flags explicitly to get consistent results in tests.

Also I fixed 3 type issues that caused deprecations or fatal errors in PHP 8.1

Codeception 5.0 change is bigger because it includes code that was previously part of phpunit-wrapper repository.


francislavoie:
@Naktibalda can we get a new release with this fix? PHP 8.1 is now released.

nforced:
Indeed, this will be very nice to have otherwise we have to wait and migrate everything to 5.x which reminds me there are some modules currently not compatible with 5.x (even in dev, redis for example) which will make it harder to further migrate.

Naktibalda:
I released it as 4.1.23

Issue #6259, opened 2021-10-07
Add links to runner output that open the file directly in my IDE


ThomasLandauer:
I started implementing what I suggested at https://github.com/Codeception/Codeception/discussions/6208 :

First step: The line https://github.com/Codeception/Codeception/blob/4.1/src/Codeception/Subscriber/Console.php#L360 needs to be replaced by something like:
```php
->append(codecept_absolute_path(Descriptor::getTestFileName($failedTest)))
```
My plan would be to define a new configuration parameter in `codeception.yml`, following phpstan's format (see https://phpstan.org/user-guide/output-format#opening-file-in-an-editor):
```yaml
editorUrl: 'phpstorm://open?file=%%file%%&line=%%line%%'
```
And if this parameter is present, replace the value of "Test" here:
```
1) FooCest: Try bar
 Test  tests/functional/FooCest.php:tryBar
 Step  See whatever
 Fail  Number of elements counted differs from expected range
Failed asserting that false is true.
```
...with the clickable link.

Questions:
1. How can I access a configuration parameter from there?
2. How can I access the failing line number? If that's not so easy, let's just omit it for now - a link to just the *file* is still better than nothing ;-)

Naktibalda:
> 1. How can I access a configuration parameter from there?

You get all settings from `settings` section of codeception.yaml file in `$this->options`  when `codeception run` command is used.  `codecept dry-run` doesn't get them, so it is always necessary to check if `$this->options['editorUrl']` is set before using it.


> 2. How can I access the failing line number?

`$failEvent->getFail()` returns exception.
You can use `getFile` and `getLine` methods to find the origin place of exception, but it won't be in the test file.
You could iterate stack trace retuned by $failEvent->getFail()->getTrace ` and find the first row matching the name of test file.

PR #6261, opened 2021-10-08
Adding clickable `editorUrl` from `codeception.yml`

ThomasLandauer:
Closes https://github.com/Codeception/Codeception/issues/6259

* I followed the syntax of https://phpstan.org/user-guide/output-format#opening-file-in-an-editor even though we wouldn't need the second `%` here. (This is due to some phpstan internals, the code ultimately reads `\str_replace(['%file%', '%line%'], ...`).
But I'd say having an *identical* `editorUrl` syntax between Codeception and phpstan by far outweighs the "cost" of some unneeded `%` characters... ;-)

* `$line = 0;` is a fallback in case the trace iteration doesn't work out. I'm guessing that putting 1 for the line is better than leaving it empty.

### DIFF
--- a/src/Codeception/Step.php
+++ b/src/Codeception/Step.php
@@ -79,10 +79,17 @@ abstract class Step
         return $this->action;
     }
 
-    public function getLine()
+    public function getFilePath()
     {
-        if ($this->line && $this->file) {
-            return codecept_relative_path($this->file) . ':' . $this->line;
+        if ($this->file) {
+            return codecept_relative_path($this->file);
+        }
+    }
+
+    public function getLineNumber()
+    {
+        if ($this->line) {
+            return $this->line;
         }
     }
 
--- a/src/Codeception/Step.php
+++ b/src/Codeception/Step.php
@@ -79,6 +79,16 @@ abstract class Step
         return $this->action;
     }
 
+    /**
+     * @deprecated To be removed in Codeception 5.0
+     */
+    public function getLine()
+    {
+        if ($this->line && $this->file) {
+            return codecept_relative_path($this->file) . ':' . $this->line;
+        }
+    }
+
     public function getFilePath()
     {
         if ($this->file) {
--- a/src/Codeception/Subscriber/Console.php
+++ b/src/Codeception/Subscriber/Console.php
@@ -367,7 +367,9 @@ class Console implements EventSubscriberInterface
                     $line = $trace['line'];
                 }
             }
-            $message = str_replace(['%%file%%', '%%line%%'], [$filePath, $line], $this->options['editorUrl']);
+            if ($line > 1) {
+                $message = str_replace(['%%file%%', '%%line%%'], [$filePath, $line], $this->options['editorUrl']);
+            }
         } else {
             $message = codecept_relative_path(Descriptor::getTestFullName($failedTest));
         }
--- a/src/Codeception/Subscriber/Console.php
+++ b/src/Codeception/Subscriber/Console.php
@@ -358,18 +358,15 @@ class Console implements EventSubscriberInterface
         $this->output->writeln('');
 
         // Clickable `editorUrl`:
-        if (isset($this->options['editorUrl']) && is_string($this->options['editorUrl'])) {
+        if (isset($this->options['editorUrl']) and is_string($this->options['editorUrl'])) {
             $filePath = codecept_absolute_path(Descriptor::getTestFileName($failedTest));
-            $message = str_replace('%%file%%', $filePath, $this->options['editorUrl']);
             $line = 1;
             foreach ($fail->getTrace() as $trace) {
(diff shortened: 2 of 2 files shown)


src/Codeception/Subscriber/Console.php
```diff
@@ -536,9 +557,17 @@ public function printScenarioTrace(ScenarioDriven $failedTest)
...
+                $filePath = $step->getFilePath();
+                if ($filePath) {
+                    // Clickable `editorUrl`:
+                    if (isset($this->options['editorUrl']) and is_string($this->options['editorUrl'])) {
+                        $lineString = str_replace(['%%file%%', '%%line%%'], [codecept_absolute_path($step->getFilePath()), $step->getLineNumber()], $this->options['editorUrl']);
+                    } else {
+                        $lineString = $step->getFilePath() . ':' . $step->getLineNumber();
+                    }
+                }
+                $message->append(" at <info>$lineString</info>");
```
Naktibalda, line 570: if `$filePath` is null, $lineString is undefined, please move this line inside `if`.

Naktibalda (changes requested):

src/Codeception/Subscriber/Console.php
ThomasLandauer: :-)

src/Codeception/Subscriber/Console.php
```diff
@@ -356,8 +356,22 @@ public function printFail(FailEvent $e)
         $this->output->write($e->getCount() . ") ");
         $this->writeCurrentTest($failedTest, false);
         $this->output->writeln('');
+
+        // Clickable `editorUrl`:
+        if (isset($this->options['editorUrl']) && is_string($this->options['editorUrl'])) {
+            $filePath = codecept_absolute_path(Descriptor::getTestFileName($failedTest));
+            $line = 1;
+            foreach ($fail->getTrace() as $trace) {
+                if (isset($trace['file']) and $filePath === $trace['file'] and isset($trace['line'])) {
```
Naktibalda, line 365: Please replace all other instances of `and` with `&&`.

src/Codeception/Subscriber/Console.php
```diff
@@ -492,7 +506,14 @@ public function printExceptionTrace($e)
                 $message->writeln();
                 continue;
             }
-            $message->append($step['file'] . ':' . $step['line']);
+
+            // Clickable `editorUrl`:
+            if (isset($this->options['editorUrl']) and is_string($this->options['editorUrl'])) {
```
Naktibalda, line 511: here too.

src/Codeception/Subscriber/Console.php
```diff
@@ -536,9 +557,17 @@ public function printScenarioTrace(ScenarioDriven $failedTest)
...
             }
 
-            $line = $step->getLine();
-            if ($line and (!$step instanceof Comment)) {
-                $message->append(" at <info>$line</info>");
+            if (!$step instanceof Comment) {
+                $filePath = $step->getFilePath();
+                if ($filePath) {
+                    // Clickable `editorUrl`:
+                    if (isset($this->options['editorUrl']) and is_string($this->options['editorUrl'])) {
```
Naktibalda, line 564: and here.

src/Codeception/Step.php
```diff
@@ -79,10 +79,17 @@ public function getAction()
         return $this->action;
     }
 
-    public function getLine()
+    public function getFilePath()
```
Naktibalda, line 82: `phpunit-wrapper` uses `getLine` in 2 places:
https://github.com/Codeception/phpunit-wrapper/search?q=getLine .

It would be the best to restore `getLine` method with original behaviour here and delete it in Codeception 5.0

Naktibalda (changes requested):

src/Codeception/Step.php
ThomasLandauer: OK. I added a note there - but I probably won't remove it myself then...

Naktibalda (approved):

TavoNiievez (approved):

Naktibalda:
@ThomasLandauer Could you document it in https://github.com/Codeception/codeception.github.com/blob/master/docs/reference/Configuration.md ?

ThomasLandauer:
@Naktibalda Just noticed that all other config options are `snake_case`. I just copied `editorUrl` from PHPStan. Should we change it to `editor_url` before it gets released?

Naktibalda:
I released it yesterday, but nobody will notice if we rename it quickly, right?

Naktibalda:
I raised https://github.com/Codeception/Codeception/pull/6303

Naktibalda:
Renamed and released as 4.1.27

ThomasLandauer:
> Could you document it

Sure :-) https://github.com/Codeception/codeception.github.com/pull/582

PR #6263, opened 2021-10-13
Move functions into separate file and autoload with composer

StuTheWebGuy:
When running Psalm and other static analysers, the functions,
for example `codecept_data_dir()`, cannot be found because
they are not autoloaded with Composer.

Now the functions have their own file, `functions.php`, and
are autoloaded with Composer as "files" in the `composer.json`.

Co-authored-by: Ade Attwood <ade@practically.io>


Naktibalda:
Thanks,

we implemented this change in 5.0 branch almost a year ago, but it is still waiting for PHPUnit 10 release.
https://github.com/Codeception/Codeception/blob/5.0/functions.php

StuTheWebGuy:
No problem - is there anything I can do to get this merged in, or is it fine as-is?

PR #6265, opened 2021-10-18
Code Coverage of PHP files only

Naktibalda:
[Code Coverage documentation](https://codeception.com/docs/11-Codecoverage#Configuration) gives examples for file patters ending with `*`.

```yml
coverage:
    enabled: true
    include:
        - app/*
    exclude:
        - app/cache/*
```

Recently [it was brought to my attention](https://stackoverflow.com/questions/69443881/include-exclude-files-for-remote-coverage-in-codeception) that such rule adds non-PHP files to code coverage report too.
I tested it myself and got feature files, yaml files and everything in tests/_output directory added to code coverage report.

image

So I made this simple patch to change `/*` to `/*.php`.
Alternatively users could use `- app/*.php` in their include list, but it is better to have sane defaults.

If someone thinks that this is a breaking change, it could be merged to 5.0 branch instead.

### DIFF
--- a/src/Codeception/Coverage/Filter.php
+++ b/src/Codeception/Coverage/Filter.php
@@ -167,6 +167,9 @@ class Filter
         $fileOrDir = str_replace('\\', '/', $pattern);
         $parts = explode('/', $fileOrDir);
         $file = array_pop($parts);
+        if ($file === '*') {
+            $file = '*.php';
+        }
         $finder->name($file);
         if (count($parts)) {
             $last_path = array_pop($parts);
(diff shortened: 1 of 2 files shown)


TavoNiievez:
Would it be fair to start excluding the `.phtml` and `.pht` extensions?
It is not difficult to imagine people writing classes and other source code in them.

I agree with this change if it includes extensions where there is usually PHP code... regardless of whether by default it is not considered part of the application domain.

Regarding `.phar` and `.phps` I would agree to exclude them.

DavertMik (approved):

DavertMik:
Let's have it with: 

```
- app/*
```

Good idea. Thanks!

PR #6273, opened 2021-11-07
Symfony 6 support

seferov:


Naktibalda:
Are you going to fix all breaking changes too?

Please don't set constraints without upper boundary.  
Even if there is no breaking changes in 6.0 (unlikely), 7.0 or 8.0 may introduce breaking changes.

CI didn't run any builds with Symfony 6 because 6.0.0-beta1 didn't match minimum stability.  
Lets wait for 6.0.0 or change composer.json to allow installation of beta1.

seferov:
> Are you going to fix all breaking changes too?

@Naktibalda yes, can you please approve the build?

TavoNiievez:
IMHO this PR should target the `5.0` branch.
Even I don't think it's correct to remove compatibility with php 5.6 -> 7.1 in Codeception 4.

Unfortunately, PHPUnit 10 has yet to be released and to have a release ready from day 1 maybe this is one of the options...

Anyway, I'm approving execution of the CI so as not to block this job.

src/Codeception/Test/Feature/IgnoreIfMetadataBlocked.php
```diff
@@ -8,7 +8,7 @@ trait IgnoreIfMetadataBlocked
     /**
      * @return Metadata
      */
-    abstract protected function getMetadata();
+    abstract public function getMetadata();
```
TavoNiievez, line 11: Why is this change necessary?

DavertMik:
Yes, I agree this should be part of 5.0 release
There are failing tests and we need to fix them before merging.

TavoNiievez (approved):

seferov:
@TavoNiievez @DavertMik thank you for the review.

it looks like after rebasing it to 5.0 nothing much left :) most of the work was already done in 5.0 before.

DavertMik:
Thanks! Let's have it merged!
At this moment we are waiting for PHPUnit 10 to be released to make Codeception 5
Not sure how long it goes but you can work from 5.x branch meanwhile

seferov:
@DavertMik thanks!

Issue #6270, opened 2021-10-26
Useless shortlink in deprecation warning for settings.bootstrap


DBX12:
#### What are you trying to achieve?
Learn about correctly removing the `settings.bootstrap` property from the codeception.yml config file. I expected to be redirected to the appropriate codeception docs section.

#### What do you get instead?
A (for me) useless Russian website.

#### How to reproduce?
Configure the `bootstrap` setting as shown below and run the tests. You should get a message like this:

> DEPRECATION: 'settings: bootstrap: _bootstrap.php' option is deprecated! Replace it with: 'bootstrap: _bootstrap.php' (not under settings section). See https://bit.ly/2YrRzVc

Or see the link in the code here https://github.com/Codeception/Codeception/blob/76ed7815922df2b958a5a8368acbb3532f7652eb/src/Codeception/Configuration.php#L259

### Details

* Codeception version: 4.1.22

```yml
# several config options
settings:
    bootstrap: _bootstrap.php
# several other config options
```

----

**Edit:** I know how where to move the bootstrap to fix my config, my issue is rather the link not pointing to the docs.

Naktibalda:
Fixed in linked pull request

PR #6274, opened 2021-11-09
Replaces outdated short URL for the bootstrap config option

p810:
Addresses #6270.

I was looking for a URL shortener that I could use, but apparently all of the big ones require you to sign up for an account nowadays. I also wasn't sure if it would be accepted since I don't know if any of these services let you change the target URL after a short URL's created, so I figured I'd just put the full link in there.

At first I was going to compensate for the length by breaking it out onto a new line, but from what I can tell the package Codeception uses for CLI output already handles this if I'm not mistaken.

If that's not the case and/or a short URL would be better, let me know and I'll update it.

### DIFF
--- a/src/Codeception/Configuration.php
+++ b/src/Codeception/Configuration.php
@@ -256,7 +256,7 @@ class Configuration
 
         if ($config['settings']['bootstrap']) {
             $bootstrap = self::$config['settings']['bootstrap'];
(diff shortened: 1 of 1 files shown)


Arhell (approved):

TavoNiievez (approved):

DavertMik:
Thanks!

Issue #6152, opened 2021-04-01
Request: --fail-slow=10


scottconnerly:
#### What are you trying to achieve?
A halfway point between --fail-fast and running the full suite

### Details
--fail-fast gives quick feedback loops
running the full suite gives all the context you need to figure out why something might be failing.

But what if there was a middle ground where, in a run where a LOT of tests were destined to fail, you still got more context on those failures than just the first one that failed? Like, maybe one could specify the number of tests that needed to fail before abandoning the rest of the suite?

Thus I propose a counter called --fail-slow that after X number of tests fail, the run quits. If < X number of tests fail, the suite completes (but still returns a failure error code, of course).

PR #6275, opened 2021-11-09
Added optional value to fail-fast option

Verest:
https://github.com/Codeception/Codeception/issues/6152

First time working with this code base. This appears to work, although ran into a few issues.
1. Not super happy about having to cache the suite object in order to potentially stop it later. Is there some form of a container present for a better way to access this?
2. Had to be a bit hacky around the failure tests. Ended out just making a separate test class/group for multiple failures.
3. I could not get 5 of the `RemoteWith...` tests to pass locally. However, I don't think this is related to my changes.
  
Aside that, let me know if any requested changes or bugs found. Intent was to keep this working with boolean values as well for `fail-fast`, in terms of the existing configuration files.

Not sure if this falls into the "merge into 4.1" or "merge into master" category, defaulting to 4.1 for the time being. Let me know if it should be changed to master.

### DIFF
--- a/src/Codeception/Codecept.php
+++ b/src/Codeception/Codecept.php
@@ -133,7 +133,7 @@ class Codecept
             $this->dispatcher->addSubscriber(new Subscriber\Console($this->options));
         }
         if ($this->options['fail-fast']) {
-            $this->dispatcher->addSubscriber(new Subscriber\FailFast());
+            $this->dispatcher->addSubscriber(new Subscriber\FailFast($this->options['fail-fast']));
         }
 
         if ($this->options['coverage']) {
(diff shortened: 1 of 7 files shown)


DavertMik:
Looks good to me! Absolutely needed feature and proper implementation!

Verest:
@DavertMik Well... it somehow is failing with windows-only php8.1. Is this a related failure to my changes?

I am using linux and cannot replicate this error currently. At the very least, this test (` SuiteManagerTest: Few tests`) does not fail on linux php8.1.

DavertMik:
PHP 8,1 and Windows.. what a combo. It is related because the only test is failing is the one you added. 
Well, you can skip this test on PHP 8.1 and Windows for now. Let's re-enable if we get more information on this cause.

Verest:
In retrospect, I am not sure if php8.1 was actually ran via ubuntu based on the details, although still passes for me locally. Technically I suppose php8.1 isn't even released yet :laughing: .

Unless I am missing something, I did not modify the test that is failing - it appears `tests\unit\Codeception\SuiteManagerTest.php:testFewTests` is the culprit. Did you want me to add logic to skip this test if windows/php8.1 is detected?

DavertMik:
> Did you want me to add logic to skip this test if windows/php8.1 is detected?

Yes, please do

Verest:
@DavertMik all green now~

PR #6278, opened 2021-11-19
Update codebase to PHP 7.4

TavoNiievez:
This PR does not include the update to PHP 7.4 of the Codeception `/tests` folder.

It was necessary to disable the `db` and `cli` tests for PHP 8.1 on Windows as they failed even without making any changes.

Last but not least, I am skipping updating the following files:

- `src/Codeception/Extension.php`
- `src/Codeception/Module.php`

Since to get a correct build I would need updates to all modules and extensions as well.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -79,6 +79,7 @@ jobs:
         run: php codecept run -c vendor/codeception/module-filesystem/
 
       - name: Run module-db sqlite tests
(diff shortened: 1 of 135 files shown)


DavertMik (approved):
Wow, this is huge! Thank you!

TavoNiievez:
@DavertMik I think this is ready to merge.

PR #6283, opened 2021-12-02
Update tests folder to PHP 7.4

TavoNiievez:


tests/data/claypit/c3.php
```diff
@@ -14,12 +14,12 @@
 use SebastianBergmann\CodeCoverage\Filter as CodeCoverageFilter;
 
 if (isset($_COOKIE['CODECEPTION_CODECOVERAGE'])) {
-    $cookie = json_decode($_COOKIE['CODECEPTION_CODECOVERAGE'], true);
+    $cookie = json_decode($_COOKIE['CODECEPTION_CODECOVERAGE'], true, 512, JSON_THROW_ON_ERROR);
```
Naktibalda, line 17: @TavoNiievez You update copy which is only used for testing, but original at https://github.com/Codeception/c3/blob/main/c3.php was not updated, so your changes made to this file will be lost eventually.

TavoNiievez: @Naktibalda you're right, I forgot.
I would like to see [#72](https://github.com/Codeception/c3/pull/72) merged before making these kinds of changes.

PR #6300, opened 2021-12-21
Patch Group Creation with Wildcard

DanielHansmannAldi:
Patch Group Creation with Wildcard to realy work with Filename and not only iterate $i


Naktibalda:
What problem does it solve?

Please add test to https://github.com/Codeception/Codeception/blob/4.1/tests/unit/Codeception/Lib/GroupManagerTest.php or another relevant test file.

DavertMik:
Thanks, @Naktibalda 
This concerns a setup in parallelization I will review it and make a release with this patch.

DavertMik:
Thanks, @DanielHansmannAldi 
I improved the implementation, see #6302

Issue #6029, opened 2020-11-03
`success` and `fail` characters (checkmark and "x") are displayed without color (i.e. grey)


ThomasLandauer:
Instead of a green ✔ and a red ✖, here's what the output looks like for me:
![Screenshot_20201103_164013](https://user-images.githubusercontent.com/1054469/98007399-a3f60580-1df3-11eb-84f7-10a12618c1ef.png)

These characters are already around since Codeception 2.4 https://github.com/Codeception/Codeception/blob/2.4/src/Codeception/Subscriber/Console.php#L104 - so my main question is: **Is anybody else observing this?**

I think this happened after upgrading from kubuntu 18.04 to 20.04. Looking closely at the screenshot, you can see that the characters are just rendered too large, and I'm guessing that this also prevents them from being colored.

Naktibalda:
It is an issue of the terminal or font that you are using.

Since 2018 I run my tests on Ubuntu installed on WSL and I was using default Windows terminal application all this time and seeing broken Unicode chacters instead of unicode success and fail characters.

![image](https://user-images.githubusercontent.com/395992/103366918-72847800-4acc-11eb-8ed3-438eeeeb8c39.png)

Thanks to your question, I did some research and found [this article](https://blog.nillsf.com/index.php/2020/02/17/setting-up-wsl2-windows-terminal-and-oh-my-zsh/) which suggests to install different app called Windows Terminal.

Now I can see these characters:

![image](https://user-images.githubusercontent.com/395992/103367027-b7a8aa00-4acc-11eb-812e-74d79b44106a.png)

ThomasLandauer:
It's in fact a bug in my terminal ("Konsole"): https://bugs.kde.org/show_bug.cgi?id=349088

Naktibalda:
@ThomasLandauer Is there anything wrong with default `+` and 'x' characters that are used in `no-ansi` mode?

You get them if you use  `codecept run --no-ansi`. Is is to cumbersome to use additional parameter?

ThomasLandauer:
Well, this removes *all* color from the output, so it's even worse than with the misrendered symbols ;-)

Naktibalda:
Try `--no-ansi --colors` combination.

PR #6306, opened 2021-12-22
Adding options `chars_success` and `chars_fail`

ThomasLandauer:
There's still no real solution for https://github.com/Codeception/Codeception/issues/6029, so in fact I'm manually changing the characters in my `vendor` directory after each Codeception release ;-)

So now I was finally looking for a more durable solution. And since this was really easy to implement, I'm suggesting: Let's allow people to just use *any* character!

### DIFF
--- a/src/Codeception/Subscriber/Console.php
+++ b/src/Codeception/Subscriber/Console.php
@@ -107,6 +107,13 @@ class Console implements EventSubscriberInterface
             $this->chars['success'] = '✔';
             $this->chars['fail'] = '✖';
         }
+        
+        if (is_string($this->options['chars_success'])) {
+            $this->chars['success'] = $this->options['chars_success'];
+        }
+        if (is_string($this->options['chars_fail'])) {
+            $this->chars['fail'] = $this->options['chars_fail'];
+        }
 
         foreach (['html', 'xml', 'phpunit-xml', 'tap', 'json'] as $report) {
             if (!$this->options[$report]) {
--- a/src/Codeception/Subscriber/Console.php
+++ b/src/Codeception/Subscriber/Console.php
@@ -107,7 +107,8 @@ class Console implements EventSubscriberInterface
             $this->chars['success'] = '✔';
             $this->chars['fail'] = '✖';
         }
-        
+
+        // Allow custom "success" and "fail" characters:
         if (is_string($this->options['chars_success'])) {
             $this->chars['success'] = $this->options['chars_success'];
         }


Naktibalda:
It looks good, but I think that option names could be improved.

ThomasLandauer:
Well, what would you suggest? `chars_test_success`?

Naktibalda:
Actually, I don't like the `chars` part.
Even full word `character` would be better.

The structure doesn't have to be flat, it could be nested like this:
```yaml
characters:
  success: 'S',
  fail: 'F'
```

But it still lacks the purpose.

How about going this deep?
```yaml
report:
  characters:
    success: 'S',
    fail: 'F'
```

Or flattening a little bit to:

```yaml
report:
    success_character: 'S',
    fail_character: 'F'
```

ThomasLandauer:
* I went for plural, cause you can pass multiple characters.
* A Deeper structure (array) is not worth it for just *two* entries. This would only make the syntax more complicated.

More self-explanatory would be `characters_for_test_success`.

DavertMik:
Can we just avoid having this and make a proper detection if output has UTF support. There are probably libraries which can do that... In non-UTF terminal we fallback to standard ANSI chars.

But to conclude:
- I'm absolutely against adding this to options... Too specific usage and too much mess with special chars
- I'm ok with adding this to config if there are no other way to detect and fallback to ANSI chars

ThomasLandauer:
@DavertMik: My terminal ("Konsole") does have UTF8 support. And (as I explained at https://bugs.kde.org/show_bug.cgi?id=349088) it also works if I add a blank: `$this->chars['success'] = '✔ ';`. However, the green color is still missing. So detecting this automatically will probably never work.

Besides: Adding another dependency which will eventually end up like `hoa/console` (i.e. unmaintained but still required by Codeception, cause nobody is taking the time to search for a replacement) is no good idea, IMO. Even more, since the same can be achieved by just 8 (!) lines of code!

And giving the user the opportunity to use whichever string they like, is an added benefit.


> and too much mess with special chars

Which mess? It's the user's responsibility to pass a character that is working in their terminal.

DavertMik:
> Which mess? It's the user's responsibility to pass a character that is working in their terminal.

The mess with options... The more we add them the harder will be to use...

Ok, if you need this, it should be in configuration...

ThomasLandauer:
> The mess with options... The more we add them the harder will be to use...

No! As long as I find "my" option quickly, I don't care if there are 3 or 300 possible others. So it's just a question of documentation - which I just overhauled :-) https://codeception.com/docs/reference/Configuration

> Ok, if you need this, it should be in configuration...

Yes, please! So what's next?

ThomasLandauer:
Ping @DavertMik

DavertMik:
@ThomasLandauer please send a new pull request which introduces changes into configuration and not into option.

sorry for delay in reply, really sorry.

And sorry that I have to close this PR. But edge case things should be available in configuration.

ThomasLandauer:
@DavertMik Sorry, what do you mean by "configuration"? I was thinking those options are the configuration... ;-)

PR #6308, opened 2021-12-31
Symfony 6 Support for Codeception 4

TavoNiievez:
We can't keep waiting for the release of Codeception 5 to do this.

Currently it is not possible to install Codeception 4 and Symfony 6 at the same time. So i'm suggesting that we release this as Codeception 4.2, Only people on PHP 8.0 or higher **and** Symfony 6 will get this version.

Doing this has a clear disadvantage... that the changes from 4.1 must be replicated in 4.2 as well as 5.0 as long as the 4.1 branch continues to receive support.
It is not the most orthodox solution but it is one of the alternatives that must be taken into account.


TavoNiievez:
@Naktibalda and @DavertMik If the answer is still "let's keep waiting" this PR could be closed without problems :)

Naktibalda:
If it requires raising lowest supported version of PHP to 8.0, it is a major release in everything but name.

Releasing it as 4.2 is a bad idea, I would rather rename existing 5.0 branch to 6.0 and release Symfony 6 support as 5.0.0.
But there is no reason not to release what we got in 5.0 branch, right?

It is very likely to be a short support branch if PHPUnit 10 is released in February.

We can tag 5.0.0 release on 5.0 branch right now, right?

TavoNiievez:
What has been roughly done in 5.0 so far is removing compatibility code, moving the required version of PHP to 7.4 and adding support for Symfony 6.

The only thing I would change again before releasing 5.0 is to move again:
`"minimal stability"` from `"beta",` to `"RC"` in `composer.json`, this was moved in [this PR](https://github.com/Codeception/Codeception/pull/6273) before Symfony 6 was released.

In everything else, `5.0` is ready to go.

TavoNiievez:
and if we have to release PHPUnit 10 support and @davertMik's PR as 6.0 in just one month, it seems better to me than keep blocking the upgrade to Symfony 6 until then.

Naktibalda:
New major version of PHPUnit was released during the first week of February for 4 years in a row, last year was exception.
It seems that development picked up speed and there is a strong probability that it will be ready for release soon.

New event system may be the biggest PHPUnit change we ever had to deal with, I will start looking at it today.

PR #5766, opened 2019-11-24
Symfony 5 support for Codeception 4.0

Naktibalda:
There is no need to wait for Codeception 5, required changes are quite small and can be accomodated by Codeception 4.

Replaces #5765
Closes #5758
Closes #5532

PR #6329, opened 2022-01-24
Improve dependency to symfony event dispatcher

athanasius-kircher:
https://github.com/Codeception/Codeception/pull/5766/commits/15ba342b6511d0557435182d815690aa6b8e6aef
introduced a change for symfony 5 compatibility by creating an alias for a new class.

While the project still uses the old class namespace with a symfony version > 5 this old namespace is always an alias to the new class.

The problem with creating this alias is, that it may have some other implications in other libraries.  

Take the symfony/messenger for example and the class SendMessageMiddleware
(LTS - https://github.com/symfony/messenger/blob/5.4/Middleware/SendMessageMiddleware.php). They design the backward compatibility the other way around.

By creating this alias, it traps the package that it is in legacy mode, but it is not.

So better replace the usage of `Symfony\Component\EventDispatcher\Event` with `Symfony\Contracts\EventDispatcher\Event` and use the alias the other way around to be backward compatible. In case this backward support is dropped, only the alias has to been removed.

### DIFF
--- a/shim.php
+++ b/shim.php
@@ -37,7 +37,7 @@ namespace {
     class_alias('Codeception\TestInterface', 'Codeception\TestCase');
 
     //Compatibility with Symfony 5
-    if (!class_exists('Symfony\Component\EventDispatcher\Event') && class_exists('Symfony\Contracts\EventDispatcher\Event')) {
-        class_alias('Symfony\Contracts\EventDispatcher\Event', 'Symfony\Component\EventDispatcher\Event');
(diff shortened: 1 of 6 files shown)


Naktibalda:
That code was good for more than 2 years, changing direction of class_alias could break backwards compatibility for some other library.

Please wait for Codeception 5 which won't have backwards compatibility and will be released soon.
https://github.com/Codeception/Codeception/discussions/6327

PR #6331, opened 2022-01-24
0 < exit-state <= 125

imme-emosol:
https://github.com/Codeception/Codeception/discussions/6330

### DIFF
--- a/src/Codeception/Subscriber/ErrorHandler.php
+++ b/src/Codeception/Subscriber/ErrorHandler.php
@@ -125,7 +125,7 @@ class ErrorHandler implements EventSubscriberInterface
             $error === null || !in_array($error['type'], [E_ERROR, E_COMPILE_ERROR, E_CORE_ERROR])
         )) {
             echo "\n\n\nCOMMAND DID NOT FINISH PROPERLY.\n";
-            exit(255);
+            exit(123);
         }
         if (!is_array($error)) {
             return;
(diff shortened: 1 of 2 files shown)


imme-emosol:
Some extra information there might not hurt either ?
Maybe when running codecept with --debug and/or -vvv ?

Naktibalda:
> Some extra information there might not hurt either ?

Usual causes for this exit is `die()` or `exit` in application code (I'm not aware of any other causes).
There is no way to tell where in the code it happened.

The code is definitely in a broken state and you can't know if failing test could pass or fail on that run.

Naktibalda:
Code which exists in untestable, so `125` would be appropriate.

imme-emosol:
Might be caused by my setting of `xdebug.halt_level=E_WARNING|E_NOTICE|E_USER_WARNING|E_USER_NOTICE` though.. So that might be an exceptional case.

Naktibalda:
yes, don't use it.

imme-emosol:
Confirmed :)

Okay, so this _should_ only be caused by exit/die.
Would it be possible to show where that exit/die happened when --debug and/or -vvv is given to codecept run ?

Naktibalda:
If it was possible, we would do it.

Naktibalda:
For the record, the reason for using 255 as exit code:

> 255 is the code used by uncaught exception, so I retained the same number.

Naktibalda:
Please update https://github.com/Codeception/Codeception/blob/5.0/tests/cli/ErrorHandlerCest.php

TavoNiievez:
Is it okay for this to target `5.0` ?

Naktibalda:
> Is it okay for this to target `5.0` ?

Fine with me. It makes exit code a bit more useful, but it doesn't look like a bugfix to me.

TavoNiievez (approved):

DavertMik (approved):

Issue #6334, opened 2022-01-27
Provide JSONSchema for *.yml configuration files


Sweetchuck:
## What are you trying to achieve?

I would like to get autocomplete assist from my IDE while I am editing `codeception.dist.yml` and `tests/*.suite.dist.yml` files.

phpunit provides an [phpunit.xsd](https://github.com/sebastianbergmann/phpunit/blob/9.5/phpunit.xsd) file, which makes a lot easier to edit the configuration files.

I think Codeception should provide JSONSchema definitions for those configurations.

PR #6335, opened 2022-01-27
Issue #6334 - Add JSONSchema definitions for configuration YAML files

Sweetchuck:
Issue #6334 - Add JSONSchema definitions for configuration YAML files

### DIFF
--- a/schema/codeception.schema.yml
+++ b/schema/codeception.schema.yml
@@ -1,0 +1,231 @@
+$id: 'https://codeception.com/1/codeception.schema.yml'
+$schema: 'https://json-schema.org/draft/2020-12/schema'
+title: 'codeception.dist.yml'
+description: 'Schema definition for Codeception main configuration YAML file.'
+
+definitions:
+    actor_suffix:
+        type: string
+        default: 'Tester'
+
+    namespace:
+        type: string
+        default: ''
+
+    include:
+        type: array
+        default: []
+
+    paths:
+        type: object
+        additionalProperties:
+            type: string
+        properties:
+            test:
+                type: string
+            log:
+                type: string
+            data:
+                type: string
+            support:
+                type: string
+
+    extends:
+        # @todo
+        oneOf:
+            -
+                type: 'null'
+            -
+                type: string
+
+    suite:
+        type: object
+        additionalProperties: true
+        properties:
+            path:
+                type: string
+
+    suites:
+        type: object
+        additionalProperties:
+            '$ref': '#/definitions/suite'
+
+    modules:
+        type: object
+        additionalProperties: true
+        properties:
+            config:
+                type: object
+                additionalProperties: true
+
+    extensions:
+        type: object
+        additionalProperties: true
+        properties:
+            enabled:
+                type: array
+                items:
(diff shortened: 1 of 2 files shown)


Sweetchuck:
Currently it is just a basic "Proof of concept" implementation.
In don't know all the configuration options, and very likely I added some deprecated options as well, such as `class_name` vs `actor` in suite definition, but I don't know which one is the correct one.

DavertMik:
Thanks @Sweetchuck, this is really good initiative. 

How can we trigger the validation for this schema?

I think we need to continue on those, deprecated properties are also ok, we don't often break compatibility on configuration level. We just need to make sure that all cases can be handled by the schema.

Sweetchuck:
@DavertMik This is how I use it with PhpStorm:

* https://www.jetbrains.com/help/phpstorm/2021.3/json.html#ws_json_schema_add_custom
* [File | Settings | Languages & Frameworks | Schemas and DTDs | JSON Schema Mappings](jetbrains://PhpStorm/settings?name=Languages+%26+Frameworks--Schemas+and+DTDs--JSON+Schema+Mappings)

Sweetchuck:
Are there any schema related differences between the `./codeception.dist.yml` and a `./tests/my-suite-01.suite.dist.yml`?

Sweetchuck:
@DavertMik  

> How can we trigger the validation for this schema?

I don't know the details, but some kind of configuration validation already happening without JSONSchema.

For example:
> DEPRECATION: 'settings: bootstrap: fake.php' option is deprecated! Replace it with: 'bootstrap: fake.php' (not under settings section). See: https://codeception.com/docs/reference/Configuration

On other projects I use [opis/json-schema](https://github.com/opis/json-schema) to validate a data structure against a JSONSchema.

Naktibalda:
> Are there any schema related differences between the ./codeception.dist.yml and a ./tests/my-suite-01.suite.dist.yml?

You could take a look at https://github.com/Codeception/Codeception/blob/5.0/src/Codeception/Configuration.php#L77-L126

Settings that can be set under `settings` key in codeception.yaml, can be set as top level settings in suite configuration files.

schema/codeception.schema.yml
```diff
@@ -0,0 +1,231 @@
...
+            data:
+                type: string
+            support:
+                type: string
+
+    extends:
+        # @todo
+        oneOf:
+            -
+                type: 'null'
```
Naktibalda, line 37: I'm pretty sure that it doesn't make sense to set optional parameters to null in yaml file.

Naktibalda:
Is there some command line tool for validation? It would be nice to have it in CI.

schema/codeception.schema.yml
Sweetchuck: The default value is `null`.
https://github.com/Codeception/Codeception/blob/5.0/src/Codeception/Configuration.php#L116

Naktibalda: Null is the default value in the PHP code, it isn't necessary to set settings to null in yaml file and it can be forbidden.
A more usual approach to emit unnecessary settings.

PR #6344, opened 2022-01-31
Update codebase to PHP 8.0

TavoNiievez:


src/Codeception/Configuration.php
```diff
@@ -628,9 +627,8 @@ public static function append(array $config = []): array
     /**
      * @param array|bool|null $a1
      * @param array|bool|null $a2
-     * @return array|bool
      */
-    public static function mergeConfigs($a1, $a2)
+    public static function mergeConfigs($a1, $a2): array|bool|null|string
```
Naktibalda, line 631: It looks like `string` type originated from `$res[$k2] = self::mergeConfigs($a1[$k2], $v2);` in line 658.

I think that it is the only place when the first parameter of `mergeConfig` can be not array.
But return type above is not sufficient, element value can be int and float too,
so we have to go full `mixed`, or make sure that `mergeConfig is not called if `$a1[$k2]` is not array.

```php
            if (!isset($a1[$k2]) || !is_array($a1[$k2]) {
                $res[$k2] = $v2;
                unset($a1[$k2]);
                continue;
            }

            $res[$k2] = self::mergeConfigs($a1[$k2], $v2);
```

If this change is made, we can set return type to `array`.

src/Codeception/Event/TestEvent.php
```diff
@@ -9,14 +9,10 @@
 
 class TestEvent extends Event
 {
-    protected PHPUnitTest $test;
+    // Time taken
```
Naktibalda, line 12: This comment should be in the docblock now.

src/Codeception/Example.php
```diff
@@ -13,14 +13,8 @@
...
-     */
-    protected $data;
-
-    public function __construct($data)
+    public function __construct(protected $data)
     {
-        $this->data = $data;
     }
 
     #[\ReturnTypeWillChange]
```
Naktibalda, line 20: We can drop these annotations, right?

src/Codeception/Extension.php
```diff
@@ -91,10 +86,7 @@ public function _initialize(): void
         $this->_reconfigure(); // hook for BC only.
     }
 
-    /**
-     * @param string|iterable $message
-     */
-    protected function write($message): void
+    protected function write(iterable|string $message): void
```
Naktibalda, line 89: `iterable` was inherited from Symfony\Component\Console\Output, we don't use it anywhere, so we can simplify type to `string`.
If we keep `iterable`, we need a docblock saying that it must be an iterable of strings.

The signature of `writeln` below must match the signature of `write`.

src/Codeception/Lib/Generator/Cest.php
```diff
@@ -31,14 +31,11 @@ public function tryToTest({{actor}} \$I)
```
Naktibalda, line 31: Template above could have `void` return types.

Naktibalda (changes requested):

src/Codeception/Lib/Generator/Cest.php
TavoNiievez: I think adding `: void` to test functions just makes them less readable and doesn't really add any value.

PR #6347, opened 2022-02-03
Fix loading .env file overwriting existing variables

mdoelker:
The v5 branch of phpdotenv is missing the `immutable` flag, changing the behavior used with v4. It is now overwriting existing environment variables with the same name.

From the docs:
image

This is preventing the following use case:

```yml
# codeception.yml
params:
  - .env # Use dev defaults from file
  - env # But prefer values set on the environment, e.g. in CI
```

This does not work, because when loading the `.env` file, the environment variables we would load in the second step were already modified.


Naktibalda:
It would work correctly if we switched the order of params, right?
```
params:
  - env
  - .env
```

I think that it would be better to leave current 4.1 behaviour as-is, because changing it would break tests for people who are relying on current behaviour.

mdoelker:
Yes, this order is working fine, just not for my use case 😀

There is a chance it might break tests, sure. It you feel this too much of a risk for 4.x, I'm fine with that.

Naktibalda:
ok, I'm closing this PR.

PR #6348, opened 2022-02-03
Return params from env file without modifying superglobals

mdoelker:
I noticed that in contrast to the other ways to load params (php, yml, xml etc), when loading from `.env` files there is a side-effect that modifies the `$_ENV` and `$_SERVER` superglobals. While this is usually what you want in the most common use-case for `phpdotenv`, in this case it is a bit odd, as `.env` files are merely used as another input format to load and merge Codeception params. That it also messes around with the values read from the real environment is unexpected.

The piece if code I used to parse the `.env` file into an array is supported in both v4 and v5 of `phpdotenv`.

I based this on the 5.0 branch as it is a breaking change. I also created a bugfix PR for 4.x that fixes environment variables being overwritten but does not remove the side-effect (#6347).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,7 +52,7 @@
     },
     "require-dev": {
         "ext-simplexml": "*",
-        "vlucas/phpdotenv": "^5.0",
+        "vlucas/phpdotenv": "^5.1",
         "symfony/process": ">=4.4 <7.0",
         "symfony/dotenv": ">=4.4 <7.0",
         "codeception/lib-innerbrowser": "*@dev",
--- a/composer.json
+++ b/composer.json
@@ -54,6 +54,7 @@
         "ext-simplexml": "*",
         "vlucas/phpdotenv": "^4.0 | ^5.0",
         "symfony/process": ">=4.4 <7.0",
+        "symfony/dotenv": "^6.0",
         "codeception/lib-innerbrowser": "*@dev",
         "codeception/module-asserts": "*@dev",
         "codeception/module-cli": "*@dev",
@@ -69,6 +70,8 @@
     },
     "suggest": {
         "ext-simplexml": "For loading params from XML files",
+        "vlucas/phpdotenv": "For loading params from .env files",
+        "symfony/dotenv": "For loading params from .env files",
         "hoa/console": "For interactive console functionality",
         "codeception/specify": "BDD-style code blocks",
         "codeception/verify": "BDD-style assertions",
--- a/composer.json
+++ b/composer.json
@@ -52,7 +52,7 @@
     },
     "require-dev": {
         "ext-simplexml": "*",
-        "vlucas/phpdotenv": "^4.0 | ^5.0",
+        "vlucas/phpdotenv": "^5.0",
         "symfony/process": ">=4.4 <7.0",
         "symfony/dotenv": "^6.0",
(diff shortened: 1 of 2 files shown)


Naktibalda:
Support for phpdotenv v4 can be dropped because the lowest supported version of PHP is 8.0.

I think that I removed fallback to `symfony/dotenv` by mistake.
```php
        } elseif (class_exists('Symfony\Component\Dotenv\Dotenv')) {
            $dotEnv = new \Symfony\Component\Dotenv\Dotenv();
            $dotEnv->load(codecept_root_dir($this->paramStorage));
            return $_SERVER;
        }
```
https://github.com/Codeception/Codeception/commit/7cbb8b962fc73e1ff2c9deabbac5b83f177caac9

Should we bring it back?

mdoelker:
Okay, didn't check that. It just happened to be implemented in v4 as well.

I think it makes sense to bring back Symfony support to have more options. I'll have a look at it and push a follow-up.

mdoelker:
Added back symfony/dotenv support and upped phpdotenv dependency to v5, as Codeception 5.0 depends on PHP 8.0 anyway as you said. That way we can also use the new parse method introduced in phpdotenv v5.

composer.json
```diff
@@ -52,8 +52,9 @@
     },
     "require-dev": {
         "ext-simplexml": "*",
-        "vlucas/phpdotenv": "^4.0 | ^5.0",
+        "vlucas/phpdotenv": "^5.0",
         "symfony/process": ">=4.4 <7.0",
+        "symfony/dotenv": "^6.0",
```
TavoNiievez, line 57: Is only `"^6.0"` compatible with these changes or could we also use `">=4.4 <7.0"`?

Naktibalda: Yes, please change this constraint and add support for older versions of symfony/dotenv or at least make sure that they don't cause crash.

Also please merge latest changes from 5.0 branch to your branch.

mdoelker: Yeah, `>= 4.4 <7.0` is totally fine, oversight on my end. Made the change and rebased on current 5.0.

mdoelker:
`PHP 8.0-ubuntu-latest` seems to install the lowest version, while the windows variant does not. Not sure if that is intentional :)

mdoelker:
Any hint on the `Function libxml_disable_entity_loader() is deprecated` error? Don't see how that is related to my change, but it seems it doesn't happen on the 5.0 branch.

Naktibalda:
I wonder if it was the fault of this change that test suite of module-symfony is failing:

https://github.com/Codeception/module-symfony/runs/5567046774?check_suite_focus=true

```
Run php vendor/bin/codecept build -c framework-tests
Building Actor classes for suites: Functional
 -> FunctionalTesterActions.php generated successfully. 275 methods added
App\Tests\FunctionalTester includes modules: Asserts, Symfony, Doctrine2, \App\Tests\Helper\Functional
Codeception PHP Testing Framework v5.0.0-RC1 https://helpukrainewin.org/
Powered by PHPUnit 9.5.19 #StandWithUkraine
[Seed] 1465289846

In EnvVarProcessor.php line 171:
                                                   
  Environment variable not found: "DATABASE_URL".
```

DATABASE_URL is set in https://github.com/Codeception/symfony-module-tests/blob/4.4/.env

PR #6350, opened 2022-02-05
[Codecept5] Updated types for Extension class properties

yesdevnull:
Minor typed property and docblock fixes for the `Extension` and `RunProcess` extension classes.

### DIFF
--- a/ext/RunProcess.php
+++ b/ext/RunProcess.php
@@ -58,7 +58,7 @@ use function sleep;
 class RunProcess extends Extension
 {
     /**
-     * @var array<string, int>
+     * @var array<int|string, mixed>
      */
     protected array $config = ['sleep' => 0];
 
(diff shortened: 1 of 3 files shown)


TavoNiievez (approved):

ext/RunProcess.php
```diff
@@ -58,7 +58,7 @@
 class RunProcess extends Extension
 {
     /**
-     * @var array<string, int>
+     * @var array<int|string, mixed>
```
Naktibalda, line 61: When would it make sense to have integer keys in configuration?

yesdevnull: This might be unique to the `RunProcess` extension, but take the following config:

```yaml
extensions:
  enabled:
    - Codeception\Extension\RunProcess:
        0: 'chromedriver --url-base=/wd/hub'
        sleep: 1
```

Which is parsed as:

```php
array (
  'sleep' => 1,
  0 => 'chromedriver --url-base=/wd/hub',
)
```

Hence why `int|string` key type is required for the `config` property (on `RunExtension`).

PR #6352, opened 2022-02-05
Update interfaces to PHP 8

TavoNiievez:


TavoNiievez:
CI is going to fail, it will be fixed when I change the interface in the modules.

src/Codeception/Lib/Interfaces/MultiSession.php
```diff
@@ -8,7 +8,7 @@ public function _initializeSession(): void;
 
     public function _loadSession($session): void;
 
-    public function _backupSession(): array;
+    public function _backupSession();
```
TavoNiievez, line 11: `[TypeError] Codeception\Module\WebDriver::_backupSession(): Return value must be of type array, Facebook\WebDriver\Remote\RemoteWebDriver returned`
See https://github.com/Codeception/module-webdriver/runs/5079691268

Naktibalda: yes, leave it untyped, `mixed` if you want so.

src/Codeception/Lib/Interfaces/ElementLocator.php
```diff
@@ -30,5 +30,5 @@ interface ElementLocator
      * @api
      * @return Crawler|array of interactive elements
      */
-    public function _findElements(array $locator);
+    public function _findElements(array $locator): array|Crawler;
```
Naktibalda, line 33: Parameter should be `mixed`, because we currently need `string|array|WebDriverBy`.

It is better to leave it more flexible.

Can return type be `iterable`?

src/Codeception/Lib/Interfaces/Web.php
```diff
@@ -47,7 +47,7 @@ public function amOnPage(string $page): void;
      *
      * @param array|string $selector optional
      */
-    public function see(string $text, $selector = null): void;
+    public function see(string $text, array|string $selector = null): void;
```
Naktibalda, line 50: module-webdriver access WebdriverBy as selector too: https://github.com/Codeception/module-webdriver/blob/57e2d8314ee757d1ff8e909afad422402e05f616/src/Codeception/Module/WebDriver.php#L3009

src/Codeception/Lib/Interfaces/ElementLocator.php
TavoNiievez: `is_iterable(\Symfony\Component\DomCrawler\Crawler::class)` returns `false`. So i don't think so.

TavoNiievez:
```php
var_dump(is_iterable(new \Symfony\Component\DomCrawler\Crawler)); // bool(true)
```

is the implementation of `\Traversable` mandatory here ?

TavoNiievez: I mean, `iterable` or `array|iterable`?

src/Codeception/Lib/Interfaces/MultiSession.php
TavoNiievez: I'll leave it untyped, thanks

src/Codeception/Lib/Interfaces/Web.php
TavoNiievez: Is it really okay to add the `WebDriverBy` union type?
It would seem a bit strange to me that an interface knows so much about its implementations.
In lib-innerbrower it is just `array|string`, so `WebDriverBy` would not make sense as a possible value there.

Is it okay if I make an exception here (and in `dontSee`) and leave it untyped?

src/Codeception/Lib/Interfaces/ElementLocator.php
TavoNiievez: i.e. do I just leave it as an `iterable`?

src/Codeception/Lib/Interfaces/Web.php
TavoNiievez: I left it untyped.

src/Codeception/Lib/Interfaces/ElementLocator.php
TavoNiievez: I changed it to `iterable`.

TavoNiievez:
I'll merge this as-is to the `5.0-interfaces` branch and use it in the modules to fix the tests.
When I get a correct build in all I will reopen a PR targeting `5.0`.

src/Codeception/Lib/Interfaces/Web.php
```diff
@@ -301,10 +301,8 @@ public function submitForm($selector, array $params, string $button = null): voi
      * // using strict locator
      * $I->click(['link' => 'Login']);
      * ```
-     *
-     * @param string|array $link
      */
-    public function click($link, $context = null): void;
+    public function click(array|string $link, $context = null): void;
```
Naktibalda, line 305: I think that `$link` and `$field` parameters must be left untyped too.

src/Codeception/Lib/Interfaces/MultiSession.php
yesdevnull: Could the `WebDriver::_backupSession()` not return the following instead?

```php
public function _backupSession(): array
{
    return [
        'driver' => $this->webDriver,
    ];
}
```

It would mean a breaking change for 5.0, but this would be the best time to do so. That way `_backupSession()` can have its return type fixed as `array` (which would be a sensible and portable value across implementors).

PR #6359, opened 2022-02-16
Fix remote code coverage

Naktibalda:


src/Codeception/Coverage/SuiteSubscriber.php
```diff
@@ -110,6 +110,6 @@ public function applyFilter(): void
 
     protected function mergeToPrint(CodeCoverage $coverage): void
     {
-        Printer::$coverage->merge($coverage);
```
Naktibalda, line 113: This is not a solution, but it fixes almost all tests.

Issue #6368, opened 2022-02-28
Drop PSR-2 coding standard and raise it to PSR-12


sebastianneubert:
### Discussed in https://github.com/Codeception/Codeception/discussions/6324

<div type='discussions-op-text'>

<sup>Originally posted by **sebastianneubert** January 17, 2022</sup>
## What

In the contribution document it's recommended to use PSR-2, which is deprecated. 

## How

Switch to PSR-12, which is current PHP standard and used by most open source PHP projects.

- add PHP code style fixer and default rules to the root project dir to easily detect code style violations
- add a composer script to easily run the code style fixer
- add a git hook which checks the code style

If you are okay with that idea, I'll like to support Codeception again and will prepare a pull request. 👋 </div>

PR #6369, opened 2022-02-28
#6368: add support for PSR-12

sebastianneubert:
- add php-cs-fixer default file for PSR-12 codeing style
- run cs fixer over all files to use PSR-12 standard
- change documentation for contribution
- only add PSR-12 ruleset to default .php-cs-fixer.dist.php file

- I do **NOT** add an extra dependency for "friendsofphp/php-cs-fixer" on purpose.
  - as dev dependency the current version is not compatible with other requirements
  - as a extra tool its to much overhead for the repository to add the complete tool/phar/vendor dir in the project


sebastianneubert:
someone has to fix the nitpick-ci rules to accept new PSR-12 codestyle

tests/data/SimpleWithNoClassCest.php
```diff
@@ -2,10 +2,9 @@
 
 class SimpleWithNoClassCest
 {
-
-    public function phpFuncitons(CodeGuy $I) {
-        $I->execute(fn() => strtoupper('hello'));
+    public function phpFuncitons(CodeGuy $I)
```
Naktibalda, line 5: I will fix this method name later.

tests/data/included/jazz/tests/functional/TestGuy.php
```diff
@@ -1,10 +1,13 @@
-<?php //[STAMP] 481f1bf8d02786cb5c4fc661670aab66
+<?php
+
+//[STAMP] 481f1bf8d02786cb5c4fc661670aab66
```
Naktibalda, line 3: This class was automatically generated, I will delete it. (all other TestGuys in tests/data/included too)

tests/support/CodeHelper.php
```diff
@@ -1,7 +1,8 @@
 <?php
+
 namespace Codeception\Module;
 
-// here you can define custom functions for CodeGuy 
+// here you can define custom functions for CodeGuy
```
Naktibalda, line 5: I will delete all `define custom functions` and `Define custom actions` comments in tests/ too.

Naktibalda (commented):
Thanks,

Comments are for myself.

Naktibalda:
@sebastianneubert How would you add PSR-12 style check to CI?

I tried to use `php php-cs-fixer list-files --config .php-cs-fixer.dist.php` but it reports many files it can't fix and the output is not useful.

`phpcs --standard=PSR12` reports different style issues.

sebastianneubert:
usually I use `php-cs-fixer fix` in the root directory. So it will fix all the files, which are not styled as expected by the config file. I do not use the list-files command.

to check, if any files are not in the correct style, there is a dry-run command:

```bash
vendor/bin/php-cs-fixer fix --dry-run
```

I also use git-hooks to check, if the file which is added to the commit is styled correct. this needs the dev dependency for `php-cs-fixer`

```bash
git status --porcelain | grep '^[AM]' | cut -c 3- | while read file; do vendor/bin/php-cs-fixer fix $file; git add $file; done
```

PR #6377, opened 2022-03-07
Fix some issues reported by phpstan

Naktibalda:


src/Codeception/Lib/Generator/Actions.php
```diff
@@ -242,15 +243,18 @@ private function stringifyType(ReflectionType $type): string
...
+
+            return sprintf(
+                '%s%s%s',
+                ($type->allowsNull() && $returnTypeString !== 'mixed') ? '?' : '',
+                $type->isBuiltin() ? '' : '\\',
+                $returnTypeString
+            );
+        } else {
+            // TODO: support ReflectionIntersectionType
+            throw new \InvalidArgumentException($type::class . ' is not supported');
```
Naktibalda, line 257: Support for `ReflectionIntersectionType` must be backported to 4.1 too.

src/PHPUnit/Constraint/WebDriverNot.php
```diff
@@ -53,5 +53,6 @@ public function toString(): string
         if ($this->string) {
             return 'that contains text "' . $this->string . '"';
         }
+        return '';
```
Naktibalda, line 56: I will try to move constraints from `src/PHPUnit/Constraint` to appropriate modules.

PR #6380, opened 2022-03-12
5.0 Support PHPUnit 9

Naktibalda:
We can't keep waiting for PHPUnit 10 release forever.

This PR introduces support for PHPUnit 9 while keeping support for the current state of PHPUnit 10.

I will release 5.0.0 RC1 after merging this PR and the final 5.0.0 version can be released in a week or two.


TavoNiievez (changes requested):
If you ask me, I would ask you to change each comment `//` you added in this PR to something like what was done in #6142

Naktibalda:
I replaced all `class_exists` conditions with `if (PHPUnitVersion::series() < 10) {`

TavoNiievez (approved):

PR #6381, opened 2022-03-12
Action generator: Handle self and parent types correctly

### LINKED ISSUE #6164: Compile Error: '\self' is an invalid class name
I'm trying to run: `vendor/bin/codecept build` with a test suite which one of its modules now has methods with a fluid interface, that is, they have as return type '`: self`' and I get this error:

```shell
PHP Fatal error:  '\self' is an invalid class name in C:\...\tests\_build\support\_
generated\FunctionalTesterActions.php on line 4474

  Compile Error: '\self' is an invalid class name  
```

Check out the failed build [here](https://github.com/Codeception/module-symfony/pull/131/checks?check_run_id=2370717573).

Naktibalda:
Fixes #6164

### DIFF
--- a/src/Codeception/Lib/Generator/Actions.php
+++ b/src/Codeception/Lib/Generator/Actions.php
@@ -159,9 +159,10 @@ EOF;
             if (PHP_VERSION_ID >= 70000) {
                 $reflectionType = $param->getType();
                 if ($reflectionType !== null) {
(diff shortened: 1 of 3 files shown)


TavoNiievez (approved):
the `parent` type was a great addition.
Thank you.

PR #6385, opened 2022-03-13
Removed HoaConsole, added PsySH as current REPL

DavertMik:
This REPL is much more powerful and less buggy than Hoa console
but it is not so well designed for running only Codeception code but arbitrary PHP code. Maybe that's for good.
Anyway, there is room for improvement.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,8 +8,8 @@
     "authors":[
         {
             "name":"Michael Bodnarchuk",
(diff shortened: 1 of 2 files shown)


Naktibalda:
Thanks, I tested it and fixed 2 bugs before merging: https://github.com/Codeception/Codeception/commit/d04726db670a308b85a83d21c8aa092eba62bf9d

PR #6386, opened 2022-03-14
Execute interactive console in the scope of paused test class

### LINKED ISSUE #6086: PHP `const` not working in Interactive Pause
When doing this in the Interactive Pause:
```php
seeElement(self::FOO);
```
... I'm getting:

> syntax error Undefined class constant 'FOO'

My `const` is defined in the very Cest file, and it happens no matter if it's `private` or `public`.

If this is a known limitation of the Pause, I'll mention it at https://codeception.com/docs/02-GettingStarted#Interactive-Pause near

> Inside Interactive Pause you can use the entire power of the PHP interpreter: variables, functions, etc.

Naktibalda:
This changes makes `self` and `$this` work as expected.

Fixes #6086

### DIFF
--- a/src/Codeception/Lib/Actor/Shared/Pause.php
+++ b/src/Codeception/Lib/Actor/Shared/Pause.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Codeception\Lib\Actor\Shared;
 
+use Codeception\Command\Console;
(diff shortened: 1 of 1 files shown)


DavertMik:
Nice, I didn't know how to implement this!
Thank you very much

PR #6387, opened 2022-03-14
Improved shell usage

DavertMik:
More pause options

* all shell settings moved to `Codecept\Lib\PauseShell`
* `codecept_pause()` function
* added `$this->pause()` for unit tests 
* all pause commands to receive `$vars` as array:

```php
codecept_pause(['user' => $user]);
$I->pause(['pageObject' => $pageObject]);
```

TODO: Documentation!

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -8,7 +8,13 @@ function codecept_debug($data)
     \Codeception\Util\Debug::debug($data);
 }
 
-function codecept_pause(array $vars = [])
+/**
+ * Executes interactive pause in ths place
+ *
+ * @param array $vars
+ * @return void
+ */
+function codecept_pause(array $vars = []): void
 {
     \Codeception\Util\Debug::pause($vars);
 }
--- a/functions.php
+++ b/functions.php
@@ -8,6 +8,11 @@ function codecept_debug($data)
     \Codeception\Util\Debug::debug($data);
 }
 
+function codecept_pause(array $vars = [])
+{
+    \Codeception\Util\Debug::pause($vars);
+}
+
 function codecept_root_dir($appendPath = '')
 {
     return \Codeception\Configuration::projectDir() . $appendPath;
--- a/src/Codeception/Test/Unit.php
+++ b/src/Codeception/Test/Unit.php
@@ -96,10 +96,10 @@ class Unit extends TestCase implements
     /**
      * Starts interactive pause in this test
      *
-     * @param array $vars
+     * @param array<string, mixed> $vars
      * @return void
      */
-    public function pause(array $vars = [])
+    public function pause(array $vars = []): void
     {
         $psy = (new PauseShell())->getShell();
         $psy->setBoundObject($this);
--- a/src/Codeception/Test/Unit.php
+++ b/src/Codeception/Test/Unit.php
@@ -13,6 +13,7 @@ use Codeception\PHPUnit\TestCase;
 use Codeception\Scenario;
 use Codeception\Test\Feature\Stub;
 use Codeception\TestInterface;
+use Codeception\Util\Debug;
 use PHPUnit\Framework\TestResult;
 
 use function get_class;
@@ -101,6 +102,9 @@ class Unit extends TestCase implements
      */
     public function pause(array $vars = []): void
     {
+        if (!Debug::isEnabled()) {
+            return;
+        }
         $psy = (new PauseShell())->getShell();
         $psy->setBoundObject($this);
         $psy->setScopeVariables($vars);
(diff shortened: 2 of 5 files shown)


src/Codeception/Test/Unit.php
```diff
@@ -92,6 +93,20 @@ public function getModule(string $module): Module
         return $modules[$module];
     }
 
+    /**
+     * Starts interactive pause in this test
+     *
+     * @param array $vars
+     * @return void
+     */
+    public function pause(array $vars = [])
```
Naktibalda, line 102: `: void`

functions.php
```diff
@@ -8,6 +8,11 @@ function codecept_debug($data)
     \Codeception\Util\Debug::debug($data);
 }
```
Naktibalda, line 10: Note to myself: all functions should be properly typed.

src/Codeception/Test/Unit.php
```diff
@@ -92,6 +93,20 @@ public function getModule(string $module): Module
         return $modules[$module];
     }
 
+    /**
+     * Starts interactive pause in this test
+     *
+     * @param array $vars
```
Naktibalda, line 99: `array<string, mixed>`

Naktibalda (changes requested):

TavoNiievez (commented):
I don't think the behavior of this new $vars parameter is intuitive. Please mention the documentation PR when you have it.
On the other hand, will there be tests for this new feature?

DavertMik:
> I don't think the behavior of this new $vars parameter is intuitive. Please mention the documentation PR when you have it.

Unfortunately, documentation is not with the source code now, so I didn't make that PR

> On the other hand, will there be tests for this new feature?

I think It can be tested in non-interactive mode. But not sure how.

TavoNiievez (approved):

src/Codeception/Util/Debug.php
```diff
@@ -38,6 +40,28 @@ public static function isEnabled(): bool
...
+
+        $pauseShell = new PauseShell();
+        $psy = $pauseShell->getShell();
+        $psy->setScopeVariables($vars);
+        $backtrace = debug_backtrace(DEBUG_BACKTRACE_PROVIDE_OBJECT, 3);
+        if (isset($backtrace[1]['object'])) {
+            // set the scope of test class
+            $pauseShell->addMessage('Use $this-> to access current object');
+            $psy->setBoundObject($backtrace[1]['object']);
+        } elseif (isset($backtrace[2]['object'])) {
```
Naktibalda, line 57: It seems we are guessing here, maybe we could check if object implements `\PHPUnit\Framework\Test` too.
Actor should be one step closer, we could detect it by looking for anan instance of `\Codeception\Actor`.
It could help to avoid code duplication too.

DavertMik: I agree on this, thanks. Could you try this?
I am not sure I understand this code good enough to handle these cases the best way

Naktibalda: I will try to improve it.

DavertMik: @Naktibalda I thought a bit for this and no, we can't use checks like you propose:

`codecept_pause` must be available from any place in the code, not only from tests.

DavertMik:
@Naktibalda I updated the behavior, please merge if you are ok with this

src/Codeception/Util/Debug.php
```diff
@@ -38,6 +40,35 @@ public static function isEnabled(): bool
...
+            return;
+        }
+
+        $pauseShell = new PauseShell();
+        $psy = $pauseShell->getShell();
+        $psy->setScopeVariables($vars);
+        $backtrace = debug_backtrace(DEBUG_BACKTRACE_PROVIDE_OBJECT, 3);
+        foreach ($backtrace as $backtraceStep) {
+            $class = $backtraceStep['class'] ?? null;
+            $fn = $backtraceStep['class'] ?? null;
```
Naktibalda, line 55: This line should be using `$backtraceStep['function']`, right?

DavertMik: Ups, right!

DavertMik, line 55:
```suggestion
            $fn = $backtraceStep['function'] ?? null;
```

src/Codeception/Test/Unit.php
```diff
@@ -92,6 +93,20 @@ public function getModule(string $module): Module
...
 
+    /**
+     * Starts interactive pause in this test
+     *
+     * @param array<string, mixed> $vars
+     * @return void
+     */
+    public function pause(array $vars = []): void
+    {
+        $psy = (new PauseShell())->getShell();
```
Naktibalda, line 104: `codecept_debug` and traint only pause in debug mode, this method pauses always.

I think that it should check debug mode too for consistency:
```php
if (!Debug::isEnabled()) {
    return;
}
```

DavertMik: Added

Naktibalda (approved):
Looks good to me

PR #6395, opened 2022-03-17
Populate $_ENV when loading .env with Symfony/DotEnv

mkilmanas:
As suggested in https://github.com/Codeception/Codeception/pull/6393#discussion_r829284908

Could not think of a reasonable way to test this - PhpDotenv seems to always take priority

### DIFF
--- a/src/Codeception/Lib/ParamsLoader.php
+++ b/src/Codeception/Lib/ParamsLoader.php
@@ -121,7 +121,10 @@ class ParamsLoader
 
         // symfony/dotenv
         if (class_exists(SymfonyDotenv::class)) {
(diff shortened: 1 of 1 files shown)


mkilmanas:
@Naktibalda @TavoNiievez - anything else I could do here? I would really love to see this (or any other fix for this issue) getting on the next RC release.

Naktibalda:
Could you confirm if this code is compatible with Symfony 4.4, Symfony 5 and Symfony 6?

mkilmanas:
Yes, I can confirm that. Just verified:
PHP: 8.1
Symfony: 4.4, 5.4 and 6.0
With 5.0-RC1 it fails to find the values from .env / .env.test, with this fix applied in all cases it finds/loads/overloads the values as expected

PR #6399, opened 2022-03-19
Added --shard, --grep, --filter options

DavertMik:
Introduced `--shard` option to execute non-intersecting subsets of tests. This allows parallelizing tests for multiple agents. The use case is slow-running browser tests that might be executed on different machines. Using shards you can easily split tests by machines and execute them independently. This feature does not aggregate reports from agents, however this can be achieved with robo-paratest.

To split tests for 3 agents call a command per agent:

```
codecept run --shard 1/3
codecept run --shard 2/3
codecept run --shard 3/3
```

Also, as a side-effect, this PR adds `--filter` and `--grep` options which I got used to while working on CodeceptJS

P.S. Updated Parallel testing guide coming soon when this is merged

### DIFF
--- a/src/Codeception/Test/Loader.php
+++ b/src/Codeception/Test/Loader.php
@@ -87,17 +87,19 @@ class Loader
                 throw new ConfigurationException("Shard must be set as --shard=CURRENT/TOTAL where CURRENT and TOTAL are number. For instance: --shard=1/3");
             }
 
-            list($shard, $totalShards) = explode('/', $this->shard);
+            [$shard, $totalShards] = explode('/', $this->shard);
+
+            if ($shard < 1) {
+                throw new ConfigurationException("Incorrect shard index. Use 1/$totalShards to start the first shard.");
+            }
 
             if ($totalShards < $shard) {
                 throw new ConfigurationException("Total shards are less than current shard.");
             }
 
             $chunks = $this->splitTestsIntoChunks($totalShards);
-            if (!isset($chunks[$shard - 1])) {
-                return [];
-            }
-            return $chunks[$shard - 1];
+
+            return $chunks[$shard - 1] ?? [];
         }
         return $this->tests;
     }
--- a/src/Codeception/Test/Loader.php
+++ b/src/Codeception/Test/Loader.php
@@ -104,7 +104,7 @@ class Loader
         return $this->tests;
     }
 
-    private function splitTestsIntoChunks($chunks)
+    private function splitTestsIntoChunks(array $chunks): array
     {
         return array_chunk($this->tests, intval(ceil(sizeof($this->tests) / $chunks)));
     }
--- a/src/Codeception/Test/Loader.php
+++ b/src/Codeception/Test/Loader.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Codeception\Test;
 
+use Codeception\Exception\ConfigurationException;
 use Codeception\Test\Loader\Cept as CeptLoader;
 use Codeception\Test\Loader\Cest as CestLoader;
 use Codeception\Test\Loader\Gherkin as GherkinLoader;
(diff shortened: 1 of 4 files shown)


src/Codeception/Test/Loader.php
```diff
@@ -76,9 +81,32 @@ public function __construct(array $suiteSettings)
 
     public function getTests(): array
     {
+        if ($this->shard) {
+            $this->shard = trim($this->shard);
+            if (!preg_match('~^\d+\/\d+$~', $this->shard)) {
+                throw new ConfigurationException("Shard must be set as --shard=CURRENT/TOTAL where CURRENT and TOTAL are number. For instance: --shard=1/3");
+            }
+
+            list($shard, $totalShards) = explode('/', $this->shard);
```
Naktibalda, line 90: I think that we can use more modern `[$shard, $totalShards] = ` syntax here.

src/Codeception/Test/Loader.php
```diff
@@ -76,9 +81,32 @@ public function __construct(array $suiteSettings)
...
+
+            if ($totalShards < $shard) {
+                throw new ConfigurationException("Total shards are less than current shard.");
+            }
+
+            $chunks = $this->splitTestsIntoChunks($totalShards);
+            if (!isset($chunks[$shard - 1])) {
+                return [];
+            }
+            return $chunks[$shard - 1];
```
Naktibalda, line 100: `!isset` is not necessary - `return $chunks[$shard - 1] ?? []`;

src/Codeception/Test/Loader.php
```diff
@@ -76,9 +81,32 @@ public function __construct(array $suiteSettings)
...
+            $this->shard = trim($this->shard);
+            if (!preg_match('~^\d+\/\d+$~', $this->shard)) {
+                throw new ConfigurationException("Shard must be set as --shard=CURRENT/TOTAL where CURRENT and TOTAL are number. For instance: --shard=1/3");
+            }
+
+            list($shard, $totalShards) = explode('/', $this->shard);
+
+            if ($totalShards < $shard) {
+                throw new ConfigurationException("Total shards are less than current shard.");
+            }
```
Naktibalda, line 94: Is `0/3` an error? :)

src/Codeception/Test/Loader.php
DavertMik: I missed this feature of PHP 😁

src/Codeception/Test/Loader.php
DavertMik: Thanks

src/Codeception/Test/Loader.php
DavertMik: Yes... I will check for this, thanks

Naktibalda:
I made some code style changes.

DavertMik:
Thanks, are we ready to merge it?

src/Codeception/Test/Loader.php
```diff
@@ -76,9 +81,34 @@ public function __construct(array $suiteSettings)
...
+
+            $chunks = $this->splitTestsIntoChunks($totalShards);
+
+            return $chunks[$shard - 1] ?? [];
+        }
         return $this->tests;
     }
 
+    private function splitTestsIntoChunks($chunks)
```
TavoNiievez, line 107:
```suggestion
    private function splitTestsIntoChunks($chunks): array
```

Naktibalda: Tests failed with `PHP Fatal error:  Uncaught TypeError: Codeception\Test\Loader::splitTestsIntoChunks(): Argument #1 ($chunks) must be of type array, string given, called in /home/runner/work/Codeception/Codeception/src/Codeception/Test/Loader.php on line 100 and defined in /home/runner/work/Codeception/Codeception/src/Codeception/Test/Loader.php:107`

PR #6403, opened 2022-03-22
Remove BC layer for php-code-coverage < 9

W0rma:
Spotted in https://github.com/Codeception/Codeception/pull/6401#issuecomment-1074834010


Naktibalda:
Thank you

PR #6406, opened 2022-03-22
Var dumper

DavertMik:
Using Symfony VarDumper for better debugg output

### DIFF
--- a/src/Codeception/Lib/Console/Output.php
+++ b/src/Codeception/Lib/Console/Output.php
@@ -9,9 +9,6 @@ use Symfony\Component\Console\Formatter\OutputFormatter;
 use Symfony\Component\Console\Formatter\OutputFormatterStyle;
 use Symfony\Component\Console\Helper\FormatterHelper as SymfonyFormatterHelper;
 use Symfony\Component\Console\Output\ConsoleOutput;
-use Symfony\Component\VarDumper\Cloner\VarCloner;
-use Symfony\Component\VarDumper\Dumper\CliDumper;
-use Symfony\Component\VarDumper\VarDumper;
 
 class Output extends ConsoleOutput
 {
--- a/src/Codeception/Lib/Console/Output.php
+++ b/src/Codeception/Lib/Console/Output.php
@@ -9,6 +9,9 @@ use Symfony\Component\Console\Formatter\OutputFormatter;
 use Symfony\Component\Console\Formatter\OutputFormatterStyle;
 use Symfony\Component\Console\Helper\FormatterHelper as SymfonyFormatterHelper;
 use Symfony\Component\Console\Output\ConsoleOutput;
+use Symfony\Component\VarDumper\Cloner\VarCloner;
+use Symfony\Component\VarDumper\Dumper\CliDumper;
+use Symfony\Component\VarDumper\VarDumper;
 
 class Output extends ConsoleOutput
 {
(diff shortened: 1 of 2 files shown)


DavertMik:
PsySH contains var-dumper as direct dependency. 
So if we are already installing is let's use its power for debugging!

src/Codeception/Lib/Console/Output.php
```diff
@@ -67,7 +67,11 @@ protected function clean(string $message): string
...
     public function debug(mixed $message): void
     {
-        $message = print_r($message, true);
+        if (!is_scalar($message)) {
+            $this->writeln("\n<debug>  DEBUG OUTPUT:</debug>");
+            dump($message);
+            return;
+        }
         $message = str_replace("\n", "\n  ", $message);
         $message = $this->clean($message);
```
Naktibalda, line 76: What? We have no unit tests for this method?

Cleanup code here could be related to result of `print_r`.

Early return breaks `waitForDebugOutput` logic below.

Naktibalda (changes requested):

src/Codeception/Lib/Console/Output.php
DavertMik: Thanks, good point. I don't know if this change can be tested as dump() prints directly to stdout
but I can try to check for side-effects, yes

Naktibalda: The point of `waitForDebugOutput` is to print the first debug message on a new line.
I think that it should be moved to the top of this method and `\n` should be removed from DEBUG OUTPUT.

Also `waitForDebugOutput` can be made private, because there is no code using it from outside this class anymore.

Naktibalda: Old code below could only handle strings, because other types produced type errors.

I removed all that code and the output looks like thi:
image

Naktibalda: Do you think that `DEBUG OUTPUT:` line is necessary?

TavoNiievez (approved):

src/Codeception/Lib/Console/Output.php
Naktibalda: strings really deserve special handling:
image

TavoNiievez: I think there must be a separator.
It could be anything from `-------` to `DEBUG OUTPUT:` although the latter are words to -ignore- when reading.

TavoNiievez: we could add the type of data being printed and its size maybe.

Naktibalda: image

Naktibalda: image

DavertMik: I think delimiters for multiline output still can be a good idea, or not? 
but I agree, we should remove DEBUG OUTPUT
and print scalar values without var dumper

Naktibalda: you can take over now, I won't make any more changes today.

Naktibalda: I removed `DEBUG OUTPUT` so the output looks like this:
image

It isn't necessary to add delimiters to string output, because when Codeception passes a string, it usually comes with meaningful prefix or format, adding delimiters would make it looks worse.
https://github.com/search?p=2&q=org%3ACodeception+codecept_debug&type=Code

PR #6409, opened 2022-03-25
Encode non-string parameters in configuration files

### LINKED ISSUE #5223: Array-parameters are passed incorrectly
#### I have environment params:
    // file env.yml
    products:
      booking:
        - 123456
        - 234567
        - 345678

#### Then I use them in the acceptance.suite.yml
    modules:
      enabled:
        - ...
        - \Helper\Acceptance
      config:
          // ......
        \Helper\Acceptance:
          products:
             booking:
               %products.booking%

#### Then in the test I try to get those params:
    class SomeCest
    {
        /** @var Acceptance */
        protected $helper;

        public function _inject(Acceptance $helper)
        {
            $this->helper = $helper;
        }

        // tests
        public function tryToTest(AcceptanceTester $I)
        {
            $products = $this->helper->_getConfig('products');
            codecept_debug($products['booking']);
        }
    }

#### But in the console I see the following:
    PHP Notice:  Array to string conversion in .../vendor/codeception/codeception/src/Codeception/Util/Template.php on line 83
    Array
    (
      [0] => Array
    )

#### The reason is:
The method `\Codeception\Util\Template::produce` tries to replace a parameter's placeholder with an array:
```    
$result = str_replace($this->placeholderStart . $placeholder . $this->placeholderEnd, $value, $result);
```
where `$value` is my array from the environment.

#### So I'm forced to use this configuration (in the acceptance.suite.yml):
      products:
        booking:
          - %products.booking.0%
          - %products.booking.1%
          - %products.booking.2%

But this is inconvenient and not the Feng Shui.

Could you fix this issue?
Or is there some other way to pass array in the test?
Thank you.

P.S. Non-array parameters are passed normal.

### LINKED ISSUE #6408: [5.0-RC1] Injected Int typed parameter stop test execution
Codeception Version: 5.0-RC1
PHP: 8.1

I tested  the newest version of Codeception 5 and hit a PHP 8.1 deprecation resulting in a test crash.
We inject some parameters into the `codeception.yml` config

```yml
modules:
  config:
    - Redis:
        host: '%redis.host%'
        port: '%redis.port%'
```

The port is an int value and this creates a deprecation here:

https://github.com/Codeception/Codeception/blob/5.0/src/Codeception/Util/Template.php#L80

A cast to string would be an easy change but not sure if this is always correct. Likely the output of Codeception 4 was already converted to a string but I did not check it.

Naktibalda:
This change allows to use non-string values in parameterized configuration files.
Most importantly objects and arrays, but also integers, floats and booleans.

Because unquoted placeholder `%var%` is not a valid value in Yaml format, this change will replace quoted placeholders with unquoted values,   e.g .  `param: "%var%"` will be rendered as `param: false`

Fixes #6408 
Fixes #5223

### DIFF
--- a/tests/data/params/tests/_support/Helper/Dummy.php
+++ b/tests/data/params/tests/_support/Helper/Dummy.php
@@ -22,4 +22,9 @@ class Dummy extends \Codeception\Module
         $this->assertContains('val1', $vars);
         $this->assertContains('val2', $vars);
     }
+
+    public function seeVarEquals($name, $value)
+    {
+        $this->assertEquals($this->config['vars'][$name], $value);
+    }
 }
--- a/tests/data/params/tests/_support/Helper/Dummy.php
+++ b/tests/data/params/tests/_support/Helper/Dummy.php
@@ -23,8 +23,8 @@ class Dummy extends \Codeception\Module
         $this->assertContains('val2', $vars);
     }
 
-    public function seeVarEquals($name, $value)
+    public function seeVarEquals(int $position, $value)
     {
-        $this->assertEquals($this->config['vars'][$name], $value);
+        $this->assertSame($value, $this->config['vars'][$position]);
     }
 }
--- a/src/Codeception/Configuration.php
+++ b/src/Codeception/Configuration.php
@@ -408,17 +408,18 @@ class Configuration
     protected static function getConfFromContents(string $contents, string $filename = '(.yml)'): array
     {
         if (self::$params) {
-            $template = new Template($contents, '%', '%');
-            $vars = [];
-            foreach (self::$params as $key => $value) {
-                if (is_string($value)) {
-                    $vars[$key] = $value;
-                } else {
-                    $vars[$key] = json_encode($value, JSON_THROW_ON_ERROR);
-                }
-            }
-            $template->setVars($vars);
-            $contents = $template->produce();
+            // replace '%var%' with encoded value
+            $singleQuoteTemplate = new Template($contents, "'%", "%'", 'json_encode');
+            $singleQuoteTemplate->setVars(self::$params);
+            $contents = $singleQuoteTemplate->produce();
+            // replace "%var%" with encoded value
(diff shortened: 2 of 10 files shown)


simonberger:
I gave it a test. It does not work in my case because [the parameter lookup](https://github.com/Codeception/Codeception/blob/5.0/src/Codeception/Util/Template.php#L72-L78) in `Template::produce` no longer works if different parameters were an array before.

Before:
```
['redis' => [
  'host' => 'redis',
  'port' => 6379
]]
```

After:

```
['redis' => '{"host":"redis","port":6379}'}
```

Naktibalda:
Good point, then we must pass context-specific encoder to `Template` and instead of calling `(string)` which wouldn't work as expected with array or Boolean, it would call `json_encode` when rendering configuration yaml file, and `var_export` when rendering PHP files.

Naktibalda:
Another problem is that `redis: "%redis%"` is rendered as `redis: "{"host":"redis","port":6379}"`.

It is possible to use unquoted placeholder `redis: %redis%` but it isn't valid YAML and `build` command currently fails, because it doesn't replace variables before processing.

Alternatively, we could do 2 passes to replace `"%var%"` and '%var%' with unquotes values for non-string values, before doing a pass to replace `%var%`.

Naktibalda:
Please try again.

After latest change `redis: "%redis%"` will be rendered as `redis: {"host":"redis","port":6379}`

simonberger:
This fixes it, thanks.

tests/data/params/tests/_support/Helper/Dummy.php
```diff
@@ -22,4 +22,9 @@ public function seeVarsAreSet()
         $this->assertContains('val1', $vars);
         $this->assertContains('val2', $vars);
     }
+
+    public function seeVarEquals(int $position, $value)
+    {
+        $this->assertEquals($value, $this->config['vars'][$position]);
```
TavoNiievez, line 28: Can we safely use `assertSame` here?

TavoNiievez (approved):

PR #6415, opened 2022-03-28
Prevent sending to STDOUT before starting tests

imme-emosol:


imme-emosol:
It appeared as if this was the fix for notifications I got from calls to header().
The dreaded "headers already sent".
But perhaps something else was the actual fix
(can't be sure as I was tinkering around a bit too much).
And perhaps PHP was wrong in its message/detection of this piece of code as the place where output was already sent.

PR #6423, opened 2022-04-02
Add more code coverage settings

### LINKED ISSUE #6258: Expose PHPUnit's --disable-coverage-ignore option
I have a large codebase with a high number of used/unreachable lines. To remove them from the coverage report, I have used the `@codeCoverageIgnore*` annotations, which Codeception honours. However, after a code change, I'd like to be able to run a coverage report while not ignoring coverage to make sure that code is still not being executed in those blocks.

PHPUnit already has a `--disable-coverage-ignore` option. Please expose this functionality in Codeception and add it to the list of options under `codecept run --help`.

Naktibalda:
We had a few feature requests to add some code coverage setting and a few attempts to add these settings to Codeception.

I reviewed configuration options of PHPCodeCoverage library and added 3 missing settings:

* strict_covers_annotation - marks test as risky if it has `@covers` annotation but executes some other code
* ignore_deprecated_code - doesn't collect code coverage for code having `@deprecated` annotation
* disable_code_coverage_ignore - ignores `@codeCoverageIgnore`, `@codeCoverageIgnoreStart` and `@codeCoverageIgnoreEnd` annotations

The only setting I left out is coverage_cache_directory, see https://github.com/sebastianbergmann/phpunit/blob/master/src/TextUI/TestRunner.php#L245-L277

Closes #6258
Replaces #5874 
Replaces #5922

### DIFF
--- a/src/Codeception/Coverage/SuiteSubscriber.php
+++ b/src/Codeception/Coverage/SuiteSubscriber.php
@@ -21,16 +21,19 @@ abstract class SuiteSubscriber implements EventSubscriberInterface
     use StaticEventsTrait;
 
     protected array $defaultSettings = [
-        'enabled'        => false,
-        'remote'         => false,
-        'local'          => false,
-        'xdebug_session' => 'codeception',
-        'remote_config'  => null,
-        'show_uncovered' => false,
-        'c3_url'         => null,
-        'work_dir'       => null,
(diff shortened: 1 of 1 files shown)


Naktibalda:
@DavertMik @TavoNiievez @SamMousa No comments?

TavoNiievez (approved):

Issue #6434, opened 2022-04-18
Improve the experience for multi-application setups / monorepos (PR included)


calvinalkan:
#### What are you trying to achieve?

Assuming the following directory structure of a monorepo:

```
├── packages
│   ├── package-a
           └── src
           └── tests
           └── codeception.yml
    └── package-b
           └── src
           └── tests
│          └── codeception.yml
├── vendor
├── codeception.yml
├── composer.json
```

Each package follows a similar structure and has a unit, functional, and browser test suite.

What I want to do now is run **all tests** for all packages by their suite. (groups are not an option here)

This is not possible currently and complicates managing our monorepo a lot.
The only way to achieve something similar right now is to exclude all suites except the one we want to run. This is not maintainable.

Ideally, there would be a way to apply to run a selected set of suites for all packages.

The multi-application analogy to `vendor/bin/codecept run unit` basically.

Naktibalda:
Fixed in [4.2.0](https://github.com/Codeception/Codeception/releases/tag/4.2.0)

PR #6435, opened 2022-04-18
Improve multi-application experience, allow including suites by name (opposite of --skip)

calvinalkan:
Fix: #6434

This PR is a minimal solution to fix/support the issue described in #6434.

This allows to run suites by name for included applications-

Test cases are included.


calvinalkan:
I don't know why some remote coverage tests would fail for 7.2 only, nothing related has been changed there.

DavertMik:
@calvinalkan thanks for the PR
I got the problem but I don't like the proposed solution and extra options added to CLI.

How about adding more verbose suite syntax instead, for instance 

* `codecept run unit` <- executes unit tests for current app
* `codecept run backed:unit` <- executes unit tests for backend app
* `codecept run *:unit` <- executes unit tests for app included apps (your case, as I understood)


This is just an idea but it is a more general solution and probably it can be implemented in a more beautiful way.

calvinalkan:
@DavertMik Thanks for the comment.

I agree the options are not ideal but this was the least change required solution.

`codecept run backed:unit` - What exactly should `backend` be here?

The relative path (from cwd) to a `codeception.yml` file I assume.

I'm using this in the context of a very big monorepo, so it would look like this for me:

`codecept run src/packages/package-a:unit`
`codecept run src/packages/package-b:unit`
`codecept run src/packages/package-a:functional`
`codecept run src/packages/package-b:functional`

Now running all tests for all packages by suite.

`codecept run *:unit`
`codecept run *:functional`

I like this solution a lot more than what I proposed.

Maybe we could even do:

`codecept run src/packages/package-a unit` (notice the space)

But this probably clashes a lot with how it currently works.

This changes should probably be done in the main method of the `Run` class right?

calvinalkan:
The only question is how should we run:

`codecept run -i unit -i functional`

Options:

1. `codecept run *:unit:functional` 
2. ...?

DavertMik:
@calvinalkan If you want to run 2 suites it would be 

```
codecept run *:unit,*:functional
```

as we have similar comma syntax for suites themselves:

```
codecept run unit,functioal
```

Sorry for delay in reply

calvinalkan:
@DavertMik Based on your feedback I implemented the following:

Runs unit and functional suites for all included apps
`
codecept run *::unit,*::functional
`

Runs unit and functional test for backend app
`codecept run backend::unit,backed::functional`

Same as above, but with nested package directory
`codecept run /path/to/nested/package::unit`

I chose using `::` instead of `:` because `:` is already used for filtering. E.g (`php vendor/bin/codecept run tests/acceptance/backend:^login`)

DavertMik:
Good idea, thank you!
That will be awesome feature.

calvinalkan:
@DavertMik Any idea what's going on on PHP <= 7.1?

DavertMik:
Please make pull request to 5.0 branch which drops php 7 support

calvinalkan:
I think I found the issue. On PHP7.1 it uses a prehistoric version of `symfony/console` which does not yet have the exit code constants.

This should work now.

I'd love to get this in 4.x as we are bound to it due to `wp-browser` not being there yet https://github.com/lucatume/wp-browser/issues/571

calvinalkan:
@DavertMik All working now

src/Codeception/Command/Run.php
```diff
@@ -405,18 +408,60 @@ public function execute(InputInterface $input, OutputInterface $output)
...
         // Run all tests of given suite or all suites
         if (!$test) {
-            $suites = $suite ? explode(',', $suite) : Configuration::suites();
-            $this->executed = $this->runSuites($suites, $this->options['skip']);
-
-            if (!empty($config['include']) and !$suite) {
+            $raw_suites = $suite ? explode(',', $suite) : Configuration::suites();
+            
+            /** @var string[] $main_app_suites */
+            $main_app_suites = [];
```
Naktibalda, line 414: Please rename all your variables to snakeCase, e.g. `$mainAppSuites`.

Naktibalda (changes requested):

src/Codeception/Command/Run.php
calvinalkan: @Naktibalda done

src/Codeception/Command/Run.php
```diff
@@ -488,8 +533,10 @@ protected function matchSingleTest($suite, $config)
      *
      * @param array $suites
      * @param string $parent_dir
+     * @param array<string,string[]> $filter_app_suites An array keyed by included app name where values are suite names to run.
+     * @param string[] $filter_suites_by_wildcard A list of suite names (applies to all included apps)
      */
-    protected function runIncludedSuites($suites, $parent_dir)
+    protected function runIncludedSuites($suites, $parent_dir, $filter_app_suites = [], $filter_suites_by_wildcard = [])
```
Naktibalda, line 539: Please rename variables in this method too.

tests/data/snapshots/tests/_data/Snapshot.NotAJsonSnapshot.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"></xs:schema>
 <note>
-    <to>Son</to>
-    <from>Vader</from>
-    <heading>Disclaimer</heading>
-    <body>I'm your father!</body>
+    <to>Tove</to>
+    <from>Jani</from>
```
Naktibalda, line 5: Please revert changes of this file.

One day I will fix the test which makes this change :)

Naktibalda:
@DavertMik This change is a feature. Should we bump version number to 4.2.0?

Naktibalda (changes requested):

calvinalkan:
> @DavertMik This change is a feature. Should we bump version number to 4.2.0?

That would be great.

calvinalkan:
Don't merge this yet. I found an issue that has no tests currently.

Or maybe it's not an issue but what about this scenario (might be an edge case):


What it the root codeception.yml file has some tests and includes other apps:

```yml
include:
  - src/Snicco/Component/*
  - src/Snicco/Bundle/*
paths:
  output: tests/_output
  tests: tests
  data: tests/_data
  support: tests/_support
settings:
  colors: true
bootstrap: bootstrap.php
reporters:
  report: "PhpStorm_Codeception_ReportPrinter"
```

This would be for example the case with a monorepo where you have some root level tests for utilities related to the monorepo and each included app has its own separate test suites.

On my fork currently when you run (with the config above)

`codecept run unit` the following happens:

- the unit suite from the root repo is run, all other suites from included apps are run. This is not desired of course and I think we can prevent this by just adding a check for a suite [here](https://github.com/Codeception/Codeception/blob/30b102e2a343e5ee6583a00b6923b2398fd2a068/src/Codeception/Command/Run.php#L439).

Ill fix that now, 

The last question is:

When I run `codecept run *::unit`, should that include the unit suite of the root application (if present). I would say yes.

calvinalkan:
@DavertMik @Naktibalda 

I added some new tests that now verify previously uncovered behavior:

Tests are pretty self-explanatory but in resume:


 - `codecept run` => Everything is run
 - `codecept run unit` => Runs unit suite from current app
 - `codecept run *::unit` => Runs unit suites from all included apps and **NOT** the root suite
 - `codecept run unit,*::unit` => Runs included unit suites **AND** root unit suite 
 - `codecept run functional,*::unit` => Runs included unit suites and root **functional** suite
 
 I think this is the most flexible approach.

Naktibalda:
Please rebase your test on 4.2 branch and fix merge conflict in IncludedCest.

src/Codeception/Command/Run.php
```diff
@@ -488,14 +541,16 @@ protected function matchSingleTest($suite, $config)
...
      */
-    protected function runIncludedSuites($suites, $parent_dir)
+    protected function runIncludedSuites($suites, $parent_dir, $filterAppSuites = [], $filterSuitesByWildcard = [])
     {
         $defaultConfig = Configuration::config();
         $absolutePath = \Codeception\Configuration::projectDir();
 
         foreach ($suites as $relativePath) {
-            $current_dir = rtrim($parent_dir, DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR . $relativePath;
+            $current_dir = rtrim($parent_dir, DIRECTORY_SEPARATOR).DIRECTORY_SEPARATOR.$relativePath;
```
Naktibalda, line 553: Please restore spaces around ` . ` in concatenation.

src/Codeception/Command/Run.php
```diff
@@ -506,7 +561,15 @@ protected function runIncludedSuites($suites, $parent_dir)
             }
 
             $suites = Configuration::suites();
-
+            
+            if(!empty($filterSuitesByWildcard)){
+                $suites = array_intersect($suites, $filterSuitesByWildcard);
+            }
+            
+            if(isset($filterAppSuites[$relativePath])) {
```
Naktibalda, line 569: And add space after `if`   - `if (`

src/Codeception/Command/Run.php
```diff
@@ -330,7 +330,7 @@ public function execute(InputInterface $input, OutputInterface $output)
 
         $suite = $input->getArgument('suite');
         $test = $input->getArgument('test');
-
+
```
Naktibalda, line 333: Please remove spaces here.

src/Codeception/Command/Run.php
```diff
@@ -385,10 +386,14 @@ public function execute(InputInterface $input, OutputInterface $output)
                 }
             }
         }
-
+
```
Naktibalda, line 389: and here.

Naktibalda (changes requested):
Code style issues

src/Codeception/Command/Run.php
calvinalkan: Ok, will be done within the next days

calvinalkan:
@Naktibalda Rebased onto 4.2

DavertMik (approved):
Great job!
Looks good to me!

calvinalkan:
@DavertMik Now we only have to fix running failed tests in multi-app and then it's pretty awesome for monorepos.

src/Codeception/Command/Run.php
```diff
@@ -3,8 +3,6 @@
 
 use Codeception\Codecept;
 use Codeception\Configuration;
-use Codeception\Lib\GroupManager;
-use Codeception\Util\PathResolver;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
```
Naktibalda, line 8: Please document new syntax in `Usage` section of class level docblock in this file.  (Line 24)

Naktibalda (changes requested):

Naktibalda:
> @DavertMik Now we only have to fix running failed tests in multi-app and then it's pretty awesome for monorepos.

Does this change break some tests?

PR #6438, opened 2022-04-20
5.0 Throw exception if actor setting is missing

Naktibalda:
If `actor` is missing, all Cest tests fail with `[Error] Class name must be a valid object or a string`.

### DIFF
--- a/src/Codeception/Test/Cest.php
+++ b/src/Codeception/Test/Cest.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Codeception\Test;
 
 use Codeception\Example;
+use Codeception\Exception\ConfigurationException;
 use Codeception\Lib\Console\Message;
 use Codeception\Lib\Parser;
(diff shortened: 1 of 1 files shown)


src/Codeception/Test/Cest.php
```diff
@@ -95,6 +96,12 @@ public function getSpecFromMethod(): string
     public function test(): void
     {
         $actorClass = $this->getMetadata()->getCurrent('actor');
+
+        if ($actorClass === null) {
+            $message = "actor setting is missing in suite configuration. Do you still use class_name?";
```
TavoNiievez, line 101: it may be that the question itself doesn't obviously imply that that setting was renamed. Could you add a link to the changelog or explain this change in another way please?
Likewise, the $message variable is not necessary.

TavoNiievez (changes requested):

TavoNiievez (approved):

PR #6439, opened 2022-04-22
Fix code coverage tests

Naktibalda:
Code coverage results are different on different versions of PHP so the test must be more flexible


Naktibalda:
I've seen line coverage of that function reported as 3/3, 3/4, 4/4 and even 4/5.

PR #6441, opened 2022-04-22
Codeception 5.0: Fix for DotReporter result printer

Orchestrator404:
Added proxy for the RESULT_PRINT_AFTER event, so the event handling is done in the configured StandardReporter.


Naktibalda:
Thank you, 

I will add a test and a few newlines to make it look the same as in Codeception 4.

Orchestrator404:
Thanks a lot!

PR #6442, opened 2022-04-22
Fix DotReporter output

Naktibalda:
Follow up to  #6441

Before:
image

After:
image

I will investigate why it reports 0 assertions later.

cc: @Orchestrator404

### DIFF
--- a/ext/DotReporter.php
+++ b/ext/DotReporter.php
@@ -55,7 +55,6 @@ class DotReporter extends Extension
 
     public function _initialize(): void
     {
-        $this->options['silent'] = false; // turn on printing for this extension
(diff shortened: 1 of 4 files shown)


Naktibalda:
I fixed assertion count too.

ext/DotReporter.php
```diff
@@ -118,6 +119,14 @@ public function printFailed(FailEvent $event): void
 
     public function afterResult(PrintResultEvent $event): void
     {
+        $this->output->writeln('');
+        $this->output->writeln('');
         $this->standardReporter->afterResult($event);
     }
+
+    public function afterTest(TestEvent $event)
```
TavoNiievez, line 127: `: void` ?

TavoNiievez (approved):

PR #6449, opened 2022-05-02
Added basic attributes support

DavertMik:
Added attributes for Cest format (some of them also work for Test)

Usage example (updated):

```php
    #[Incomplete]
    #[Skip]
    #[Group('ue1')]
    #[Example([1, 1], [2, 2])]
    #[Depends('test1')]
    #[Before('test1')]
    #[Before('test2')]
    public function myTest($I) 
    {
    }
```

### DIFF
--- a/src/Codeception/Test/Metadata.php
+++ b/src/Codeception/Test/Metadata.php
@@ -196,6 +196,19 @@ class Metadata
         }
     }
 
+    public function setParamsFromAttributes($attributes): void
+    {
+        $params = [];
+        foreach ($attributes as $attribute) {
+            $params[$attribute->getName()] = $attribute->getArguments();
+        }
+        $this->params = array_merge_recursive($this->params, $params);
+        // set singular value for some params
+        foreach (['skip', 'incomplete'] as $single) {
+            $this->params[$single] = empty($this->params[$single]) ? null : (string)$this->params[$single][0];
+        }
+    }
+
     public function setParams(array $params): void
     {
         $this->params = array_merge_recursive($this->params, $params);
--- a/src/Codeception/Test/Metadata.php
+++ b/src/Codeception/Test/Metadata.php
@@ -200,7 +200,8 @@ class Metadata
     {
         $params = [];
         foreach ($attributes as $attribute) {
-            $params[$attribute->getName()] = $attribute->getArguments();
+            $name = lcfirst(str_replace('Codeception\\Attribute\\', '', $attribute->getName()));
+            $params[$name] = $attribute->getArguments();
         }
         $this->params = array_merge_recursive($this->params, $params);
         // set singular value for some params
--- a/src/Codeception/Util/Annotation.php
+++ b/src/Codeception/Util/Annotation.php
@@ -110,6 +110,11 @@ class Annotation
 
     public function fetch(string $annotation): ?string
     {
+        $attr = $this->attribute($annotation);
+        if ($attr) {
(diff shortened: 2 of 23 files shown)


TavoNiievez (commented):
I'm not sure modifying `src/Codeception/Util/Annotation.php` is the best way to add this.
Is it possible to create a separate class just for handling the attributes?
Are you sure you want attributes to start with lowercase?

It looks good, what doesn't convince me is where it was added.

Naktibalda:
> While not strictly required it is recommended to create an actual class for every attribute.

https://www.php.net/manual/en/language.attributes.classes.php

PHPUnit has 49 attribute classes: https://github.com/sebastianbergmann/phpunit/tree/main/src/Framework/Attributes

DavertMik:
@TavoNiievez 
> I'm not sure modifying src/Codeception/Util/Annotation.php is the best way to add this.
Is it possible to create a separate class just for handling the attributes?

I think it is the right way as 90% of logic is the same, to get the reflection for a method and parse out data from it.
At this point, attributes are just syntax sugar for annotations. I don't see why we need a new class for it. Maybe rename the current one?

> Are you sure you want attributes to start with lowercase?

Good point, I'm also not so happy to set a root namespace for all of them... 
Lowercase was chosen to follow annotations but I agree that this can be updated.

@Naktibalda 

> While not strictly required it is recommended to create an actual class for every attribute.

Well, we can add it. For instance, this may differentiate Codeception's attributes from PHPUnit's attributes, as some of them are quite different like `Before`. I don't like one extra Import line here but the overall idea looks good. We don't have 49 attributes so it should not be a problem to add them.

src/Codeception/Util/Annotation.php
```diff
@@ -119,13 +124,39 @@ public function fetch(string $annotation): ?string
...
+            return $attr->getArguments();
+        }
         $docBlock = (string)$this->currentReflectedItem->getDocComment();
         if (preg_match_all(sprintf(self::$regex, $annotation), $docBlock, $matched)) {
             return $matched[1];
         }
         return [];
     }
 
+    public function attributes(): array
```
TavoNiievez, line 138: Awesome!

TavoNiievez (approved):

Arhell (approved):

src/Codeception/Util/Annotation.php
```diff
@@ -119,13 +124,39 @@ public function fetch(string $annotation): ?string
...
+    {
+        $attrs = $this->currentReflectedItem->getAttributes();
+        $attrs = array_filter($attrs);
+        $attrs = array_filter($attrs, fn (\ReflectionAttribute $a) => str_starts_with($a->getName(), 'Codeception\\Attribute\\'));
+        return $attrs;
+    }
+
+    public function attribute($name): ?\ReflectionAttribute
```
TavoNiievez, line 146: `string $name`

src/Codeception/Test/Metadata.php
```diff
@@ -196,6 +196,20 @@ public function setParamsFromAnnotations($annotations): void
         }
     }
 
+    public function setParamsFromAttributes($attributes): void
```
TavoNiievez, line 199: `array $attributes`

mbrodala:
Thanks a lot for this, especially the now unified `Examples` attribute is pure gold. :star_struck: 

If only there was a suitable Rector ... :sweat_smile:

PR #6452, opened 2022-05-06
use overwriting config when running included apps

calvinalkan:
See: #5978

### DIFF
--- a/tests/cli/IncludedCest.php
+++ b/tests/cli/IncludedCest.php
@@ -205,4 +205,19 @@ class IncludedCest
         $I->seeInShellOutput('2 tests');
         $I->dontSeeInShellOutput('4 tests');
     }
+    
+    /**
+     * @param CliGuy $I
+     * @before moveToIncluded
+     */
+    public function overwrittenConfigurationIsAlsoAppliedWhenRunningAnIncludedAppFromTheRootApp(CliGuy $I)
+    {
+        if(PHP_VERSION_ID < 700100) {
+            $I->markTestSkipped('This test can only run on PHP >= 7.1 because of reporter type declarations.');
+        }
+        $I->executeCommand('run jazz/tests/functional --report -o "reporters: report: Jazz\CustomReporter1"');
+        $I->seeInShellOutput('CustomReporter1');
+    }
+    
+    
 }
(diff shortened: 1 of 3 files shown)


tests/data/included/jazz/tests/_helpers/CustomReporter1.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Jazz;
+
+use Codeception\PHPUnit\ResultPrinter\Report;
+
+final class CustomReporter1 extends Report
+{
+    public function endTest(\PHPUnit\Framework\Test $test, float $time) :void
```
Naktibalda, line 9: Please come up with the test which doesn't use custom reporters, this test won't work on Codeception 5.

Naktibalda (changes requested):

tests/data/included/jazz/tests/_helpers/CustomReporter1.php
calvinalkan: Any suggestion? Colors?

calvinalkan: What else could be asserted in output?

Naktibalda: Fine, you will need it for Codeception 4 only anyway :)

tests/cli/IncludedCest.php
```diff
@@ -205,4 +205,19 @@ public function includedSuitesAreNotRunTwice (CliGuy $I) {
...
         $I->dontSeeInShellOutput('4 tests');
     }
+    
+    /**
+     * @param CliGuy $I
+     * @before moveToIncluded
+     */
+    public function overwrittenConfigurationIsAlsoAppliedWhenRunningAnIncludedAppFromTheRootApp(CliGuy $I)
+    {
+        if(PHP_VERSION_ID < 700100) {
```
Naktibalda, line 215: Correct version id is 70100

Naktibalda:
Merged to 4.1 https://github.com/Codeception/Codeception/commit/b815ea3fa7c2cf54b96cb68bf93b528e67ffb63f

Naktibalda:
This change will be released as 4.2.0 when #6435 is merged.

PR #6461, opened 2022-05-19
fix compatibility with PHPUnit error expectations

calvinalkan:
I'm surprised that nobody noticed this before, but PHPUnits `expectNotice()` `expectWarning()`, `expectError()` methods dont seem to work at all.

This is because the `ErrorHandler` transforms all errors into exceptions instead of conditionally looking at the error code (like the PHPUnit error handler).

I created tests that prove this.

For deprecations, it's the same issue and I created a failing test. I don't now how to fix that tho because I don't understand how deprecations are currently handled here.

### DIFF
--- a/src/Codeception/Subscriber/ErrorHandler.php
+++ b/src/Codeception/Subscriber/ErrorHandler.php
@@ -4,7 +4,14 @@ namespace Codeception\Subscriber;
 use Codeception\Event\SuiteEvent;
 use Codeception\Events;
 use Codeception\Lib\Notification;
+use PHPUnit\Framework\Error\Error;
+use PHPUnit\Framework\Error\Notice;
+use PHPUnit\Framework\Error\Warning;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+use const E_USER_ERROR;
+use const E_USER_NOTICE;
+use const E_USER_WARNING;
 
 class ErrorHandler implements EventSubscriberInterface
 {
@@ -80,6 +87,14 @@ class ErrorHandler implements EventSubscriberInterface
             return false;
         }
 
+        if($errno === E_USER_NOTICE) {
+            throw new Notice($errstr, $errno,  $errfile, $errline);
+        }elseif ($errno === E_USER_WARNING) {
+            throw new Warning($errstr, $errno,  $errfile, $errline);
+        }elseif ($errno === E_USER_ERROR) {
+            throw new Error($errstr, $errno,  $errfile, $errline);
+        }
+        
         $relativePath = codecept_relative_path($errfile);
         throw new \PHPUnit\Framework\Exception("$errstr at $relativePath:$errline", $errno);
     }
--- a/tests/cli/ErrorExpectationsCest.php
+++ b/tests/cli/ErrorExpectationsCest.php
@@ -1,0 +1,52 @@
+<?php
+
+class ErrorExpectationsCest
+{
+    
+    public function _before(\CliGuy $I)
+    {
+        $I->amInPath('tests/data/error_handling');
+    }
+    
+    public function expectNoticeWorks(\CliGuy $I)
+    {
+        $this->skipIfNot72($I);
+        
+        $I->executeCommand('run tests/unit/ErrorExceptionTest.php:test_notice');
+        $I->seeInShellOutput("OK (");
+    }
+    
+    public function expectWarningWorks(\CliGuy $I)
+    {
+        $this->skipIfNot72($I);
+        
+        $I->executeCommand('run tests/unit/ErrorExceptionTest.php:test_warning');
+        $I->seeInShellOutput('OK (');
+    }
+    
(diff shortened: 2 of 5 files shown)


calvinalkan:
Tests are failing because of `tests/unit/Codeception/Subscriber/ErrorHandlerTest.php::testShowsLocationOfWarning` which is a test that IMO is only proving that things don't work with the PHPUnit API.

But maybe that's desired..?

Naktibalda:
There methods are quite recent, they there introduced in PHPUnit 8.4,

src/Codeception/Subscriber/ErrorHandler.php
```diff
@@ -80,6 +87,14 @@ public function errorHandler($errno, $errstr, $errfile, $errline, $context = arr
...
         }
 
+        if($errno === E_USER_NOTICE) {
+            throw new Notice($errstr, $errno,  $errfile, $errline);
+        }elseif ($errno === E_USER_WARNING) {
+            throw new Warning($errstr, $errno,  $errfile, $errline);
+        }elseif ($errno === E_USER_ERROR) {
+            throw new Error($errstr, $errno,  $errfile, $errline);
+        }
+
```
Naktibalda, line 97: To get this change into Codeception 4, it must be compatible with PHPUnit 5.7.
Classes you used here were renamed in PHPUnit 6.0, so this code would cause fatal error.
Please add some test that runs on all versions of PHP.
Then add class aliases to https://github.com/Codeception/phpunit-wrapper/blob/6.0/src/shim.php

tests/data/error_handling/tests/unit/ErrorExceptionTest.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+class ErrorExceptionTest extends \PHPUnit\Framework\TestCase
+{
+    
+    public function test_notice()
```
Naktibalda, line 6: `testNotice`

Naktibalda (changes requested):

src/Codeception/Subscriber/ErrorHandler.php
calvinalkan: @Naktibalda What about the deprecation stuff? 

Codeception should support `expectDeprecation` aswell.

But I don't understand what is done there currently. Looks like something symfony-related.

Naktibalda: Codeception doesn't have to support any of this.

Deprecations are counted and summary is printed below test results:
image

tests/cli/ErrorExpectationsCest.php
```diff
@@ -0,0 +1,52 @@
...
+        $this->skipIfNot72($I);
+        $I->markTestSkipped('This test is just to reproduce that is doesnt work. It will fail because nothing has been implemented');
+        
+        $I->executeCommand('run tests/unit/ErrorExceptionTest.php:test_deprecation');
+        $I->seeInShellOutput('OK (');
+    }
+    
+    private function skipIfNot72(CliGuy $I)
+    {
+        if(PHP_VERSION_ID < 70200) {
```
Naktibalda, line 46: And change this condition to 
```php
if (version_compare(\PHPUnit\Runner\Version::id(), '8.4.0', '<')) {
```

src/Codeception/Subscriber/ErrorHandler.php
calvinalkan: > Codeception doesn't have to support any of this.
> 
> Deprecations are counted and summary is printed below test results: image

But doesn't this just show the deprecations that occurred while running the test?

What I'm talking about (and what PHPUnit supports in the `TestCase` class) is making assertions about deprecations.

```php

class Foo {

   public function foo(): {
     trigger_error('foo is deprecated', E_USER_DEPRECATED);
   }

}

class FooTest extends Codeception\Unit {

  public function test_that_foo_triggers_deprecation() {
  
     $this->expectDeprecation('foo is deprecated');
     
     (new Foo())->foo();
     
  }


}

```

Naktibalda: Since 8.5.21 PHPUnit doesn't convert deprecations to exceptions by default, it requires setting configuration option.

https://github.com/sebastianbergmann/phpunit/blob/8.5/ChangeLog-8.5.md#8521---2021-09-25

You could implement `convert_deprecations_to_exceptions` option too.

* See how `error_level` option is handled in `handle`.
* Add new option to `$defaultSuiteSettings` https://github.com/Codeception/Codeception/blob/4.2/src/Codeception/Configuration.php#L116-L137 (the default value should be false too).
* Document at https://github.com/Codeception/codeception.github.com/blob/master/docs/reference/Configuration.md#settings

Naktibalda:
I implemented all proposed changes myself in https://github.com/Codeception/Codeception/pull/6469

calvinalkan:
@Naktibalda Great. Thanks a lot.

PR #6469, opened 2022-06-10
Fix compatibility with PHPUnit error expectations

Naktibalda:
Replaces #6461

### DIFF
--- a/src/Codeception/Configuration.php
+++ b/src/Codeception/Configuration.php
@@ -133,6 +133,7 @@ class Configuration
             'config' => [],
         ],
         'error_level' => 'E_ALL & ~E_STRICT & ~E_DEPRECATED',
+        'convert_deprecations_to_exceptions' => false
     ];
 
     protected static $params;
--- a/src/Codeception/Subscriber/ErrorHandler.php
+++ b/src/Codeception/Subscriber/ErrorHandler.php
@@ -4,7 +4,14 @@ namespace Codeception\Subscriber;
 use Codeception\Event\SuiteEvent;
 use Codeception\Events;
 use Codeception\Lib\Notification;
+use PHPUnit\Framework\Error\Error;
+use PHPUnit\Framework\Error\Notice;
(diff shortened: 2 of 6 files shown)


calvinalkan:
@Naktibalda Can we release a 4.2 version once this is merged? Together with the other PRs merged lately.

Naktibalda:
that's the plan.

Naktibalda:
@calvinalkan Most likely this functionality won't work with PHPUnit 10, unless we implement error expectations in [Codeception\Test\Unit](https://github.com/Codeception/Codeception/blob/5.0/src/Codeception/Test/Unit.php) but in that case it will be limited to tests extending `Codeception\Test\Unit`.

See https://github.com/sebastianbergmann/phpunit/commit/d4b3d9d89fa82ae38712ad96b64b1cd4500eb991

Naktibalda:
We can't override `expectNotice` and other methods in Unit class, because they are declared as final in TestCase.
If this functionality is important, we will have to use different method names.

Emitting PHPUnit event is not an option too, because it fixes `expect` functionality, but unexpected errors won't be reported by Codeception.

calvinalkan:
> @calvinalkan Most likely this functionality won't work with PHPUnit 10, unless we implement error expectations in [Codeception\Test\Unit](https://github.com/Codeception/Codeception/blob/5.0/src/Codeception/Test/Unit.php) but in that case it will be limited to tests extending `Codeception\Test\Unit`.
> 
> See [sebastianbergmann/phpunit@d4b3d9d](https://github.com/sebastianbergmann/phpunit/commit/d4b3d9d89fa82ae38712ad96b64b1cd4500eb991)

Are there other usecase other than in classes extening `Unit`?

Of the top of my head I cant think of any.

In Browser and cli tests you cant make assertions about errors this way no matter what

Naktibalda:
Tests extending \PHPUnit\Framework\TestCase directly.

PR #6470, opened 2022-06-10
Make dry-run work with module methods having return types

### LINKED ISSUE #6463: Dry-run on throws Return value must be of type int, Codeception\Util\Maybe returned in ...
Dryrun my tests

#### What do you get instead?
Any Actions with a return type specified will throw:

`PHP Fatal error:  Uncaught TypeError: FunctionalTester::haveInDatabase(): Return value must be of type int, Codeception\Util\Maybe returned in ... tests/_support/_generated/FunctionalTesterActions.php:91`

> command

```bash
vendor/bin/codecept dry-run functional
```
> result

```bash
Functional Tests (1) ------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Modules: \Helper\Functional, Db, \First4Figures\Magento\Codeception\Helper\Magento\Driver, \First4Figures\Magento\Codeception\Helper\Magento
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
TestCest: Test
Signature: TestCest:Test
Test: tests/functional/TestCest.php:Test
Scenario --
 I have in database "x",[]
PHP Fatal error:  Uncaught TypeError: FunctionalTester::haveInDatabase(): Return value must be of type int, Codeception\Util\Maybe returned in /mnt/d/Projects/vc2/mage/001.www/tests/_support/_generated/FunctionalTesterActions.php:91
```
> tests/functional/TestCest.php
```php
<?php
class TestCest
{
    public function Test(FunctionalTester $I){
        $I->haveInDatabase('x', []);
    }
}
```

> tests/_support/_generated/FunctionalTesterActions.php (excerpt: `return $this-> ...` is on line 91) 
```php
    /**
     * [!] Method is generated. Documentation taken from corresponding module.
     *
     * Inserts an SQL record into a database. This record will be erased after the test.
     *
     * ```php
     * <?php
     * $I->haveInDatabase('users', array('name' => 'miles', 'email' => 'miles@davis.com'));
     * ```
     * @see \Codeception\Module\Db::haveInDatabase()
     */
    public function haveInDatabase(string $table, array $data): int {
        return $this->getScenario()->runStep(new \Codeception\Step\Action('haveInDatabase', func_get_args()));
    }
```

### Details

* Codeception version: 4.1.29
* PHP Version: 8.1.5
* Operating System: ubuntu on window subsystem for linux
* Installation type: Composer

Naktibalda:
Closes  #6463

### DIFF
--- a/src/Codeception/Command/DryRun.php
+++ b/src/Codeception/Command/DryRun.php
@@ -6,16 +6,25 @@ use Codeception\Event\DispatcherWrapper;
 use Codeception\Event\SuiteEvent;
 use Codeception\Event\TestEvent;
 use Codeception\Events;
+use Codeception\Lib\Generator\Actions;
+use Codeception\Lib\ModuleContainer;
+use Codeception\Module;
+use Codeception\Step;
+use Codeception\Stub;
 use Codeception\Subscriber\Bootstrap as BootstrapLoader;
 use Codeception\Subscriber\Console as ConsolePrinter;
 use Codeception\SuiteManager;
 use Codeception\Test\Interfaces\ScenarioDriven;
 use Codeception\Test\Test;
 use Codeception\Util\Maybe;
+use PHPUnit\Framework\MockObject\MockObject;
+use ReflectionIntersectionType;
+use ReflectionMethod;
+use ReflectionType;
+use ReflectionUnionType;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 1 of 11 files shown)


TavoNiievez (approved):
ask for my review when you go to integrate to the 5.0 branch, I have a couple of suggestions for typing

Naktibalda:
Now I think that it would be better to mock all methods of all modules to return correct types.

The main problem is that methods implemented in Tester classes are not steps, so they won't be mocked and they will call methods of real modules.

Naktibalda:
I updated code to mock all modules method, so dry-run works as before, but methods return correct types.

PR #6474, opened 2022-06-17
Merge 4.2 changes to 5.0

Naktibalda:
@TavoNiievez Please review.

I will fix PHPUnit 10 compatibility separately.

### DIFF
--- a/tests/cli/IncludedCest.php
+++ b/tests/cli/IncludedCest.php
@@ -207,4 +207,142 @@ final class IncludedCest
         $I->seeInShellOutput('2 tests');
         $I->dontSeeInShellOutput('4 tests');
     }
+
+    /**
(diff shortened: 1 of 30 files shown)


tests/cli/IncludedCest.php
```diff
@@ -207,4 +207,142 @@ public function includedSuitesAreNotRunTwice(CliGuy $I)
         $I->seeInShellOutput('2 tests');
         $I->dontSeeInShellOutput('4 tests');
     }
+
+    /**
+     * @before moveToIncluded
```
TavoNiievez, line 212: We can use attributes here right?

Naktibalda: yes

TavoNiievez (approved):

PR #6475, opened 2022-06-19
WIP: 5.0 Reduce dependency on PHPUnit

Naktibalda:
[PHPUnit\Framework\TestResult](https://github.com/sebastianbergmann/phpunit/blob/cbd0ac18623faaa7ab738db180871d3166b6dbb5/src/Framework/TestResult.php) class was signifficantly modified recently and is marked for removal in the future.

It is the best to stop relying on PHPUnit functionality for aggregating test results.


samdark (approved):

src/Codeception/Test/Filter.php
```diff
@@ -1,70 +1,56 @@
 <?php
 
-declare(strict_types=1);
+namespace Codeception\Test;
```
TavoNiievez, line 3: please keep strict types in this class.

Naktibalda: This is actually a new class, it just contains some code from the old file :)

Naktibalda: Fixed, anything else?

TavoNiievez (approved):

src/Codeception/Test/Filter.php
TavoNiievez: that's all
Thank you

PR #6476, opened 2022-06-19
Replace annotations with attributes

TavoNiievez:


tests/data/claypit/tests/_support/DumbGuy.php
```diff
@@ -4,13 +4,13 @@
  * Inherited Methods
  * @method void wantToTest($text)
  * @method void wantTo($text)
- * @method void execute($callable)
- * @method void expectTo($prediction)
- * @method void expect($prediction)
- * @method void amGoingTo($argumentation)
```
Naktibalda, line 10: This docblock was autogenerated, generator still doesn't generate any comments with `self` return type - https://github.com/Codeception/Codeception/blob/522c7f0f1be5b700ffb8ef8342ba6595f4e16242/src/Codeception/Lib/Generator/Actor.php#L93-L103

Naktibalda (commented):
Now we need test for annotations :)

DavertMik:
@TavoNiievez thank you! This one is really important to get merged!
Please check my PR #6487 as it also adds support for repeatable attributes so we could have multiple `#[Group]` attributes

Naktibalda:
Are we dropping support for annotations now?
If we keep support for annotations, we must have some tests using or (even better) testing them.

tests/unit/Codeception/Util/PathResolverTest.php
```diff
@@ -4,21 +4,21 @@
...
+use Codeception\Test\Unit;
+
+final class PathResolverTest extends Unit
 {
-    /**
-     * @dataProvider getRelativeDirTestData
-     * @group core
-     */
+    #[DataProvider('getRelativeDirTestData')]
+    #[Group('core')]
```
Naktibalda, line 14: Codeception DataProvider attribute does not work in Unit tests.

```
  [ArgumentCountError] Too few arguments to function Codeception\Util\PathResolverTest::testGetRelativeDir(), 0 passed in /.../Codeception/vendor/phpunit/phpunit/src/Framework/TestCase.php on line 1545 and exactly 4 expected


#1  /.../Codeception/tests/unit/Codeception/Util/PathResolverTest.php:15
```

DavertMik:
> Are we dropping support for annotations now?
If we are keeping support for annotations, we must have some tests using or (even better) testing them.

No, we are not dropping, so I agree that some tests with annotation should also be kept

TavoNiievez:
Thank you @Naktibalda

Naktibalda:
and https://github.com/Codeception/Codeception/commit/d43250940289c94a48ea46526702bd0daac96495

Issue #5164, opened 2018-09-11
Setup before the entire Cest


cabiste007:
#### What are you trying to achieve?
I want to make prepare data and fixtures before the entire Cest and only one time same thing for the after
#### What do you get instead?
Using the _before and _after  function in Cest it run before and after each function in my Cest 
(I've seen something like _beforeSuite but I don't know how to use it and if it really what I need. If yes could you explain it more please )

reinholdfuereder:
I think there are different approaches:
* With extension:
  * https://github.com/Codeception/Codeception/issues/4245#issuecomment-303549081
    * https://github.com/oprudkyi/codeception-events-scripting
* With helper class:
  * https://codeception.com/docs/06-ModulesAndHelpers#Extending-a-Module
    > In this helper you can replace the parent’s methods with your own implementation. You can also replace the _before and _after hooks, which might be an option when you need to customize starting and stopping of a testing session.
  * https://codeception.com/docs/06-ModulesAndHelpers#Hooks
  * https://github.com/Codeception/Codeception/issues/5244
* Maybe even in bootstrap code


**Potentially the easiest approach is using a helper class:**
* "\<root folder>/...suite.yml"
```
...
modules:
  enabled:
    ...
    - \Helper\Acceptance
...
```
* "\<root folder>/Helper/Acceptance.php"
```
<?php
declare(strict_types=1);

namespace Helper;

class Acceptance extends \Codeception\Module {

  public function _beforeSuite($settings = []) {
    if ($this->isModuleLoaded('WebDriver')) {
      $enabledModulesSettings = $settings['modules']['enabled'];
      $webDriverSettings = array_column($enabledModulesSettings, 'WebDriver')[0];

      echo "Testing WebDriver URL '" . $webDriverSettings['url'] . "' with "
        . "Selenium at '" . $webDriverSettings['host'] . ":" . $webDriverSettings['port'] . "'" . PHP_EOL;
    }
  }

}
```

xepozz:
Any news? I have the same issue.

PR #6480, opened 2022-06-21
Execute beforeClass and afterClass hooks for Cest format

Naktibalda:
I was looking at [Subscriber\BeforeAfterTest](https://github.com/Codeception/Codeception/blob/0355833627f27bcf1b8e995bdd8de20ecaf3bd8b/src/Codeception/Subscriber/BeforeAfterTest.php) for another issue and realized that it wouldn't be too difficult to implement beforeClass and afterClass hooks for Cest format.

It would probably be good to implement attributes too.

Closes https://github.com/Codeception/Codeception/issues/5164

### DIFF
--- a/src/Codeception/Subscriber/BeforeAfterTest.php
+++ b/src/Codeception/Subscriber/BeforeAfterTest.php
@@ -4,8 +4,10 @@ declare(strict_types=1);
 
 namespace Codeception\Subscriber;
 
+use Codeception\Actor;
 use Codeception\Event\SuiteEvent;
 use Codeception\Events;
+use Codeception\Test\Cest;
 use Codeception\Test\Test;
 use Codeception\Test\TestCaseWrapper;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
@@ -39,15 +41,34 @@ class BeforeAfterTest implements EventSubscriberInterface
         }
     }
 
+    /**
+     * @param string[] $methods
+     */
     private function executeMethods(Test $test, array $methods): void
     {
+        if ($methods === []) {
+            return;
+        }
         if ($test instanceof TestCaseWrapper) {
             $test = $test->getTestCase();
         }
 
-        foreach ($methods as $method) {
-            if (is_callable([$test, $method])) {
-                call_user_func([$test, $method]);
+        if ($test instanceof Cest) {
+            $actorClass = $test->getMetadata()->getCurrent('actor');
+            $scenario = $test->getScenario();
+            $actor = new $actorClass($scenario);
+            $testInstance = $test->getTestInstance();
+
+            foreach ($methods as $method) {
+                if (is_callable([$testInstance, $method])) {
+                    call_user_func([$testInstance, $method], $actor, $scenario);
+                }
+            }
+        } else {
+            foreach ($methods as $method) {
+                if (is_callable([$test, $method])) {
+                    call_user_func([$test,$method]);
+                }
             }
         }
     }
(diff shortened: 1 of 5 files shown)


src/Codeception/Subscriber/BeforeAfterTest.php
```diff
@@ -30,16 +33,31 @@ class BeforeAfterTest implements EventSubscriberInterface
...
+    private ?Actor $actor = null;
+
+    private ?Scenario $scenario = null;
+
     public function beforeClass(SuiteEvent $event): void
     {
         foreach ($event->getSuite()->getTests() as $test) {
-            $testClass = $test::class;
+            if ($test instanceof Cest) {
+                $testClass = $test->getTestClass()::class;
```
TavoNiievez, line 44: `getTestClass()::class`
https://github.com/Codeception/Codeception/blob/522c7f0f1be5b700ffb8ef8342ba6595f4e16242/src/Codeception/Test/Cest.php#L230-L233
Could we consider renaming that method?

tests/cli/ConfigParamsCest.php
```diff
@@ -6,17 +6,27 @@
 
 final class ConfigParamsCest
 {
-    private function moveToTestDir(CliGuy $I): void
+    /**
+     * @beforeClass
+     */
+    public static function _beforeClass(CliGuy $I)
```
TavoNiievez, line 12: the name of these functions could be anything as long as they include the `@beforeClass` annotation, right?
why are they `static`?

Naktibalda: PHPUnit convention - only static methods can be executed before and after class.

We would have to replace `TestUtil::getHookMethods($testClass);` with our own code, which is actually a good idea.

Naktibalda: I will try to replace TestUtil with Codeception annotation parsing code.

DavertMik (approved):

Naktibalda:
I have no time to finish this PR now, so I will leave it for 5.1.0.

I think that Cest format must support `#[\Codeception\Attribute\BeforeClass]` on non-static methods instead of relying on PHPUnit's annotation parser/attributes.

src/Codeception/Subscriber/BeforeAfterTest.php
Naktibalda: Renamed

tests/cli/ConfigParamsCest.php
Naktibalda: instance methods are supported now

Naktibalda:
Interesting, test don't fail for me locally.

Naktibalda:
I moved code cleanup to https://github.com/Codeception/Codeception/pull/6499

tests/support/CliHelper.php
```diff
@@ -20,7 +20,7 @@ public function _after(\Codeception\TestInterface $test)
     {
         codecept_debug('deleting dirs');
         $this->getModule('Filesystem')->deleteDir(codecept_data_dir() . 'sandbox');
-        chdir(\Codeception\Configuration::projectDir());
+        $this->getModule('Filesystem')->amInPath(codecept_root_dir());
```
Naktibalda, line 23: This change broke some unit tests that were relying on matching relative tests when `unit` suite is executed after `cli`

Naktibalda: Actually the problem was that `amInPath` called in `ConfigParamsCest::_beforeClass` changed working directory for all tests.

Naktibalda:
Maybe later.

PR #6487, opened 2022-06-29
Attributes improvements

DavertMik:
* Added `Prepare` attribute
* Added support for repeatable attributes (Group, Env)
* Updated tests 

```php
// define groups in 1 attribute
#[Group('g1', 'g2')]

// or in multiple
#[Group('g1')]
#[Group('g2')]
```
* Added support for Given/When/Then attributes (repeatable) for BDD:

```php
<?php
namespace Tests\Support;

use Codeception\Attribute\Given;
use Codeception\Attribute\When;
use Codeception\Attribute\Then;

class AcceptanceTester extends \Codeception\Actor
{
    use _generated\AcceptanceTesterActions;

     #[Given('I have product with :num1 price in my cart')]
     public function iHaveProductWithPriceInMyCart($num1)
     {
        throw new \PHPUnit\Framework\IncompleteTestError("Step `I have product with :num1 price in my cart` is not defined");
     }

     #[When('I go to checkout process')]
     public function iGoToCheckoutProcess()
     {
        throw new \PHPUnit\Framework\IncompleteTestError("Step `I go to checkout process` is not defined");
     }
```

Corresponding changes were reflected in documentation):

https://github.com/Codeception/codeception.github.com/pull/611


src/Codeception/Test/Loader/Gherkin.php
```diff
@@ -140,6 +140,7 @@ protected function addSteps(array $contexts, string $group = 'default'): void
                 $annotation = Annotation::forMethod($context, $method);
                 foreach (['Given', 'When', 'Then'] as $type) {
                     $patterns = $annotation->fetchAll($type);
+                    codecept_debug($patterns);
```
Naktibalda, line 143: Probably this debug statement is not necessary.

DavertMik: ups, sorry :)

Naktibalda (approved):

PR #6492, opened 2022-07-03
Added a sample tests within new structure

DavertMik:
As we don't have a new structure shown in any codebase I thought of adding a new folder with examples

Currently it contains only new generated tests

The main purpose of it at this moment is to be a reference for maintainers for the new structure. 
Probably we will add more valuable examples soon


examples/tests/Support/_generated/UnitTesterActions.php
```diff
@@ -0,0 +1,1701 @@
...
+     * $I->expectThrowable(new MyError("Don't do bad things"), function() {
+     *     $this->doSomethingBad();
+     * });
+     * ```
+     *
+     * @param \Throwable|string $throwable
+     * @see \Codeception\Module\Asserts::expectThrowable()
+     */
+    public function expectThrowable($throwable, callable $callback): void {
+        $this->getScenario()->runStep(new \Codeception\Step\Action('expectThrowable', func_get_args()));
```
Naktibalda, line 45: It would be great if `codecept bootstrap` could automatically create `test/Support/_generated/.gitignore` file containing:
```
*.php
```

Please delete all Actions files from this pull request.

Naktibalda (changes requested):

examples/tests/Functional/LoginCest.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace Tests\Functional;
+
+use Codeception\Attribute\Before;
+use \Tests\Support\FunctionalTester;
```
TavoNiievez, line 6: the trailing slash is not necessary, and we should avoid adding it here and anywhere else.

examples/tests/Support/AcceptanceTester.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+namespace Tests\Support;
```
TavoNiievez, line 4: there should be a space between those two lines, and in the other Actor classes as well.

examples/tests/Functional.suite.yml
```diff
@@ -0,0 +1,12 @@
+# Codeception Test Suite Configuration
+#
+# Suite for functional tests
+# Emulate web requests and make application process them
+# Include one of framework modules (Symfony, Yii2, Laravel, Phalcon4) to use it
```
TavoNiievez, line 5: [`Phalcon5`](https://github.com/Codeception/module-phalcon5)*

Naktibalda:
I fixed all issues in #6522

PR #6497, opened 2022-07-06
Wrap PHPUnit TestCase to behave like native Codeception test type

Naktibalda:
This change will allow to use Codeception attributes in unit tests and remove many specials cases handling of Exception.


TavoNiievez (approved):
LGTM

DavertMik (approved):

PR #6498, opened 2022-07-06
Add DoesNotPerformAssertions attribute

Naktibalda:


src/Codeception/Exception/UselessTestException.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Codeception\Exception;
+
+use PHPUnit\Framework\AssertionFailedError;
+
+class UselessTestException extends AssertionFailedError
```
TavoNiievez, line 9: Why didn't you call it `UselessTestError` if it extends from `AssertionFailedError`?

Naktibalda: Because it is Codeception exception and we don't have any `Error` classes.

Arhell (approved):

Naktibalda:
@DavertMik thinks that this change isn't necessary.

PR #6499, opened 2022-07-07
Load class level attributes

Naktibalda:


src/Codeception/Test/Cest.php
```diff
@@ -228,9 +235,9 @@ public function getSignature(): string
         return get_class($this->getTestClass()) . ":" . $this->getTestMethod();
     }
 
-    public function getTestClass(): object|string
+    public function getTestClass(): object
     {
-        return $this->testClassInstance;
+        return $this->testInstance;
     }
```
TavoNiievez, line 241: If you renamed the property to `$testInstance`, you must also rename this getter.
`getTestClass` would not make sense anymore.
On the other hand, I see that you created a property: `$testClass`, is it necessary? would a getter suffice?
I didn't see that you used it in this class or that you created any getter for it,
does it make sense to use it internally only for classes that inherit from `Cest.php`?
The calls to `get_class` should no longer exist, as in line 252.

TavoNiievez (changes requested):

src/Codeception/Test/Cest.php
Naktibalda: It looks like I missed some things when I moved this change from another branch, will update it later.

DavertMik (approved):

src/Codeception/Test/Cest.php
Naktibalda: Fixed

TavoNiievez (approved):

Arhell (approved):

PR #6505, opened 2022-07-15
/ Lib / Di / (enable) variadic defaultability

xethub:
dependency injection is missing a check for variadic support as a variadic can just be skipped and just not being passed to the constructor

### DIFF
--- a/src/Codeception/Lib/Di.php
+++ b/src/Codeception/Lib/Di.php
@@ -148,6 +148,9 @@ class Di
             } else {
                 $arg = $this->instantiate($dependency);
                 if (is_null($arg)) {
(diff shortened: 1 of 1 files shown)


src/Codeception/Lib/Di.php
```diff
@@ -148,6 +148,9 @@ protected function prepareArgs(ReflectionMethod $method, array $defaults = []):
             } else {
```
Naktibalda, line 148: I think, that `$parameter->isVariadic()` should be checked when parameter has no typehint too, above.

Naktibalda: I expected you to make that change :)

PR #6517, opened 2022-07-23
/ Lib / Di / <PHP> (add) named parameter support

xethub:
we can support named parameter through reflection too

### DIFF
--- a/src/Codeception/Lib/Di.php
+++ b/src/Codeception/Lib/Di.php
@@ -141,10 +141,13 @@ class Di
                     continue;
                 }
                 if (!$parameter->isOptional()) {
-                    if (!isset($defaults[$k])) {
+                    if (
+                        (!isset($defaults[$k])) &&
+                        (!isset($defaults[$parameter->name]))
+                    ) {
                         throw new InjectionException("Parameter '{$parameter->name}' must have default value.");
                     }
-                    $args[] = $defaults[$k];
+                    $args[] = $defaults[$k] ?? $defaults[$parameter->name];
                     continue;
                 }
                 $args[] = $parameter->getDefaultValue();


xethub:
sorry for my late responses and none interactions this is not my daily job

Naktibalda:
What's the use case?

Only [Cest format](https://github.com/Codeception/Codeception/blob/3b1b241a2aa51cb0f364553e70aba97c9792f019/src/Codeception/Test/Cest.php#L180) uses `$defaults` and it always passes positional array - `[$I, $this->scenario]` in 3 places and `$I, $this->scenario, new Example($this->getMetadata()->getCurrent('example'))]` in 1 place.

xethub:
@Naktibalda the use is general a php feature:

https://github.com/php/php-src/pull/5357
https://wiki.php.net/rfc/named_params

specified a readability and maintenance approach:
```php
$this->di->injectDependencies($form, Di::DEFAULT_INJECT_METHOD_NAME, [
   'fields' => $fields
]);
```
also my order of parameters begins with `class`es and ends with `primitive`s, so if I add another class which happens often I need to change the order of my defaults

for maintenance it is important to got a unique key and the order position is not that unique

Naktibalda:
Can you show your code that uses this feature with injectDependencies?

Naktibalda:
So the possible usecase could be to allow out-of-order parameters for Cest methods `function testMe(PageObjectX $pageObject, $example, $scenario, $I)`, but named defaults would have to be added to `$defaults` too.

There is an actual bug here, that `$defaults` isn't used if argument has type declarations.

Naktibalda:
Maybe it isn't a bug. It looks like workaround for people who don't declare parameter types in their Cest methods.

Naktibalda:
Thanks for pointing this out, I will deprecate the use case of defaults in Codeception 5.0.0 and remove defaults in 6.0.0

https://github.com/Codeception/Codeception/pull/6521

Issue #6542, opened 2022-08-06
Codeception 5 templates generate composer.json requiring incompatible versions of modules


Naktibalda:
#### What are you trying to achieve?

Get working test suite after bootstrap

#### What do you get instead?

```json
    "require": {
        "codeception/codeception": "^5.0",
        "codeception/module-asserts": "^1.0.0"
    }
```

> Provide console output if related. Use `-vvv` mode for more details.

```bash
 Adding codeception/module-asserts for Asserts to composer.json
1 new packages added to require
? composer.json updated. Do you want to run "composer update"? (y/n) y
 Running composer update
Loading composer repositories with package information
Info from https://repo.packagist.org: #StandWithUkraine
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - codeception/module-asserts[1.0.0, ..., 1.3.0] require php >=5.6.0 <8.0 -> your php version (8.1.8) does not satisfy that requirement.
    - codeception/lib-asserts[1.13.0, ..., 1.13.1] require php >=5.6.0 <8.0 -> your php version (8.1.8) does not satisfy that requirement.
    - codeception/codeception[5.0.0-alpha1, ..., 5.0.0-alpha3] require phpunit/phpunit dev-master as 9.5.99 -> found phpunit/phpunit[dev-main, 3.7.0, ..., 3.7.38, 4.0.0, ..., 4.8.36, 5.0.0, ..., 5.7.27, 6.0.0, ..., 6.5.14, 7.0.0, ..., 7.5.20, 8.0.0, ..., 8.5.x-dev, 9.0.0, ..., 9.5.x-dev, 10.0.x-dev (alias of dev-main)] but it does not match the constraint. Perhaps dev-master was renamed to dev-main?
    - codeception/module-asserts 1.3.1 requires codeception/codeception *@dev -> satisfiable by codeception/codeception[5.0.0-alpha1, ..., 5.0.x-dev].
    - codeception/codeception[5.0.0-RC1, ..., 5.0.x-dev] require codeception/lib-asserts 2.0.*@dev -> satisfiable by codeception/lib-asserts[2.0.0].
    - You can only install one version of a package, so only one of these can be installed: codeception/lib-asserts[1.0.0, ..., 1.13.2, 2.0.0].
    - codeception/module-asserts 1.3.1 requires codeception/lib-asserts ^1.13.1 -> satisfiable by codeception/lib-asserts[1.13.1, 1.13.2].
[...]

PR #6545, opened 2022-08-08
Trying to fix `vendor/bin/codecept bootstrap` command for Codeception 5

ThomasLandauer:
I don't really know what I'm doing here, so this is probably not the right way to fix it!

Anyway - when following https://codeception.com/quickstart, I'm getting v1.0.0 of module-phpbrowser and module-asserts added to `composer.json`, which aren't compatible with Codeception 5.

Output:

```
$ php vendor/bin/codecept bootstrap
 Bootstrapping Codeception 

File codeception.yml created       <- global configuration
 Adding codeception/module-phpbrowser for PhpBrowser to composer.json
 Adding codeception/module-asserts for Asserts to composer.json
2 new packages added to require-dev
? composer.json updated. Do you want to run "composer update"? (y/n) y
 Running composer update
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - codeception/module-phpbrowser[1.0.0, ..., 1.0.1] require php >=5.6.0 <8.0 -> your php version (8.1.8) does not satisfy that requirement.
    - codeception/module-phpbrowser 1.0.2 requires codeception/codeception ^4.0 -> found codeception/codeception[4.0.0, ..., 4.2.1] but it conflicts with your root composer.json require (^5.0).
    - codeception/module-phpbrowser 1.0.3 requires codeception/codeception ^4.1 -> found codeception/codeception[4.1.0, ..., 4.2.1] but it conflicts with your root composer.json require (^5.0).
    - Root composer.json requires codeception/module-phpbrowser ^1.0.0 -> satisfiable by codeception/module-phpbrowser[1.0.0, 1.0.1, 1.0.2, 1.0.3].

Use the option --with-all-dependencies (-W) to allow upgrades, downgrades and removals for packages currently locked to specific versions.
 Composer installation failed. Please check composer.json and try to run "composer update" manually

In ModuleContainer.php line 100:
                                                     
  Module Asserts is not installed.                   
  Use Composer to install corresponding package:     
                                                     
  composer require codeception/module-asserts --dev  
                                                     

bootstrap [-s|--namespace [NAMESPACE]] [-a|--actor [ACTOR]] [-e|--empty] [--] [<path>]
```

### DIFF
--- a/src/Codeception/InitTemplate.php
+++ b/src/Codeception/InitTemplate.php
@@ -248,8 +248,8 @@ abstract class InitTemplate
                     continue;
                 }
                 $package = $packages[$module];
-                $this->say(sprintf('"%s": "%s"', $package, "^1.0.0"));
-                $composer[$section][$package] = "^1.0.0";
+                $this->say(sprintf('"%s": "%s"', $package, "^3.0.0"));
+                $composer[$section][$package] = "^3.0.0";
             }
             $this->say('');
             return null;
@@ -279,7 +279,7 @@ abstract class InitTemplate
                 continue;
             }
             $this->sayInfo("Adding {$package} for {$module} to composer.json");
-            $composer[$section][$package] = "^1.0.0";
+            $composer[$section][$package] = "^3.0.0";
             ++$packageCounter;
         }
 


Naktibalda:
I noticed this issue too and raised #6542 

`^3.0` is a better constraint for Codeception 5, but not all modules have versions 3 yet, for example module-cli is at 2.0.0 still, because no changes were required to support Codeception 5, and module-yii2 is 1.1.4.

It is possible that none of these modules are used in templates, but we will certainly have to release module-webdriver 4.0.0 to support PHPUnit 10.

The simplest and most flexible solution would be to use `*` in constrait,
alternatively we could specify version constraint of each module in template:
```php
    $this->addModulesToComposer(['WebDriver' => '^3.0 | ^4.0']);
```

ThomasLandauer:
What's the purpose of having this code at all? Is this just used in `bootstrap` or somewhere else too?

Cause it looks like a nightmare to maintain - so maybe the best solution is getting rid of it...?

Naktibalda:
See https://codeception.com/docs/Customization#Installation-Templates

Naktibalda:
I fixed this and other template issues in https://github.com/Codeception/Codeception/pull/6552
please review it.

ThomasLandauer:
Great! :-)

PR #6554, opened 2022-08-14
Fix PHPUnit 10 compatibility

Naktibalda:
Error exceptions have been removed from PHPUnit 10,

Error expectations (`expectDeprecation*()`, `expectError*()`, `expectNotice*()`, and `expectWarning*()`) won't be supported
`expectException*()` can be used instead.

P.S. There is non-zero possibility that `expectException` will be broken later, if it happens we can reimplement all these methods in `Codeception\Test\Unit`, but method names must be different.
P.P.S. There is non-zero possibility that some future change will be so big that Codeception won't be able to support PHPUnit 10.

### DIFF
--- a/src/Codeception/Exception/Deprecation.php
+++ b/src/Codeception/Exception/Deprecation.php
@@ -1,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Codeception\Exception;
+
+class Deprecation extends Error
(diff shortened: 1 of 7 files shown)


TavoNiievez (approved):
I'm fine with the changes.

Arhell (approved):

PR #6557, opened 2022-08-18
Fixed issue when include groups and test groups empty

geega:
Had some bug in my own custom command 
When we have included groups as an empty array and the test don't have any groups isGroupAccepted is always false. 
Example
```
count(array_intersect([], [])) === 0) 
```

### DIFF
--- a/src/Codeception/Test/Filter.php
+++ b/src/Codeception/Test/Filter.php
@@ -96,7 +96,7 @@ class Filter
 
     public function isGroupAccepted(Test $test, array $groups): bool
     {
-        if ($this->includeGroups !== null && count(\array_intersect($groups, $this->includeGroups)) === 0) {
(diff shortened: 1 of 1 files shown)


src/Codeception/Test/Filter.php
```diff
@@ -96,7 +96,7 @@ public function isNameAccepted(Test $test): bool
 
     public function isGroupAccepted(Test $test, array $groups): bool
     {
-        if ($this->includeGroups !== null && count(\array_intersect($groups, $this->includeGroups)) === 0) {
+        if ($this->includeGroups !== null && $this->includeGroups !== [] && count(\array_intersect($groups, $this->includeGroups)) === 0) {
             return false;
         }
         if ($this->excludeGroups !== null && count(\array_intersect($groups, $this->excludeGroups)) > 0) {
```
Naktibalda, line 102: Isn't the same patch necessary in this line?

Issue #6558, opened 2022-08-18
Cannot instantiate abstract class in Codeception/Test/DataProvider.php on line 66


p-golovin:
#### What are you trying to achieve?

Use abstract base class for cest and inherited cest. 

```php
abstract class AbstractRouteCest
{
    /**
     * @dataProvider returnCodeData
     */
    public function returnCode(AcceptanceTester $I, Example $example): void
    {
            $I->checkHttpCode($example);
    }

    abstract protected function returnCodeData(): array;
}

class VideoCest extends AbstractRouteCest
{
    protected function returnCodeData(): array
    {
        return [
            [
                'url' => '/video/231935',
                'code' => 200
            ]
        ];
    }
}
```

#### What do you get instead?

Got error:
```bash
Error: Cannot instantiate abstract class AbstractRouteCest in Codeception/Test/DataProvider.php on line 66
```

### Details

* Codeception version: 5.0.1
* PHP Version: 8.1.9
* Operating System: macOs 12.2.1
* Installation type: Composer
* List of installed packages (`composer show`)
* Suite configuration:

```yml
actor: AcceptanceTester
modules:
    enabled:
        - PhpBrowser:
              url: https://dev.domain/
        - \Helper\Acceptance
    step_decorators: ~
```

p-golovin:
If I make base class not abstract
```php
class RouteCest
{
    /**
     * @dataProvider returnCodeData
     */
    public function returnCode(AcceptanceTester $I, Example $example): void
    {
            $I->checkHttpCode($example);
    }

    protected function returnCodeData(): array
    {
        return [];
    }

class VideoCest extends RouteCest
{
    protected function returnCodeData(): array
    {
        return [
            [
                'url' => '/video/231935',
                'code' => 200
            ]
        ];
    }
}
```
then no data used from provider:
```
No tests executed!
```

p-golovin:
Add push request to solve this issue: https://github.com/Codeception/Codeception/pull/6560

Naktibalda:
Fixed in 5.0.2

PR #6560, opened 2022-08-19
Fix dataProvider work for abstract Cest class (#6558)

p-golovin:


Naktibalda:
Thanks, it looks good to me.

PR #6593, opened 2022-10-22
Support for Attributes in Actions

### LINKED ISSUE #6592: Support for Attributes in Actions Generation

I'd like to use attributes, like Jetbrains' [PhpStorm attributes](https://packagist.org/packages/jetbrains/phpstorm-attributes) on methods within Codeception helpers, particularly in seeding methods that involve entities with no direct relationship.

#### What do you propose?

I have a [working PoC](https://github.com/yesdevnull/Codeception/tree/5.0-generators) that handles the building of attributes on the Actions' generator. If this is something the Codeception team would be interested in I am happy to submit a PR to get it finalised and merged in.

As Codeception 5 requires PHP 8.0 as a minimum and attributes were released in PHP 8.0 also we do not have to do any version/compatibility checks, much to my relief.

A quick example (yes I know using the example attribute here is incorrect, it's purely for demonstration) is as follows:

```php
// tests/support/CliHelper.php

#[Examples("hello")]
public function seeDirFound($dir)
{
    $this->assertTrue(is_dir($dir) && file_exists($dir), "Directory does not exist");
}
```

And the following is generated:

```php
// tests/support/_generated/CliGuyActions.php

/**
 * [!] Method is generated. Documentation taken from corresponding module.
 *
 *
 * @see \Codeception\Module\CliHelper::seeDirFound()
 */
#[\Codeception\Attribute\Examples("hello")]
public function seeDirFound($dir) {
    return $this->getScenario()->runStep(new \Codeception\Step\Assertion('seeDirFound', func_get_args()));
}
/**
 * [!] Method is generated. Documentation taken from corresponding module.
 *
 * [!] Conditional Assertion: Test won't be stopped on fail
 *
 * @see \Codeception\Module\CliHelper::seeDirFound()
 */
#[\Codeception\Attribute\Examples("hello")]
public function canSeeDirFound($dir) {
    return $this->getScenario()->runStep(new \Codeception\Step\ConditionalAssertion('seeDirFound', func_get_args()));
}
```

yesdevnull:
This adds support for attributes in generated Action traits as discussed [here](https://github.com/Codeception/Codeception/issues/6592). This closes #6592.

- ~I've also tidied up a few CI-related things~ see #6604 instead.
- ~Added parameters to the attributes so IDEs don't complain about parameters being there when they think there shouldn't be~ see #6600 instead.

### TODO

- [x] Handle parameter attributes.
- [x] Tests for parameter attributes.
- [x] Update docs, including note about named parameters ([in my branch](https://github.com/yesdevnull/codeception-website/commits/add-attributes)).

### DIFF
--- a/tests/cli/BuildCest.php
+++ b/tests/cli/BuildCest.php
@@ -141,7 +141,7 @@ final class BuildCest
         $I->seeInThisFile('public function seeDirFound(mixed $dir = NULL): mixed');
     }
 
-    public function generateCorrectTypeWhenSelfTypeIsUsed(CliGuy $I, Scenario $scenario)
+    public function generateCorrectTypeWhenSelfTypeIsUsed(CliGuy $I, Scenario $scenario): void
     {
         if (PHP_MAJOR_VERSION < 7) {
             $scenario->skip('Does not work in PHP < 7');
@@ -159,7 +159,7 @@ final class BuildCest
         $I->seeInThisFile('public function seeDirFound(\Codeception\Module\CliHelper $dir): \Codeception\Module\CliHelper');
     }
 
-    public function generateCorrectTypeWhenParentTypeIsUsed(CliGuy $I, Scenario $scenario)
+    public function generateCorrectTypeWhenParentTypeIsUsed(CliGuy $I, Scenario $scenario): void
     {
         if (PHP_MAJOR_VERSION < 7) {
             $scenario->skip('Does not work in PHP < 7');
@@ -177,7 +177,7 @@ final class BuildCest
         $I->seeInThisFile('public function seeDirFound(\Codeception\Module $dir): \Codeception\Module');
     }
 
-    public function noReturnForNeverType(CliGuy $I, Scenario $scenario)
+    public function noReturnForNeverType(CliGuy $I, Scenario $scenario): void
     {
         if (PHP_VERSION_ID < 80100) {
             $scenario->skip('Does not work in PHP < 8.1');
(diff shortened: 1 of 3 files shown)


tests/cli/BuildCest.php
```diff
@@ -199,4 +199,224 @@ public function noReturnForNeverType(CliGuy $I, Scenario $scenario)
         $I->seeInThisFile('$this->getScenario()->runStep(new \Codeception\Step\ConditionalAssertion(\'seeDirFound\', func_get_args()));');
         $I->dontSeeInThisFile('return $this->getScenario()->runStep(new \Codeception\Step\ConditionalAssertion(\'seeDirFound\', func_get_args()));');
     }
+
+    public function generateAttributeForMethodAttributeWithNoParametersAndNoBrackets(CliGuy $I, Scenario $scenario): void
+    {
+        if (PHP_MAJOR_VERSION < 8) {
+            $scenario->skip('Does not work in PHP < 8');
```
Naktibalda, line 206: Skips for `< ū` and `<7` aren't necessaty, because the lowest supported version is 8.0.
It is only necessary to skip tests that require PHP 8.1 or 8.2,

src/Codeception/Attribute/After.php
```diff
@@ -7,4 +7,7 @@
 #[Attribute(Attribute::TARGET_METHOD | Attribute::IS_REPEATABLE)]
 final class After
 {
+    public function __construct(string $methodName)
```
Naktibalda, line 10: Repeatable attribute should use `...$methodNames` , like in #6600 , right?

yesdevnull: I'm going to drop my commit in favour of @mdoelker's work in #6600 as that looks better than mine. I'll add a handful of suggestions to that PR for a few tweaks to make it perfect.

PR #6595, opened 2022-10-28
Print failure/error/warning/skipped/incomplete messages in HTML reports

### LINKED ISSUE #6594: Failure block has disappeared from the html report since Codeception 5

Have the same html report as on Codeception 4
Before, in codeception/phpunit-wrapper/src/ResultPrinter/HTML.php, we had this method 
```
/**
     * A failure occurred.
     *
     * @param \PHPUnit\Framework\Test                 $test
     * @param \PHPUnit\Framework\AssertionFailedError $e
     * @param float                                  $time
     */
    public function addFailure(\PHPUnit\Framework\Test $test, \PHPUnit\Framework\AssertionFailedError $e, float $time) : void
    {
        $this->failures[Descriptor::getTestSignatureUnique($test)][] = $this->cleanMessage($e);
        var_dump($this->failures);
        parent::addFailure($test, $e, $time);
    }
```
In new file codeception/codeception/src/Codeception/Reporter/HtmlReporter.php, it is no longer there

Maybe related to this message 
Codeception 5.0.0-RC4: 5.0.0-RC4
> Replaced PHPUnit\Framework\TestResult with ResultAggregator

How to find this block? Is it necessary to make an additional call?

#### What do you get instead?

The failure block is missing 

Codeception 4
Capture d’écran 2022-10-28 à 13 46 13

Codeception 5
Capture d’écran 2022-10-28 à 13 46 20

### Details

* Codeception version: 5.0.3
* PHP Version: 8.1

Naktibalda:
Fixes #6594

### DIFF
--- a/src/Codeception/Reporter/HtmlReporter.php
+++ b/src/Codeception/Reporter/HtmlReporter.php
@@ -20,6 +20,8 @@ use SebastianBergmann\Template\Template;
 use SebastianBergmann\Timer\Timer;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 
+use function trigger_error;
+
 class HtmlReporter implements EventSubscriberInterface
 {
     use StaticEventsTrait;
@@ -44,8 +46,6 @@ class HtmlReporter implements EventSubscriberInterface
     protected string $scenarios = '';
 
     protected string $templatePath;
-
-    protected array $failures = [];
 
     private string $reportFile;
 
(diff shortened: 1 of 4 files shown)


Naktibalda:
I noticed that HTML Reporter reports incorrect count of successful tests, because ResultAggregator::addSuccessful is called after all tests, I will fix it later.

Arhell (approved):

PR #6596, opened 2022-10-28
Fix parsing of skip annotation

Naktibalda:


src/Codeception/Test/Metadata.php
```diff
@@ -200,7 +200,11 @@ public function setParamsFromAnnotations($annotations): void
```
Naktibalda, line 200: I noticed that after `array_merge_recursive`, `$this->params['skip']` contains array `[null, 'message']`

Arhell (approved):

PR #6600, opened 2022-11-11
Provide typed signatures for attributes

mdoelker:
Hey there.

The current implementation of attributes does not include any signatures and thus IDEs like PhpStorm report that attributes accept no arguments and also provide no hints on the type or number of arguments supported.

image

I have added constructors with proper signatures to fix this.

image

Due to the way attributes are read, there is no further change needed.

#### Out of scope
Looking forward, it could be beneficial to make the passed in values available via getters instead of using `ReflectionAttribute::getArguments` in the code processing attributes. But that is for another time.

### DIFF
--- a/src/Codeception/Attribute/Depends.php
+++ b/src/Codeception/Attribute/Depends.php
@@ -9,7 +9,7 @@ use Attribute;
 #[Attribute(Attribute::TARGET_CLASS | Attribute::TARGET_METHOD | Attribute::IS_REPEATABLE)]
 final class Depends
 {
-    public function __construct(string $testName)
+    public function __construct(string ...$testNames)
     {
     }
 }
--- a/src/Codeception/Attribute/Depends.php
+++ b/src/Codeception/Attribute/Depends.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Codeception\Attribute;
 
 use Attribute;
@@ -7,4 +9,7 @@ use Attribute;
 #[Attribute(Attribute::TARGET_CLASS | Attribute::TARGET_METHOD | Attribute::IS_REPEATABLE)]
 final class Depends
 {
+    public function __construct(string $testName)
+    {
+    }
 }
--- a/src/Codeception/Attribute/Given.php
+++ b/src/Codeception/Attribute/Given.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Codeception\Attribute;
 
 use Attribute;
@@ -7,4 +9,7 @@ use Attribute;
 #[Attribute(Attribute::TARGET_METHOD | Attribute::IS_REPEATABLE)]
 final class Given
 {
+    public function __construct(string $description)
+    {
+    }
 }
--- a/src/Codeception/Attribute/After.php
+++ b/src/Codeception/Attribute/After.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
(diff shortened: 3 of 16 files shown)


Arhell (commented):
cc @Naktibalda

Naktibalda:
Thanks, but this pull request overlaps with #6593 a lot and he did it first.

src/Codeception/Attribute/Depends.php
```diff
@@ -1,10 +1,15 @@
...
+declare(strict_types=1);
+
 namespace Codeception\Attribute;
 
 use Attribute;
 
 #[Attribute(Attribute::TARGET_CLASS | Attribute::TARGET_METHOD | Attribute::IS_REPEATABLE)]
 final class Depends
 {
+    public function __construct(string $testName)
```
yesdevnull, line 12: `Depends` is repeatable so the constructor params can be changed to `string ...$testNames`

src/Codeception/Attribute/Given.php
```diff
@@ -1,10 +1,15 @@
...
+declare(strict_types=1);
+
 namespace Codeception\Attribute;
 
 use Attribute;
 
 #[Attribute(Attribute::TARGET_METHOD | Attribute::IS_REPEATABLE)]
 final class Given
 {
+    public function __construct(string $description)
```
yesdevnull, line 12: Also repeatable, but I haven't tested this attribute. Might be one to change to `string ...$descriptions`.

yesdevnull (changes requested):
Just some minor suggestions for you. Looks better than what I had, so let's get this merged instead.

yesdevnull:
> Thanks, but this pull request overlaps with #6593 a lot and he did it first.

As mentioned in #6593, I'm in favour of this PR over the attribute changes in mine. Will drop my commit once this is merged/discussed further.

src/Codeception/Attribute/Given.php
mdoelker: Multiple descriptions per Given don't seem to be a thing.

src/Codeception/Attribute/Depends.php
mdoelker: Digging through the code it seems that depends should be flattened as well. I'm not even sure the current code works properly for Depends. I'll have to test this a bit.

yesdevnull: Sorry, yep, repeatable was a bad way for me to describe it. I should've said it can take multiple parameters instead. You're correct in that there's no tests which cover multiple test names to the `Depends` attribute, however manually testing this I confirmed that it could take multiple test names and correctly act accordingly to whether those tests passed or failed. If I get time today/tonight I'll write a test for multiple methods in `Depends` and add the patch for it here.

src/Codeception/Attribute/Given.php
yesdevnull: Yep, a misunderstanding by me re: the `IS_REPEATABLE` flag for an attribute. Case closed 🔨

src/Codeception/Attribute/Depends.php
mdoelker: I'll change it like you suggested then

mdoelker: Thanks. I've added the change to the MR 👍

mdoelker:
@Naktibalda @yesdevnull Looks like we are done here. Failed checks look unrelated.

Naktibalda:
thank you

PR #6602, opened 2022-11-16
Fixed 'expectNotToPerformAssertions override in unit tests

yesdevnull:
I noticed that the PHPUnit TestCase method `expectNotToPerformAssertions()` wasn't being correctly applied by the TestCaseWrapper in Codeception. Unit tests with `expectNotToPerformAssertions` were being identified as useless tests, despite having informed the test runner that it does not perform assertions.

What I propose fixes the issue however it may not be the preferred way to do it. Happy to make alterations if required.

Side note, the test `RunUselessTestsCest:checkOutput` is failing because the `UselessCest` and `UselessCept` output isn't showing correctly in the test. Doesn't appear to be related but I'll make sure and fix in a subsequent commit if it it my fault.

### DIFF
--- a/src/Codeception/Test/Test.php
+++ b/src/Codeception/Test/Test.php
@@ -199,7 +199,12 @@ abstract class Test extends TestWrapper implements TestInterface, Interfaces\Des
             $this->assertionCount = Assert::getCount();
             $result->addToAssertionCount($this->assertionCount);
 
-            if ($this->reportUselessTests && $this->assertionCount === 0 && $eventType === Events::TEST_SUCCESS) {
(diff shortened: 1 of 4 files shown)


Naktibalda:
Looks good to me.
Exception message in TestCaseWrapper `This test is annotated with "@doesNotPerformAssertions" but performed 1 assertions` must be changed to something more suitable, because Codeception doesn't support this annotation,

yesdevnull:
Good pickup! I have changed that exception message to the following:

> This test indicates it does not perform assertions but %d assertions were performed

PR #6614, opened 2022-12-08
tryTo methods must always return Boolean result

### LINKED ISSUE #6559: The generated TryTo* function does not have a return statement
Use the TryTo step decorator to check some condition

#### What do you get instead?
The generated TryTo* function does not have a return statement

> Expected:

```php
    public function tryToSeeElement($selector, array $attributes = []): bool {
        return $this->getScenario()->runStep(new \Codeception\Step\TryTo('seeElement', func_get_args()));
    }
```
> Actual:

```php
    public function tryToSeeElement($selector, array $attributes = []): void {
        $this->getScenario()->runStep(new \Codeception\Step\TryTo('seeElement', func_get_args()));
    }
```
### Details

* Codeception version: 5.0.1
* PHP Version: 8.1.8
* Operating System: Windows WSL/ubuntu

Naktibalda:
Fixes #6559

### DIFF
--- a/src/Codeception/Step/TryTo.php
+++ b/src/Codeception/Step/TryTo.php
@@ -46,6 +46,8 @@ class TryTo extends Assertion implements GeneratedStep
         return $template
             ->place('doc', $conditionalDoc)
             ->place('action', 'tryTo' . ucfirst($action))
+            ->place('return', 'return ')
+            ->place('return_type', ': bool')
             ->place('step', 'TryTo');
     }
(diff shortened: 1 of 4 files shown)


Arhell (approved):

jmrieger (approved):
Running into this same issue when updating to Codeception 5. The proposed changes definitely look like they will solve the attached issue, and I also appreciate including the logical check for this in tests.

PR #6618, opened 2022-12-19
Update Run.php (Fix depricated messages when running tests)

shtiher-pp:
Fix depricated messages when running tests:
Deprecated: Using ${var} in strings is deprecated, use {$var} instead in ../vendor/codeception/codeception/src/Codeception/Command/Run.php on line 422
Deprecated: Using ${var} in strings is deprecated, use {$var} instead in ../vendor/codeception/codeception/src/Codeception/Command/Run.php on line 505

### DIFF
--- a/src/Codeception/Command/Run.php
+++ b/src/Codeception/Command/Run.php
@@ -419,7 +419,7 @@ class Run extends Command
 
         if ($this->options['shard']) {
             $this->output->writeln(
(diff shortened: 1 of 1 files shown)


Arhell (commented):
cc @Naktibalda

Naktibalda:
thanks

Issue #6652, opened 2023-03-12
Class "BaseCommandRunner" not found


cybd:
Trying to run unit test locally. I'm using https://github.com/Codeception/Codeception/blob/5.0/tests/README.md notes to run Dockerized testing.

Steps to reproduce:
git clone git@github.com:MY-GITHUB-USERNAME/Codeception.git
cd tests
docker-compose build
docker-compose up -d
docker-compose run --rm codecept run unit Codeception/Command

I've got in console output:
**Fatal error: Uncaught Error: Class "BaseCommandRunner" not found in /repo/tests/unit/Codeception/Command/BuildTest.php on line 5**

**Error: Class "BaseCommandRunner" not found in /repo/tests/unit/Codeception/Command/BuildTest.php on line 5**

Full console output with call stack:

```bash
Codeception PHP Testing Framework v5.0.9 https://helpukrainewin.org
[Seed] 1404964118

Fatal error: Uncaught Error: Class "BaseCommandRunner" not found in /repo/tests/unit/Codeception/Command/BuildTest.php on line 5

Error: Class "BaseCommandRunner" not found in /repo/tests/unit/Codeception/Command/BuildTest.php on line 5

Call Stack:
    0.0001     409488   1. {main}() /codecept/vendor/bin/codecept:0
    0.0002     410048   2. include('/codecept/vendor/codeception/codeception/codecept') /codecept/vendor/bin/codecept:120
    0.0003     427768   3. require('/codecept/vendor/codeception/codeception/app.php') /codecept/vendor/codeception/codeception/codecept:7
    0.0061    1823504   4. {closure:/codecept/vendor/codeception/codeception/app.php:7-46}() /codecept/vendor/codeception/codeception/app.php:46
    0.0175    4366624   5. Codeception\Application->run($input = ???, $output = ???) /codecept/vendor/codeception/codeception/app.php:45
[...]

PR #6655, opened 2023-03-13
[BUGFIX] #6652 Class BaseCommandRunner not found

cybd:
Fixed autoloader for unit tests.
Fixed xdebug mode warning message.

Command runs with all green:
docker-compose run --rm codecept run unit Codeception/Command

Known issues:
1. `docker-compose run --rm codecept run unit Codeception/Lib/DiTest` fail with **Failed asserting that exception of type "ReflectionException" matches expected exception "Codeception\Exception\InjectionException".**
2. `docker-compose run --rm codecept run unit Codeception/Coverage/FilterTest::testWhitelistFilterApplied` fails with **[Symfony\Component\Finder\Exception\DirectoryNotFoundException] The "/repo/vendor" directory does not exist.** 
3. `docker-compose run --rm codecept run cli` fail with **Could not open input file: /repo/codecept**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,6 +24,7 @@ RUN set -eux; \
     echo "date.timezone = UTC" >> /usr/local/etc/php/php.ini;
 
 ENV COMPOSER_ALLOW_SUPERUSER '1'
+ENV XDEBUG_MODE 'coverage'
 
 WORKDIR /codecept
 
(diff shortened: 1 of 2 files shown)


Naktibalda:
Thanks, but I won't merge it. 

The main purpose of this Dockerfile is building images that are used to test other applications, so it is better to keep image smaller.
https://hub.docker.com/r/codeception/codeception

cybd:
I just tried to make another pull request to fix issue in module-db. Wished to run tests before that. It's weird to propose fix the issue and then decline it.

DavertMik:
I'm sorry but I agree, I prefer to close this as we do not use docker for running internal tests and I'd prefer not to pass that config into userland

PR #6684, opened 2023-08-16
Differentiate parsing errors in a parent class from the ones in the test class

martin-rueegg:
Currently, when a test inherits from a parent class, and there is a parsing error happening while parsing this parent class, then the error shows the parse error and line number from the parent class, but the file name from the (child) test class.

This patch will show the file name in which the error actually happens. And if it's not in the test class it self, it will additionally indicate to the name of the test class.

### DIFF
--- a/tests/unit/Codeception/Lib/ParserTest.php
+++ b/tests/unit/Codeception/Lib/ParserTest.php
@@ -3,6 +3,7 @@
 declare(strict_types=1);
 
 use Codeception\Attribute\Group;
+use Codeception\Exception\TestParseException;
 use Codeception\Lib\Parser;
 use Codeception\Scenario;
 use Codeception\Test\Cept;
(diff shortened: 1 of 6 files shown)


martin-rueegg:
New tests have been added, but I was unable to run them with reasonable effort. 

If they fail, please help me adjusting them as I've already spent way to much time on this little fix ... :-)

Thank you!

tests/unit/Codeception/Lib/ParserTest.php
```diff
@@ -146,13 +146,80 @@ public function testParseFileWhichUnsetsFileVariable()
...
+        $this->expectException(\Codeception\Exception\TestParseException::class);
+        $this->expectExceptionMessage("Couldn't parse test 'test.file'");
+        throw new \Codeception\Exception\TestParseException('test.file');
+    }
+
+    #[Group('core')]
+    public function testParseExceptionWithErrors()
+    {
+        $this->expectException(\Codeception\Exception\TestParseException::class);
+        $this->expectExceptionMessage("Couldn't parse test 'test.file':\nFunny error");
```
Naktibalda, line 161: Your tests fail because there is no `:` in error exception message.

martin-rueegg: @Naktibalda thank you so much. Yes, when I was adapting to the latest 5.0 branch, I forgot to re-add the colon there.

I've now fixed that and also used `PHP_EOL` rather than `\n` in the test, so that they should not fail on other OS types.

martin-rueegg:
Thank you, @Naktibalda 

Apparently, `codecept_data_dir()` does not return a normalized file path, but includes `test/data` rather than `test\data` also on Windows. So I'm now normalizing the file paths before comparison.

martin-rueegg:
@Naktibalda I'm so sorry, I thought I had tested if null input works ...

martin-rueegg:
@Naktibalda Thanks, I guess we're getting there eventually ... :-)

Issue #4087, opened 2017-03-28
No way to use helpers/modules inside dataprovider methods


Morjodrom:
**What are you trying to achieve?**
Need to populate dataprovider with data from a database. 

**What do you get instead?**
Unfortunately, it's not possible as neither _inject method is called before the dataproviders nor the Actor or Scenario is accessible in it.

Perhaps, it's a bad testing pattern to use DB in dataprovider, but I think it's rather ubiquitous use case for them. So, it'd be nice to see such a feature in the framework or at least mention in the docs that such approach is discouraged.

edno:
Yes, this is currently a known limitation. Please see my comment in #4008.

For solving both use cases, we should change the sequence for loading test cases in Codeception. 
@DavertMik, what do you about processing `@example` and `@dataprovider` at the end of event `beforeSuite`?

In short term, I can extend the [dataprovider module](https://github.com/edno/codeception-dataprovider) with a new annotation `@usemodule`. This will allows to use `@dataprovider` processed by the module instead of core Codeception,  hence it will make the extension compatible again with current version of Codeception. 
@Fuin and @Morjodrom, would that be useful for you?

Fuin:
For me it would be totally useful, your answer in #4008 sadly wont work for me ;)
But if you could adjust it like that I can work with it.

Morjodrom:
@edno , thank you!
I made a workaround with custom singleton DB connection, rather dirty way, but quick. I like the idea of following the vanilla codeception for this case, hence, if you feel like improving [Dataprovider Module](https://github.com/edno/codeception-dataprovider), I'd be glad to test it in my environment and use before it's implemented in the framework.  But now it's not extremely crucial for me.

keithcrooks:
Did anything progress with this?

I have some helpers to create common Stubs. These work fine in the test methods. However, as above, these cannot be used in the data provider methods.

I was hoping to move some of the code out of the tests and into the data providers to keep the tests focused. However, I have encountered the issue above.

darkbluesun:
+1 this would be super useful

mlambley:
Yes please! I'm hitting this issue right now.
I have a module which does various things including fetching data, and it is used by all of my cests.
I can call it everywhere except from within my data providers.

mikolajzieba:
Is there any possibility for that to be implemented in the future or everyone should look for workaround on this?

mlambley:
[...]

PR #6696, opened 2023-10-02
Add actor to Cest tests dataProviders

weeg:
It makes possible to use actors inside dataProviders. It's a very useful functionality, during testing it happens that data need to be retrieved ie.: from the database, but it's not possible to use it with the current solution. There was a discussion about it in the past: #4087 

I am not sure if I implemented it in the correct way, I am new to Codeception and open source contribution as well. What do you think of it @DavertMik?

### DIFF
--- a/src/Codeception/Command/Shared/ActorTrait.php
+++ b/src/Codeception/Command/Shared/ActorTrait.php
@@ -4,9 +4,11 @@ declare(strict_types=1);
 
 namespace Codeception\Command\Shared;
 
+use Codeception\Scenario;
+
 trait ActorTrait
 {
-    protected function getActor(): ?string
+    protected function getActorClassName(): ?string
     {
         if (empty($this->settings['actor'])) {
             return null;
@@ -21,8 +23,16 @@ trait ActorTrait
         if (isset($this->settings['support_namespace'])) {
             $namespace .= '\\' . $this->settings['support_namespace'];
         }
+
         $namespace = rtrim($namespace, '\\') . '\\';
 
         return $namespace . $this->settings['actor'];
+    }
+
+    private function getActor($test)
+    {
+        $actorClass = $this->getActorClassName();
+
+        return $actorClass ? new $actorClass(new Scenario($test)) : null;
     }
 }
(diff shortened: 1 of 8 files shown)


Arhell (approved):

src/Codeception/Command/Shared/ActorTrait.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Codeception\Command\Shared;
+
+trait ActorTrait
+{
+    protected function getActor(): ?string
```
DavertMik, line 9: I would expect this function to actually return an Actor
Can we rename it to `getActorClassName`

src/Codeception/Command/Shared/ActorTrait.php
```diff
@@ -0,0 +1,28 @@
...
+            $namespace .= '\\' . $this->settings['namespace'];
+        }
+
+        if (isset($this->settings['support_namespace'])) {
+            $namespace .= '\\' . $this->settings['support_namespace'];
+        }
+        $namespace = rtrim($namespace, '\\') . '\\';
+
+        return $namespace . $this->settings['actor'];
+    }
```
DavertMik, line 27: Method `getActor` can also be implemented here:

```php
private function getActor($test) 
{
    $actorClass = $this->getActor();
    return $actorClass ? new $actorClass(new Scenario($test)) : null;
}
```

src/Codeception/Test/Loader/Cest.php
```diff
@@ -51,9 +62,14 @@ public function loadTests(string $filename): void
                     continue;
                 }
 
+                $test = new CestFormat($unit, $method, $filename);
+                $actorClass = $this->getActor();
```
DavertMik, line 66:
```suggestion
                $I = $this->getActor($test);
```

DavertMik (changes requested):
@weeg Thanks for the PR
I like it in general and I have some ideas to improve it
Please tell what you think

W0rma:
@DavertMik @Naktibalda Are you okay with the latest changes in this PR?

Seems like the author addressed all comments with the latest commit.

Naktibalda:
@weeg @W0rma @DavertMik Could someone document this new feature somewhere?

https://github.com/Codeception/codeception.github.com/tree/master/docs

PR #6710, opened 2023-11-08
Avoid infinite loop while waiting for all running tests to finish

### LINKED ISSUE #6709: Code coverage stopped working after upgrading 4 to 5

Trying to upgrade to the latest Codeception from codeception 4.

#### What do you get instead?

Everything works as it did on Codeception 4, however the code coverage just hands after suite is finished with no output.

Example of command I run:
```bash
    codecept run app-suite --coverage --coverage-xml
```

This only hangs after suite is finished if coverage option is passed, after digging with xdebug I found why it hangs, this is related to commit https://github.com/Codeception/Codeception/commit/f99620c5fb36c6c7dd25c6c9cc876a5329b4a9f2 in codeception, if I remove the code added in this commit, coverage just calculates fine, so I just need help understanding the nature of this issue and why is it happening.

### Details

* Codeception version: 5.0.12
* PHP Version: 8.2.10
* Operating System: Ubuntu 20.04
* Installation type:  Composer
* Suite configuration:

My codeception.yml file:
```yml
paths:
    tests: tests/integration
    output: tests/integration/_output
    data: tests/integration/_data
    support: tests/integration/_support
    envs: tests/integration/_envs
bootstrap: bootstrap.php
actor_suffix: Tester
extensions:
    enabled:
# To use code coverage `codecept run --coverage`, check --help to see the full list of available output types
coverage:
    enabled: true
    remote: false
    include:
        - application/modules/*/code/*.php
        - library/pi-ext/lambda-utils/src/*.php
        - library/pi-ext/s3-client/src/*.php
        - library/pi-ext/search-data/src/*.php
        - library/pi-ext/site-utils/src/*.php
        - library/pi-ext/time/src/*.php
        - library/pi-ext/url-utils/src/*.php
        - library/iputils/*.php
    exclude:
        - application/modules/tractor/code/Model/TrackingStateManager.php
        - library/pi-ext/search-data/tests/*
        - library/pi-ext/site-utils/tests/*
        - library/pi-ext/time/tests/*
        - library/pi-ext/url-utils/tests/*

modules:
    config:
        PhpBrowser:
            url: http://localhost
            curl:
                CURLOPT_TIMEOUT: 90 # https://stackoverflow.com/questions/43181920/codeception-rest-timeout-issue
        REST:
            depends: PhpBrowser
            shortDebugResponse: 300 # only the first 300 chars of the response
            timeout: 90
        Db:
            dsn: 'mysql:host=db;dbname=pi'
            user: 'root'
            password: 'root'
            dump: 'tests/integration/dump.sql'
            populate: true
            cleanup: false
            populator: 'mysql -u $user -p$password -h $host $dbname < $dump'
            initial_queries:
                - 'CREATE DATABASE IF NOT EXISTS pi;'
                - 'USE pi;'
                - 'SET NAMES utf8;'```

MarcelBolten:
This could potentially close #6709 but on the other hand it might result in truncated reports again. This would happen if the accumulated runtime of the [registered shutdown functions in c3.php](https://github.com/Codeception/c3/blob/e23298a1cd5e7745973ea26a53572a3d9b013439/c3.php#L416) exceeds 10 * 0.25 = 2.5 sec which seems possible.
Maybe a bigger value for `retries` is necessary.

### DIFF
--- a/src/Codeception/Coverage/Subscriber/LocalServer.php
+++ b/src/Codeception/Coverage/Subscriber/LocalServer.php
@@ -151,7 +151,8 @@ class LocalServer extends SuiteSubscriber
         // wait for all running tests to finish
         $blockfilename = Configuration::outputDir() . 'c3tmp/block_report';
         if (file_exists($blockfilename) && filesize($blockfilename) !== 0) {
-            while (file_get_contents($blockfilename) !== '0') {
+            $retries = 10;
+            while (file_get_contents($blockfilename) !== '0' && --$retries >= 0) {
                 usleep(250_000); // 0.25 sec
             }
         }
--- a/src/Codeception/Coverage/Subscriber/LocalServer.php
+++ b/src/Codeception/Coverage/Subscriber/LocalServer.php
@@ -13,6 +13,7 @@ use Codeception\Events;
 use Codeception\Exception\ModuleException;
 use Codeception\Exception\RemoteException;
 use Codeception\Lib\Interfaces\Web as WebInterface;
+use Codeception\Lib\Notification;
 use Codeception\Module\WebDriver as WebDriverModule;
 use Facebook\WebDriver\Exception\NoSuchAlertException;
 use RuntimeException;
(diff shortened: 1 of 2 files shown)


MarcelBolten:
It runs fine for our tests at github.com/elabftw/elabftw, no infinite loops.

However, I tried to enable path coverage and ran out of memory during (un)serializing in the registered shutdown function in c3.php. In this case, there was an infinite loop because the shutdown function did not finish, resulting in "unfinished" tests. Hence, blocked reports.

I would argue that in this case, it is good to realize there is a problem. Even when it is not caused by the code that is tested.

MarcelBolten:
Let me change `$retries` so the loop runs for 30s.
I will push it later today.

MarcelBolten:
Wait period is set to 30s (`$retries = 120;`).
Also, a `Notification::warning()` is added and will be shown when 30s are reached.
`tests/data/claypit/c3.php` is updated to version 2.9.0

PR #6721, opened 2023-11-30
Updating `composer.json` for Symfony 7.0

ThomasLandauer:
I didn't check if it really works...


Arhell (approved):

composer.json
```diff
@@ -28,12 +28,12 @@
...
-        "symfony/event-dispatcher": ">=4.4.24 <7.0",
-        "symfony/finder": ">=4.4.24 <7.0",
-        "symfony/yaml": ">=4.4.24 <7.0",
-        "symfony/var-dumper": ">=4.4.24 < 7.0",
+        "symfony/console": ">=4.4.24",
+        "symfony/css-selector": ">=4.4.24",
+        "symfony/event-dispatcher": ">=4.4.24",
+        "symfony/finder": ">=4.4.24",
+        "symfony/yaml": ">=4.4.24",
+        "symfony/var-dumper": ">=4.4.24",
```
Naktibalda, line 36: Please add `<8.0` to all constraints.
We don't want to make a false impression that Codeception 5.0.13 supports Symfony 20.

ThomasLandauer: Wouldn't it be better to switch the "default" from incompatible to compatible? I guess it's more probably that Symfony 8 will be supported than not supported.
And right now the version constraint is preventing people from upgrading to Symfony 7, until somebody (=you) is taking action. If we omit the constraint, anybody will be able to upgrade to Symfony 8 immediately, and if it's not working, somebody will report/fix it anyway...

What do you think?

Naktibalda: How about keeping the upper bound for now, but it will be up to you to decide what to do when Symfony 8 is released? :)
I updated internal dependencies (lib-web, lib-xml, module-filesystems) also merged changes to lib-innerbrowser and module-phpbrowser, but we still can't upgrade all Codeception symfony dependencies to v7, until a new version of `psy/psysh` package is released.

With psysh constraint set to `"psy/psysh": "dev-main#c65d832"` a few errors happen in cli tests:

> PHP Fatal error:  Declaration of Project\Command\YourCustomCommand::execute(Symfony\Component\Console\Input\InputInterface $input, Symfony\Component\Console\Output\OutputInterface $output) must be compatible with Symfony\Component\Console\Command\Command::execute(Symfony\Component\Console\Input\InputInterface $input, Symfony\Component\Console\Output\OutputInterface $output): int in /.../Codeception/tests/data/register_command/examples/YourCustomCommand.php on line 70

W0rma: > With psysh constraint set to "psy/psysh": "dev-main#c65d832" a few errors happen in cli tests:

ThomasLandauer: Sorry, I didn't receive notification that somebody answered here...

I added `<8.0` now - the psysh issue convinced me: There will always be some dependency that needs some update...
Besides that, I don't understand what you meant here:
> but it will be up to you to decide what to do when Symfony 8 is released? :)

@W0rma: Could you continue with the return types?

PR #6728, opened 2024-02-02
Support PHPUnit 11

Naktibalda:


DavertMik (approved):
:rocket:

src/Codeception/Coverage/Filter.php
```diff
@@ -74,18 +75,22 @@ public function whiteList(array $config): self
...
+                            $filter->excludeFile((string)$file);
+                        }
+                    } catch (DirectoryNotFoundException) {
+                        continue;
                     }
-                } catch (DirectoryNotFoundException) {
-                    continue;
                 }
+            } else {
+                Notification::warning('`coverage: whitelist: exclude` option has no effect since PHPUnit 11.0', '');
```
Naktibalda, line 93: @DavertMik What do you think about this change?

W0rma (approved):

Naktibalda:
I released this change as 5.1.0 - https://github.com/Codeception/Codeception/releases/tag/5.1.0

Usually I'm reluctant to bump minor version because it requires updating scripts in https://github.com/Codeception/codeception.github.com/blob/master/RoboFile.php and possibly some changes I don't even remember now.

I will update RoboFile.php later and document required changes in Wiki.

PR #6741, opened 2024-02-22
Replace Doctrine2 with Doctrine module

W0rma:
The Doctrine2 module was replaced by the Doctrine module which adds support for newer Doctrine versions while keeping compatibility with Doctrine ORMv2/DBALv3.

See https://github.com/Codeception/module-doctrine/pull/28

### DIFF
--- a/src/Codeception/Lib/ModuleContainer.php
+++ b/src/Codeception/Lib/ModuleContainer.php
@@ -45,7 +45,7 @@ class ModuleContainer
         'Cli' => 'codeception/module-cli',
         'DataFactory' => 'codeception/module-datafactory',
         'Db' => 'codeception/module-db',
-        'Doctrine2' => "codeception/module-doctrine2",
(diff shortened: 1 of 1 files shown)


TavoNiievez (approved):

src/Codeception/Lib/ModuleContainer.php
```diff
@@ -45,7 +45,7 @@ class ModuleContainer
         'Cli' => 'codeception/module-cli',
         'DataFactory' => 'codeception/module-datafactory',
         'Db' => 'codeception/module-db',
-        'Doctrine2' => "codeception/module-doctrine2",
+        'Doctrine' => "codeception/module-doctrine",
         'Filesystem' => 'codeception/module-filesystem',
         'FTP' => 'codeception/module-ftp',
         'Laravel5' => 'codeception/module-laravel5',
```
TavoNiievez, line 51: now that I see it, this should also be renamed

Issue #6716, opened 2023-11-26
Try-catch in the second test catches fail in the first test


erolatex:
#### What are you trying to achieve?
Migrated from codeception 4 to 5
Try to use try-catches in tests and page methods and got strange behaviour:
If first test failed. Second test with try catch fails too, but all asserts is passed.
if first test passed. Second test works normal. (

#### What do you get instead?

> Run Positive scenario: first test good, second test with try-catch good.

```bash
# php vendor/bin/codecept run tests/PositiveCest -vvv 
```
> Provide test source code if related 

```php
<?php

namespace Tests;

use Tests\Support\AcceptanceTester;

class PositiveCest 
{    
    public function _before(AcceptanceTester $I)
    {
        $I->amOnUrl('https://google.com');
    }

    public function loginSuccessfully(AcceptanceTester $I)
    {
        $I->wait(1);
    }
    
    public function loginWithInvalidPassword(AcceptanceTester $I)
    {
        try {
            $I->waitForElementVisible('//membercat') | RuntimeException;
        } catch (\Exception){}
    }       
}
```
<img width="570" alt="image" src="https://github.com/Codeception/Codeception/assets/41438388/05d86a83-844c-473f-ae0d-da9f5a0b979d">

> Run Negative scenario: first test failed, second test with try-catch should be passed.

```bash
# php vendor/bin/codecept run tests/NegativeCest -vvv 
```
> Provide test source code if related 

```php
<?php

namespace Tests;

use Tests\Support\AcceptanceTester;

class NegativeCest 
{    
    public function _before(AcceptanceTester $I)
    {
        $I->amOnUrl('https://google.com');
    }

    public function loginSuccessfully(AcceptanceTester $I)
    {
        $I->wait(1);
        $I->fail('Test one fail message');
    }
    
    public function loginWithInvalidPassword(AcceptanceTester $I)
    {
        try {
            $I->waitForElementVisible('//membercat') | RuntimeException;
        } catch (\Exception){}
    }       
}
```
<img width="745" alt="image" src="https://github.com/Codeception/Codeception/assets/41438388/28b25499-ce6d-4ff4-a4d8-d2b28f934ce1">

trace
```
Scenario Steps:

 2. $I->waitForElementVisible("//membercat") at tests/NegativeCest.php:23
 1. $I->amOnUrl("https://google.com") at tests/NegativeCest.php:11

[...]

PR #6743, opened 2024-02-29
Prevent unrelated error from being displayed if a scenario step has failed

craig-mcmahon:
This fixes https://github.com/Codeception/Codeception/issues/6716 which results in incorrect test failures being reported where a scenario has a failed assertion that has been caught within the test itself or via `expectException`

### DIFF
--- a/src/Codeception/ResultAggregator.php
+++ b/src/Codeception/ResultAggregator.php
@@ -219,4 +219,9 @@ class ResultAggregator
     {
         return array_pop($this->failures);
     }
+
+    public function getLastFailure(): ?FailEvent
+    {
+        return end($this->failures) ?: null;
+    }
 }
--- a/src/Codeception/Test/Test.php
+++ b/src/Codeception/Test/Test.php
@@ -169,17 +169,8 @@ abstract class Test extends TestWrapper implements TestInterface, Interfaces\Des
                 $status = self::STATUS_OK;
                 $eventType = Events::TEST_SUCCESS;
 
-                if (method_exists($this, 'getScenario')) {
-                    foreach ($this->getScenario()?->getSteps() ?? [] as $step) {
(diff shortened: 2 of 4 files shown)


Naktibalda:
Your change broke [conditional assertions](https://codeception.com/docs/AcceptanceTests#Conditional-Assertions)
Probably because condition assertion implementation passes a clone of test to FailEvent:
https://github.com/Codeception/Codeception/blob/fcfa1d0ec2eb7439e5290fd3ea685b5d1f90075a/src/Codeception/Scenario.php#L79

Naktibalda:
I think that the best way to compare if 2 variables contain the same test is to compare results of `\Codeception\Test\Descriptor::getTestSignatureUnique()`.

```php
if ($lastFailure && Descriptor::getTestSignatureUnique($lastFailure->getTest()) === Descriptor::getTestSignatureUnique($this)
```

For ideas how to implement test for your change see https://github.com/Codeception/Codeception/blob/fcfa1d0ec2eb7439e5290fd3ea685b5d1f90075a/tests/cli/RunCest.php#L606-L669

https://github.com/Codeception/Codeception/blob/fcfa1d0ec2eb7439e5290fd3ea685b5d1f90075a/tests/cli/OrderCest.php#L16-L75 contains some similar tests too, but they are more difficult to understand.

craig-mcmahon:
Thanks for the feedback, your suggestion fixed the broken test. I have also refactored to avoid the loop when not needed.

I will look at adding a new test in the next few days

craig-mcmahon:
I have added a test to cover the new change, and fixed the code style check that failed in the last run

craig-mcmahon:
And fix pushed for the code style in the tests

Naktibalda:
Released as 5.1.2

PR #6754, opened 2024-05-08
Fix  FAIL message color highlighting

### LINKED ISSUE #6740: Running with --no-ansi still uses color

When running codecept with `--report --no-ansi` there is still color output appearing:

image

This appears to be done by both Recorder and by the FAIL result

### Details

* Codeception version: 5.0.12
* PHP Version: 8.1.27
* Operating System: Windows

antonvolokha:
Closes #6740

When I run `codecept run --report --no-ansi` there is still color output  
image

So I fixed it by removing extra chars in `ReportPrinter::testFailure`, I think it just typo

### DIFF
--- a/src/Codeception/Reporter/ReportPrinter.php
+++ b/src/Codeception/Reporter/ReportPrinter.php
@@ -64,7 +64,7 @@ class ReportPrinter implements ConsolePrinter
 
     public function testFailure(FailEvent $event): void
     {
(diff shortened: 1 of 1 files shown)


Naktibalda:
Cool, nobody questioned necessity of these ANSI codes since initial commit: https://github.com/Codeception/Codeception/commit/111d08289a16493068c9d62b7b6c6c58daffc758#diff-10133a5d6a2cd3c004acbefa34ae0455eacd4febda3b7915e9945808158f6021R17

PR #6796, opened 2024-09-18
Prevent serialization of non-backed enums

innerfly:
Fixes an https://github.com/Codeception/Codeception/issues/6753, when json_encode fails to serialize non-backed enum.   
Follow up https://github.com/Codeception/Codeception/pull/6762#issuecomment-2343641842

### DIFF
--- a/src/Codeception/Test/Descriptor.php
+++ b/src/Codeception/Test/Descriptor.php
@@ -7,6 +7,7 @@ namespace Codeception\Test;
 use Codeception\Test\Interfaces\Descriptive;
 use Codeception\Test\Interfaces\Plain;
 use Codeception\TestInterface;
+use JsonException;
 use PHPUnit\Framework\SelfDescribing;
 
 use function codecept_relative_path;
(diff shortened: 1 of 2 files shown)


Arhell (commented):
@Naktibalda look when you have time, thanks

W0rma:
@innerfly Can you try to add a test case?

src/Codeception/Test/Descriptor.php
```diff
@@ -33,9 +34,13 @@ public static function getTestSignatureUnique(SelfDescribing $testCase): string
         if (method_exists($testCase, 'getScenario') && $env = $testCase->getScenario()?->current('env')) {
             $signature .= ':' . $env;
         }
-        if (method_exists($testCase, 'getMetadata') && $example = $testCase->getMetadata()->getCurrent('example')) {
-            $encoded = json_encode($example, JSON_THROW_ON_ERROR | JSON_INVALID_UTF8_SUBSTITUTE);
-            $signature .= ':' . substr(sha1($encoded), 0, 7);
+        try {
+            if (method_exists($testCase, 'getMetadata') && $example = $testCase->getMetadata()->getCurrent('example')) {
+                $encoded = json_encode($example, JSON_THROW_ON_ERROR | JSON_INVALID_UTF8_SUBSTITUTE);
```
burned42: This will avoid breaking the whole execution, however this might result in non-unique signatures being returned as the error will just get ignored. I don't know if this is an issue or not.

But I think we could improve this even further by avoiding an exception entirely. We could convert enums manually before passing the example to `json_encode`. Maybe something like this:


```suggestion
            $example = self::normalizeForJsonEncoding($example);
            $encoded = json_encode($example, JSON_THROW_ON_ERROR | JSON_INVALID_UTF8_SUBSTITUTE);
        // ...
    }
                
    private static function normalizeForJsonEncoding(mixed $value): mixed
    {
        return match (true) {
            is_array($value) => array_map(self::normalizeForJsonEncoding(...), $value),
            $value instanceof UnitEnum => $value->name,
            default => $value,
        };
    }
```

innerfly: @burned42 I have added manual enum convertions following your idea, and test cases to cover this

W0rma:
In one of our projects we recently stumbled over the problem fixed by this PR.

The implementation LGTM.

@TavoNiievez Would you mind taking a look?
In the first step it would be great to approve the workflow in order to run the tests in the CI.

PR #6828, opened 2025-02-17
Fix missing absolute path resolving in ParamsLoader

garvinhicking:
This is a regression bugfix for release 5.2.0.

With https://github.com/Codeception/Codeception/pull/6767 the ParamsLoader was streamlined, but when previously `$paramsFile` was passed to the loader methods, now only `$paramStorage` is passed, which misses getting converted to an absolute path via `codecept_absolute_path()`.

This patch restores utilizing `$paramsFile` and also uses that to emit warnings when a file cannot be loaded.

This bug caused missing file resolving, because software like TYPO3 for example executes tests from the repo directory `typo3/sysext/core/Tests/codeception.yml` and uses:

```
params:
  - parameters.yml
  - env.yml
```

and expects that the file is found next to the `codeception.yml` file in the same directory, and thus a breaking change.

### DIFF
--- a/src/Codeception/Lib/ParamsLoader.php
+++ b/src/Codeception/Lib/ParamsLoader.php
@@ -51,14 +51,14 @@ class ParamsLoader
         foreach ($loaderMappings as $method => $pattern) {
             if (preg_match($pattern, $paramStorage)) {
                 try {
-                    return self::$method($paramStorage);
+                    return self::$method($paramsFile);
                 } catch (Exception $e) {
-                    throw new ConfigurationException("Failed loading params from {$paramStorage}\n" . $e->getMessage());
+                    throw new ConfigurationException("Failed loading params from {$paramsFile}\n" . $e->getMessage());
                 }
             }
         }
 
-        throw new ConfigurationException("Params can't be loaded from `{$paramStorage}`.");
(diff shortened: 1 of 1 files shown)


garvinhicking:
Thanks for the trust in advance. :) I'll see what I can do for some tests, I haven't yet checked this out, but should be possible. Give me a day or two.

garvinhicking:
@TavoNiievez I tried to get the tests up and running, according to the dockerized README in `tests`:

```
git clone ...
cd tests
docker-compose build
docker-compose up -d
docker-composer run --rm codecept run cli
```

which just yields:

```
Codeception PHP Testing Framework v5.2.1 https://stand-with-ukraine.pp.ua/
[Seed] 1553456819

In Assert.php line 2286:

  Result code was 1.

  Could not open input file: /repo/codecept
```

Seems that `/repo/` is a mount to my cwd `tests` but there's no `codecept` directory and I wonder where it could stem from. What step could I be missing?

I assume that I would add tests with some dummy codeception.yaml files plus configured `params: ...` files with dummy files too, and then do assertions on non/failure variants, and do that in `tests/cli/GlobalCommandOptionCest.php`. However I need to get the tests to actually run, and my experience with codeception is just one as a consumer right now. If you have some docs to point me to to get this up and running, I'd be more than happy to contribute.

Issue #6833, opened 2025-02-26
Incorrect path for behat i18n.php file


santilin:
**Describe the bug**
i18n.php file cannot be loaded in Gherkin tests due to incorrect file path resolution in Gherkin.php loader.

**To Reproduce**
1. Configure Codeception 5.0 with Gherkin and i18n support
2. Create Gherkin feature files with localized steps
3. Run tests using `codecept run`
4. See error: "Failed opening required '.../i18n.php'"

**Expected behavior**
Successful loading of i18n.php for Gherkin localization support.

**Screenshots/Code Snippets**
Problematic code in `vendor/codeception/codeception/src/Codeception/Test/Loader/Gherkin.php:77`:

    $i18n = require $gherkinClassPath . '/../../../i18n.php';

**Proposed Fix**

     $i18n = require $gherkinClassPath . '/../i18n.php';

**Environment**
- Codeception version: 5.0.x
- PHP Version: 8.2
- Operating System: debian 12
- Installation type: Composer

**Additional context**
The incorrect relative path breaks Gherkin's internationalization support. The correct path should resolve to:
`src/Codeception/Test/Loader/../i18n.php` instead of navigating three directories up. This matches the actual file structure in the Codeception repository.

DanaLuther:
Just ran into the same issue after upgrading packages. Took a while to figure out this was the cause of the suite completely breaking.

PR #6836, opened 2025-03-11
Fix include path for behat/gherkin v4.12

W0rma:
Fixes https://github.com/Codeception/Codeception/issues/6833

https://getcomposer.org/doc/07-runtime.md#knowing-the-path-in-which-a-package-is-installed is used to determine the vendor directory of the `behat/gherkin` composer package.
This way it is no longer necessary to work with paths relative to some source files which was broken because `behat/gherkin:4.12` relocated some source files.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,6 +22,7 @@
         "behat/gherkin": "^4.6.2",
         "codeception/lib-asserts": "^2.0",
         "codeception/stub": "^4.1",
+        "composer-runtime-api": "^2.1",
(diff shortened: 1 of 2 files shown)


W0rma:
@Naktibalda @TavoNiievez What's your opinion here?

Currently, test pipelines for some codeception modules fail if the latest `behat/gherkin` version is installed, f.e. https://github.com/Codeception/module-doctrine/actions/runs/13508538018/job/37743638356

TavoNiievez:
@W0rma If this is the only place in the code where `composer-runtime-api` is going to be used I would prefer to create a simple if with `class_exists`.

PR #6839, opened 2025-03-30
Fix loading keywords in behat/gherkin v4.12

W0rma:
Fixes https://github.com/Codeception/Codeception/issues/6833

Replaces https://github.com/Codeception/Codeception/pull/6836

I overlooked that the release notes https://github.com/Behat/Gherkin/releases/tag/v4.12.0 already suggest a simple way to load the keywords independently from the file path:

> Provide CachedArrayKeywords::withDefaultKeywords() to create an instance without an external dependency on the path to the i18n.php file in this repo. NOTE that paths to source files will change in the next Gherkin release - use the new constructor to avoid any impact.

This PR uses this approach and bumps the minimum version of `behat/gherkin` to 4.12 which should be okay because this version does still support PHP 8.1 which is also the case for codeception v5.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "ext-curl": "*",
         "ext-json": "*",
         "ext-mbstring": "*",
-        "behat/gherkin": "^4.6.2",
+        "behat/gherkin": "^4.12",
         "codeception/lib-asserts": "^2.0",
         "codeception/stub": "^4.1",
         "phpunit/phpunit": "^9.5.20 || ^10.0 || ^11.0 || ^12.0",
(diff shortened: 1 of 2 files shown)


TavoNiievez (approved):

TavoNiievez:
Thank you @W0rma

francislavoie:
@TavoNiievez can we get a release for this? I'd rather not have to stick a crappy workaround in my `composer.json` :(

TavoNiievez:
I released Codeception version [5.3.0](https://github.com/Codeception/Codeception/releases/tag/5.3.0), with these changes included,

but, a few hours ago a new version of Gherkin [4.13](https://github.com/Behat/Gherkin/releases/tag/v4.13.0) was also released which seems to be giving problems again.

cc. @W0rma

francislavoie:
Thanks, 5.3.0 seems to work fine for me (I don't use Gherkin at all)

PR #6850, opened 2025-04-29
Migrate commands to use AsCommand attribute

TavoNiievez:


src/Codeception/Command/GeneratePageObject.php
```diff
@@ -21,14 +22,18 @@
  * * `codecept g:page Registration`
  * * `codecept g:page acceptance Login`
  */
+#[AsCommand(
+    name: 'generate:page',
+    description: 'Generates empty PageObject class'
+)]
 class GeneratePageObject extends Command
```
W0rma, line 29: I'm not familiar with the autogenerated console name which was created before but wasn't it `generate:pageobject` instead of `generate:page`?

`generate:pageobject` is shown if I run `vendor/bin/codecept list` in my local app.

W0rma (approved):

src/Codeception/Command/GeneratePageObject.php
TavoNiievez: don't worry, you're right.

I fixed the following names: `generate:environment`, `generate:groupobject`, `generate:pageobject`

`self-update` is not listed, but I never used that and I don't even know if it is working.

TavoNiievez: The interesting thing about this is that the tests did not complain despite the change, suggesting that we are not testing this as we should ;-)

PR #6887, opened 2025-10-23
chore: allow installation of lib-asserts v3

W0rma:
https://github.com/Codeception/module-asserts/pull/33 should be merged (and tagged) before.


W0rma:
@TavoNiievez It would be great if you could review this PR and the related https://github.com/Codeception/module-asserts/pull/33

Issue #6888, opened 2025-10-23
Codeception 5.3.2 dry-run "In DryRun.php line 194: Unsupported return type iterable"


paulcdejean:
This seems like its a bug, as it knows how to handle arrays but panics if it gets an iterable.

paulcdejean:
After applying #6889 

> ~/longhornphp/codeception-demo $ php vendor/bin/codecept dry-run acceptance sponsors.feature           
> 
> Tests.acceptance Tests (1) --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
> Modules: PhpBrowser
> -----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
> sponsors: try sponsors
> Signature: sponsors:try sponsors
> Test: tests/sponsors.feature:try sponsors
> Scenario --
>  In order to thank our sponsors
>  As a organizer
>  I need to make sure all my sponsors are on the sponsor page
>  Given i am on the sponsor page 
>  Then i should see "vehikl" as a gold sponsor 
> 
>  INCOMPLETE 
> Step definition for `I am on the sponsor page` not found in contexts
> Step definition for `I should see "vehikl" as a gold sponsor` not found in contexts
> Run gherkin:snippets to define missing steps

PR #6889, opened 2025-10-23
Add iterable to list of defaultable types for DryRun

paulcdejean:
Fixes https://github.com/Codeception/Codeception/issues/6888


W0rma:
@paulcdejean Would you mind adding a test case which demonstrates the bugfix in order to avoid future regressions?

PR #6916, opened 2025-12-19
Removing "hidden" `memory_limit`

### LINKED ISSUE #6840: Test runner sometimes just stops, sometimes gives "Allowed memory size exhausted" error
Since my latest run (~2 weeks ago), the Codeception runner is suddenly hanging strangely:
* Sometimes, in the middle of the suite, it just stops with message "COMMAND DID NOT FINISH PROPERLY."
* Sometimes (at another test), it crashes with:
  > PHP Fatal error:  Allowed memory size of 1073741824 bytes exhausted (tried to allocate 20480 bytes) in ...vendor/symfony/var-dumper/Caster/ClassStub.php on line 52

  However, the `memory_limit` in my `php.ini`s (for CLI and FPM) is `3000M`. So where is the 1G limit coming from?

This is not caused by the tests themselves, cause when I run just the affected files alone, everything's fine.

Any ideas? Is somebody else experiencing this? Were there recent changes regarding memory management (couldn't find anything)?

ThomasLandauer:
Fixes https://github.com/Codeception/Codeception/issues/6840

This line was introduced in 2013 by https://github.com/Codeception/Codeception/commit/558498efbf8f125c14e9a0bc55791b26b1b44162

Overriding what the user has in `php.ini` doesn't make sense to me.

### DIFF
--- a/src/Codeception/Codecept.php
+++ b/src/Codeception/Codecept.php
@@ -197,8 +197,6 @@ class Codecept
 
     public function run(string $suite, ?string $test = null, ?array $config = null): void
     {
-        ini_set('memory_limit', $this->config['settings']['memory_limit'] ?? '1024M');
-
         $config = Configuration::suiteSettings($suite, $config ?: Configuration::config());
         $selectedEnvironments = $this->options['env'];
 


ArbitraryCritter:
Wouldn't the backwards compatible way of handling this be to do something like this:


Note: I haven't tested the code below. It probably at the very least needs some more null checks.

```
$hardCodedMemoryLimit = ini_parse_quantity('1024M');
$systemMemoryLimit = ini_parse_quantity(ini_get('memory_limit'));
$configuredMemoryLimit =  ini_parse_quantity($this->config['settings']['memory_limit']);

$calculatedMemoryLimit = max($hardCodedMemoryLimit, $systemMemoryLimit, $configuredMemoryLimit);
ini_set('memory_limit', $calculatedMemoryLimit);
```

Then deprecate "$this->config['settings']['memory_limit']" with a message asking people to just set the PHP memory limit and then remove the hardcoded minimum and "$this->config['settings']['memory_limit']" in the next major release?

s1lver:
If this option is removed, many people's tests will fail. For example, I use the limitation in the settings because there is no way to change php.ini. Therefore, the configuration in codeception.yaml is very convenient.

PR #6924, opened 2026-01-29
PHP 8.5: Avoid accessing deprecated $http_response_header

W0rma:
Follow-up of https://github.com/Codeception/Codeception/pull/6919

The current check does still trigger a deprecation warning in PHP 8.5.

This is fixed by using the check as suggested in https://wiki.php.net/rfc/deprecations_php_8_5#deprecate_the_http_response_header_predefined_variable


TavoNiievez (approved):
LGTM

PR #6929, opened 2026-03-11
catch general Throwable for before test execution code

applyACS:
We had an error in a seeder and this was not caught by codeception. No test was running but codeception return 0 so it looked like everything was fine.
This is a fix this that case.

### DIFF
--- a/src/Codeception/Test/Test.php
+++ b/src/Codeception/Test/Test.php
@@ -144,7 +144,7 @@ abstract class Test extends TestWrapper implements TestInterface, Interfaces\Des
             $this->fire(Events::TEST_BEFORE, new TestEvent($this));
             $this->runHooks('Start');
             $failedToStart = false;
-        } catch (\Exception $e) {
+        } catch (\Throwable $e) {
             $failedToStart = true;
             $this->dispatchOutcome(Events::TEST_ERROR, $e, $time);
         }


W0rma (approved):

burned42:
I just tried throwing an `\Error` in the try block (because that would be caught with your change, but not without) and codeception exited with a code 255 and the output contained a stack trace for `Uncaught Error`. No test was executed, but codeception also didn't exit with code 0 for me.

Applying your change makes codeception exit with a code of 1 and it does list all the tests as failing.

So the change looks ok to me, but I couldn't reproduce your issue with codeception exiting with code 0 in this case.

applyACS:
my error was: Call to a member function foo() on null
this was in one of the observers
maybe that help reproducing it

burned42:
I tried it with

```php
$foo = null;
$foo->foo();
```
in the try block and got
> PHP Fatal error:  Uncaught Error: Call to a member function foo() on null

but still got exit code 255.

I don't think this is a blocker for this PR though.

TavoNiievez:
I'm okay with this change, but I'm not sure it's the only place where it should be made.

applyACS:
Not sure if it's the only place. It's the place I've had a problem with.
Is this going to be approved?

TavoNiievez (approved):

PR #6938, opened 2026-04-16
Add support for phpunit/php-code-coverage 14.0

michalbundyra:


W0rma:
According to https://github.com/sebastianbergmann/php-code-coverage/blob/14.0.0/ChangeLog-14.0.md#removed the class `SebastianBergmann\CodeCoverage\Report\PHP` was removed but it is used f.e. in https://github.com/michalbundyra/Codeception/blob/dc433f018c3889397825f909debd693177f02120/src/Codeception/Coverage/Subscriber/Printer.php#L20

boesing:
I wonder why there are so many versions supported via constraints yet only the latest version(s) are tested?

W0rma:
@michalbundyra Thank you! I tried your changes in one of my projects and it worked like a charm.

W0rma:
@TavoNiievez Would you mind approving the workflow?

W0rma (approved):

W0rma:
> I wonder why there are so many versions supported via constraints yet only the latest version(s) are tested?

I don't think that only the latest version(s) are tested.
The CI currently tests against PHP 8.2 - 8.5.

`phpunit/php-code-coverage:` v13 and v14 require at least PHP 8.4, v12 requires PHP 8.3 and v11 requires PHP 8.2. Therefore the CI already covers some versions but not all.

We could think about adding additional CI jobs to cover more versions, but I'm not sure if this is a blocker for this PR.

TavoNiievez (approved):

PR #6939, opened 2026-05-16
Update Configuration.php: Adding path to missing config file error message

ThomasLandauer:


src/Codeception/Configuration.php
```diff
@@ -145,7 +145,7 @@ public static function config(?string $configFile = null): array
 
         $configDistFile = ($dir !== false ? $dir : dirname($configFile)) . DIRECTORY_SEPARATOR . 'codeception.dist.yml';
         if (!file_exists($configFile) && !file_exists($configDistFile)) {
-            throw new ConfigurationException("Configuration file could not be found.\nRun bootstrap to initialize Codeception.", 404);
+            throw new ConfigurationException("Configuration file $configFile could not be found.\nRun bootstrap to initialize Codeception.", 404);
```
W0rma, line 148: @ThomasLandauer The exception is thrown if neither `$configFile` nor `$configDistFile` were found. Should this be reflected in the the message, f.e. by using `Configuration files $configFile and $configDistFile could not be found`?

W0rma (changes requested):

src/Codeception/Configuration.php
ThomasLandauer: I forgot, which config files are we talking about here? Probably `codeception.yml` or `Functional.suite.yml`? I've never seen a `.yml,dist` file of those (where would they come from?), so I figured that's not no important.
Besides, IMO, the idea of `.dist` is to contain factory default values, so if you don't have one, the solution is not to create one, but rather to create your own `*.yml` file - isn't it?
And lastly, I didn't want to cover all (exotic) use cases, but just point the user to the right direction (i.e. which file).

